Issue #8, opened 2019-07-15
feat: add DidComm HTTP transport


Baha-sk:
Add the capability for Aries Agents to safely communicate over HTTPS transport.

This implements the HTTP(S) **Aries-RFC 25**: https://github.com/hyperledger/aries-rfcs/tree/master/features/0025-didcomm-transports#https

Only HTTPS will be supported for secured communications.

And must support the DID Exchange Protocol 1.0 as per **Aries-RFC 23**: https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange

To complete transport security, future work (after this issue) will include the Pack/Unpack messaging capability as described in **Aries-RFC 19** to encode/decode & authenticate senders and receivers: https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope

troyronda:
Part of Epic #43

PR #9, opened 2019-07-15
feat: Add Outbound HTTPS transport layer

Baha-sk:
Add HTTPS Outbound capability for Aries Agents to safely communicate between each other.

	This implements the HTTP(S) Aries-RFC 25:
	https://github.com/hyperledger/aries-rfcs/tree/master/features/0025-didcomm-transports#https

	Only HTTPS will be supported for secured communications.

	It support the DID Exchange Protocol 1.0 as per Aries-RFC 23:
	https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange

	To support unit-tests, sample test keys were generated using the openssl tool

	Future work (to be implemented):
	* Inbound HTTPS capability
	* Implement the Pack/Unpack messaging capability
	as described in Aries-RFC 19:
	https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope

closes issue #8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,13 @@ license:
 	@scripts/check_license.sh
 
 .PHONY: unit-test
-unit-test:
+unit-test: generate-test-keys
 	@scripts/check_unit.sh
+
+generate-test-keys: clean
+	@scripts/openssl_env.sh scripts/generate_test_keys.sh
 
 .PHONY: clean
 clean:
+	rm -Rf test/fixtures/keys
 	rm -f coverage.txt
--- a/pkg/didcomm/transport/http/outbound.go
+++ b/pkg/didcomm/transport/http/outbound.go
@@ -1,0 +1,108 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package http
+
+import (
+	"bytes"
+	"crypto/tls"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+const commContentType = "application/didcomm-envelope-enc"
+
+// outboundCommHTTPOpts holds options for the HTTP transport implementation of CommTransport
+// it has an http.Client instance
+type outboundCommHTTPOpts struct {
+	client *http.Client
+}
+
+// OutboundHTTPOpt is an outbound HTTP transport option
+type OutboundHTTPOpt func(opts *outboundCommHTTPOpts)
+
+// WithOutboundHTTPClient option is for creating an Outbound HTTP transport using an http.Client instance
+func WithOutboundHTTPClient(client *http.Client) OutboundHTTPOpt {
+	return func(opts *outboundCommHTTPOpts) {
+		opts.client = client
+	}
+}
+
+// WithOutboundTimeout option is for creating an Outbound HTTP transport using a client timeout value
+func WithOutboundTimeout(timeout time.Duration) OutboundHTTPOpt {
+	return func(opts *outboundCommHTTPOpts) {
+		opts.client.Timeout = timeout
+	}
+}
+
+// WithOutboundTLSConfig option is for creating an Outbound HTTP transport using a tls.Config instance
+func WithOutboundTLSConfig(tlsConfig *tls.Config) OutboundHTTPOpt {
+	return func(opts *outboundCommHTTPOpts) {
+		opts.client = &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: tlsConfig,
+			},
+		}
+	}
+}
+
+// OutboundHTTPClient represents the Outbound HTTP transport instance
+type OutboundHTTPClient struct {
+	client *http.Client
+}
+
+// NewOutbound creates a new instance of Outbound HTTP transport to Post requests to other Agents.
+// An http.Client or tls.Config options is mandatory to create a transport instance.
+func NewOutbound(opts ...OutboundHTTPOpt) (*OutboundHTTPClient, error) {
+	clOpts := &outboundCommHTTPOpts{}
+	// Apply options
+	for _, opt := range opts {
+		opt(clOpts)
+	}
+
+	if clOpts.client == nil {
+		return nil, errors.New("Can't create an outbound transport without an HTTP client")
+	}
+
+	cs := &OutboundHTTPClient{
+		client: clOpts.client,
+	}
+	return cs, nil
+}
+
+// Send sends a2a exchange data via HTTP (client side)
+func (cs *OutboundHTTPClient) Send(data string, url string) (string, error) {
+	resp, err := cs.client.Post(url, commContentType, bytes.NewBuffer([]byte(data)))
+	if err != nil {
+		log.Printf("HTTP Transport - Error posting did envelope to agent at [%s]: %v", url, err)
+		return "", err
+	}
+
+	var respData string
+	if resp != nil {
+		isStatusSuccess := resp.StatusCode == http.StatusAccepted || resp.StatusCode == http.StatusOK
+		if !isStatusSuccess {
+			return "", errors.Errorf("Warning - Received non success POST HTTP status from agent at [%s]: status : %v", url, resp.Status)
+		}
+		// handle response
+		defer func() {
+			e := resp.Body.Close()
+			if e != nil {
+				log.Printf("HTTP Transport - Error closing response body: %v", e)
+			}
+		}()
+		buf := new(bytes.Buffer)
+		_, e := buf.ReadFrom(resp.Body)
+		if e != nil {
+			return "", e
+		}
+		respData = buf.String()
+	}
+	return respData, nil
+}
--- a/pkg/didcomm/transport/http/outbound_test.go
+++ b/pkg/didcomm/transport/http/outbound_test.go
@@ -1,0 +1,179 @@
+/*
+	Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+	SPDX-License-Identifier: Apache-2.0
+*/
+
+package http
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+)
+
+const certPrefix = "../../../../test/fixtures/keys/"
+
+func TestWithOutboundOpts(t *testing.T) {
+	opt := WithOutboundHTTPClient(nil)
+	require.NotNil(t, opt)
+	clOpts := &outboundCommHTTPOpts{}
+	opt(clOpts)
+
+	opt = WithOutboundTimeout(10 * time.Second)
+	require.NotNil(t, opt)
+	clOpts = &outboundCommHTTPOpts{}
+	// opt.client is nil, so setting timeout should panic
+	require.Panics(t, func() { opt(clOpts) })
+
+	opt = WithOutboundTLSConfig(nil)
+	require.NotNil(t, opt)
+	clOpts = &outboundCommHTTPOpts{}
+	opt(clOpts)
+}
+
+func TestOutboundHTTPTransport(t *testing.T) {
+	// prepare http server
+	server := startMockServer()
+	port := server.Addr().(*net.TCPAddr).Port
+	serverUrl := fmt.Sprintf("https://localhost:%d", port)
+	defer func() {
+		err := server.Close()
+		if err != nil {
+			log.Fatalf("Failed to stop server: %s", err)
+		}
+	}()
+
+	//build a mock cert pool
+	cp := x509.NewCertPool()
+	err := addCertsToCertPool(cp)
+	require.NoError(t, err)
+
+	// build a tls.Config instance to be used by the outbound transport
+	tlsConfig := &tls.Config{
+		RootCAs:      cp,
+		Certificates: nil,
+	}
+	// create a new invalid Outbound transport instance
+	ot, err := NewOutbound()
+	require.Error(t, err)
+	require.EqualError(t, err, "Can't create an outbound transport without an HTTP client")
+
+	// now create a new valid Outbound transport instance and test its Send() call
+	ot, err = NewOutbound(WithOutboundTLSConfig(tlsConfig), WithOutboundTimeout(1*time.Second))
+	require.NoError(t, err)
+	require.NotNil(t, ot)
+
+	// test Outbound transport's api
+	// first with an empty url
+	r, e := ot.Send("Hello World", "")
+	require.Error(t, e)
+	require.Empty(t, r)
+
+	// now try a bad url
+	r, e = ot.Send("Hello World", "https://badurl")
+	require.Error(t, e)
+	require.Empty(t, r)
+
+	// and try with a 'bad' payload with a valid url..
+	r, e = ot.Send("bad", serverUrl)
+	require.Error(t, e)
+	require.Empty(t, r)
+
+	// finally using a valid url
+	r, e = ot.Send("Hello World", serverUrl)
+	require.NoError(t, e)
+	require.NotEmpty(t, r)
+
+}
+
+func addCertsToCertPool(pool *x509.CertPool) error {
+	var rawCerts []string
+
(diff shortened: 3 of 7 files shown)


Makefile
```diff
@@ -20,9 +20,13 @@ license:
 	@scripts/check_license.sh
 
 .PHONY: unit-test
-unit-test:
+unit-test: generate-test-keys
 	@scripts/check_unit.sh
 
+generate-test-keys: clean
```
troyronda: I think it's better to have artifacts for unit tests be generated with the go generate conventions.

go.mod
```diff
@@ -7,4 +7,5 @@ module github.com/hyperledger/aries-framework-go
 require (
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.3.0
+	github.com/trustbloc/aries-framework-go v0.0.0-20190712135907-d5047213aa2c
```
troyronda: This is incorrect.

pkg/didcomm/protocol/transport/http/http_transport.go
```diff
@@ -0,0 +1,262 @@
+/*
```
troyronda: transport is a sibling, not a child of protocol.
https://wiki.hyperledger.org/display/ARIES/Framework+Go+Package+Hierarchy

Makefile
troyronda: (also - it feels a bit odd to regenerate every time we do make unit-test).

scripts/generate_test_keys.sh
```diff
@@ -0,0 +1,50 @@
...
+#
+# Copyright SecureKey Technologies Inc. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+echo "Generating aries-framework-go Test PKI"
+mkdir -p test/fixtures/keys/tls
```
troyronda: Suggest a more generic name than "key"/"keys", since I see x509 certificates here too.

pkg/didcomm/protocol/transport/http/http_transport.go
```diff
@@ -0,0 +1,262 @@
...
+		case commHandler.IntroductionResponse.Path:
+			processPOSTRequest(w, r, commHandler.IntroductionResponse.HandlerFunc)
+			return
+		}
+
+		handler.ServeHTTP(w, r)
+	})
+}
+
+// TODO Log error message with common trustbloc/logger-lib
```
troyronda: This comment will need to be updated to work within Hyperledger.

pkg/didcomm/protocol/transport/http/http_transport.go
Baha-sk: moved transport at the same level as protocol

scripts/generate_test_keys.sh
Baha-sk: 'pki' maybe?

pkg/didcomm/protocol/transport/http/http_transport.go
Baha-sk: will update comment..

go.mod
Baha-sk: good catch, will update dependency..

Makefile
Baha-sk: will try to incorporate `//go:generate openssl....` in the test code

pkg/didcomm/protocol/transport/transport_interface.go
```diff
@@ -0,0 +1,30 @@
...
+// This is the client side of the agent
+type OutboundTransport interface {
+	// Send send a2a exchange data
+	Send(data string, destination string) (string, error)
+}
+
+// RequestRouter struct for path and handler function
+type RequestRouter struct {
+	Path        string
+	HandlerFunc func([]byte) error
```
troyronda: I'm uncertain about this pattern in this context.

If you create function signatures, they should be created as a type with a comment. e.g., https://golang.org/src/net/http/server.go?s=60931:60978#L1981

pkg/didcomm/transport/transport_interface.go
```diff
@@ -0,0 +1,30 @@
...
+
+// RequestRouter struct for path and handler function
+type RequestRouter struct {
+	Path        string
+	HandlerFunc func([]byte) error
+}
+
+// DIDCommHandler struct to pass details for request handlers
+type DIDCommHandler struct {
+	RecieveInvitation    *RequestRouter
```
troyronda: This should not be tightly coupled to specific message types.

pkg/didcomm/protocol/transport/http/http_transport.go
```diff
@@ -0,0 +1,262 @@
...
+		// if RootCAs is nil, client will use the host's root CA instead
+		RootCAs:      cp,
+		Certificates: nil,
+	}
+	return tlsConfig, nil
+}
+
+// DIDCommRequestHandler will create a new handler to enforce Did-Comm HTTP transport specs
+// then routes processing to the passed in handler argument
+func DIDCommRequestHandler(handler http.Handler, commHandler *transport.DIDCommHandler) http.Handler {
```
troyronda: Normally, we would use an option pattern to pass one or more handlers.

pkg/didcomm/protocol/transport/http/http_transport.go
```diff
@@ -0,0 +1,262 @@
...
+
+	validateHandler(commHandler)
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
```
troyronda: This should loop through a slice of passed arguments. This code is too tightly coupled.

pkg/didcomm/protocol/transport/http/http_transport.go
Baha-sk: Inbound request handling will be added in a separate PR

pkg/didcomm/protocol/transport/http/http_transport.go
Baha-sk: inbound handling will be done in a separate PR

pkg/didcomm/transport/transport_interface.go
Baha-sk: Inbound handling will be done in a separate PR

pkg/didcomm/protocol/transport/transport_interface.go
Baha-sk: actually, we can use `http.HandlerFunc` as type. Anyway, inbound handling will be done in a separate PR

pkg/didcomm/transport/http/outbound.go
```diff
@@ -0,0 +1,170 @@
...
+		client: client,
+	}
+	return cm, nil
+}
+
+// Send sends a2a exchange data via HTTP (client side)
+func (cs *OutboundCommHTTP) Send(data string, url string) (string, error) {
+	resp, err := cs.client.Post(url, commContentType, bytes.NewBuffer([]byte(data)))
+	if err != nil {
+		log.Printf("HTTP Transport - Error posting did envelope to agent at [%s]: %v", url, err)
```
rolsonquadras: This should be error logger. May be we can add a TODO here to change once we have a logging framework in the project.

Baha-sk: the intent is to replace this by a logger framework (and add possibly a few more logs around the code). This can be done a separate PR

Makefile
Baha-sk: done in a separate PR and will be merged right after this PR

pkg/didcomm/transport/http/tls/certpool.go
```diff
@@ -0,0 +1,197 @@
...
+	certPool       *x509.CertPool
+	certs          []*x509.Certificate
+	certsByName    map[string][]int
+	lock           sync.RWMutex
+	dirty          int32
+	systemCertPool bool
+}
+
+// NewCertPool new CertPool implementation
+func NewCertPool(useSystemCertPool bool) (CertPool, error) {
```
llorllale: @Baha-sk why not return a pointer?

pkg/didcomm/transport/http/tls/certpool.go
```diff
@@ -0,0 +1,197 @@
...
+// certPool is a thread safe wrapper around the x509 standard library
+// cert pool implementation.
+// It optionally allows loading the system trust store.
+type certPool struct {
+	certPool       *x509.CertPool
+	certs          []*x509.Certificate
+	certsByName    map[string][]int
+	lock           sync.RWMutex
+	dirty          int32
+	systemCertPool bool
```
llorllale: @Baha-sk this attribute does not seem to be required at all

pkg/didcomm/transport/http/tls/certpool.go
llorllale: @Baha-sk any reason why the user cannot simply provide their own certpool? Seems pretty easy to do, and I'd argue that if a user wants a custom cert pool then they must know what they're doing.

Having the user pass in a cert pool to the constructor function in the transport layer would void all of this code to avoid memory inconsistencies (swapping cert pools, mutexes, filtering duplicates, etc. etc)

pkg/didcomm/transport/http/tls/certpool.go
Baha-sk: this attribute designates what its name implies, it's to use the system cert pool as opposed to using new and empty cert pool. It's only used if a user does not pass in a cert pool path during client agent creation.

pkg/didcomm/transport/http/tls/certpool.go
Baha-sk: this can be done in a separate PR, we can add another options to the new outbound client that accepts a cert pool instance, path or no options if they wish to use the system cert pool.. 


this initial HTTP client Agent PR assumes the user of the library will pass in a path to the certpool, enhancements can be added on top of it.

pkg/didcomm/transport/http/tls/certpool.go
llorllale: @Baha-sk I had missed that it's used inside that swapCertPool function

scripts/generate_test_keys.sh
troyronda: yes, pki seems okay (or possibly crypto).

troyronda: Aslo, if these are fixtures for the http transport testing then they should be underneath a testate folder (under transport/http)

pkg/didcomm/transport/http/tls/certpool.go
llorllale: @Baha-sk we incur additional maintenance by adding all these unnecessary features. For now, let's just have the user provide their own cert pool and ff they don't, we can reasonably use the system cert pool by default. Right now it's not clear if we need dynamic cert pools.

pkg/didcomm/transport/http/outbound.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		respData = buf.String()
+	}
+	return respData, nil
+}
+
+// creates a new instance of HTTP transport as a client
+func newHTTPClient(cfg *OutboundCommConfig) (*http.Client, error) {
+	var err error
+	var caCertPool tlsCertPool.CertPool
```
troyronda: TLS Cert Pool builder logic does not belong in this package.

The client should accept an interface for a cert pool, and builder utilities should all be outside.

Baha-sk: will refactor to add options support and allow for a certpool option .. should this dynamic cert pool creation code move to a separate "utility" package? or removed from this PR then we can add it later, if needed?

troyronda: A dynamic cert pool package seems like it could be its own PR.

scripts/generate_test_keys.sh
Baha-sk: sure will rename to `crypto` and move folder under transport/http/.. as opposed to test/fixtures/...

troyronda: (note: under transport/http/testdata)

Baha-sk: sure.. this will be done in another pr though

pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"net/http/httptest"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+)
+
+const certPrefix = "../../../../test/fixtures/keys/"
```
troyronda: this will need to be updated in the next PR.

pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+func TestWithOutboundOpts(t *testing.T) {
+	opt := WithOutboundCommHTTP(nil)
+	require.NotNil(t, opt)
+	clOpts := &outboundCommHTTPOpts{}
+	opt(clOpts)
+
+	opt = WithOutboundTimeout(10 * time.Second)
```
troyronda: Prefer time-based arguments to be pulled into more visible constants.

pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -0,0 +1,183 @@
...
+	for i := range certs {
+		pool.AddCert(certs[i])
+	}
+	return nil
+}
+
+func startMockServer() *http.Server {
+	testHandler := mockHttpHandler{}
+	httpServer := &http.Server{
+		Addr:    ":8090",
```
troyronda: This can fail if the port is already used. You should have the http.Server supply the port back and then use that returned port in the tests.

pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -0,0 +1,183 @@
...
+	require.Error(t, err)
+	require.EqualError(t, err, "Can't create an outbound transport without an HTTP client or tls.Config option")
+
+	// now create a new valid Outbound transport instance and test its Send() call
+	ot, err = NewOutbound(WithOutboundTLSConfig(tlsConfig), WithOutboundTimeout(1*time.Second))
+	require.NoError(t, err)
+	require.NotNil(t, ot)
+
+	// test mock HTTPHandler
+	req, err := http.NewRequest("POST", "https://localhost:8090", nil)
```
troyronda: All "valid" instance of port 8090 should be replaced with the port returned by spinning up http server.

Baha-sk: done

pkg/didcomm/transport/http/outbound_test.go
Baha-sk: updated to get dynamic port with net.Listener ..

pkg/didcomm/transport/http/outbound_test.go
Baha-sk: this is for test... should we add a default timeout?

pkg/didcomm/transport/http/outbound_test.go
Baha-sk: yes, next PR has the go generate command with updated path..

pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	opt = WithOutboundTLSConfig(nil)
+	require.NotNil(t, opt)
+	clOpts = &outboundCommHTTPOpts{}
+	opt(clOpts)
+}
+
+func TestOutboundHTTPTransport(t *testing.T) {
+	// prepare http l
+	l := startMockServer()
```
llorllale: @Baha-sk let's rename this variable to `server` for readability

Baha-sk: done

pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -0,0 +1,179 @@
...
+	r, e = ot.Send("Hello World", serverUrl)
+	require.NoError(t, e)
+	require.NotEmpty(t, r)
+
+}
+
+func addCertsToCertPool(pool *x509.CertPool) error {
+	var rawCerts []string
+
+	for i := 1; i <= 3; i++ {
```
troyronda: this loop should be commented. i.e., why 3 iterations?

Baha-sk: 3 iterations for three certs.. I will add comment in a future pr

troyronda:
Part of Epic #43

PR #11, opened 2019-07-16
feat: Add DID Document data model

### LINKED ISSUE #10: Add DID Document data model
did document spec: https://w3c-ccg.github.io/did-spec/#did-documents

fqutishat:
closes #10

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -1,0 +1,208 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package did
+
+import (
+	"encoding/json"
+	"encoding/pem"
+	"time"
+
+	"encoding/hex"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/pkg/errors"
+)
+
+const (
+	jsonldType         = "type"
+	jsonldID           = "id"
+	jsonldServicePoint = "serviceEndpoint"
+	jsonldController   = "controller"
+
+	// various public key encodings
+	jsonldPublicKeyBase58 = "publicKeyBase58"
+	jsonldPublicKeyHex    = "publicKeyHex"
+	jsonldPublicKeyPem    = "publicKeyPem"
+)
+
+// Doc DID Document definition
+type Doc struct {
+	Context        []string
+	ID             string
+	PublicKey      []PublicKey
+	Service        []Service
+	Authentication []VerificationMethod
+	Created        time.Time
+	Updated        time.Time
+	Proof          Proof
+}
+
+// PublicKey DID doc public key
+type PublicKey struct {
+	ID         string
+	Type       string
+	Controller string
+	Value      []byte
+}
+
+// Service DID doc service
+type Service struct {
+	ID              string
+	Type            string
+	ServiceEndpoint string
+	Properties      map[string]interface{}
+}
+
+// VerificationMethod authentication verification method
+type VerificationMethod struct {
+	PublicKey PublicKey
+}
+
+type rawDoc struct {
+	Context        []string                 `json:"@context,omitempty"`
+	ID             string                   `json:"id,omitempty"`
+	PublicKey      []map[string]interface{} `json:"publicKey,omitempty"`
+	Service        []map[string]interface{} `json:"service,omitempty"`
+	Authentication []interface{}            `json:"authentication,omitempty"`
+	Created        time.Time                `json:"created,omitempty"`
+	Updated        time.Time                `json:"updated,omitempty"`
+	Proof          Proof                    `json:"proof,omitempty"`
+}
+
+// Proof is cryptographic proof of the integrity of the DID Document
+type Proof struct {
+	Type           string    `json:"type,omitempty"`
+	Created        time.Time `json:"created,omitempty"`
+	Creator        string    `json:"creator,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/doc/did/doc.go
```diff
@@ -0,0 +1,81 @@
...
+	// validate did document
+	if err := doc.validate(); err != nil {
+		return nil, err
+	}
+	return doc, nil
+}
+
+// validate did doc
+func (doc *Doc) validate() error {
+	// TODO Validate that the output DID Document conforms to the serialization of the DID Document data model (https://w3c-ccg.github.io/did-spec/#did-documents)
```
troyronda: Please create issue.

pkg/doc/did/doc.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+// stringEntry
+func stringEntry(entry interface{}) string {
+	if entry == nil {
+		return ""
+	}
+	id, ok := entry.(string)
+	if !ok {
```
troyronda: when does this happen?

pkg/doc/did/publickey.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package did
+
+// PublicKey must include id and type properties, and exactly one value property
+type PublicKey map[string]interface{}
```
troyronda: Is there a case for a non-string storage in PublicKey?

Are there mandatary fields that would be best represented as a struct?

pkg/doc/did/service.go
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package did
+
+// Service represents any type of service the entity wishes to advertise
+type Service map[string]interface{}
```
troyronda: Is there a case for a non-string storage in Service?

Are there mandatary fields that would be best represented as a struct?

pkg/doc/did/doc.go
llorllale: This function is reused in several places to extract values for attributes of type string. I'm concerned about how this function assumes it's ok to return an empty string if the input is not a string. This function should just not be used if the input were not a string (alternatively, it could provide a boolean return parameter that would indicate as much)

pkg/doc/did/doc.go
issue: #15

pkg/doc/did/publickey.go
fqutishat: done

pkg/doc/did/service.go
fqutishat: done

pkg/doc/did/doc.go
```diff
@@ -0,0 +1,202 @@
+/*
```
troyronda: @llorllale @fqutishat @rolsonquadras Any thoughts on the directory structure?

this is now pkg/doc/did.Doc

pkg/doc/did/doc.go
```diff
@@ -0,0 +1,202 @@
...
+	jsonldPublicKeyPem    = "publicKeyPem"
+)
+
+// Doc DID Document definition
+type Doc struct {
+	Context        []string
+	ID             string
+	PublicKey      []PublicKey
+	Service        []Service
+	Authentication []PublicKey
```
troyronda: I thought this was technically a Verification Method (and a Verification Method can be a PublicKey)?

pkg/doc/did/doc.go
llorllale: @troyronda I'm not sure of the need for that top-level `doc`, but I care far more about referencing `did.Doc` as the type all throughout client code, so I'm generally OK with this

pkg/doc/did/doc.go
```diff
@@ -0,0 +1,207 @@
...
+type Service struct {
+	ID              string
+	Type            string
+	ServiceEndpoint string
+	Properties      map[string]interface{}
+}
+
+// VerificationMethod authentication verification method
+type VerificationMethod struct {
+	PublicKey PublicKey
```
troyronda: We should revisit later - should think about the appropriate pattern to access this field.

pkg/doc/did/doc.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+// Doc DID Document definition
+type Doc struct {
+	Context        []string
+	ID             string
+	PublicKey      []PublicKey
+	Service        []Service
+	Authentication []VerificationMethod
+	Created        string
```
llorllale: @fqutishat `Created` and `Updated` should be `time.Time`

pkg/doc/did/doc.go
```diff
@@ -0,0 +1,207 @@
...
+	Updated        string
+	Proof          Proof
+}
+
+// PublicKey DID doc public key
+type PublicKey struct {
+	ID         string
+	Type       string
+	Controller string
+	Value      []byte
```
troyronda: I wonder if Value or Bytes makes more sense as the field name?

pkg/doc/did/doc.go
rolsonquadras: @troyronda I agree with @llorllale's comments

pkg/doc/did/doc.go

pkg/doc/did/doc.go
fqutishat: ok

Issue #13, opened 2019-07-16
Use go:generate instead of make generate-test-keys


Baha-sk:
users of the library should not be forced to call `make` to test the library, a simpler `go generate` mechanism for the library authors needs to be introduced instead

PR #14, opened 2019-07-16
chore: use go generate to generate test PKIs

Baha-sk:
Currently to test the project, users are forced to call `make generate-test-keys`
    
This target is not intended to be used by users of the library.
    
Instead a more general `go generate` command can be called to create test keys

closes issue #13


genTestPKI.go
```diff
@@ -0,0 +1,11 @@
+/*
```
troyronda: Also, if these are fixtures for the http transport testing then this go:generate should be in the transport/http package.

scripts/openssl_env.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+#
+# Copyright SecureKey Technologies Inc. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+OPENSSL_CMD=${OPENSSL_CMD:-openssl}
```
llorllale: @Baha-sk does this command need to be configurable?

llorllale:
@Baha-sk @troyronda if this PR is merged as-is, and #9 is also merged, it will leave the `make generate-test-keys` target

scripts/openssl_env.sh
troyronda: I generally think it’s a good pattern to allow overriding of commands in these scripts.

genTestPKI.go
Baha-sk: this means one will have to run `go generate` in transport/http package.

with the current change, we can run `go generate` from the root of the project and the artifacts will be generated in /test folder

Baha-sk:
> @Baha-sk @troyronda if this PR is merged as-is, and #9 is also merged, it will leave the `make generate-test-keys` target

yes this is why this PR should be merged after #9 is merged, I created branch off of #9, I should have created it off of master to remove the dependency

genTestPKI.go
troyronda: The generate goes with the code that needs it, not at the base of the project.

scripts/openssl_env.sh
llorllale: I think it's a preemptive solution to a rare, user-specific problem, with the side effect of adding clutter to our scripts. Our build certainly doesn't require it - we have more control over this by using `frapsoft/openssl`.

That said, it's just a one-liner, although it leaves one (uselessly) wondering why it's configurable.

genTestPKI.go
Baha-sk: file removed, generate is now under the http transport package

README.md
```diff
@@ -10,3 +10,16 @@
 ## License
 
 Hyperledger Aries Framework Go is licensed under the [Apache License Version 2.0](LICENSE).
+
```
rolsonquadras: This should go under proper sub-heading. Currently, it's inside Licence.

scripts/check_unit.sh
```diff
@@ -12,4 +12,11 @@ echo "Running $0"
 PKGS=`go list github.com/hyperledger/aries-framework-go/... 2> /dev/null | \
                                                  grep -v /mocks`
 
+CWD=$PWD
+if [[ ! -d "pkg/didcomm/transport/http/testdata" ]]; then
```
fqutishat: do we need to add every package that use go generate here ?

fqutishat: use go generate ./...

README.md
Baha-sk: added sub-heading

scripts/check_unit.sh
Baha-sk: done

scripts/generate_test_keys.sh
```diff
@@ -8,43 +8,43 @@
...
 #extract pubkey 2
-openssl x509 -inform pem -in test/fixtures/keys/ec-pubCert2.pem -pubkey -noout > test/fixtures/keys/ec-pubKey2.pem
+openssl x509 -inform pem -in testdata/ec-pubCert2.pem -pubkey -noout > testdata/ec-pubKey2.pem
 
 #create creds 3
-openssl ecparam -name prime256v1 -genkey -noout -out test/fixtures/keys/ec-key3.pem
-openssl req -new -key test/fixtures/keys/ec-key3.pem -subj "/C=CA/ST=ON/O=Example Inc.:aries-framework-go/OU=aries-framework-go/CN=*.example.com" -reqexts SAN -config ./test/fixtures/openssl/openssl.cnf -out test/fixtures/keys/ec-key3.csr
-openssl x509 -req -in test/fixtures/keys/ec-key3.csr -extensions SAN -extfile ./test/fixtures/openssl/openssl.cnf -CA test/fixtures/keys/ec-cacert.pem -CAkey test/fixtures/keys/ec-cakey.pem -CAcreateserial -out test/fixtures/keys/ec-pubCert3.pem -days 365
+openssl ecparam -name prime256v1 -genkey -noout -out testdata/ec-key3.pem
+openssl req -new -key testdata/ec-key3.pem -subj "/C=CA/ST=ON/O=Example Inc.:aries-framework-go/OU=aries-framework-go/CN=*.example.com" -reqexts SAN -config ../../../../test/fixtures/openssl/openssl.cnf -out testdata/ec-key3.csr
```
fqutishat: why openssl.cnf in test/fixtures/openssl can we create dynamically

troyronda: (Or put it in testdata folder)

.gitignore
```diff
@@ -17,7 +17,8 @@ vendor
 
 # Test binary, build with `go test -c`
 *.test
-test/fixtures/keys
+# Generate test PKIs with `go generate`
+pkg/didcomm/transport/http/testdata
```
troyronda: I would put the generated certs in a subdirectory of testdata and only ignore that directory.

Baha-sk: this means testdata will be committed to git...

Baha-sk: done, new subdirectory called crypto/ is added

scripts/generate_test_keys.sh
Baha-sk: deleted openssl.cnf file .. it is now temporarily generated in the script (and deleted at the end)

pkg/didcomm/transport/http/outbound.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+//go:generate ../../../../scripts/openssl_env.sh ../../../../scripts/generate_test_keys.sh
```
troyronda: Why not put this under the testdata directory too?

Baha-sk: sure, might as well since they're used for http transport..

PR #18, opened 2019-07-18
feat: Add Generic DID Resolution

### LINKED ISSUE #17: Add Generic DID Resolution
Add Generic did resolver framework to handle multiple DID methods

spec: https://w3c-ccg.github.io/did-resolution/

fqutishat:
closes #17

### DIFF
--- a/pkg/framework/didresolver/api.go
+++ b/pkg/framework/didresolver/api.go
@@ -1,0 +1,77 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didresolver
+
+import "time"
+
+// ResultType input option can be used to request a certain type of result.
+type ResultType int
+
+const (
+	// DidDocumentResult Request a DID Document as output
+	DidDocumentResult ResultType = iota
+	// ResolutionResult Request a DID Resolution Result
+	ResolutionResult
+)
+
+// DidMethod operations
+type DidMethod interface {
+	Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error)
+}
+
+// resolveOpts holds the options for did resolve
+type resolveOpts struct {
+	resultType  ResultType
+	versionID   interface{}
+	versionTime string
+	noCache     bool
+}
+
+// ResolveOpt is a did resolve option
+type ResolveOpt func(opts *resolveOpts)
+
+// WithResultType the result type input option can be used to request a certain type of result
+func WithResultType(resultType ResultType) ResolveOpt {
+	return func(opts *resolveOpts) {
+		opts.resultType = resultType
+	}
+}
+
+// WithVersionID the version id input option can be used to request a specific version of a DID Document
+func WithVersionID(versionID interface{}) ResolveOpt {
+	return func(opts *resolveOpts) {
+		opts.versionID = versionID
+	}
+}
+
+// WithVersionTime the version time input option can used to request a specific version of a DID Document
+func WithVersionTime(versionTime time.Time) ResolveOpt {
+	return func(opts *resolveOpts) {
+		opts.versionTime = versionTime.Format(time.RFC3339)
+	}
+}
+
+// WithNoCache the no-cache input option can be used to turn cache on or off
+func WithNoCache(noCache bool) ResolveOpt {
+	return func(opts *resolveOpts) {
+		opts.noCache = noCache
+	}
+}
+
+// didResolverOpts holds the options for resolver instance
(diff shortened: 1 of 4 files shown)


pkg/framework/didresolver/api.go
```diff
@@ -0,0 +1,77 @@
...
+
+const (
+	// DidDocumentResult Request a DID Document as output
+	DidDocumentResult ResultType = iota
+	// ResolutionResult Request a DID Resolution Result
+	ResolutionResult
+)
+
+// didMethod operations
+type didMethod interface {
```
llorllale: @fqutishat how will users inject their own did methods if this interface is not exported?

fqutishat: @llorllale func WithDidMethod(id string, method didMethod) Opt {

pkg/framework/didresolver/resolver.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"strings"
+
+	diddoc "github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/pkg/errors"
+)
+
+// Resolver did resolver
+type Resolver struct {
```
llorllale: @troyronda @fqutishat this stutters: `didresolver.Resolver`

pkg/framework/didresolver/api.go
troyronda: the interface doesn't need to be exported for the exported function to accept it.

llorllale: @troyronda correct, but the user needs to know the shape of the interface in order to implement it

pkg/framework/didresolver/resolver.go
llorllale: @troyronda @fqutishat this should ideally be typed as `did.Resolver`

fqutishat: should I change didresolver to did

pkg/framework/didresolver/api.go
```diff
@@ -0,0 +1,77 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package resolver
```
troyronda: isn't the folder name "didresolver"?

pkg/framework/didresolver/api.go
fqutishat: I make it exported

pkg/framework/didresolver/api.go
fqutishat: yes

pkg/framework/didresolver/resolver.go
troyronda: It can be okay for the main struct of a dedicated package to stutter with the package itself.
(But in that case, I would have expected an exact match like didresolver.DIDResolver).

pkg/framework/didresolver/api.go
fqutishat: it's bug

fqutishat: done

pkg/framework/didresolver/resolver.go
troyronda: (Note that the usage is actually didresolver.New in this case).

fqutishat: change to DIDResolver

pkg/framework/didresolver/didresolver.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Resolve did document
+func (r *DIDResolver) Resolve(did string, opts ...ResolveOpt) (*diddoc.Doc, error) {
+	resolveOpts := &resolveOpts{}
+	// Apply options
+	for _, opt := range opts {
+		opt(resolveOpts)
+	}
+
+	// TODO Validate that the input DID conforms to the did rule of the Generic DID Syntax (https://w3c-ccg.github.io/did-spec/#generic-did-syntax)
```
troyronda: Issue created?

pkg/framework/didresolver/api.go
```diff
@@ -0,0 +1,77 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didresolver
```
troyronda: perhaps we should revisit (in the future), a shorter name.

pkg/framework/didresolver/didresolver.go
fqutishat: #20

PR #27, opened 2019-07-22
feat: add HTTP Inbound transport capability

Baha-sk:
Similar to the Outbound transport capability
	This change adds support for the Inbound transport cabaility
	as per the Aries Agent exchange communcation RFC-23 and 25

Closes: #8

### DIFF
--- a/pkg/didcomm/transport/http/inbound.go
+++ b/pkg/didcomm/transport/http/inbound.go
@@ -1,0 +1,85 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package http
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+// MessageHandler is a function that handles the inbound request payload
+// the payload will be unpacked prior to calling this function.
+type MessageHandler func(payload []byte)
+
+// NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
+// then routes processing to the mandatory 'msgHandler' argument.
+//
+// Arguments:
+// * 'msgHandler' is the handler function that will be executed with the inbound request payload.
+//    Users of this library must manage the handling of all inbound payloads in this function.
+func NewInboundHandler(msgHandler MessageHandler) (http.Handler, error) {
+	if msgHandler == nil {
+		log.Println("Error creating a new inbound handler: message handler function is nil")
+		return nil, errors.New("Failed to create NewInboundHandler")
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		processPOSTRequest(w, r, msgHandler)
+	}), nil
+}
+
+// TODO Log error message with a common logger library for aries-framework-go
+func processPOSTRequest(w http.ResponseWriter, r *http.Request, messageHandler MessageHandler) {
+	if valid := validateHTTPMethod(w, r); !valid {
+		return
+	}
+	if valid := validatePayload(r, w); !valid {
+		return
+	}
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("Error reading request body: %s - returning Code: %d", err, http.StatusInternalServerError)
+		http.Error(w, "Failed to read payload", http.StatusInternalServerError)
+	}
+
+	// TODO add Unpack(body) call here
+	//...
+
+	w.WriteHeader(http.StatusAccepted)
+
+	go messageHandler(body)
+}
+
+// validatePayload validate and get the payload from the request
+func validatePayload(r *http.Request, w http.ResponseWriter) bool {
+	if r.ContentLength == 0 { // empty payload should not be accepted
+		http.Error(w, "Empty payload", http.StatusBadRequest)
+		return false
+	}
+	return true
+}
+
+// validateHTTPMethod validate HTTP method and content-type
+func validateHTTPMethod(w http.ResponseWriter, r *http.Request) bool {
+	if r.Method != "POST" {
+		http.Error(w, "HTTP Method not allowed", http.StatusMethodNotAllowed)
+		return false
+	}
+
+	ct := r.Header.Get("Content-type")
+	if ct != commContentType {
+		http.Error(w, fmt.Sprintf("Unsupported Content-type \"%s\"", ct), http.StatusUnsupportedMediaType)
+		return false
+	}
+
+	return true
+}
--- a/pkg/didcomm/transport/http/inbound_test.go
+++ b/pkg/didcomm/transport/http/inbound_test.go
@@ -1,0 +1,89 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package http
+
+import (
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"log"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// mockMsgHandler is an http.msgHandler type, it is similar to MockHandler struct
+// but will be injected in WithInboundSetting() directly, will be used by each transport comm handle function
+func mockMsgHandler(payload []byte) {
+	log.Printf("Payload received is %s", payload)
+}
+
+func TestInboundHandler(t *testing.T) {
+	// test inboundHandler with empty args should fail
+	inHandler, err := NewInboundHandler(nil)
+	require.Error(t, err)
+	require.Nil(t, inHandler)
+
+	// now create a valid inboundHandler to continue testing..
+	inHandler, err = NewInboundHandler(mockMsgHandler)
+
+	require.NoError(t, err)
+	require.NotNil(t, inHandler)
+	server := startMockServer(inHandler)
+	port := getServerPort(server)
+	serverUrl := fmt.Sprintf("https://localhost:%d", port)
+	defer func() {
+		e := server.Close()
+		if e != nil {
+			log.Fatalf("Failed to stop server: %s", e)
+		}
+	}()
+
+	//build a mock cert pool
+	cp := x509.NewCertPool()
+	err = addCertsToCertPool(cp)
+	require.NoError(t, err)
+
+	// build a tls.Config instance to be used by the outbound transport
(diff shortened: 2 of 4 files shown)


pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	commHandler := &DIDCommHandler{routersOpts: routersOpts}
+	err := validateHandler(commHandler)
+	if err != nil {
+		// TODO add log message here
+		return nil, errors.Wrapf(err, "Failed to create NewInboundHandler")
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
llorllale: @Baha-sk this design would force the user to bind this handler to the root path `/` or some other path that is parent of the specific handler paths. A much simpler design would be to simply provide handlers and let the user take care of binding them to the paths. It also avoids the potential DoS due to the `return` on line 65 - this inbound connection should not be dropped because a request was made to an unregistered path - that is outside the user's control.

Note: for the demo, we definitely need to serve non-DIDComm paths

Baha-sk: since DidComm should only have 1 URL path for inbound communication, the code was refactored to allow for only 1 path and its corresponding http handler.

as discussed offline @llorllale, did exchange routing should be done at a higher level where the payload envelope is unmarshalled as it has the routing info required by DidComm.

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,115 @@
...
+// InboundHTTPSetting is an inbound HTTP transport options setting function
+type InboundHTTPSetting func(opts *inboundCommHTTPOpts)
+
+// InboundDIDCommHandler struct to pass details for request handlers
+type InboundDIDCommHandler struct {
+	handlerOpts *inboundCommHTTPOpts
+}
+
+// WithInboundSetting option is for creating an Inbounnd HTTP transport using an http.Handler
+func WithInboundSetting(inboundPath string, handlerFunc http.HandlerFunc) InboundHTTPSetting {
```
troyronda: I don't think we typically use the term "setting"

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,115 @@
...
+	return func(opts *inboundCommHTTPOpts) {
+		opts.Path = inboundPath
+		opts.HandlerFunc = handlerFunc
+	}
+}
+
+// NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
+// then routes processing to the passed 'in' handler argument
+// 'in' is mandatory as it contains the url Path and the HTTP handler needed by inbound comm.
+func NewInboundHandler(in InboundHTTPSetting) (http.Handler, error) {
```
troyronda: The purpose of an Options struct/method pattern is to allow 0 or more option methods. Why does the constructor only take one? Is this argument optional or required?

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+// InboundDIDCommHandler struct to pass details for request handlers
+type InboundDIDCommHandler struct {
```
troyronda: where is this struct type used externally? I see an internal usage within NewInboundHandler but this type is not returned.

pkg/didcomm/transport/http/inbound.go
Baha-sk: removed function

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,94 @@
...
+	Path        string
+	HandlerFunc http.HandlerFunc
+}
+
+// NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
+// then routes processing for the passed 'inURL' and 'handlerFunc' arguments.
+// Both 'inURL' and 'handlerFunc' are mandatory for inbound communications.
+func NewInboundHandler(inURL string, handlerFunc http.HandlerFunc) (http.Handler, error) {
+	commHandler := &InboundDIDCommHandler{Path: inURL, HandlerFunc: handlerFunc}
+	err := validateHandler(commHandler)
```
troyronda: it seems we are creating a struct just so it can be passed here? are you planning more extensive usage of this type?

pkg/didcomm/transport/http/inbound.go
Baha-sk: removed the struct altogether, now the function accepts only a url and an http handlerFunc

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,94 @@
...
+// InboundDIDCommHandler struct to pass details for request handlers
+type InboundDIDCommHandler struct {
+	Path        string
+	HandlerFunc http.HandlerFunc
+}
+
+// NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
+// then routes processing for the passed 'inURL' and 'handlerFunc' arguments.
+// Both 'inURL' and 'handlerFunc' are mandatory for inbound communications.
+func NewInboundHandler(inURL string, handlerFunc http.HandlerFunc) (http.Handler, error) {
```
troyronda: Your previous version of this PR had an option style argument. Are you saying that these two arguments are now always needed (and as this type)?

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,94 @@
...
+
+// InboundDIDCommHandler struct to pass details for request handlers
+type InboundDIDCommHandler struct {
+	Path        string
+	HandlerFunc http.HandlerFunc
+}
+
+// NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
+// then routes processing for the passed 'inURL' and 'handlerFunc' arguments.
+// Both 'inURL' and 'handlerFunc' are mandatory for inbound communications.
```
troyronda: It would help to provide a description of these arguments - esp in the constructor.

pkg/didcomm/transport/transport_interface.go
```diff
@@ -12,3 +12,6 @@ type OutboundTransport interface {
 	// Send send a2a exchange data
 	Send(data string, destination string) (string, error)
 }
+
+// for Inbound Transport, use of dynamic handler is sufficient, no interface is needed
```
llorllale: @Baha-sk this comment is out of place. godocs should be attached to a package or a type

pkg/didcomm/transport/http/inbound.go
Baha-sk: struct is now not exported

pkg/didcomm/transport/http/inbound.go
Baha-sk: for now there's no plan to extend this type, it's been unexported

pkg/didcomm/transport/http/inbound.go
Baha-sk: yes, there needs to be 1 service endpoint and 1 http handler for inbound communication to happen as discussed with @llorllale

llorllale: @Baha-sk `inURL` is not required. Let the user worry about that in their layer.

`handlerFunc` should not be an `http.HandlerFunc` - it should be a custom interface that accepts a decrypted blob (ie. `[]byte`). This custom interface will be aware of the different DIDComm messages and know how to dispatch them

pkg/didcomm/transport/http/inbound.go
Baha-sk: 'inURL' depicts the inbound http path in the POST request
'handlerFunc' is the HTTP handlerFunc that handles the inbound request

I can add these in the comments

pkg/didcomm/transport/transport_interface.go
Baha-sk: removed comment

pkg/didcomm/transport/http/inbound.go
Baha-sk: update, it seems no URL path is required as agents will communicate via 1 endpoint (regardless of the path). As mentioned above, DID Exchange flow routing will be done at a higher level. 

So all that's required here is a simple msgHandler function that will manage the payload (envelope) handling.

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,87 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+// MessageHandler is a function that handles the inbound request payload
+// the payload will be unpacked prior to calling this function.
+// this function will be called for all incoming requests
```
llorllale: @Baha-sk I think comments on lines 20 and 21 can be removed

pkg/didcomm/transport/http/inbound.go
```diff
@@ -0,0 +1,87 @@
...
+		return
+	}
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("Error reading request body: %s - returning Code: %d", err, http.StatusInternalServerError)
+		http.Error(w, "Failed to read payload", http.StatusInternalServerError)
+	}
+
+	// TODO add Unpack(body) call here
```
troyronda: Does Unpack happen within the transport code?

llorllale: @troyronda based on our discussion, I assume the answer is "yes". My thoughts are that unpacking has to be done across all DIDComm messages, so the transport layer seems like a natural choice. This decision can be revisited later on.

pkg/didcomm/transport/http/inbound.go
Baha-sk: ok..

pkg/didcomm/transport/http/inbound.go
Baha-sk: now the function accepts a msgHandler function

pkg/didcomm/transport/http/inbound.go
Baha-sk: removed

pkg/didcomm/transport/http/inbound.go
Baha-sk: technically we can call Pack/Unpack at DidComm (upper level),  but at the transport level, it will be handled in one place for all calls.

troyronda:
Part of Epic #43

Issue #26, opened 2019-07-22
Contents should be opaque to the did store and connection store


llorllale:
I just realized after we closed #22: the store implementation will be able to read all the DIDs, including the pairwise mappings. This could result in a breach of privacy in the case where the storage is an external service.

Ideally, no stores should know what DIDs I am interacting with. This could be uniformly enforced across all implementations by a change in API. For example, a K-V store:

```go
// The layer that consumes this interface is responsible to 
// encrypting the data such that providers cannot interpret it.
// That layer remains secured in aries-framework-go.
type Store interface {
    Put(k string, v []byte)  error
    Get(k string) ([]byte, error)
}
```

See [RFC 0050 - Wallets](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0050-wallets/README.md#composition)

@rolsonquadras @troyronda thoughts?

llorllale:
The main idea is that the Store should not be able to read what the Agent is persisting in it. This means the DID documents need to be encrypted **and then** given to the Store for persistence.

**We search by DID**
* A DID document's key is simply its DID. To prevent the Store from even reading the DID, we should pass in a hash of the DID as key
* The `kid` in incoming [encryption envelopes](https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope) is a [DID fragment](https://w3c-ccg.github.io/did-spec/#fragment) - searching for the document with that key should be straightforward (following from previous point: simply extract the DID from the `kid`, hash it, and use it as search parameter).

llorllale:
The above is the bare minimum for the did-exchange demo. I suggest we start here and figure out details (like adding tags to the search capabilities) as we move forward.

rolsonquadras:
@llorllale The above interface satisfies the document store defined in https://github.com/hyperledger/aries-framework-go/tree/master/pkg/storage/docstore. 

But we need to come up generic search function, particularly for https://github.com/hyperledger/aries-framework-go/blob/master/pkg/storage/connection/connstore.go#L41. By encrypting the value, there is no way to search based on the content. May be as you mentioned, we can defer the implementation to later.

llorllale:
@rolsonquadras this method...

```go
LookupByTheirPublicKey(pubKey []byte) (*Connection, bool)
```

... can be replaced by using `Get(k string) ([]byte, error)` defined above in `Store`.

The way it would work is as follows:

[...]

Issue #22, opened 2019-07-19
Doc Storage - Interface design


rolsonquadras:
The Aries Framework should support pluggable storage(e.g file store, embedded db store, cloud store, etc.) for DID documents along with Connections(pairwise mapping).

llorllale:
@rolsonquadras agreed

These are the key things an Agent needs to do:

1. Store pairwise mappings of DIDs (ours and theirs)
2. Look up a pairwise mapping of DIDs from just the sender's public key (when receiving a message)
3. Look up pairwise mappings in order to then send a message to the other party.

Rough draft:
```go
package did

// Connection is a pairwise relation between two DIDs (ours and theirs)
type Connection struct {
   Mine *did.Doc
   Theirs *did.Doc
   Label string // human readable string set by the user, such as "Bob"
}

// Store is the storage interface for DIDs
type Store interface {
    Iter() *Iterator  // <--- iterator of Connections
    Put(c *Connection) error
    LookupByTheirPublicKey(k crypto.PublicKey) (c *Connection, found bool)
}
```

rolsonquadras:
@llorllale Based on our offline discussion, i have updated interface design. 

docstore to store only peer did docs
```
package docstore

type Store interface {
	// Put stores the DID Doc
	Put(*did.Doc) error

	// GetAll fetches the DID Doc based on DID
	Get(string) (*did.Doc, error)

	// GetAll fetches all the DID Docs
	Iter() *DocIterator
}
```

connstore to store the mapping between my DID and their DID along with their public key
```
package connstore

// Connection is a pairwise relation between two DIDs (ours and theirs)
type Connection struct {
	MineDID        string
	TheirDID       string
	TheirPublicKey []byte
	Label          string
}

type Store interface {
	Iter() *ConnIterator // <--- iterator of Connections
	Put(c *Connection) error
	LookupByTheirPublicKey(pubKey *[]byte) (*Connection, bool)
}
```

Primary consideration to have separate the doc store from connection store is to have a clear scope for each items rather than mixing both, which might result in maintainability issues later on. Also, peer did resolution will be simpler with a separate DID Doc store as we can just pass the DID to the method.

Please Let me know your thoughts. @troyronda FYI.

llorllale:
@rolsonquadras sounds good

PR #30, opened 2019-07-24
refactor: Storage interface update to support encrypted data

rolsonquadras:
Refactor storage interface based on conversion in #26 


llorllale:
@rolsonquadras 

> based on conversion in #22

You mean #26

rolsonquadras:
> @rolsonquadras
> 
> > based on conversion in #22
> 
> You mean #26

Yes @llorllale. Updated the PR comment.

troyronda:
I'm wondering about the end-state of the deleted interfaces. Are they going to be used later as a middle layer?

rolsonquadras:
> I'm wondering about the end-state of the deleted interfaces. Are they going to be used later as a middle layer?

@troyronda I feel we can delete now and restore later based on use case. (those were just interfaces, also later we might have more clarity on the use cases when we explore each feature). For now, we can focus just on Storage.

PR #31, opened 2019-07-24
feat: Local storage implementation (leveldb)

### LINKED ISSUE #25: Doc/Connection Storage - Default Implementation
The Agent Framework should provide a default storage for Documents and Connections. The interfaces for these storages were defined in #22 and merged as part of #23.

The following requirement are considered for the default implementation.
- No network call
- Pure golang implementation
- Backed by file system

The Golang LevelDB provider(based on lib https://github.com/syndtr/goleveldb) looks to be the best fit considering the above requirements.

rolsonquadras:
leveldb implementation for storage interfaces defined in #26

Closes #25


troyronda:
I suggest that we remove the term default from this PR / Commit description.

rolsonquadras:
> I suggest that we remove the term default from this PR / Commit description. It is more the case that we are using local storage to facilitate simpler demos / tests in this repository.

@troyronda updated the commit and PR message (default -> local).

Issue #21, opened 2019-07-19
feat: logging interface and default implementation


sudeshrshetty:
Define generic logging interface for aries framework and provide default implementation.

llorllale:
@sudeshrshetty what do you have in mind?

sudeshrshetty:
//Logger - Standard logger interface
type Logger interface {

	//Fatalf is critical fatal logging, should possibly followed by system shutdown
	Fatalf(msg string, args ...interface{})

	//Panicf is critical logging, should possibly followed by panic
	Panicf(msg string, args ...interface{})

	//Debugf is for logging verbose messages
	Debugf(msg string, args ...interface{})

	//Infof for logging general logging messages
	Infof(msg string, args ...interface{})

	//Warnf is for logging messages about possible issues
	Warnf(msg string, args ...interface{})

	//Errorf is for logging errors
	Errorf(msg string, args ...interface{})
}

Custom implementation can be provided by calling `_logging.Initialize(customProvider)_` before using (before logging any lines).

Logging levels will be maintained by aries framework and it will be module based.
Like, each modules can have their own logging levels enabled.
Default logging level is _INFO_

troyronda:
the discussion here looks useful:

- https://github.com/golang/go/issues/28412
-  https://github.com/golang/go/issues/13182

PR #32, opened 2019-07-24
feat: logging interface and default implementation

sudeshrshetty:
- provided logging interface
- log levels supported - CRITICAL, ERROR, WARNING, INFO, DEBUG, where
INFO is default logging level
-logging levels will be managed by framework which is module based,
where each modules can maintain their own logging levels
- refers issue #21 where more details can be found

### DIFF
--- a/pkg/internal/common/logging/metadata/callerInfo.go
+++ b/pkg/internal/common/logging/metadata/callerInfo.go
@@ -1,0 +1,65 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package metadata
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/common/logging/api"
+)
+
+type callerInfoKey struct {
+	module string
+	level  api.Level
+}
+
+//CallerInfo maintains module-level based information to show or hide caller info
+type CallerInfo struct {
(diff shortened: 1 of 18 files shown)


pkg/internal/common/logging/metadata/callerInfo.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+//CallerInfo maintains module-level based information to show or hide caller info
+type CallerInfo struct {
+	showcaller map[callerInfoKey]bool
+}
+
+//ShowCallerInfo enables caller info for given module and level
+func (l *CallerInfo) ShowCallerInfo(module string, level api.Level) {
+	if l.showcaller == nil {
```
bstasyszyn: This is done in three places - maybe create a func or initialize in a NewCallerInfo() func?

pkg/internal/common/logging/metadata/level.go
```diff
@@ -0,0 +1,45 @@
...
+			return defaultLogLevel
+		}
+	}
+	return level
+}
+
+// SetLevel sets the log level for given module and level.
+func (l *ModuleLevels) SetLevel(module string, level api.Level) {
+	if l.levels == nil {
+		l.levels = make(map[string]api.Level)
```
bstasyszyn: Is this thread-safe?

sudeshrshetty:
Due to huge patch set, will be pushed as multiple small PRs.

PR #33, opened 2019-07-24
feat: logging interface

sudeshrshetty:
-provided logging interface
-log levels supported - CRITICAL, ERROR, WARNING, INFO, DEBUG, where
INFO is default logging level
-logging provider implementation can be used to provide custom logging
provider
-Story: #21

### DIFF
--- a/pkg/common/log/api.go
+++ b/pkg/common/log/api.go
@@ -1,0 +1,46 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package log
+
+// Level defines all available log levels for logging messages.
+type Level int
+
+// Log levels.
+const (
+	CRITICAL Level = iota
+	ERROR
+	WARNING
+	INFO //default logging level
+	DEBUG
+)
+
+//Logger - Standard logger interface
+type Logger interface {
+
+	//Fatalf is critical fatal logging, should possibly followed by a call to os.Exit(1)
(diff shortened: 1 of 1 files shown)


pkg/common/logging/api/loggerapi.go
```diff
@@ -0,0 +1,46 @@
+/*
```
troyronda: is the api subdirectory needed? is there a more descriptive name for the package?

pkg/common/logging/api/loggerapi.go
```diff
@@ -0,0 +1,46 @@
...
+	ERROR
+	WARNING
+	INFO //default logging level
+	DEBUG
+)
+
+//Logger - Standard logger interface
+type Logger interface {
+
+	//Fatalf is critical fatal logging, should possibly followed by system shutdown
```
rolsonquadras: "should possibly followed by system shutdown" is bit confusing. May be we can use standard Go library doc message (to be consistent with the standard lib).

"Fatalf is equivalent to Printf() followed by a call to os.Exit(1)" - https://godoc.org/log#Fatalf

pkg/common/logging/api/loggerapi.go
sudeshrshetty: renamed 'common/logging' to 'common/log', and moved logger interfaces right inside 'common/log'.

pkg/common/logging/api/loggerapi.go
sudeshrshetty: Fixed

PR #52, opened 2019-07-29
feat: moduled levels and caller infos for logging

sudeshrshetty:
- added module based log levels implementation where default log level is INFO
- added module and log level based caller info feature which shows caller function
names in log lines for given module and level configured.
- logger opts implementation is added which can be used by logger implementation
- Story: #21

### DIFF
--- a/pkg/internal/common/logging/metadata/level.go
+++ b/pkg/internal/common/logging/metadata/level.go
@@ -1,0 +1,48 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package metadata
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+const (
+	defaultLogLevel   = log.INFO
+	defaultModuleName = ""
+)
+
+func newModuledLevels() *moduleLevels {
+	return &moduleLevels{levels: make(map[string]log.Level)}
+}
+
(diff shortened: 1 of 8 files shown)


pkg/internal/common/logging/metadata/level.go
```diff
@@ -0,0 +1,54 @@
...
+	l.levels()[module] = level
+}
+
+// IsEnabledFor will return true if logging is enabled for given module and level.
+func (l *moduleLevels) IsEnabledFor(module string, level log.Level) bool {
+	return level <= l.GetLevel(module)
+}
+
+func (l *moduleLevels) levels() map[string]log.Level {
+	l.once.Do(func() {
```
bstasyszyn: Does the map have to be created lazily? You can create it when moduleLevels is created.

levels := &moduleLevels{mLevels: make(map[string]log.Level}

sudeshrshetty: It doesn't have to be lazy, will add NewXXX() functions and will remove sync.once.

pkg/internal/common/logging/metadata/opts.go
```diff
@@ -0,0 +1,75 @@
+/*
```
troyronda: How do you intend to expose the functions in a later PR?

sudeshrshetty: I just use metadata.GetLoggerOpts() 
I am going to add one more package under internal `internal/common/logging/modlog` which will use this package.

PR #53, opened 2019-07-29
feat: default logger implementation

sudeshrshetty:
- defLog : default implementation of moduled logger built on top
of go 'log' package.
- modLog: moduled level based logger implementation wrapping custom logger implementation.
- Story: #21

### DIFF
--- a/pkg/internal/common/logging/modlog/deflog.go
+++ b/pkg/internal/common/logging/modlog/deflog.go
@@ -1,0 +1,137 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package modlog
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/internal/common/logging/metadata"
+
+	logapi "github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+const (
+	logLevelFormatter   = "UTC %s-> %s "
+	logPrefixFormatter  = " [%s] "
+	callerInfoFormatter = "- %s "
+)
+
+// defLog is a logger implementation built on top of standard go log.
+// There is a  configurable caller info feature which displays caller function information name in logged lines.
+// caller info can be configured by log levels and modules. By default it is enabled.
+// Log Format : [<MODULE NAME>] <TIME IN UTC> - <CALLER INFO> -> <LOG LEVEL> <LOG TEXT>
+type defLog struct {
+	logger *log.Logger
+	module string
+}
+
+//Fatalf is CRITICAL log formatted followed by a call to os.Exit(1).
+func (l *defLog) Fatalf(format string, args ...interface{}) {
+	l.logf(logapi.CRITICAL, format, args...)
+	os.Exit(1)
+}
+
+//Panicf is CRITICAL log formatted followed by a call to panic()
+func (l *defLog) Panicf(format string, args ...interface{}) {
+	l.logf(logapi.CRITICAL, format, args...)
+	panic(fmt.Sprintf(format, args...))
+}
+
+//Debugf calls go 'log.Output' and can be used for logging verbose messages.
+// Arguments are handled in the manner of fmt.Printf.
+func (l *defLog) Debugf(format string, args ...interface{}) {
+	l.logf(logapi.DEBUG, format, args...)
+}
+
+//Infof calls go 'log.Output' and can be used for logging general information messages.
+//INFO is default logging level
+// Arguments are handled in the manner of fmt.Printf.
+func (l *defLog) Infof(format string, args ...interface{}) {
+	l.logf(logapi.INFO, format, args...)
+}
+
+// Warnf calls go log.Output and can be used for logging possible errors.
+// Arguments are handled in the manner of fmt.Printf.
+func (l *defLog) Warnf(format string, args ...interface{}) {
+	l.logf(logapi.WARNING, format, args...)
+}
+
+// Errorf calls go 'log.Output' and can be used for logging errors.
+// Arguments are handled in the manner of fmt.Printf.
+func (l *defLog) Errorf(format string, args ...interface{}) {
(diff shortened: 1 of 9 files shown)


pkg/internal/common/logging/modlog/modlog.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+//modLog is a moduled wrapper for api.Logger implementation
+//This adds functionality of module based levels on top of provider logger implementation
+type modLog struct {
+	logger log.Logger
+	module string
+}
+
+//Fatalf calls underlying logger.Fatal
```
troyronda: Why different method comments than defLog?

pkg/internal/common/logging/modlog/deflog.go
```diff
@@ -0,0 +1,145 @@
...
+	logapi "github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+const (
+	logLevelFormatter   = "UTC %s-> %s "
+	logPrefixFormatter  = " [%s] "
+	callerInfoFormatter = "- %s "
+)
+
+//defLog is a standard default logger implementation
```
troyronda: It would be helpful to comment further on:
- what this package is doing (It appears you are wrapping the std library logger to support ...).
- a "default logger implementation" means. (it's a bit of a funny name for a wrapper package -- in an isolated PR anyways).
- why modLog wasn't leveraged (as there appears to be overlap) - performance?

pkg/internal/common/logging/modlog/deflog.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+func (l *defLog) getCallerInfo(opts *metadata.LoggerOpts) string {
+
+	if !opts.CallerInfoEnabled {
+		return ""
+	}
+
+	const MAXCALLERS = 6  // search MAXCALLERS frames for the real caller
```
troyronda: It would be helpful to further comment on why these particular values for the contents (i.e., MAXCALLERS = 6 and SKIPCALLERS = 4).

pkg/internal/common/logging/modlog/modlog.go
sudeshrshetty: defLog is implementation built on top of go standard logging,
modLog is built on top of logger injected through provider, mod log is wrapper added to control logger levels of logging provider injected.

pkg/internal/common/logging/modlog/deflog.go
sudeshrshetty: Will fix comments.
Using different separate logger implementation for default and custom to avoid using locks and performance of course.

pkg/internal/common/logging/modlog/deflog.go
sudeshrshetty: Will fix it

pkg/internal/common/logging/modlog/modlog.go
troyronda: Ok. Hopefully "underlying" becomes more clear to the reader once the struct comment has more details.

pkg/internal/common/logging/modlog/deflog.go
```diff
@@ -0,0 +1,145 @@
...
+func (l *defLog) Errorf(format string, args ...interface{}) {
+	opts := metadata.GetLoggerOpts(l.module, logapi.ERROR)
+	if !opts.LevelEnabled {
+		return
+	}
+	l.logf(opts, logapi.ERROR, format, args...)
+}
+
+//ChangeOutput for changing output destination for the logger.
+func (l *defLog) ChangeOutput(output io.Writer) {
```
rolsonquadras: Any specific reason to not use the same function name as std library wrapped function? ie, ChangeOutput -> SetOutput. Looks like this function just wraps the stdlib logger function. Better to maintain consistent name.

pkg/internal/common/logging/modlog/deflog.go
sudeshrshetty: fixed it

pkg/internal/common/logging/modlog/deflog.go
sudeshrshetty: Changed it to SetOutput

pkg/internal/common/logging/modlog/provider.go
```diff
@@ -0,0 +1,55 @@
...
+type ModuledLoggerOpts func(opts *providerOpts)
+
+//WithCustomProvider can be used to provide custom logger provider for moduled logger
+func WithCustomProvider(customProvider logapi.LoggerProvider) ModuledLoggerOpts {
+	return func(opts *providerOpts) {
+		opts.customProvider = customProvider
+	}
+}
+
+//ModuledLoggerProvider returns logger provider for moduled level based logger
```
troyronda: Isn’t this NewProvider? Also why return interface instead of *provider?

pkg/internal/common/logging/modlog/provider.go
```diff
@@ -0,0 +1,55 @@
...
+// provider is the default logger implementation
+type provider struct {
+	customProvider logapi.LoggerProvider
+}
+
+//GetLogger returns moduled logger implementation.
+// default logging built on top go log will be used if custom logger provider is not provided.
+// if custom logging provider is provided then moduled logger wrapper of custom logger will be returned.
+func (p *provider) GetLogger(module string) logapi.Logger {
+	if p.customProvider != nil {
```
troyronda: instead of an if condition, i would always set p.provider. 

i.e., in the case of defLog, you should have a defProvider.

pkg/internal/common/logging/modlog/provider.go
sudeshrshetty: spilt it into --> NewModLogProvider & NewDefLogProvider

pkg/internal/common/logging/modlog/provider.go
sudeshrshetty: spilt it into --> NewModLogProvider & NewDefLogProvider

pkg/internal/common/logging/modlog/provider.go
troyronda: Why not return *provider?

troyronda:
We should revisit the defLog and modLog separation in a followup.

PR #55, opened 2019-07-31
feat: Peer DID Document store

rolsonquadras:
Implementation of Peer DID backing store spec (https://openssi.github.io/peer-did-method-spec/index.html#backing-storage) without conflict resolution

Implements #47

Signed-off-by: Rolson Quadras <rolson.quadras@securekey.com>

Issue #46, opened 2019-07-29
Peer DID: Generate Identifier


rolsonquadras:
Peer DID should have an identifier as defined in the [Spec](https://openssi.github.io/peer-did-method-spec/index.html#method-specific-identifier).

ABNF Syntax for Peer DID:
```
peer-did = "did:peer:" numalgo encalgo "-" numbasis 
numalgo = "1" 
encalgo = "1" 
numbasis = 64*HEXDIGCI 
HEXDIGCI = HEXDIG / "a" / "b" / "c" / "d" / "e" / "f"
```

The _numbasis_ is generated from [Genesis Version](https://openssi.github.io/peer-did-method-spec/index.html#dfn-genesis-version) of the Peer DID.

rolsonquadras:
Part of epic #35

PR #56, opened 2019-07-31
feat: Generate Peer Did Identifier

talwinder50:
As defined in the [Spec.](https://openssi.github.io/peer-did-method-spec/index.html#method-specific-identifier), Peer did Identifier is generated. 

ABNF Syntax for Peer DID:

peer-did = "did:peer:" numalgo encalgo "-" numbasis 
numalgo = "1" 
encalgo = "1" 
numbasis = 64*HEXDIGCI 
HEXDIGCI = HEXDIG / "a" / "b" / "c" / "d" / "e" / "f"
The numbasis is generated from [Genesis Version](https://openssi.github.io/peer-did-method-spec/index.html#dfn-genesis-version) of the Peer DID. 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,11 +5,11 @@
 module github.com/hyperledger/aries-framework-go
 
 require (
-	github.com/btcsuite/btcutil v0.0.0-20170419141449-a5ecb5d9547a
+	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
 	github.com/stretchr/testify v1.3.0
 	github.com/syndtr/goleveldb v1.0.0
 	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
-	github.com/xeipuuv/gojsonschema v1.1.1-0.20190626083556-16a6735db808
+	github.com/xeipuuv/gojsonschema v1.1.0
 	golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7
 )
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,5 @@
-github.com/btcsuite/btcutil v0.0.0-20170419141449-a5ecb5d9547a h1:UwuNC3d8iGIAwMyAO92ZGuVQYmdGkWXtQEGZdiIrJQs=
-github.com/btcsuite/btcutil v0.0.0-20170419141449-a5ecb5d9547a/go.mod h1:+5NJ2+qvTyV9exUAL/rxXi3DcLg2Ts+ymUAY5y4NvMg=
+github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d h1:yJzD/yFppdVCf6ApMkVy8cUxV0XrxdP9rVf6D87/Mng=
+github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d/go.mod h1:+5NJ2+qvTyV9exUAL/rxXi3DcLg2Ts+ymUAY5y4NvMg=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/fsnotify/fsnotify v1.4.7 h1:IXs+QLmnXW2CcXuY+8Mzv/fWEsPGWxqefPtCP5CnV9I=
@@ -26,8 +26,8 @@ github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f h1:J9EGpcZtP
 github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f/go.mod h1:N2zxlSyiKSe5eX1tZViRH5QA0qijqEDrYZiPEAiq3wU=
 github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 h1:EzJWgHovont7NscjpAxXsDA8S8BMYve8Y5+7cuRE7R0=
 github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415/go.mod h1:GwrjFmJcFw6At/Gs6z4yjiIwzuJ1/+UwLxMQDVQXShQ=
-github.com/xeipuuv/gojsonschema v1.1.1-0.20190626083556-16a6735db808 h1:RySiKkoBBLdM0Jis6C+ZBvU998pPB2niXl7UwvZ4L/4=
-github.com/xeipuuv/gojsonschema v1.1.1-0.20190626083556-16a6735db808/go.mod h1:5yf86TLmAcydyeJq5YvxkGPE2fm/u4myDekKRoLuqhs=
+github.com/xeipuuv/gojsonschema v1.1.0 h1:ngVtJC9TY/lg0AA/1k48FYhBrhRoFlEmWzsehpNAaZg=
+github.com/xeipuuv/gojsonschema v1.1.0/go.mod h1:5yf86TLmAcydyeJq5YvxkGPE2fm/u4myDekKRoLuqhs=
 golang.org/x/net v0.0.0-20180906233101-161cd47e91fd h1:nTDtHvHSdCn1m6ITfMRqtOd/9+7a3s8RBNOZ3eYZzJA=
 golang.org/x/net v0.0.0-20180906233101-161cd47e91fd/go.mod h1:mL1N/T3taQHkDXs73rZJwtUhF3w3ftmwwsq0BUmARs4=
 golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f h1:wMNYb4v58l5UBM7MYRLPG6ZhfOqbKu7X5eyFl8ZhKvA=
--- a/pkg/didmethod/peer/did.go
+++ b/pkg/didmethod/peer/did.go
@@ -1,0 +1,61 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package peer
+
+import (
+	"crypto"
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	errors "golang.org/x/xerrors"
+)
+
+const (
+	//numAlgo is the algorithm for choosing a numeric basis Reference: https://openssi.github.io/peer-did-method-spec/index.html#method-specific-identifier
+	numAlgo = "1"
+	//encAlgo  is the algorithm for encoding
+	encAlgo = "1"
+
+	peerPrefix = "did:peer:"
+)
+
+//newDid creates the peer DID. For example : did:peer:11-479cbc07c3f991725836a3aa2a581ca2029198aa420b9d99bc0e131d9f3e2cbe
+func newDid(doc *did.Doc) (string, error) {
+
+	if doc.PublicKey == nil || doc.Authentication == nil {
+		return "", errors.New("the genesis version must include public keys and authentication")
+	}
+
+	docBytes, err := json.Marshal(doc)
+	if err != nil {
+		return "", err
+	}
+
+	hash, err := computeHash(docBytes)
+	if err != nil {
+		return "", err
+	}
+	numBasis := base64.URLEncoding.EncodeToString(hash)
+	messageIdentifier := []string{peerPrefix, numAlgo, encAlgo, "-", numBasis}
+	peerDID := strings.Join(messageIdentifier, "")
+
+	return peerDID, nil
+}
+
+// computeHash will compute the hash for the supplied bytes
+func computeHash(bytes []byte) ([]byte, error) {
+
+	if len(bytes) == 0 {
+		return nil, errors.New("empty bytes")
+	}
+
+	h := crypto.SHA256.New()
+	hash := h.Sum(bytes)
+	return hash, nil
+
+}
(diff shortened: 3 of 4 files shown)


pkg/didmethod/peer/generate.go
```diff
@@ -0,0 +1,68 @@
...
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+const numalgo = "1"
+const encalgo = "1"
+
+//GeneratePeerDid creates the peer DID. For example : did:peer:11-479cbc07c3f991725836a3aa2a581ca2029198aa420b9d99bc0e131d9f3e2cbe
+func GeneratePeerDid(doc *did.Doc) (string, error) {
```
rolsonquadras: I feel the function should take the publicKey, authentication details as the input param. ie have the signature of "generateGenesis" (line 36). The function can then create the DID document and return this Genesis DID document along with ID. 

peerDID, genesisDoc, err := peer.GenerateDID([]did.PublicKey, []did.VerificationMethod)

Framework should define how the Genesis Doc is created as same doc need to be validated at the peer agent (deterministic).

pkg/didmethod/peer/generate.go
```diff
@@ -0,0 +1,68 @@
...
+	"crypto"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+const numalgo = "1"
```
rolsonquadras: Better to have some comments on these as it's difficult to read the code (names doesn't give much info). Also, these 2 constants can be wrapped.

troyronda:
The commit and PR is missing a description.

pkg/didmethod/peer/generate.go
talwinder50: Added comments and wrapped the constants

troyronda: Also, should be camel cased.

pkg/didmethod/peer/generate.go
llorllale: @talwinder50 `NewDid` would be a better name, ie. `peer.NewDid()`

pkg/didmethod/peer/generate.go
talwinder50: done

pkg/didmethod/peer/generate.go
talwinder50: Done

pkg/didmethod/peer/generate_test.go
```diff
@@ -0,0 +1,87 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package peer
+
+import (
+	"github.com/stretchr/testify/require"
```
troyronda: it is nice to have groupings of imports.

pkg/didmethod/peer/generate.go
troyronda: BTW - I would expect a function named “NewDoc” to return some doc struct. Your comment says you are creating a DID?

pkg/didmethod/peer/generate.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	//numAlgo is the algorithm for choosing a numeric basis Reference: https://openssi.github.io/peer-did-method-spec/index.html#method-specific-identifier
+	numAlgo = "1"
+	//encAlgo  is the algorithm for encoding
+	encAlgo = "1"
+)
+
+//NewDid creates the peer DID. For example : did:peer:11-479cbc07c3f991725836a3aa2a581ca2029198aa420b9d99bc0e131d9f3e2cbe
+func NewDid(doc *did.Doc) (string, error) {
```
llorllale: @talwinder50 it would be better if the DID is an attribute of the `did.Doc`. As in, `did := doc.DID()` where `doc` is a peer DID document

Basically, all DID documents and their attributes should be under a single unified model

pkg/didmethod/peer/generate.go
```diff
@@ -0,0 +1,63 @@
...
+	var peerDID string
+	peerPrefix := "did:peer:"
+	// doc MUST NOT include the DID
+	doc.ID = ""
+
+	if doc.PublicKey == nil || doc.Authentication == nil {
+		return "", fmt.Errorf("the genesis version must include public keys and authentication")
```
troyronda: should now use the xerrors package. see prior PR for an example.

rolsonquadras: @talwinder50 here is the reference - https://github.com/hyperledger/aries-framework-go/pull/55/files#diff-d9e817deb772fd4e55da4b57afbca798R16

pkg/didmethod/peer/generate.go
talwinder50: @rolsonquadras @llorllale when you say peer document you mean genesis document ?

rolsonquadras: @talwinder50 genesis doc(version) is the first version of peer doc.

pkg/didmethod/peer/generate.go
talwinder50: done

pkg/didmethod/peer/generate.go
talwinder50: done

pkg/doc/did/doc.go
```diff
@@ -212,6 +212,16 @@ type Proof struct {
 	Nonce          string     `json:"nonce,omitempty"`
 }
 
+//DID returns the new did
+func (doc *Doc) DID() (string, error) {
```
llorllale: @talwinder50 should be refactored - not all DID documents calculate their DID the same way.

talwinder50: then DID() func needs to have a peer document  as an input from somewhere (as we are using public key and authentication (Which are must to have) for creating did.

llorllale: Yes, in the case of peer DIDs the `DID()` function needs as input the DID document itself. That is not the case for other DID methods.

This suggests to me that `did.Doc.ID` should be a function, not a string:

```go
type Doc struct {
	Context        []string
	ID             func() string       /// function
	PublicKey      []PublicKey
	Service        []Service
	Authentication []VerificationMethod
	Created        *time.Time
	Updated        *time.Time
	Proof          *Proof
}
```

talwinder50: As per our discussion, when reading from storage the genesis document can be from any DID method (peer DID, sidtree, etc). But we dont know what is the differentiating factor to  compute the specific method DID. The current implementation ties the DOC struct to create the PEER dids only. I have added TODO that we do need to think about how this can be generalized for all other did methods as well.

go.mod
```diff
@@ -6,10 +6,13 @@ module github.com/hyperledger/aries-framework-go
 
 require (
 	github.com/btcsuite/btcutil v0.0.0-20170419141449-a5ecb5d9547a
+	github.com/magiconair/properties v1.8.1
+	github.com/pkg/errors v0.8.1
```
rolsonquadras: Looks like this import is not used. Also, we should be using xerrors.

pkg/doc/did/doc_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"encoding/pem"
+	"github.com/magiconair/properties/assert"
```
rolsonquadras: Can we use "require.Equals" as used in others rather than importing a new lib ? ref - https://github.com/hyperledger/aries-framework-go/blob/master/pkg/doc/did/doc_test.go#L82

talwinder50: sure

pkg/doc/did/doc.go
rolsonquadras: @talwinder50 Generating a DID is a one time task based on Genesis Version of the Document. Not sure it should be fit into a member function of Doc.

talwinder50: @rolsonquadras .. I added into a member function of doc. Based on the below comment. 

 @talwinder50 it would be better if the DID is an attribute of the did.Doc. As in, did := doc.DID() where doc is a peer DID document

Basically, all DID documents and their attributes should be under a single unified model

rolsonquadras: I assume you are referring to [this](https://github.com/hyperledger/aries-framework-go/pull/56#discussion_r309697639) comment. But I personally feel one time DID calculation should't part of a member function. It looks like the code here calculates DID dynamically whenever doc.DID is invoked. Documents can be updated the consumers and this will result in different DID on subsequent calls.

talwinder50: yea @rolsonquadras i agree. It should be one time thing

talwinder50: @llorllale can you please share your insights as well, if there was something else in your mind when you suggested that

llorllale: @rolsonquadras @talwinder50 you're correct - it's not required for `did.Doc` to have an accessor function for its DID. I am naturally inclined towards accessor functions.

To me this seems like a classic case of the Strategy pattern, but I will wait and see how much overlap there is between a parser in package `peer` and other packages.

llorllale: But to be clear - I am strongly opposed to the original design: `did := GetPeerDid(&doc)` - there's no reason why `did.Doc` itself cannot provide the DID. The DID is part of its model.

rolsonquadras: @llorllale The [original version](https://github.com/hyperledger/aries-framework-go/compare/6800c15cc516bf920a81ce2475ab53ee48881a50..f68ebb4b0edf73dec6ac74a4292e3a4f3822206c#diff-32a7b3a4704464a13c867cdbf299342eL22) had ```GeneratePeerDid(doc *did.Doc)```. It was Generate DID and not Get DID(as in, fetch existing DID). You still have reservation on this design?

llorllale: Yes - my objection was not on the name ("GeneratePeerDid") - but the fact that `doc.ID` didn't just give me the DID. And since `GeneratePeerDid` is exported, I assumed that the design intended for *users* to call `GeneratePeerDid(doc)` - which is something I strongly oppose

llorllale:
@rolsonquadras @troyronda 

To be clear: @talwinder50 un-exported the function that calculates the peer DID. Work remains to actually wire that into the process that creates new peer DID documents.

Note: we figure that the DID on incoming peer DID docs from other parties cannot be validated against the function's output because there is no way the Recipient can be sure that the version they're getting is the genesis version.

pkg/didmethod/peer/did.go
```diff
@@ -0,0 +1,63 @@
...
+	//encAlgo  is the algorithm for encoding
+	encAlgo = "1"
+)
+
+//newDid creates the peer DID. For example : did:peer:11-479cbc07c3f991725836a3aa2a581ca2029198aa420b9d99bc0e131d9f3e2cbe
+func newDid(doc *did.Doc) (string, error) {
+
+	var peerDID string
+
+	peerPrefix := "did:peer:"
```
troyronda: were you unable to use a const here?

pkg/didmethod/peer/did.go
```diff
@@ -0,0 +1,63 @@
...
+	//numAlgo is the algorithm for choosing a numeric basis Reference: https://openssi.github.io/peer-did-method-spec/index.html#method-specific-identifier
+	numAlgo = "1"
+	//encAlgo  is the algorithm for encoding
+	encAlgo = "1"
+)
+
+//newDid creates the peer DID. For example : did:peer:11-479cbc07c3f991725836a3aa2a581ca2029198aa420b9d99bc0e131d9f3e2cbe
+func newDid(doc *did.Doc) (string, error) {
+
+	var peerDID string
```
troyronda: this line is unnecessary.

pkg/didmethod/peer/did.go
talwinder50: will edit it to constant

go.sum
```diff
@@ -10,11 +12,14 @@ github.com/golang/snappy v0.0.0-20180518054509-2e65f85255db h1:woRePGFeVFfLKN/pO
...
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/magiconair/properties v1.8.1 h1:ZC2Vc7/ZFkGmsVC9KvOjumD+G5lXy2RtTKyzRKO2BQ4=
+github.com/magiconair/properties v1.8.1/go.mod h1:PppfXfuXeibc/6YijjN8zIbojt8czPbwD3XqdrwzmxQ=
 github.com/onsi/ginkgo v1.6.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
 github.com/onsi/ginkgo v1.7.0 h1:WSHQ+IS43OoUrWtD1/bbclrwK8TTH5hzp+umCiuxHgs=
 github.com/onsi/ginkgo v1.7.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
 github.com/onsi/gomega v1.4.3 h1:RE1xgDvH7imwFD45h+u2SgIfERHlS2yNG4DObb5BSKU=
 github.com/onsi/gomega v1.4.3/go.mod h1:ex+gbHU/CVuBBDIJjb2X0qEXbFg53c61hWP/1CpauHY=
+github.com/pkg/errors v0.8.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
```
rolsonquadras: Looks like this is a left-over from [previous change](https://github.com/hyperledger/aries-framework-go/pull/56#discussion_r310222231). Please run 'go mod tidy'.

pkg/doc/did/doc_test.go
```diff
@@ -156,7 +156,6 @@ func TestPublicKeys(t *testing.T) {
 		require.Contains(t, err.Error(), "public key encoding not supported")
 	})
 }
-
```
rolsonquadras: Looks like unintended change.

rolsonquadras:
Tagging the issue #46

PR #57, opened 2019-07-31
feat: logger initialization and customization

sudeshrshetty:
- added implementation for logger provider injection using
log.Initialize() function, which should be called before logging any
line to use custom loger implementation
- implemented log.NewLogger() which returns new instance of logger based
on logger provider initialized
- exposed APIs for log level and caller info management
- migrated existing logs to aries logger implementation
- closes:#21


pkg/internal/common/logging/metadata/level.go
```diff
@@ -6,26 +6,34 @@ SPDX-License-Identifier: Apache-2.0
 
 package metadata
 
-import (
-	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+// Level defines all available log levels for logging messages.
+type Level int
+
+// Log levels.
+const (
```
troyronda: I assume this was duplicated to avoid circular references?

This should have a comment that it needs to stay in sync with the other package.

sudeshrshetty: Will add those comments.

pkg/common/log/logger.go
```diff
@@ -0,0 +1,150 @@
...
+	instance Logger
+	module   string
+	once     sync.Once
+}
+
+// NewLogger creates and returns a Logger object based on the given module name.
+// note: the underlying logger instance is lazy initialized on first use.
+// To use your own logger implementation provide logger provider in 'Initialize()' before logging any line.
+// If 'Initialize()' is not called before logging any line then default logging implementation will be used.
+func NewLogger(module string) Logger {
```
troyronda: Interesting. It is log.NewLogger since it returns a Logger.

But under the covers it is actually log.NewLog (which could be returning *Log).

i.e., should we have the extra abstraction overhead?

troyronda: CC: @bstasyszyn

sudeshrshetty: Since this is returning implementation of 'Logger' interface, I named in NewLogger (also maintaining naming conventions used in other kind of loggers)

troyronda: actually i should have said log.New rather than log.NewLog (like the stdlib)

sudeshrshetty: Fixed it, changed return to *Log

pkg/common/log/logger.go
```diff
@@ -0,0 +1,150 @@
...
+	instance Logger
+	module   string
+	once     sync.Once
+}
+
+// NewLog creates and returns a Logger implementation based on given module name.
+// note: the underlying logger instance is lazy initialized on first use.
+// To use your own logger implementation provide logger provider in 'Initialize()' before logging any line.
+// If 'Initialize()' is not called before logging any line then default logging implementation will be used.
+func NewLog(module string) *Log {
```
troyronda: Since this is the main constructor, suggest shortening to New

PR #63, opened 2019-08-02
test: remove parallel flag from unit test

### LINKED ISSUE #62: Unit test - remove parallel flag
The parallel flag(maximum number of tests to run simultaneously) is set in [unit-test](https://github.com/hyperledger/aries-framework-go/blob/master/scripts/check_unit.sh#L17) to 1. This will result in running tests in sequence. If this flag is not present, it is set to the value of GOMAXPROCS. At this moment there is no need to set this to 1. By setting to default, we can decrease the test execution time.

rolsonquadras:
Removed parallel flag from go test command to run unit-tests.

Closes #62

### DIFF
--- a/scripts/check_unit.sh
+++ b/scripts/check_unit.sh
@@ -14,4 +14,4 @@ PKGS=`go list github.com/hyperledger/aries-framework-go/... 2> /dev/null | \
 
 go generate ./...
 
-go test $PKGS -count=1 -race -coverprofile=coverage.txt -covermode=atomic  -p 1 -timeout=10m
+go test $PKGS -count=1 -race -coverprofile=coverage.txt -covermode=atomic -timeout=10m


troyronda:
Thoughts about the golang test cache?

rolsonquadras:
> Thoughts about the golang test cache?
> 
> count=1 disables it.

@troyronda looks like the default value is 1. 

https://golang.org/cmd/go/#hdr-Testing_flags
-count n
    Run each test and benchmark n times (default 1).
    If -cpu is set, run n times for each GOMAXPROCS value.
    Examples are always run once.

Baha-sk:
> > Thoughts about the golang test cache?
> > count=1 disables it.
> 
> @troyronda looks like the default value is 1.
> 
> https://golang.org/cmd/go/#hdr-Testing_flags
> -count n
> Run each test and benchmark n times (default 1).
> If -cpu is set, run n times for each GOMAXPROCS value.
> Examples are always run once.

@rolsonquadras there's no need to set the default value

troyronda:
@rolsonquadras @Baha-sk @llorllale To
disable test caching, use any test flag or argument other than the
cacheable flags. The idiomatic way to disable test caching explicitly
is to use -count=1.

https://golang.org/pkg/cmd/go/internal/test/

PR #64, opened 2019-08-06
feat: Peer DID Resolver

### LINKED ISSUE #58: Peer DID: Resolver
The framework provides a [DID Resolver](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/didresolver/api.go#L22) API. 

Framework needs to provide a Peer DID Resolver against the [Backing Storage](#47).

rolsonquadras:
- Resolver against configurable backing store
- ToJSON function to convert DID Document to json format

Closes #58 

### DIFF
--- a/pkg/didmethod/peer/resolver.go
+++ b/pkg/didmethod/peer/resolver.go
@@ -1,0 +1,38 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package peer
+
+import (
+	errors "golang.org/x/xerrors"
+)
+
+// DIDResolver resolver
+type DIDResolver struct {
+	store *DIDStore
+}
+
+// NewDIDResolver new Peer DID resolver instance
+func NewDIDResolver(store *DIDStore) *DIDResolver {
+	return &DIDResolver{store: store}
+}
+
+// Read implements didresolver.DidMethod.Read interface (https://w3c-ccg.github.io/did-resolution/#resolving-input)
+func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+	// get the document from the store
+	doc, err := resl.store.Get(did)
+	if err != nil {
+		return nil, errors.Errorf("Fetching data from store failed: %w", err)
+	}
+
+	// convert the doc to JSON as DID Resolver expects byte result.
+	jsonDoc, err := doc.JSONBytes()
+	if err != nil {
+		return nil, errors.Errorf("Json marshalling of document failed: %w", err)
+	}
+
+	return jsonDoc, nil
+}
--- a/pkg/didmethod/peer/resolver_test.go
+++ b/pkg/didmethod/peer/resolver_test.go
@@ -1,0 +1,126 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package peer
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	diddoc "github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/didresolver"
+	"github.com/hyperledger/aries-framework-go/pkg/storage/leveldb"
+	"github.com/stretchr/testify/require"
+)
+
+const peerDIDDoc = `{
+  "@context": ["https://w3id.org/did/v1","https://w3id.org/did/v2"],
+  "id": "did:peer:1234",
+  "publicKey": [
+    {
+      "id": "did:peer:1234#keys-1",
+      "type": "Secp256k1VerificationKey2018",
+      "controller": "did:peer:1234",
+      "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
+    },
+    {
+      "id": "did:example:123456789abcdefghw#key2",
+      "type": "RsaVerificationKey2018",
+      "controller": "did:example:123456789abcdefghw",
+      "publicKeyPem": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAryQICCl6NZ5gDKrnSztO\n3Hy8PEUcuyvg/ikC+VcIo2SFFSf18a3IMYldIugqqqZCs4/4uVW3sbdLs/6PfgdX\n7O9D22ZiFWHPYA2k2N744MNiCD1UE+tJyllUhSblK48bn+v1oZHCM0nYQ2NqUkvS\nj+hwUU3RiWl7x3D2s9wSdNt7XUtW05a/FXehsPSiJfKvHJJnGOX0BgTvkLnkAOTd\nOrUZ/wK69Dzu4IvrN4vs9Nes8vbwPa/ddZEzGR0cQMt0JBkhk9kU/qwqUseP1QRJ\n5I1jR4g8aYPL/ke9K35PxZWuDp3U0UPAZ3PjFAh+5T+fc7gzCs9dPzSHloruU+gl\nFQIDAQAB\n-----END PUBLIC KEY-----"
+    }
+  ],
+  "authentication": [
+    "did:peer:1234#keys-1",
+    {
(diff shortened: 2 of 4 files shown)


pkg/didmethod/peer/resolver.go
```diff
@@ -0,0 +1,43 @@
...
+
+// NewDIDResolver new Peer DID resolver instance
+func NewDIDResolver(store *DIDStore) *DIDResolver {
+	return &DIDResolver{store: store}
+}
+
+// Read implements DidMethod.Read function
+func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+	// validate the input
+	if did == "" {
```
troyronda: Will store.Get fail in some strange way if did == ""?

Or produce a normal error that DID was not found. If so, I wonder why this check is needed?

pkg/didmethod/peer/resolver_test.go
```diff
@@ -0,0 +1,126 @@
...
+	document := &did.Doc{}
+	err = json.Unmarshal(doc, document)
+	require.NoError(t, err)
+	require.Equal(t, did1, document.ID)
+
+	// nil DID
+	doc, err = resl.Read("", nil, "", false)
+	require.Error(t, err)
+
+	// missing DID
```
llorllale: @rolsonquadras is this test asserting that the resolver returns an error if the DID is not found? It should not return an error - a zero value (empty byte slice) suffices

pkg/didmethod/peer/resolver.go
```diff
@@ -0,0 +1,43 @@
...
+	store *DIDStore
+}
+
+// NewDIDResolver new Peer DID resolver instance
+func NewDIDResolver(store *DIDStore) *DIDResolver {
+	return &DIDResolver{store: store}
+}
+
+// Read implements DidMethod.Read function
+func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
```
troyronda: There are extra unused arguments. I assume there should be a TODO describing how they will be used.

Also are they intended to be mandatory or optional?

pkg/didmethod/peer/resolver.go
```diff
@@ -0,0 +1,43 @@
...
+type DIDResolver struct {
+	store *DIDStore
+}
+
+// NewDIDResolver new Peer DID resolver instance
+func NewDIDResolver(store *DIDStore) *DIDResolver {
+	return &DIDResolver{store: store}
+}
+
+// Read implements DidMethod.Read function
```
troyronda: Can we please provide a fuller description?

pkg/didmethod/peer/resolver_test.go
```diff
@@ -0,0 +1,126 @@
...
+	resl := NewDIDResolver(store)
+	doc, err := resl.Read(did1, nil, "", false)
+	require.NoError(t, err)
+
+	document := &did.Doc{}
+	err = json.Unmarshal(doc, document)
+	require.NoError(t, err)
+	require.Equal(t, did1, document.ID)
+
+	// nil DID
```
troyronda: technically the zero value (empty) not nil.

pkg/didmethod/peer/resolver.go
llorllale: These are all [optional](https://w3c-ccg.github.io/did-resolution/#resolving-input) (except for the DID of course)

pkg/doc/did/doc.go
```diff
@@ -345,3 +352,73 @@ func stringEntry(entry interface{}) string {
 	}
 	return entry.(string)
 }
+
+// ToJSON converts document to raw json
+func (doc *Doc) ToJSON() ([]byte, error) {
```
llorllale: This method's name does not match what it produces. Might be clearer to rename it `RawJson`

troyronda: Out of curiosity, why raw instead of say.. Bytes?

pkg/didmethod/peer/resolver.go
rolsonquadras: If we remove the validation, the error that we get is not easily readable.

Error description when we remove it.
```
Fetching data from store failed:
     github.com/hyperledger/aries-framework-go/pkg/didmethod/peer.(*DIDResolver).Read
         /Users/rolson.quadras/gows/src/github.com/hyperledger/aries-framework-go/pkg/didmethod/peer/resolver.go:33
   - ID is mandatory:
     github.com/hyperledger/aries-framework-go/pkg/didmethod/peer.(*DIDStore).Get
         /Users/rolson.quadras/gows/src/github.com/hyperledger/aries-framework-go/pkg/didmethod/peer/store.go:78
```

pkg/didmethod/peer/resolver.go
llorllale: Opened #65 to refactor this

pkg/doc/did/doc.go
llorllale: That works too.

pkg/didmethod/peer/resolver.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/476126814875769cf2630645a8567f84753b0219..0dc9081c1b8917fad36c8e7230cbf4a4f67693f4#diff-85c3d889e3df4373e3188d4932b12f3fR23) the description and added link to the spec.

pkg/didmethod/peer/resolver_test.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/476126814875769cf2630645a8567f84753b0219..0dc9081c1b8917fad36c8e7230cbf4a4f67693f4#diff-70e7a24fcba1ead3adbfa0488f9586adR91) the comments in the test file.

pkg/doc/did/doc.go
rolsonquadras: @troyronda @llorllale [Updated](https://github.com/hyperledger/aries-framework-go/compare/0dc9081c1b8917fad36c8e7230cbf4a4f67693f4..2edd9c351304abb26a71e1406933eea923d0c228#diff-ca45993820e1e4b4338acd4e485ae699R357) the name of the function to JSONBytes.

pkg/didmethod/peer/resolver.go
troyronda: I'd argue (IMO):
- This doesn't seem like a common scenario - why would an empty DID be provided anyways?
- we should minimize input checks that don't result in serious issues like a panic or causing corruption.
- error checking should be handled as close as possible to where the real check is needed, and bubble it back up (rather than duplicating at every level).
- parameters should be documented in the code (and, hence, GoDoc) comments: what is each argument for? Is it required? If not, how do I not include?
  - I generally think "just in case" checks for a human doing something very silly should be minimized (i.e., is this a check against a "silly" human error or logic error elsewhere in the code?).
-  Ideally zero-values are treated as defaults, when possible.

pkg/didmethod/peer/resolver_test.go
rolsonquadras: @llorllale Yes, thats what the test is doing. The [DID Resolver API](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/didresolver/api.go#L22) doesn't allow to return bool value.

pkg/didmethod/peer/resolver.go
rolsonquadras: agreed with most of the points. [Removed](https://github.com/hyperledger/aries-framework-go/compare/2edd9c351304abb26a71e1406933eea923d0c228..2e43c02af5c1920a233d76a7fb6f3430ef827bc6) the validation.

pkg/didmethod/peer/resolver_test.go
llorllale: Deferred to #65

pkg/doc/did/doc.go
```diff
@@ -345,3 +352,73 @@ func stringEntry(entry interface{}) string {
...
+
+	byteDoc, err := json.Marshal(raw)
+	if err != nil {
+		return nil, errors.Errorf("Json unmarshalling of document failed: %w", err)
+	}
+
+	return byteDoc, nil
+}
+
+func populateRawServices(services []Service) []map[string]interface{} {
```
troyronda: What makes something "Raw"?

rolsonquadras: Existing code has a struct named `rawDoc` - https://github.com/hyperledger/aries-framework-go/blob/master/pkg/doc/did/doc.go#L194. And many operations/variable names have `raw` as the prefix

PR #66, opened 2019-08-06
Validate peer DID identifier 

### LINKED ISSUE #48: Peer DID: Validate Doc against DID Identifier
The agents exchange Peer DIDs using [DID Exchange Protocol](https://github.com/hyperledger/aries-rfcs/blob/master/features/0023-did-exchange/README.md). 

The receiving agent should have the capability to [validate](https://openssi.github.io/peer-did-method-spec/index.html#create-register) the DID Document against the name-string component of the DID Identifier. 

The generation of Identifier is described in #46.

talwinder50:
The receiving agent should have the capability to validate the DID Document against the name-string component of the DID Identifier.

closes #48

### DIFF
--- a/pkg/didmethod/peer/did.go
+++ b/pkg/didmethod/peer/did.go
@@ -7,8 +7,9 @@ package peer
 
 import (
 	"crypto"
-	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
+	"regexp"
 	"strings"
 
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
@@ -30,7 +31,61 @@ func newDid(doc *did.Doc) (string, error) {
 	if doc.PublicKey == nil || doc.Authentication == nil {
 		return "", errors.New("the genesis version must include public keys and authentication")
 	}
+	numBasis, err := numBasis(doc)
+	if err != nil {
+		return "", err
+	}
+	messageIdentifier := []string{peerPrefix, numAlgo, encAlgo, "-", numBasis}
+	peerDID := strings.Join(messageIdentifier, "")
 
+	return peerDID, nil
+}
+
+/*
+ validateDID checks the format of the doc's DID and checks that the DID's 'namestring' matches against its numeric basis as per the
+ Namestring Generation Method: https://openssi.github.io/peer-did-method-spec/index.html#namestring-generation-method
+ Note: this check should be done only on the resolved variant of the genesis version of Peer DID documents.
+*/
+func validateDID(doc *did.Doc) error {
+
+	peerDid := doc.ID
+
+	matched, _ := regexp.MatchString(`did:peer:11-([a-fA-F0-9]){64}`, peerDid)
+	if !matched {
+		return errors.Errorf("did doesnt follow matching regex")
+	}
+
+	//extracting numbasis from the validated did
+	splitDid := strings.FieldsFunc(peerDid, func(r rune) bool { return r == '-' })
+	extractedNumBasis := splitDid[1]
+
+	// genesis version(no did) of the peer DID doc
+	genesisDoc := &did.Doc{
+		Context:        doc.Context,
+		ID:             "",
+		PublicKey:      doc.PublicKey,
+		Service:        doc.Service,
+		Authentication: doc.Authentication,
+		Created:        doc.Created,
+		Updated:        doc.Updated,
+		Proof:          doc.Proof,
+	}
+
+	//calculate the numbasis of the genesis version of the peer DID doc
+	numBas, err := numBasis(genesisDoc)
+	if err != nil {
+		return err
+	}
+
+	if !(numBas == extractedNumBasis) {
+		return errors.New("hash of the doc doesnt match the computed hash")
+	}
+	return nil
+}
+
+/* numBasis is numeric basis. The spec requires a 256-bit (encoded using 64 hexadecimal digits) numBasis generated from a hash of the initial content of a DID doc i.e genesis doc.
+   Reference : https://dhh1128.github.io/peer-did-method-spec/#matching-regex */
+func numBasis(doc *did.Doc) (string, error) {
 	docBytes, err := json.Marshal(doc)
 	if err != nil {
 		return "", err
@@ -40,11 +95,9 @@ func newDid(doc *did.Doc) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	numBasis := base64.URLEncoding.EncodeToString(hash)
-	messageIdentifier := []string{peerPrefix, numAlgo, encAlgo, "-", numBasis}
-	peerDID := strings.Join(messageIdentifier, "")
+	numBasis := hex.EncodeToString(hash)
 
-	return peerDID, nil
+	return numBasis, nil
 }
 
 // computeHash will compute the hash for the supplied bytes
--- a/pkg/didmethod/peer/did_test.go
+++ b/pkg/didmethod/peer/did_test.go
@@ -15,18 +15,16 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestNewDid(t *testing.T) {
-	storedDoc, err := GenerateGenesisDoc()
-	require.NoError(t, err)
+func TestNewDID(t *testing.T) {
+	storedDoc := genesisDoc()
 	require.NotNil(t, storedDoc)
-
 	peerDID, err := newDid(storedDoc)
 	require.NoError(t, err)
 	require.NotNil(t, peerDID)
 	assert.Contains(t, peerDID, "did:peer:11")
 }
 
-func TestNewDidError(t *testing.T) {
+func TestNewDIDError(t *testing.T) {
 	storedDoc := &did.Doc{ID: "did:peer:11"}
 	_, err := newDid(storedDoc)
 	require.Error(t, err)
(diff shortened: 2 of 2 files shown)


pkg/didmethod/peer/did_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func TestComputeHashError(t *testing.T) {
+	hash, err := computeHash([]byte(""))
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "empty bytes")
+	assert.Nil(t, hash)
+}
+
+// GenerateGenesisDoc creates the doc without an id
+func GenerateGenesisDoc() (*did.Doc, error) {
```
llorllale: Please un-export this

pkg/didmethod/peer/did.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	//numAlgo is the algorithm for choosing a numeric basis Reference: https://openssi.github.io/peer-did-method-spec/index.html#method-specific-identifier
+	numAlgo = "1"
+	//encAlgo  is the algorithm for encoding
+	encAlgo = "1"
+)
+
+//newDid creates the peer DID. For example : did:peer:11-479cbc07c3f991725836a3aa2a581ca2029198aa420b9d99bc0e131d9f3e2cbe
+func newDid(doc *did.Doc) (string, error) {
```
llorllale: This is already being delivered in PR #56

talwinder50: yes i think it is not merged thats why its showing two commits .. please ignore.

pkg/didmethod/peer/validate.go
```diff
@@ -0,0 +1,33 @@
...
+package peer
+
+import (
+	"regexp"
+	"unicode"
+
+	errors "golang.org/x/xerrors"
+)
+
+func validate(did string) error {
```
rolsonquadras: In case of Peer DIDs, the validation should be done against the genesis version of the document.

Quoting the [spec](https://openssi.github.io/peer-did-method-spec/index.html#create-register)

> The inviter must verify that the received DID was generated according to the namestring generation method from the stored variant of the received genesis DID Doc

llorllale: Interesting - I previously thought [here](https://github.com/hyperledger/aries-framework-go/pull/56#issuecomment-517830852) the Recipient of a peer DID should not attempt this validation. Turns out that by enforcing this validation we help prevent MiTM attacks as explained [here](https://openssi.github.io/peer-did-method-spec/index.html#proof-of-control).

The implication is that a party that wants to reuse the same peer DID for another connection must replay all changes back to the new recipient.

troyronda: @talwinder50 any updates on this topic and this PR?

talwinder50: @troyronda pushing patch set in next 15 min

llorllale:
@talwinder50 can you amend the message on your commit to say "peer DID" rather than just "DID"?

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +32,59 @@ func newDid(doc *did.Doc) (string, error) {
...
+}
+
+func validate(doc *did.Doc) error {
+	did := doc.ID
+	hashedNumBasis, err := validateDid(did)
+	if err != nil {
+		return err
+	}
+	//removing did to calculate the hash of genesis doc
+	doc.ID = ""
```
llorllale: It's a bad idea to mutate input args on validation functions. Make a copy of the input doc instead.

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +32,59 @@ func newDid(doc *did.Doc) (string, error) {
...
+	}
+
+	hashed := strings.FieldsFunc(did, split)
+
+	numBasis := hashed[1]
 
+	return numBasis, nil
+}
+
+func hashString(doc *did.Doc) (string, error) {
```
llorllale: This can be renamed to "numBasis" to mirror the terms used in the spec.

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +32,59 @@ func newDid(doc *did.Doc) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	messageIdentifier := []string{peerPrefix, numAlgo, encAlgo, "-", numBasis}
+	peerDID := strings.Join(messageIdentifier, "")
+
+	return peerDID, nil
+}
+
+func validate(doc *did.Doc) error {
```
llorllale: This needs some godocs explaining what it does and, most importantly, the requirements it has on the input (it works only on the genesis version of the peer DID doc)

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +32,59 @@ func newDid(doc *did.Doc) (string, error) {
...
+		return err
+	}
+
+	if !(numBasis == hashedNumBasis) {
+		return errors.New("hash of the doc doesnt match the computed hash")
+	}
+	return nil
+}
+
+func validateDid(did string) (string, error) {
```
llorllale: This function is doing two things: it is validating the peer DID format, and it's also returning its `numBasis`. I suggest three things:

* Add godocs explaining what it does
* Extract the portion that returns the numBasis of a given peer DID into its own function and use that one

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +32,59 @@ func newDid(doc *did.Doc) (string, error) {
...
+	//no whitespace
+	for _, v := range did {
+		// test each character to see if it is whitespace.
+		if unicode.IsSpace(v) {
+			if v == 32 {
```
llorllale: I think this `if` check is not required given the use of `unicode.IsSpace()`

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +32,59 @@ func newDid(doc *did.Doc) (string, error) {
...
+				return "", errors.New("white space is not allowed in did")
+			}
+		}
+	}
+	// matching regex
+	matched, _ := regexp.MatchString(`did:peer:(1)(1)-([a-fA-F0-9]{64})*`, did)
```
llorllale: This regex is wrong - take a look at https://dhh1128.github.io/peer-did-method-spec/#matching-regex

pkg/didmethod/peer/did_test.go
```diff
@@ -81,5 +131,24 @@ func GenerateGenesisDoc() (*did.Doc, error) {
 		Created:        &time.Time{},
 	}
 	return &did.Doc{Context: doc.Context, PublicKey: doc.PublicKey, Authentication: doc.Authentication,
-		Created: doc.Created}, nil
+		Created: doc.Created}
+}
+
+func generatePeerDoc() (*did.Doc, error){
```
llorllale: Suggest we remove the "generate" prefix from all these functions.

talwinder50: removed

pkg/didmethod/peer/did.go
talwinder50: fixed removed * from the end

pkg/didmethod/peer/did.go
talwinder50: removed

pkg/didmethod/peer/did.go
talwinder50: added comments and splited the function as suggested

pkg/didmethod/peer/did.go
talwinder50: added comments

pkg/didmethod/peer/did.go
talwinder50: refactored

pkg/didmethod/peer/did.go
talwinder50: created a copy of the input doc

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +31,77 @@ func newDid(doc *did.Doc) (string, error) {
...
+	if !matched {
+		return errors.Errorf("did doesnt follow matching regex")
+	}
+	return nil
+}
+
+// extractNumbasis extract the portion that returns the numBasis of a given peer DID
+func extractNumBasis(peerDid string) string {
+	splitDid := strings.FieldsFunc(peerDid, func(r rune) bool { return r == '-' })
+	numBasis := splitDid[1]
```
llorllale: Because this may result in a panic if the input argument is malformed, I suggest we remove this function and inline the splitting where it's needed (right after the format is validated).

pkg/didmethod/peer/did.go
llorllale: Sorry - I think we should rename this function to `validateDid` (below I suggested to remove the existing "validateDid" function) and I think a simpler comment could be:

```
validateDid checks the format of the doc's DID and checks that the DID's 'namestring' matches against its numeric basis as per the Namestring Generation Method: https://openssi.github.io/peer-did-method-spec/index.html#namestring-generation-method
Note: this check should be done only on the resolved variant of the genesis version of Peer DID documents.
```

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +31,77 @@ func newDid(doc *did.Doc) (string, error) {
...
+
+	if !(numBasis == extractedNumBasis) {
+		return errors.New("hash of the doc doesnt match the computed hash")
+	}
+	return nil
+}
+
+/* validateDid validates the did using the regular expression mentioned in the spec.
+   Reference: https://dhh1128.github.io/peer-did-method-spec/#matching-regex */
+func validateDid(did string) error {
```
llorllale: This function can be removed and instead inlined into the one above

pkg/didmethod/peer/did_test.go
```diff
@@ -81,5 +115,24 @@ func GenerateGenesisDoc() (*did.Doc, error) {
...
+		Created: doc.Created}
+}
+
+func peerDidDoc() (*did.Doc, error){
+
+	doc :=  genesisDoc()
+    did, err := newDid(doc)
+    if err!= nil {
+    	return nil,  err
+	}
```
troyronda: this go file needs to be formatted.

pkg/didmethod/peer/did_test.go
```diff
@@ -81,5 +115,24 @@ func GenerateGenesisDoc() (*did.Doc, error) {
...
+    doc.ID = did
+	return doc , nil
+}
+
+func peerDidDocError() *did.Doc {
```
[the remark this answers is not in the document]
talwinder50: done

pkg/didmethod/peer/did_test.go
talwinder50: i formatted in local

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +31,60 @@ func newDid(doc *did.Doc) (string, error) {
...
 
+	return peerDID, nil
+}
+
+/*
+ validateDid checks the format of the doc's DID and checks that the DID's 'namestring' matches against its numeric basis as per the
+ Namestring Generation Method: https://openssi.github.io/peer-did-method-spec/index.html#namestring-generation-method
+ Note: this check should be done only on the resolved variant of the genesis version of Peer DID documents.
+*/
+func validateDid(doc *did.Doc) error {
```
troyronda: DID should be capitals: validateDID

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +31,60 @@ func newDid(doc *did.Doc) (string, error) {
...
+	// genesis version(no did) of the peer DID doc
+	genesisDoc := &did.Doc{
+		Context:        doc.Context,
+		ID:             "",
+		PublicKey:      doc.PublicKey,
+		Service:        doc.Service,
+		Authentication: doc.Authentication,
+		Created:        doc.Created,
+		Updated:        doc.Updated,
+		Proof:          doc.Proof}
```
troyronda: prefer the closing brace on next line.

pkg/didmethod/peer/did.go
```diff
@@ -30,7 +31,60 @@ func newDid(doc *did.Doc) (string, error) {
...
+
+	//calculate the numbasis of the genesis version of the peer DID doc
+	numBasis, err := numBasis(genesisDoc)
```
troyronda: unusual (to me) that variable names are matching method names.

PR #88, opened 2019-08-12
feat: Framework Init - base template with Outbound transport config

### LINKED ISSUE #78: Framework Init - base template with Outbound transport config
As part of the requirements in #74, the consumer of the lib should have the ability to initialize the framework and pass in different configuration options.

As part of this issue, build a template for framework initialization.
- Add New(opts) function which takes in the options to configuration
- Support outbound transport configuration

rolsonquadras:
- Add New(opts) function which takes in the options to configuration
- Support outbound transport configuration

Closes #78

### DIFF
--- a/pkg/didcomm/protocol/exchange/exchange.go
+++ b/pkg/didcomm/protocol/exchange/exchange.go
@@ -68,6 +68,11 @@ type Connection struct {
 	DIDDoc *did.Doc `json:"did_doc,omitempty"`
 }
 
+// provider contains dependencies for the Exchange protocol and is typically created by using aries.Context()
+type provider interface {
+	OutboundTransport() transport.OutboundTransport
+}
+
 // GenerateInviteWithPublicDID generates the DID exchange invitation string with public DID
 func GenerateInviteWithPublicDID(invite *Invitation) (string, error) {
 	if invite.ID == "" || invite.DID == "" {
@@ -86,8 +91,20 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
 	return encodedExchangeInvitation(invite)
 }
 
+// Protocol for exchange protocol
+type Protocol struct {
+	outboundTransport transport.OutboundTransport
+}
+
+// New instanstiated new exchange client
+// The argument takes a implementation of transport.OutboundTransport (dependencies required for Exchange protocol) and
+// this is typically called by using aries.Context()
+func New(prov provider) *Protocol {
+	return &Protocol{prov.OutboundTransport()}
+}
+
 // SendExchangeRequest sends exchange request
-func SendExchangeRequest(exchangeRequest *Request, destination string, transport transport.OutboundTransport) error {
+func (p *Protocol) SendExchangeRequest(exchangeRequest *Request, destination string) error {
 	if exchangeRequest == nil {
 		return errors.New("exchangeRequest cannot be nil")
 	}
@@ -95,12 +112,12 @@ func SendExchangeRequest(exchangeRequest *Request, destination string, transport
 	exchangeRequest.Type = connectionRequest
 
 	// ignore response data as it is not used in this communication mode as defined in the spec
-	_, err := marshalAndSend(exchangeRequest, "Error Marshalling Exchange Request", destination, transport)
+	_, err := p.marshalAndSend(exchangeRequest, "Error Marshalling Exchange Request", destination)
 	return err
 }
 
 // SendExchangeResponse sends exchange response
-func SendExchangeResponse(exchangeResponse *Response, destination string, transport transport.OutboundTransport) error {
+func (p *Protocol) SendExchangeResponse(exchangeResponse *Response, destination string) error {
 	if exchangeResponse == nil {
 		return errors.New("exchangeResponse cannot be nil")
 	}
@@ -108,7 +125,7 @@ func SendExchangeResponse(exchangeResponse *Response, destination string, transp
 	exchangeResponse.Type = connectionResponse
 
 	// ignore response data as it is not used in this communication mode as defined in the spec
-	_, err := marshalAndSend(exchangeResponse, "Error Marshalling Exchange Response", destination, transport)
+	_, err := p.marshalAndSend(exchangeResponse, "Error Marshalling Exchange Response", destination)
 	return err
 }
 
@@ -123,10 +140,10 @@ func encodedExchangeInvitation(inviteMessage *Invitation) (string, error) {
 	return base64.URLEncoding.EncodeToString(invitationJSON), nil
 }
 
-func marshalAndSend(data interface{}, errorMsg, destination string, transport transport.OutboundTransport) (string, error) {
+func (p *Protocol) marshalAndSend(data interface{}, errorMsg, destination string) (string, error) {
 	jsonString, err := json.Marshal(data)
 	if err != nil {
 		return "", errors.Errorf("%s : %w", errorMsg, err)
 	}
-	return transport.Send(string(jsonString), destination)
+	return p.outboundTransport.Send(string(jsonString), destination)
 }
--- a/pkg/framework/aries/api/factory.go
+++ b/pkg/framework/aries/api/factory.go
@@ -1,0 +1,16 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package api
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+)
+
+// TransportProviderFactory allows overriding of aries protocol providers
+type TransportProviderFactory interface {
+	CreateOutboundTransport() transport.OutboundTransport
+}
--- a/pkg/framework/aries/default.go
+++ b/pkg/framework/aries/default.go
@@ -1,0 +1,33 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package aries
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/factory/transport"
+)
+
+// defFramework provides default framework configs
+type defFramework struct{}
+
+// transportProviderFactory provides default Outbound Transport provider factory
+func (d defFramework) transportProviderFactory() api.TransportProviderFactory {
+	return transport.NewProviderFactory()
+}
+
+// defFrameworkOpts provides default framework options
+func defFrameworkOpts() []Option {
+	// get the default framework configs
+	def := defFramework{}
+
+	var opts []Option
+	// protocol provider factory
+	opt := WithTransportProviderFactory(def.transportProviderFactory())
+	opts = append(opts, opt)
+
+	return opts
+}
--- a/pkg/framework/aries/framework.go
+++ b/pkg/framework/aries/framework.go
@@ -1,0 +1,52 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package aries
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/context"
+	errors "golang.org/x/xerrors"
+)
+
+// Aries provides access to clients being managed by the framework.
+type Aries struct {
+	transport api.TransportProviderFactory
+}
+
+// Option configures the framework.
+type Option func(opts *Aries) error
+
+// New initializes the Aries framework based on the set of options provided.
+func New(opts ...Option) (*Aries, error) {
+	// get the default framework options
+	defOpts := defFrameworkOpts()
+
+	frameworkOpts := &Aries{}
+
+	// generate framework configs from options
+	for _, option := range append(defOpts, opts...) {
+		err := option(frameworkOpts)
+		if err != nil {
+			return nil, errors.Errorf("Error in option passed to New: %w", err)
+		}
+	}
+
+	return frameworkOpts, nil
+}
+
+// WithTransportProviderFactory injects a protocol provider factory interface to Aries
+func WithTransportProviderFactory(ot api.TransportProviderFactory) Option {
+	return func(opts *Aries) error {
+		opts.transport = ot
+		return nil
+	}
+}
+
+// Context provides handle to framework context
+func (a *Aries) Context() (*context.Provider, error) {
+	return context.New(context.WithOutboundTransport(a.transport.CreateOutboundTransport()))
+}
--- a/pkg/framework/aries/framework_test.go
+++ b/pkg/framework/aries/framework_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package aries
+
+import (
+	"testing"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/exchange"
+	mocktransport "github.com/hyperledger/aries-framework-go/pkg/internal/didcomm/transport/mock"
+	"github.com/stretchr/testify/require"
+	errors "golang.org/x/xerrors"
+)
+
+func TestFramework(t *testing.T) {
+	// framework new - error
+	_, err := New(func(opts *Aries) error {
+		return errors.New("error creating the framework option")
+	})
+	require.Error(t, err)
+
+	// framework new - success
+	aries, err := New(WithTransportProviderFactory(&mockTransportProviderFactory{}))
+	require.NoError(t, err)
+
+	// context
+	ctx, err := aries.Context()
+	require.NoError(t, err)
+
+	// exchange client
+	exClient := exchange.New(ctx)
+	require.NoError(t, err)
+
+	req := &exchange.Request{
+		ID:    "5678876542345",
+		Label: "Bob",
+	}
+	require.NoError(t, exClient.SendExchangeRequest(req, "http://example/didexchange"))
+	require.Error(t, exClient.SendExchangeRequest(req, ""))
+}
+
+type mockTransportProviderFactory struct {
+}
+
+func (f *mockTransportProviderFactory) CreateOutboundTransport() transport.OutboundTransport {
+	return mocktransport.NewOutboundTransport("success")
+}
--- a/pkg/framework/context/context.go
+++ b/pkg/framework/context/context.go
@@ -1,0 +1,45 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package context
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	errors "golang.org/x/xerrors"
+)
+
+// Provider supplies the framework configuration to client objects.
+type Provider struct {
+	outboundTransport transport.OutboundTransport
+}
+
+// New instantiated new context provider
+func New(opts ...ProviderOption) (*Provider, error) {
+	ctxProvider := Provider{}
+	for _, opt := range opts {
+		err := opt(&ctxProvider)
+		if err != nil {
+			return nil, errors.Errorf("Error in option passed to New: %w", err)
+		}
+	}
+	return &ctxProvider, nil
+}
+
+// OutboundTransport returns the outbound transport provider
+func (p *Provider) OutboundTransport() transport.OutboundTransport {
+	return p.outboundTransport
+}
+
+// ProviderOption configures the framework.
+type ProviderOption func(opts *Provider) error
+
+// WithOutboundTransport injects transport provider into the framework
+func WithOutboundTransport(ot transport.OutboundTransport) ProviderOption {
+	return func(opts *Provider) error {
+		opts.outboundTransport = ot
+		return nil
+	}
+}
--- a/pkg/didcomm/protocol/exchange/exchange_test.go
+++ b/pkg/didcomm/protocol/exchange/exchange_test.go
@@ -9,6 +9,8 @@ package exchange
 import (
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+
 	mocktransport "github.com/hyperledger/aries-framework-go/pkg/internal/didcomm/transport/mock"
 	"github.com/stretchr/testify/require"
 )
@@ -83,19 +85,19 @@ func TestGenerateInviteWithKeyAndEndpoint(t *testing.T) {
 }
 
 func TestSendRequest(t *testing.T) {
-	oTr := mocktransport.NewOutboundTransport(successResponse)
+	prov := New(&mockProvider{})
 
 	req := &Request{
 		ID:    "5678876542345",
 		Label: "Bob",
 	}
 
-	require.NoError(t, SendExchangeRequest(req, destinationURL, oTr))
-	require.Error(t, SendExchangeRequest(nil, destinationURL, oTr))
+	require.NoError(t, prov.SendExchangeRequest(req, destinationURL))
+	require.Error(t, prov.SendExchangeRequest(nil, destinationURL))
 }
 
 func TestSendResponse(t *testing.T) {
-	oTr := mocktransport.NewOutboundTransport(successResponse)
+	prov := New(&mockProvider{})
 
 	resp := &Response{
 		ID: "12345678900987654321",
@@ -104,6 +106,13 @@ func TestSendResponse(t *testing.T) {
 		},
 	}
 
-	require.NoError(t, SendExchangeResponse(resp, destinationURL, oTr))
-	require.Error(t, SendExchangeResponse(nil, destinationURL, oTr))
+	require.NoError(t, prov.SendExchangeResponse(resp, destinationURL))
+	require.Error(t, prov.SendExchangeResponse(nil, destinationURL))
+}
+
+type mockProvider struct {
+}
+
+func (p *mockProvider) OutboundTransport() transport.OutboundTransport {
+	return mocktransport.NewOutboundTransport(successResponse)
 }
--- a/pkg/framework/aries/factory/transport/transport_factory.go
+++ b/pkg/framework/aries/factory/transport/transport_factory.go
@@ -1,0 +1,27 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package transport
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+)
+
+// ProviderFactory represents the default transport provider factory.
+type ProviderFactory struct {
+}
+
+// NewProviderFactory returns the default transport provider factory.
+func NewProviderFactory() *ProviderFactory {
+	f := ProviderFactory{}
+	return &f
+}
+
+// CreateOutboundTransport returns a new default implementation of outbound transport provider
+func (f *ProviderFactory) CreateOutboundTransport() transport.OutboundTransport {
+	// TODO - https://github.com/hyperledger/aries-framework-go/issues/83
+	return nil
+}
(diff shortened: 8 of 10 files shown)


pkg/framework/framework.go
```diff
@@ -0,0 +1,66 @@
...
+
+package framework
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	errors "golang.org/x/xerrors"
+)
+
+// AriesFramework provides access to clients being managed by the Framework.
+type AriesFramework struct {
```
llorllale: Might be a good idea to name this `Aries` instead

pkg/framework/framework.go
```diff
@@ -0,0 +1,66 @@
...
+	OutboundTransport transport.OutboundTransport
+}
+
+// Option configures the framework.
+type Option func(opts *options) error
+
+// New initializes the Framework based on the set of options provided.
+func New(opts ...Option) (*AriesFramework, error) {
+	// get the default framework configs
+	def := defFramework{}
```
troyronda: You could alternatively represent defFramework as, itself, an Option that is processed first.

pkg/framework/framework.go
rolsonquadras: @troyronda any thoughts on the name?

rolsonquadras: [Moved](https://github.com/hyperledger/aries-framework-go/compare/9b194c723afdf936279c58a665113ec88f3239e0..94efeb32066ee1f587c67c8642c5e2ff588c5efa#diff-da8e5280b0328cb616f0b42390a35b00R7) the package inside aries folder. The framework can be instantiated with aries.New(opts).

pkg/framework/framework.go
rolsonquadras: [Assigned](https://github.com/hyperledger/aries-framework-go/compare/9b194c723afdf936279c58a665113ec88f3239e0..94efeb32066ee1f587c67c8642c5e2ff588c5efa#diff-7b2448fdc2181299533bc944239b9744R29) defaults directly to options.

pkg/framework/aries/framework.go
```diff
@@ -0,0 +1,63 @@
...
+// WithOutboundTransport injects a OutboundTransport interface to the Framework
+func WithOutboundTransport(ot transport.OutboundTransport) Option {
+	return func(opts *options) error {
+		opts.OutboundTransport = ot
+		return nil
+	}
+}
+
+// OutboundTransport provides configured OutboundTransport
+func (f *Framework) OutboundTransport() transport.OutboundTransport {
```
troyronda: Why does this need to be exposed? It should be injected into instantiated packages.

pkg/framework/aries/default.go
```diff
@@ -0,0 +1,40 @@
...
+// defFramework provides default framework configs
+type defFramework struct{}
+
+// outboundTransport provides default OutboundTransport
+func (d defFramework) outboundTransport() (transport.OutboundTransport, error) {
+	// TODO - https://github.com/hyperledger/aries-framework-go/issues/83
+	return nil, nil
+}
+
+// defFrameworkOpts provides default framework optios
```
troyronda: typo "optios"

pkg/framework/aries/framework.go
```diff
@@ -0,0 +1,63 @@
...
+package aries
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	errors "golang.org/x/xerrors"
+)
+
+// Framework provides access to clients being managed by the Framework.
+type Framework struct {
+	opts options
```
troyronda: Why a substructure of options? i.e., Why not have OutboundTransport as part of the Framework struct?

pkg/framework/aries/framework.go
```diff
@@ -0,0 +1,63 @@
...
+
+package aries
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	errors "golang.org/x/xerrors"
+)
+
+// Framework provides access to clients being managed by the Framework.
+type Framework struct {
```
troyronda: I would have thought the aries package's New would return a *Aries?

Also the comment is a bit hard to grok (Framework ... Framework): // Framework provides access to clients being managed by the Framework.

pkg/framework/aries/default.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/94efeb32066ee1f587c67c8642c5e2ff588c5efa..b6b4c89d93e1cd85b02b49cf33b3cd03858e5bba#diff-da8e5280b0328cb616f0b42390a35b00R23)

pkg/framework/aries/framework.go
rolsonquadras: It would stutter if we have `Aries` as the name, as we already have package name as `aries`.

pkg/framework/aries/framework.go
rolsonquadras: [DIDExchange Send Request/Response](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/exchange/exchange.go#L90) takes transport has one of the function arguments. The consumer would need to get the transport from framework to pass into this. (Unless we update DID Exchange to use instance of transport).

pkg/framework/aries/framework.go
rolsonquadras: I thought it would be cleaner to have config options inside its own struct than at the top level. In case we need something is future which are different from options/config, we can form a new group at top level. Anyways, opts is not exposed and we can refactor as we want. Let me know your thoughts, I can move it if its required.

pkg/framework/aries/framework.go
troyronda: Yes but the Framework should be instantiating a DID exchange instance and supplying a transport to use. The point of a framework is so consumers don't need to manage everything...

pkg/framework/aries/framework.go
troyronda: This isn't stutter.

troyronda: It's normal for a package that supplies/constructs a single type to have the same name as the struct.

For an explanation, I like the first answer here: https://stackoverflow.com/questions/45885985/ways-to-avoid-stuttering-in-go-package-and-struct-names

> What you want to avoid is unnecessary stuttering

pkg/framework/aries/framework.go
troyronda: To me, it doesn't seem like "options" once the constructor has completed...

pkg/framework/aries/framework.go
troyronda: Note: I don't think that only having *one* transport at a time in the framework works either.

pkg/framework/aries/framework.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/b6b4c89d93e1cd85b02b49cf33b3cd03858e5bba..e7224c2166bc28f01e32148341d7f003d8fe83c0#diff-7b2448fdc2181299533bc944239b9744R15).

pkg/framework/aries/framework.go
rolsonquadras: [Moved](https://github.com/hyperledger/aries-framework-go/compare/e7224c2166bc28f01e32148341d7f003d8fe83c0..226512799fa5cca78500bfcc4f336200df37bea8#diff-7b2448fdc2181299533bc944239b9744R15).

pkg/framework/aries/framework.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/e7224c2166bc28f01e32148341d7f003d8fe83c0..226512799fa5cca78500bfcc4f336200df37bea8) based on our discussion (except Context name).

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -68,6 +68,11 @@ type Connection struct {
 	DIDDoc *did.Doc `json:"did_doc,omitempty"`
 }
 
+// Context context for exchange protocol
```
troyronda: Context context :)

pkg/framework/aries/context.go
```diff
@@ -0,0 +1,22 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package aries
+
+import "github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+
+// Context provides handles to various configs to be used by the clients
+type Context struct {
```
troyronda: FYI: in Fabric SDK Go, we had the context implementation in its own [package](https://github.com/hyperledger/fabric-sdk-go/tree/master/pkg/context).

Thoughts?

CC: @sudeshrshetty @Baha-sk @fqutishat @bstasyszyn

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/226512799fa5cca78500bfcc4f336200df37bea8..cf2367d04a6dbd7e0196b81fcf24ec5a1df59ae9).

pkg/framework/aries/context.go
troyronda: Another example is the AWS v2 SDK.

The config object has some similarity: https://github.com/aws/aws-sdk-go-v2/blob/master/aws/config.go

(and is kept in the same package).

troyronda: Note: this Context is an Application Context as opposed to the stdlib context.Context, which is a Request Context / Net Context (for cancellation).

pkg/framework/aries/context.go
```diff
@@ -0,0 +1,22 @@
...
+	outboundTransport transport.OutboundTransport
```
troyronda: In a future PR, it would be nice to get rid of the logging singleton. We could instead pass it around in the Context.
@sudeshrshetty @rolsonquadras

pkg/framework/aries/context.go
rolsonquadras: @troyronda  @sudeshrshetty [updated code](https://github.com/hyperledger/aries-framework-go/compare/cf2367d04a6dbd7e0196b81fcf24ec5a1df59ae9..10f3e059cd136f4f1f0dd4d06539cd51a8b4f3e4) to follow fab-sdk structure.

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -68,6 +68,11 @@ type Connection struct {
 	DIDDoc *did.Doc `json:"did_doc,omitempty"`
 }
 
+// Provider for exchange protocol
```
troyronda: Suggest augmenting the comment with something like: Provider contains dependencies for the Exchange protocol and is typically created by using aries.Context().

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -86,29 +91,39 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
 	return encodedExchangeInvitation(invite)
 }
 
+// Client for exchange protocol
+type Client struct {
```
troyronda: I think recycling this pattern from the Fabric SDK is good (where each subfolder of Client contained a Client). 

I'm wondering if Client is the right word in this case (vs say Protocol or ...)? Thoughts?

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -86,29 +91,39 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
...
+}
+
+// New instanstiated new exchange client
+func New(prov Provider) *Client {
+	return &Client{prov: prov}
+}
+
 // SendExchangeRequest sends exchange request
-func SendExchangeRequest(exchangeRequest *Request, destination string, transport transport.OutboundTransport) error {
+func (c *Client) SendExchangeRequest(exchangeRequest *Request, destination string) error {
```
troyronda: As a followup, we should see if we should be using request contexts for Cancellables (I haven't thought too much about the implications of simply using the http client timeout duration yet).

pkg/framework/aries/api/factory.go
```diff
@@ -0,0 +1,16 @@
...
+
+package api
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/exchange"
+)
+
+// ProtocolProviderFactory allows overriding of aries protocol providers
+type ProtocolProviderFactory interface {
+	CreateExchangeProvider() exchange.Provider
```
troyronda: This is not the right abstraction. There cannot be a dependency back to "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/..."

troyronda: You should be creating generic factories/providers here - like Transports.

pkg/framework/aries/factory/protocol/protocol_factory.go
```diff
@@ -0,0 +1,27 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package protocol
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/exchange"
```
troyronda: ditto: https://github.com/hyperledger/aries-framework-go/pull/88/files#r313588529

pkg/framework/aries/framework.go
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package aries
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
```
troyronda: Is this due to a circular dependency?

pkg/framework/context/context.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package context
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/exchange"
```
troyronda: ditto: https://github.com/hyperledger/aries-framework-go/pull/88/files#r313588529

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -68,6 +68,11 @@ type Connection struct {
 	DIDDoc *did.Doc `json:"did_doc,omitempty"`
 }
 
+// Provider for exchange protocol
+type Provider interface {
```
troyronda: This should not be imported outside this package. It is the dependencies of this package.
Context should contain the needed methods that match up with the interface naming.

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/f4a9ffc938e46fb640514954eef882c1518e1871..463122c87b11e27a303f59b42906628162e71b73#diff-4d3ca323f9b237cfa9940c12efa9b615R71)

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/f4a9ffc938e46fb640514954eef882c1518e1871..463122c87b11e27a303f59b42906628162e71b73#diff-4d3ca323f9b237cfa9940c12efa9b615R72)

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: Makes sense to use Protocol. [Updated](https://github.com/hyperledger/aries-framework-go/compare/f4a9ffc938e46fb640514954eef882c1518e1871..463122c87b11e27a303f59b42906628162e71b73#diff-4d3ca323f9b237cfa9940c12efa9b615R95) the code.

pkg/framework/aries/api/factory.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/f4a9ffc938e46fb640514954eef882c1518e1871..463122c87b11e27a303f59b42906628162e71b73#diff-9988f5820bf05228e5657f2842fa071eR16) the the abstraction.

pkg/framework/aries/factory/protocol/protocol_factory.go
rolsonquadras: [changed the abstraction](https://github.com/hyperledger/aries-framework-go/pull/88#discussion_r313604198)

pkg/framework/context/context.go
rolsonquadras: @troyronda [changed the abstraction](https://github.com/hyperledger/aries-framework-go/pull/88#discussion_r313604198)

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: can we do this in separate issue/pr ?

pkg/framework/aries/framework.go
troyronda: By "this", I mean the sub-package "api" @rolsonquadras

pkg/framework/aries/api/factory.go
```diff
@@ -0,0 +1,16 @@
...
+
+package api
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+)
+
+// TransportProviderFactory allows overriding of aries protocol providers
+type TransportProviderFactory interface {
+	CreateOutboundTransportProvider() transport.OutboundTransport
```
troyronda: I'm a bit confused by this method name: `CreateOutboundTransportProvider`

Does this method create an OutboundTransport or a Factory/Provider of OutboundTransports?

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -86,29 +91,39 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
...
+}
+
+// New instanstiated new exchange client
+func New(prov provider) *Protocol {
+	return &Protocol{prov: prov}
+}
+
 // SendExchangeRequest sends exchange request
-func SendExchangeRequest(exchangeRequest *Request, destination string, transport transport.OutboundTransport) error {
+func (c *Protocol) SendExchangeRequest(exchangeRequest *Request, destination string) error {
```
troyronda: `c` as a variable name no longer matches with `Protocol`

pkg/framework/aries/framework.go
rolsonquadras: sorry @troyronda missed this comment.

yes, @sudeshrshetty mentioned the possibility of circular dependency (same is done in fabric-sdk-go as well).

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/463122c87b11e27a303f59b42906628162e71b73..a31807a201229cd5ab417413af54914995b98b87#diff-4d3ca323f9b237cfa9940c12efa9b615R105). missed during refactoring.

pkg/framework/aries/api/factory.go
rolsonquadras: suffix is not valid as it creates outbound transport. [Updated](https://github.com/hyperledger/aries-framework-go/compare/463122c87b11e27a303f59b42906628162e71b73..a31807a201229cd5ab417413af54914995b98b87#diff-cbf47775202e3791da3aefcade22c7b8R15).

pkg/didcomm/protocol/exchange/exchange.go
troyronda: Yes. Created https://github.com/hyperledger/aries-framework-go/issues/109

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -86,29 +91,39 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
 	return encodedExchangeInvitation(invite)
 }
 
+// Protocol for exchange protocol
+type Protocol struct {
+	prov transport.OutboundTransport
```
Baha-sk: prov is not meaningful here, outTransport makes more sense, see comment below

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -123,10 +138,10 @@ func encodedExchangeInvitation(inviteMessage *Invitation) (string, error) {
...
 }
 
-func marshalAndSend(data interface{}, errorMsg, destination string, transport transport.OutboundTransport) (string, error) {
+func (p *Protocol) marshalAndSend(data interface{}, errorMsg, destination string) (string, error) {
 	jsonString, err := json.Marshal(data)
 	if err != nil {
 		return "", errors.Errorf("%s : %w", errorMsg, err)
 	}
-	return transport.Send(string(jsonString), destination)
+	return p.prov.Send(string(jsonString), destination)
```
Baha-sk: it makes more sense to have p.outTransport.Send() instead of p.prov.Send() since the struct acting on Send() is a transport, not a provider

pkg/framework/aries/framework_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	req := &exchange.Request{
+		ID:    "5678876542345",
+		Label: "Bob",
+	}
+	require.NoError(t, exClient.SendExchangeRequest(req, "http://example/didexchange"))
+	require.Error(t, exClient.SendExchangeRequest(req, ""))
+}
+
+type mockTransportProviderFactory struct {
```
sudeshrshetty: can we move all mocks to common pkg if possible?

pkg/framework/aries/framework.go
```diff
@@ -0,0 +1,52 @@
...
+	transport api.TransportProviderFactory
+}
+
+// Option configures the framework.
+type Option func(opts *Aries) error
+
+// New initializes the Aries framework based on the set of options provided.
+func New(opts ...Option) (*Aries, error) {
+	// get the default framework options
+	defOpts := defFrameworkOpts()
```
sudeshrshetty: Just a thought.
Not sure about having default factory settings as opts, if someone passes custom opts then we may have duplicate opts. It may work here in this case due to order of opts. Need proper test cases to see if overriding actually works (I can see in one of the test you passed mock transport factory), just to be at safe side we shouldn't pass default settings as opts. 

Instead at line number 28, initialize Aries struct with default opts.

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -86,29 +91,39 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
 	return encodedExchangeInvitation(invite)
 }
 
+// Protocol for exchange protocol
+type Protocol struct {
+	prov transport.OutboundTransport
+}
+
+// New instanstiated new exchange client
+func New(prov provider) *Protocol {
```
sudeshrshetty: May be we need a public interface here, if in case someone wants to use 'exchange.Protocol' in standalone fashion (without using aries.New(...) )?  In future if 'exchange.Provider' interface grows? @troyronda @Baha-sk

pkg/framework/aries/framework.go
```diff
@@ -0,0 +1,52 @@
...
+
+	frameworkOpts := &Aries{}
+
+	// generate framework configs from options
+	for _, option := range append(defOpts, opts...) {
```
Baha-sk: defOpts has default and append opts to it may duplicate options, this can lead to unexpected behaviour

pkg/framework/aries/framework.go
rolsonquadras: This was done based on [comments](https://github.com/hyperledger/aries-framework-go/pull/88#discussion_r312992096).

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: This was changed based on [comments](https://github.com/hyperledger/aries-framework-go/pull/88#discussion_r313590180).

pkg/framework/aries/framework.go
sudeshrshetty: I will mark it as resolved

pkg/framework/aries/framework.go
rolsonquadras: The second opts(user opts) will overwrite the defOpts.

pkg/framework/aries/framework_test.go
rolsonquadras: @troyronda @Baha-sk @fqutishat thoughts?

pkg/didcomm/protocol/exchange/exchange.go
sudeshrshetty: Add some notes/comments if we are continuing with private interface.

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/a31807a201229cd5ab417413af54914995b98b87..a26c159c381cbbb380ef97f616c8f9b9bddd476a#diff-4d3ca323f9b237cfa9940c12efa9b615R100).

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/a31807a201229cd5ab417413af54914995b98b87..a26c159c381cbbb380ef97f616c8f9b9bddd476a#diff-4d3ca323f9b237cfa9940c12efa9b615R96).

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/a31807a201229cd5ab417413af54914995b98b87..a26c159c381cbbb380ef97f616c8f9b9bddd476a#diff-4d3ca323f9b237cfa9940c12efa9b615R148).

pkg/framework/aries/framework_test.go
troyronda: What is meant by “all mocks”? I generally think mocks should be near its real implementation.

pkg/framework/aries/framework.go
troyronda: Yes order of opts needs to be kept intact.

pkg/framework/aries/framework.go
troyronda: Yes - second opt can override.

The other way around this type of problem seems to be to force framework users to pass the default option, rather than implicitly include as the first option. This has its own negative trade-off, of course.

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -86,29 +91,41 @@ func GenerateInviteWithKeyAndEndpoint(invite *Invitation) (string, error) {
 	return encodedExchangeInvitation(invite)
 }
 
+// Protocol for exchange protocol
+type Protocol struct {
+	transport.OutboundTransport
```
troyronda: This should be a named field. The Protocol struct should not be exposing everything that is in its members.

pkg/didcomm/protocol/exchange/exchange.go
```diff
@@ -68,6 +68,11 @@ type Connection struct {
 	DIDDoc *did.Doc `json:"did_doc,omitempty"`
 }
 
+// provider contains dependencies for the Exchange protocol and is typically created by using aries.Context()
+type provider interface {
+	transport.OutboundTransport
```
troyronda: This should be a named field. We do not want to worry about name conflicts underneath.

pkg/framework/aries/framework.go
troyronda: Perhaps we could also introduce a function argument like WithoutDefaults()

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/a26c159c381cbbb380ef97f616c8f9b9bddd476a..2002cc47490c6dea1ede21a12a1f5ebee834f700#diff-4d3ca323f9b237cfa9940c12efa9b615R73).

pkg/didcomm/protocol/exchange/exchange.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/a26c159c381cbbb380ef97f616c8f9b9bddd476a..2002cc47490c6dea1ede21a12a1f5ebee834f700#diff-4d3ca323f9b237cfa9940c12efa9b615R96).

troyronda:
BTW: I am wondering if we should have this package directly under pkg (since it is effectively the default entry point for developers). The argument for being under framework is also good - i.e., it is a framework :). Something to think about anyways.

pkg/framework/aries/context.go
troyronda: Created #115.

Issue #37, opened 2019-07-25
Create a new Package to support Chacha20/Poly1035 encryptions/authentication (in JWE format)


Baha-sk:
Standard JWE encryption/decryption operations does not support Chacha20 algorithm with Poly1035 authentication.

There are talks to update the JWE spec with this new addition. An IETF draft can be found here:
https://datatracker.ietf.org/doc/draft-amringer-jose-chacha/

The aim of this issue is to create a new package that will support this new JWE spec.

finally, for reference, Aries-rfc has an issue to update its spec using this new format:
https://github.com/hyperledger/aries-rfcs/issues/133

A PHP example is found here:
https://github.com/gamringer/php-authcrypt

the task here is to follow the same format in Go

Part of #36

troyronda:
How do we plan to incorporate the modifications of go-jose into this project? Is a fork required or are we able to wrap the go-jose library and contain the modifications into this repository?

Baha-sk:
> How do we plan to incorporate the modifications of go-jose into this project? Is a fork required or are we able to wrap the go-jose library and contain the modifications into this repository?

The changes in Go-Jose are intertwined within the framework, changes are done on a fork (and we can potentially submit a PR to merge the change but we will need the IETF draft to be come part of the JWE standard to support the merge)

Baha-sk:
New update to this issue, forking Go-Jose is not suitable, we will instead build our own package that encrypts/decrypts JWE formatted messages using Chacha20/Poly1035 instead.

Issue #96, opened 2019-08-13
Add support for Ed25519VerificationKey2018


llorllale:
Add support for the `Ed25519VerificationKey2018` signature suite such that `publicKey` entries can be added to the DID documents and the owners of those DIDs can prove ownership to a Verifier.

This is the first in a series of cryptographic suites that will need to be implemented by the framework in order to be interoperable.

Links:
* [Ed25519 Signature 2018](https://w3c-dvcg.github.io/lds-ed25519-2018/)
* [Linked Data Signatures](https://w3c-dvcg.github.io/ld-signatures/)
* [Linked Data Cryptographic Suite Registry](https://w3c-ccg.github.io/ld-cryptosuite-registry)

llorllale:
Part of epic #35

dmitrizagidulin:
This is great! 
(Also, the LD Crypto Suite related specs are currently being updated / refreshed, so please feel free to open issues and ask questions if something is unclear or looks out of date.)

PR #107, opened 2019-08-13
feat: Authcrypt Encrypt Using (X)Chach20Poly1035

Baha-sk:
This change adds support to encrypt agent's payloads
	for the Pack() call at the transport layer

**Title:**
Authcrypt Encrypt Using (X)Chach20Poly1035

**Description:**
This is part of #37 and allows encrypting payloads at the transport layer of DidComm Agent

Summary:

This PR adds support to encrypt payloads using (X)Chacha20Poly35 IETF authenticated encryption.

Closes: #150


pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+
+	c := &Crypter{
+		sender,
+		recipients,
+		alg,
+		nonceSize,
+	}
+
+	typeofS := reflect.TypeOf(sender)
+	if typeofS != reflect.TypeOf(&ecdsa.PrivateKey{}) {
```
llorllale: We'll need support for `ed25519.PrivateKey` as well, see #96

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+
+	return c, nil
+}
+
+// Encrypt will JWE encode the payload argument for the sender and recipients
+// Using (X)Chacha20Poly1035 encryption & authentication (Authcrypt)
+func (c *Crypter) Encrypt(payload string) (string, error) {
+	headers := jweHeaders{
+		"typ": "prs.hyperledger.aries-auth-message",
+		"alg": "ECDH-SS+XC20PKW",
```
llorllale: This isn't the `alg` I see defined [in the RFC](https://tools.ietf.org/html/draft-amringer-jose-chacha-01#section-3)

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+type Crypter struct {
+	Sender     interface{}
+	Recipients []interface{}
+	Alg        string
+	NonceSize  int
+}
+
+// JweEnvelope represents a JWE envelope as per the Aries Encryption envelop specs
+type JweEnvelope struct {
+	Protected  jweHeaders  `json:"Protected,omitempty"`
```
llorllale: Lower case "protected"

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	apuEncoded := make([]byte, base64.URLEncoding.EncodedLen(len(apu)))
+	base64.URLEncoding.Encode(apuEncoded, apu)
+
+	// create authentication key for symmetric shared key
+	kek := josecipher.DeriveECDHES(c.Alg, apuEncoded, nil, senderEphemeralKey, recipientKey, chacha20poly1305.KeySize)
+
+	var cipher cipher.AEAD
+	switch c.NonceSize {
```
llorllale: This `switch` appears twice. Can probably be extracted into a function.

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+
+	tag := kekOutput[len(kekOutput)-poly1305.TagSize:]
+	symKeyCipher := kekOutput[0 : len(kekOutput)-poly1305.TagSize]
+
+	return c.buildRecipient(symKeyCipher, apu, nonce, tag, recipientKey)
+}
+
+func (c *Crypter) buildRecipient(symKeyCipher []byte, apu []byte, nonce []byte, tag []byte, recipientKey *ecdsa.PublicKey) (*Recipient, error) {
+
+	/*senderKey := c.Sender.(*ecdsa.PrivateKey).PublicKey
```
llorllale: Delete all this commented-out code

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, err
+	}
+
+	oid := cipher.Seal(nil, nonce,
+		[]byte(base64.URLEncoding.EncodeToString(senderKeyBytes)), nil) // TODO double check oid is encrypted correctly (using anoncrypt seal)
+	*/
+	senderKey := c.Sender.(*ecdsa.PrivateKey).PublicKey
+	oid := elliptic.Marshal(senderKey, senderKey.X, senderKey.Y)
+	recipientHeaders := RecipientHeaders{
+		base64.URLEncoding.EncodeToString(apu),
```
llorllale: Let's make explicit the attributes to which we're assigning all these values

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+		base64.URLEncoding.EncodeToString(nonce),
+		base64.URLEncoding.EncodeToString(tag),
+		base58.Encode(elliptic.Marshal(recipientKey, recipientKey.X, recipientKey.Y)),
+		base64.URLEncoding.EncodeToString(oid),
+	}
+
+	recipient := &Recipient{
+		base64.URLEncoding.EncodeToString(symKeyCipher),
```
llorllale: Same here (explicit attribute assignment)

troyronda:
Test coverage is a bit low.

pkg/didcomm/crypto/authcrypt/encrypt_test.go
```diff
@@ -0,0 +1,71 @@
...
+	recipient2Key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+
+	recipient3Key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+
+	_, err = NewAuthCrypter(sendEcKey, []interface{}{&recipient1Key.PublicKey, &recipient2Key.PublicKey, &recipient3Key.PublicKey}, "BAD")
+	require.Error(t, err)
+	badKey := ""
+	_, err = NewAuthCrypter(badKey, []interface{}{&recipient1Key.PublicKey, &recipient2Key.PublicKey, &recipient3Key.PublicKey}, "XC20P")
```
llorllale: Let's break these test cases into separate tests such that the focus of the test becomes clearer. Right now it's rather hard to see what exactly is being tested in each assertion.

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: according to @gamringer, this encryption is static-static

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: good catch

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: oid value is temporary and will need to be changed.. this comment is temporary till the value is properly encrypted.

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: yes, eventually, not sure if we can derive ECDH keys from ed25519 keys though.. this needs to be investigated ..

pkg/didcomm/crypto/authcrypt/encrypt_test.go
Baha-sk: will separate them, or add comment for each test case, probably not worth creating a test function for each test case

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: > Test coverage is a bit low.

the remaining uncovered lines are related to JSON marshal errors, or derive keys.. I will try to increase the coverage a bit.

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,288 @@
...
+		cipher, err = chacha20poly1305.NewX(symKey)
+	}
+	if err != nil {
+		return "", err
+	}
+
+	pld := []byte(payload)
+
+	// encrypt payload
+	symOutput := cipher.Seal(nil, nonce, pld, []byte(pldAad))
```
llorllale: Missing the `dst` arg for `Seal()`

pkg/didcomm/crypto/authcrypt/encrypt_test.go
Baha-sk: added comments for each test case

llorllale: If we're going to stick all scenarios into a single method then I'd prefer we have them like in [this example](https://github.com/hyperledger/aries-framework-go/blob/4380aaed9d8282b324bb9d69a904b0888ab5d28b/pkg/framework/didresolver/didresolver_test.go#L66), which has several virtues:

* aesthetics: much cleaner separation of test cases
* each anonymous function declares its own variables which reside in the anonymous scope. This avoids coupling between tests

Baha-sk: will update..

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: found solution for encryption oid: `sodium_crypto_box_seal` is actually nacl's `box.Seal()` (using ed25519 + Salsa20 + Poly1035) so will replace oid to call the x/crypto/nacl package instead..

pkg/didcomm/crypto/authcrypt/encrypt_test.go
Baha-sk: done

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: removed

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: alg as per the wip draft discussed with @gamringer:
```
Baha-sk
12:23 PM
another question.. is this right ECDH-SS+XC20PKW or you meant ECDH-ES+XC20PKW in the headers?
gamringer
1:26 PM
I did mean ECDH-SS+XC20PKW, because with ECDH-ES*, you have to put the whole ephemeral public key in the header, whereas the requirement for authcrypt is to have only a key identifier
```
so the draft for this JWE will use SS in the algorithm and should be updated.

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: for now, we're supporting keys used by Chacha20Poly1035 and Salsa20, these are 32 byte keys.

pkg/didcomm/crypto/authcrypt/encrypt.go
```diff
@@ -0,0 +1,308 @@
...
+	Tag string `json:"tag,omitempty"`
+	Kid string `json:"kid,omitempty"`
+	Oid string `json:"oid,omitempty"`
+}
+
+// NewAuthCrypter will create an encrypter instance to 'AuthCrypt' payloads for the given sender and recipients arguments
+// and the encryption alg argument. Possible algorithms supported are:
+// C20P (chacha20-poly1035 ietf)
+// XC20P (xchacha20-poly1035 ietf)
+func NewAuthCrypter(sender keyPair, recipients []keyPair, alg string) (*Crypter, error) {
```
llorllale: `recipients` shouldn't be key pairs

pkg/didcomm/crypto/authcrypt/encrypt_test.go
```diff
@@ -0,0 +1,103 @@
...
+		require.NotEmpty(t, crypter)
+		enc, e := crypter.Encrypt("lorem ipsum dolor sit amet")
+		require.NoError(t, e)
+		require.NotEmpty(t, enc)
+
+		m, e := prettyPrint(enc)
+		require.NoError(t, e)
+		t.Logf("Encryption with C20P: %s", m)
+
+		t.Run("Error test Case: use a valid AuthCrypter but scramble the nonce size", func(t *testing.T) {
```
llorllale: This test case should be a sibling of the parent

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: sure, they can be a list of `[]byte`

pkg/didcomm/crypto/authcrypt/encrypt_test.go
Baha-sk: it's re-using the crypter

pkg/didcomm/crypto/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/authcrypt/encrypt_test.go
Baha-sk: I've unexpected all crypter's fields, so nonce testing is internal and should not be exposed outside the library. It should be fine to re-use the crypter for 'internally' testing the nonceSize value.

troyronda:
@Baha-sk Does this close #37 now?

pkg/didcomm/crypto/cypter.go
```diff
@@ -0,0 +1,22 @@
+/*
```
Moopli: filename typo, `crypter.go`

Baha-sk: done

doc.go
```diff
@@ -0,0 +1,8 @@
+/*
```
troyronda: please rebase / squash - should only have 1 commit in the PR.

pkg/didcomm/crypto/crypter.go
```diff
@@ -0,0 +1,22 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
+
+// Crypter is an Aries envelop encrypter to support
+// secure DIDComm exchange of envelops between Aries agents
```
troyronda: typo: envelops

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
```diff
@@ -0,0 +1,40 @@
...
+// decryptOID will decrypt a recipient's encrypted OID (in the case of this package, it is represented as
+// ephemeral key concatenated with the sender's public key) using the recipient's privKey/pubKey keypair,
+// this is equivalent to libsodium's C function: crypto_box_seal()
+// https://libsodium.gitbook.io/doc/public-key_cryptography/sealed_boxes#usage
+func decryptOID(privKey, pubKey *[chacha20poly1305.KeySize]byte, encrypted []byte) []byte {
+	var epk [32]byte
+	var nonce [24]byte
+	copy(epk[:], encrypted[:chacha20poly1305.KeySize])
+
+	nonceWriter, _ := blake2b.New(24, nil)
```
troyronda: error is being ignored here.

I have seen this in several places in the PR (mostly without comments).

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"golang.org/x/crypto/blake2b"
+	"golang.org/x/crypto/chacha20poly1305"
+	"golang.org/x/crypto/nacl/box"
+)
+
+// Decrypt will JWE decode the envelop argument for the sender and recipients
+// Using (X)Chacha20 encryption algorithm and Poly1035 authenticator
+func (c *Crypter) Decrypt(envelope []byte) (string, error) {
+	decryptOID(nil, nil, nil)
```
troyronda: I assume this is a TODO? If so, please add code comments and create issue.

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
```diff
@@ -0,0 +1,40 @@
...
+	return "", nil
+}
+
+// decryptOID will decrypt a recipient's encrypted OID (in the case of this package, it is represented as
+// ephemeral key concatenated with the sender's public key) using the recipient's privKey/pubKey keypair,
+// this is equivalent to libsodium's C function: crypto_box_seal()
```
troyronda: It would be good to have some test cases that demonstrate this equivalence.

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,368 @@
...
+
+// Crypter represents an Authcrypt Encrypter (Decrypter) that outputs/reads JWE envelopes
+type Crypter struct {
+	sender     keyPair
+	recipients []*[chacha20poly1305.KeySize]byte
+	alg        string
+	nonceSize  int
+}
+
+// Envelope represents a JWE envelope as per the Aries Encryption envelop specs
```
troyronda: typo: envelop

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,368 @@
...
+	Aad        string      `json:"aad,omitempty"`
+	Iv         string      `json:"iv,omitempty"`
+	Tag        string      `json:"tag,omitempty"`
+	CipherText string      `json:"ciphertext,omitempty"`
+}
+
+// jweHeaders are the Protected JWE headers in a map format
+type jweHeaders map[string]string
+
+// Recipient is a recipient of an envelop including the shared encryption key
```
troyronda: typo: envelop

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,368 @@
...
+type Recipient struct {
+	EncryptedKey string           `json:"encrypted_key,omitempty"`
+	Header       RecipientHeaders `json:"header,omitempty"`
+}
+
+// RecipientHeaders are the recipient headers
+type RecipientHeaders struct {
```
troyronda: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,368 @@
...
+	Kid string `json:"kid,omitempty"`
+	Oid string `json:"oid,omitempty"`
+}
+
+// New will create an encrypter instance to 'AuthCrypt' payloads for the given sender and recipients arguments
+// and the encryption alg argument. Possible algorithms supported are:
+// C20P (chacha20-poly1035 ietf)
+// XC20P (xchacha20-poly1035 ietf)
+// The returned crypter contains all the information required to encrypt payloads.
+func New(sender keyPair, recipients []*[chacha20poly1305.KeySize]byte, alg string) (*Crypter, error) {
```
troyronda: Recommending creating a type for alg for better documentation.

(also to conform better with go-jose: https://github.com/square/go-jose/blob/8bad6148bd0a8da57f9840693f6f611ac4483d09/shared.go#L36)

pkg/didcomm/crypto/jwe/authcrypt/encrypt_test.go
```diff
@@ -0,0 +1,104 @@
+/*
```
troyronda: We need test cases.

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,368 @@
...
+
+	if !isKeyPairValid(sender) {
+		return nil, errors.New(fmt.Sprintf("sender keyPair not supported, it must have %d bytes keys", chacha20poly1305.KeySize))
+	}
+
+	return c, nil
+}
+
+func isKeyPairValid(kp keyPair) bool {
+	if kp.priv == nil || kp.pub == nil || len(kp.priv) != chacha20poly1305.KeySize || len(kp.pub) != chacha20poly1305.KeySize {
```
troyronda: It doesn't seem like this needs to be such a long line.
https://github.com/golang/go/wiki/CodeReviewComments#line-length

pkg/didcomm/crypto/jwe/authcrypt/encrypt_test.go
Baha-sk: I do have a test case that generates a JWE using the PHP example keys, but we can't compare the cipher output as it's based on cryptographically random keys/nonce generation

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: should I reference go-jose's type or create a new internal type?

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: rename to RecipientHEADERS?

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
troyronda: I had thought a new type.

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,373 @@
...
+
+	return c, nil
+}
+
+func isKeyPairValid(kp keyPair) bool {
+	if kp.priv == nil || kp.pub == nil {
+		return false
+	}
+
+	if len(kp.priv) != chacha20poly1305.KeySize ||
```
Moopli: Dead code, given current keyPair definition

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
troyronda: No. I was referring to some of the fields.

troyronda: Also you should check Envelope fields.

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: true, will remove

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: reduced... will remove the len() check as it's not possible (I thought the linter will catch it)

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: I was thinking to reference go-jose as they already have a type for this.

I'll create a new type then..

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: will update ..

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
troyronda: I had meant break it into multiple lines, but dead code removal is even better ;).

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
Baha-sk: will add a todo to add more tests to ensure interoperability between libsodium and and box.Seal

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
Baha-sk: yes will add a TODO

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
Baha-sk: will add error handling..

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done :)

pkg/didcomm/crypto/jwe/authcrypt/encrypt_test.go
Baha-sk: pushed a patch with a comment... will definitely need to test interoperability with other libraries (like the reference php one: https://github.com/gamringer/php-authcrypt)

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done..

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
Baha-sk: done

pkg/didcomm/crypto/crypter.go
Baha-sk: will update...

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
llorllale: I'm also a bit curious about the equivalence between a "decrypt" function and `crypto_box_seal`

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,392 @@
...
+type ContentEncryption string
+
+// C20P Chacha20Poly1035 algorithm
+const C20P = ContentEncryption("C20P") // Chacha20 encryption + Poly1035 authenticator cipher (96 bits nonce)
+
+// XC20P XChacha20Poly1035 algorithm
+const XC20P = ContentEncryption("XC20P") // XChacha20 encryption + Poly1035 authenticator cipher (192 bits nonce)
+
+// RandReader is a cryptographically secure random number generator.
+var RandReader = rand.Reader
```
llorllale: suggest we un-export this

pkg/didcomm/crypto/crypter.go
```diff
@@ -0,0 +1,23 @@
...
+package crypto
+
+// Crypter is an Aries envelope encrypter to support
+// secure DIDComm exchange of envelopes between Aries agents
+type Crypter interface {
+	// Encrypt a payload in an Aries compliant format
+	// returns:
+	// 		[]byte containing the encrypted envelope
+	//		error if encryption failed
+	Encrypt(payload string) ([]byte, error)
```
troyronda: I think payload is better represented as []byte.

See https://github.com/square/go-jose/blob/8bad6148bd0a8da57f9840693f6f611ac4483d09/crypter.go#L31

pkg/didcomm/crypto/crypter.go
```diff
@@ -0,0 +1,23 @@
...
+	// Decrypt an envelope in an Aries compliant format
+	// returns:
+	// 		string containing the decrypted payload
+	//		error if encryption failed
+	// TODO add recipient's private key as an argument of Decrypt as it's needed for decrypting the message
+	Decrypt(envelope []byte) (string, error)
```
troyronda: I think the return value is better represented as []byte.

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,392 @@
...
+	_, err = RandReader.Read(nonce)
+	if err != nil {
+		return nil, err
+	}
+	nonceEncoded := base64.URLEncoding.EncodeToString(nonce)
+
+	var cek *[chacha20poly1305.KeySize]byte
+
+	// generate a cek for encryption (it will be treated as a symmetric key)
+	_, cek, err = box.GenerateKey(RandReader)
```
llorllale: curious choice. Why not use `RandReader` directly?

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"errors"
+
+	"golang.org/x/crypto/blake2b"
+	"golang.org/x/crypto/chacha20poly1305"
+	"golang.org/x/crypto/nacl/box"
+)
+
+// Decrypt will JWE decode the envelop argument for the sender and recipients
```
troyronda: Typo: envelop

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
troyronda: This remains fishy to me - you should be able to generate a symmetric key just from reading from secure rand. Using an asymmetric generator is slower and confusing to the reader. Also how do you know the key is the right size?

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil, err
+	}
+
+	// create a cipher for the given nonceSize and generated cek above
+	cipher, err := createCipher(c.nonceSize, cek[:])
+	if err != nil {
+		return nil, err
+	}
+
+	pld := []byte(payload)
```
troyronda: Just take a []byte argument.

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,392 @@
...
+		Header:       recipientHeaders,
+	}
+
+	return recipient, nil
+}
+
+// generateRecipientCek will generate an ephemeral symmetric key for the recipientKey to
+// be used for encrypting the cek.
+// it will return this new key along with the corresponding APU or an error if it fails.
+func (c *Crypter) generateRecipientCek(apu []byte, recipientKey *[chacha20poly1305.KeySize]byte) ([]byte, error) {
```
troyronda: generateRecipientCEK

troyronda: This function will need further review and more tests. It is hard to follow. Assume this is coming in a followup.

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
troyronda: Also suggest using chacha20poly1305.KeySize constant when you generate. This makes it more obvious :).

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil, err
+	}
+
+	pld := []byte(payload)
+
+	// encrypt payload using generated nonce, payload and its AAD
+	// the output is a []byte containing the cipherText + tag
+	symOutput := cipher.Seal(nil, nonce, pld, []byte(pldAad))
+
+	// symOutput has a length of len(pld) + poly1035.TagSize
```
troyronda: I'd personally move this tag manipulation stuff to another method.
(to make things more obvious and less cluttered).

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: sure I'll un-export this

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: yes.. .refactoring this..

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: converting payload type..

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
troyronda: Similar to https://github.com/hyperledger/aries-framework-go/pull/107#discussion_r316849605

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: moved both tag and ciphertext into separate functions as they're reused for recipient keys as well

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: yes, this is following the php example instructions, interop testing is required to ensure this works with libraries in other packages.

I will rename the function to CEK

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done... it's not an asymmetric key.. removed this old logic

Baha-sk: replaced

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/didcomm/crypto/crypter.go
Baha-sk: will update

pkg/didcomm/crypto/crypter.go
Baha-sk: will update

pkg/didcomm/crypto/jwe/authcrypt/decrypt.go
Baha-sk: will update

pkg/didcomm/crypto/crypter.go
Baha-sk: I think I've already updated this to []byte

PR #108, opened 2019-08-14
feat: Allow DID method to Accept supported DID prefixes

### LINKED ISSUE #101: Allow DID method implementations to Accept supported DID prefixes.
- Each DID implementation exposes an Accept method that determines if the DID string prefix is supported by that implementation.
  - Example: the implementation can allow for simple string comparison, reg. exp. (e.g., wildcard) and catch-all.
- Due to flexible Accept pattern, execution needs to be ordered.

epic #86

fqutishat:
closes #101

### DIFF
--- a/pkg/didmethod/peer/resolver.go
+++ b/pkg/didmethod/peer/resolver.go
@@ -36,3 +36,8 @@ func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime str
 
 	return jsonDoc, nil
 }
+
+// Accept did method
+func (resl *DIDResolver) Accept(method string) bool {
+	return "peer" == method
+}
--- a/pkg/framework/didresolver/api.go
+++ b/pkg/framework/didresolver/api.go
@@ -5,7 +5,9 @@ SPDX-License-Identifier: Apache-2.0
 
 package didresolver
 
-import "time"
+import (
+	"time"
+)
 
 // ResultType input option can be used to request a certain type of result.
 type ResultType int
@@ -20,6 +22,7 @@ const (
 // DidMethod operations
 type DidMethod interface {
 	Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error)
+	Accept(method string) bool
 }
 
 // resolveOpts holds the options for did resolve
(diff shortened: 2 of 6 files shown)


pkg/framework/didresolver/api.go
```diff
@@ -72,6 +72,6 @@ type Opt func(opts *didResolverOpts)
 // WithDidMethod to add did method
```
troyronda: Should update the package docs to indicate that regular expressions are supported and DID methods are checked in the order added.

fqutishat: I add comment

pkg/framework/didresolver/didresolver.go
```diff
@@ -73,3 +81,14 @@ func (r *DIDResolver) Resolve(did string, opts ...ResolveOpt) (*diddoc.Doc, erro
 
 	return didDoc, nil
 }
+
+// resolveDidMethod resolve did method
+func (r *DIDResolver) resolveDidMethod(method string) (DidMethod, error) {
+	for _, v := range r.didMethods {
+		match, _ := regexp.MatchString(v.id, method)
```
troyronda: The regular expression should be pre-compiled rather than compiled on demand during resolve.

troyronda:
It would be good to add some description to the commit message so that the reader of the git logs has a better idea of the feature.

pkg/framework/didresolver/api.go
```diff
@@ -72,6 +72,6 @@ type Opt func(opts *didResolverOpts)
...
 func WithDidMethod(id string, method DidMethod) Opt {
 	return func(opts *didResolverOpts) {
-		opts.didMethods[id] = method
+		opts.didMethods = append(opts.didMethods, didMethod{id: id, DidMethod: method})
```
llorllale: This makes using `WithDidMethod` awkward. The user now must specify a regex instead of just a DID method. Eg. the regex "rie" matches the string "aries" yet they clearly are not the same strings. To lock down the definition, the user is forced to provide "^rie$", which is counter intuitive.

One way to enable "catch-alls" is to invert control of the pattern matching such that the `DidMethod` is responsible for it. Add an `Accept(method string) bool` method to `DidMethod`, thereby freeing the user from such awkwardness and, incidentally, simplifying `WithDidMethod` by removing the `id` arg:

```go
resolver := didresolver.New(
    WithDidMethod(peer.NewDIDResolver(store)),
    WithDidMethod(universal.Resolver(url)),
    WithDidMethod(sidetree.Resolver(method, url)),
)
```

pkg/framework/didresolver/api.go
fqutishat: done

pkg/framework/didresolver/didresolver.go
fqutishat: done

pkg/framework/didresolver/api.go
fqutishat: I agree

fqutishat: done

pkg/didmethod/peer/resolver.go
```diff
@@ -12,12 +12,13 @@ import (
 
 // DIDResolver resolver
 type DIDResolver struct {
-	store *DIDStore
+	store  *DIDStore
+	method string
```
llorllale: In the case of peer DIDs the method is always "peer", so this can be hard coded to `peer.DIDResolver.Accept()`

fqutishat: done

Issue #99, opened 2019-08-13
Create DID Resolver HTTP Binding (from W3C DID resolver spec) - AKA Universal Resolver


fqutishat:
Write universal resolver that use http client to resolve did from specific URL.

epic #86

troyronda:
More generally we should support:
> DID Resolution functionality over the [HTTP(S) binding](https://w3c-ccg.github.io/did-resolution/#bindings-https)

Rather than calling this specific to the "Universal Resolver". I will update the title.

Issue #87, opened 2019-08-12
ci: add project-specific initialisms


troyronda:
We should configure the linter to additionally include our project-specific [Initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) (if possible). For example: DID.

PR #117, opened 2019-08-15
feat: Universal HTTP DID Resolver

kdimak:
```
This change adds support of Universal  HTTP DID Resolver
```

**Title:**
Universal  HTTP DID Resolver

**Description:**
This is a part of #99 and allows for Universal HTTP DID Resolver.

**Summary:**
This PR adds a definition of Universal  HTTP(s) DID Resolver which is based on standard `http.Client` HTTP client. It's used for making HTTP GET request and reading the DID document from the reply.

### DIFF
--- a/pkg/didmethod/httpbinding/resolver.go
+++ b/pkg/didmethod/httpbinding/resolver.go
@@ -1,0 +1,121 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package httpbinding
+
+import (
+	"crypto/tls"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	errors "golang.org/x/xerrors"
+)
+
+var logger = log.New("aries-framework/didmethod/httpbinding")
+
+// resolverOpts holds options for the DID Resolver
+// it has a http.Client instance initialized with default parameters
+type resolverOpts struct {
+	client *http.Client
+}
+
+// ResolverOpt is the DID Resolver option
+type ResolverOpt func(opts *resolverOpts)
+
+// WithTimeout option is for definition of HTTP(s) timeout value of DID Resolver
+func WithTimeout(timeout time.Duration) ResolverOpt {
+	return func(opts *resolverOpts) {
+		opts.client.Timeout = timeout
+	}
+}
+
+// WithTLSConfig option is for definition of secured HTTP transport using a tls.Config instance
+func WithTLSConfig(tlsConfig *tls.Config) ResolverOpt {
+	return func(opts *resolverOpts) {
+		opts.client.Transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+	}
+}
+
+// resolveDID makes DID resolution via HTTP
+func (res *DIDResolver) resolveDID(url string) ([]byte, error) {
+	resp, err := res.client.Get(url)
+	if err != nil {
+		return nil, errors.Errorf("HTTP Get request failed: %w", err)
+	}
+
+	defer func() {
+		e := resp.Body.Close()
+		if e != nil {
+			logger.Errorf("Failed to close response body: %v", e)
+		}
+	}()
+
+	// TODO support for service endpoint URL resolution
+	if containsDIDDocument(resp) {
+		var gotBody []byte
+		gotBody, err = ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, errors.Errorf("Failed to read response body: %w", err)
+		}
+
+		return gotBody, nil
+
+	} else if notExistentDID(resp) {
+		return nil, errors.Errorf("Input DID does not exist: %w", err)
+	}
+
+	return nil, errors.Errorf("Unsupported response from DID Resolver with status code: %v", resp.StatusCode)
+}
+
+// notExistentDID checks if requested DID is not found on remote DID resolver
+func notExistentDID(resp *http.Response) bool {
+	return resp.StatusCode == http.StatusNotFound
+}
+
+// containsDIDDocument checks weather reply from remote DID resolver contains DID document
+func containsDIDDocument(resp *http.Response) bool {
+	return resp.StatusCode == http.StatusOK && resp.Header.Get("Content-type") == "application/did+ld+json"
+}
+
+// New creates new DID Resolver
+func New(endpointURL string, opts ...ResolverOpt) (*DIDResolver, error) {
+	// Apply options
+	clOpts := &resolverOpts{client: &http.Client{}}
+	for _, opt := range opts {
+		opt(clOpts)
+	}
+
+	// Validate host
+	_, err := url.ParseRequestURI(endpointURL)
+	if err != nil {
+		return nil, errors.Errorf("Invalid base url: %w", err)
+	}
+
+	return &DIDResolver{
+		endpointURL: endpointURL,
+		client:      clOpts.client,
+	}, nil
+}
+
+// Read implements didresolver.DidMethod.Read interface (https://w3c-ccg.github.io/did-resolution/#resolving-input)
+func (res *DIDResolver) Read(DID string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+	reqURL, _ := url.ParseRequestURI(res.endpointURL)
+
+	reqURL.Path = path.Join(reqURL.Path, DID)
+
+	return res.resolveDID(reqURL.String())
+}
+
+// DIDResolver DID Resolver via HTTP(s) endpoint
+type DIDResolver struct {
+	endpointURL string
+	client      *http.Client
+}
--- a/pkg/didmethod/httpbinding/resolver_test.go
+++ b/pkg/didmethod/httpbinding/resolver_test.go
@@ -1,0 +1,138 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+package httpbinding
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithOutboundOpts(t *testing.T) {
+	opt := WithTimeout(1 * time.Second)
+	require.NotNil(t, opt)
+	clOpts := &resolverOpts{}
+	// opt.client is nil, so setting timeout should panic
+	require.Panics(t, func() { opt(clOpts) })
+
+	opt = WithTLSConfig(nil)
+	require.NotNil(t, opt)
+	clOpts = &resolverOpts{}
+	// opt.client is nil, so setting TLS config should panic
+	require.Panics(t, func() { opt(clOpts) })
+}
+
+func TestNew(t *testing.T) {
+	var err error
+
+	// OK with no options
+	_, err = New("https://uniresolver.io/")
+	require.NoError(t, err)
+
+	// All options are applied
+	i := 0
+	_, err = New("https://uniresolver.io/",
+		func(opts *resolverOpts) {
+			i += 1 // nolint
+		},
+		func(opts *resolverOpts) {
+			i += 2
+		},
+	)
+	require.NoError(t, err)
+	require.Equal(t, 1+2, i)
+
+	// Invalid URL
+	_, err = New("invalid url")
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "Invalid base url")
+}
+
+func TestRead_DIDDoc(t *testing.T) {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		require.Equal(t, "/did:example:334455", req.URL.String())
+		res.Header().Add("Content-type", "application/did+ld+json")
+		res.WriteHeader(http.StatusOK)
+		_, _ = res.Write([]byte("did doc body"))
+	}))
+	defer func() { testServer.Close() }()
+
+	resolver, _ := New(testServer.URL)
+	gotDocument, err := resolver.Read("did:example:334455", nil, "", false)
+	require.NoError(t, err)
+	require.Equal(t, []byte("did doc body"), gotDocument)
+}
+
+func TestRead_DIDDocWithBasePath(t *testing.T) {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		require.Equal(t, "/document/did:example:334455", req.URL.String())
+		res.Header().Add("Content-type", "application/did+ld+json")
+		res.WriteHeader(http.StatusOK)
+		_, _ = res.Write([]byte("did doc body"))
+	}))
+	defer func() { testServer.Close() }()
+
+	resolver, _ := New(testServer.URL + "/document")
+	gotDocument, err := resolver.Read("did:example:334455", nil, "", false)
+	require.NoError(t, err)
+	require.Equal(t, []byte("did doc body"), gotDocument)
+}
+
+func TestRead_DIDDocWithBasePathWithSlashes(t *testing.T) {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		require.Equal(t, "/document/did:example:334455", req.URL.String())
+		res.Header().Add("Content-type", "application/did+ld+json")
+		res.WriteHeader(http.StatusOK)
+		_, _ = res.Write([]byte("did doc body"))
+	}))
+	defer func() { testServer.Close() }()
+
+	resolver, _ := New(testServer.URL + "/document/")
+	gotDocument, err := resolver.Read("did:example:334455", nil, "", false)
+	require.NoError(t, err)
+	require.Equal(t, []byte("did doc body"), gotDocument)
+}
+
+func TestRead_DIDDocNotFound(t *testing.T) {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		require.Equal(t, "/did:example:334455", req.URL.String())
+		res.WriteHeader(http.StatusNotFound)
+		_, _ = res.Write([]byte("did doc body"))
+	}))
+	defer func() { testServer.Close() }()
+
+	resolver, _ := New(testServer.URL)
+	_, err := resolver.Read("did:example:334455", nil, "", false)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "Input DID does not exist")
+}
+
+func TestRead_UnsupportedStatus(t *testing.T) {
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		res.WriteHeader(http.StatusForbidden)
+	}))
+	defer func() { testServer.Close() }()
+
+	resolver, _ := New(testServer.URL)
+	_, err := resolver.Read("did:example:334455", nil, "", false)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "Unsupported response from DID Resolver with status code")
+}
+
+func TestRead_HTTPGetFailed(t *testing.T) {
+	// HTTP GET failed
+	testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+		res.WriteHeader(http.StatusSeeOther)
+	}))
+	defer func() { testServer.Close() }()
+
+	resolver, _ := New(testServer.URL)
+	_, err := resolver.Read("did:example:334455", nil, "", false)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "HTTP Get request failed")
+}


troyronda:
Thanks @kdimak 

Can I suggest that we update the PR and Commit description to refer to the [DID resolution HTTPS binding](https://w3c-ccg.github.io/did-resolution/#bindings-https) and also confirm the current PR is compatible with this spec.

troyronda:
> This is a part of #99 and allows for Universal HTTP DID Resolver.

Does this close #99 or you have follow-up PRs? If it closes #99, then please use the closes keyword.

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
...
+	Get(url string) (body []byte, err error)
+}
+
+//StandardHTTPClient is based on http.Client
+type StandardHTTPClient struct {
+	client *http.Client
+}
+
+//Get implements HTTP Get operation on standard http client
+func (shc StandardHTTPClient) Get(url string) (body []byte, err error) {
```
llorllale: Although not really needed in this case, let's always use pointer receivers for consistency (see [Go FAQ](https://golang.org/doc/faq#methods_on_values_or_pointers)).

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, errors.Errorf("Failed to close response body: %w", err)
+	}
+	return b, nil
+}
+
+// URLBuilderCreator is a factory of URL builder based on base url and did
+type URLBuilderCreator func(baseURL, did string) (string, error)
+
+// QueryParamURLBuilder provides a factory of URL builder using query parameters
+func QueryParamURLBuilder(paramName string) URLBuilderCreator {
```
llorllale: I don't think we need this additional level of indirection now, do we? I don't see definitions for different strategies for building URLs

troyronda:
Please also include a short within your git commit description. You should also only have one Signed-off-by: line :).

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
+/*
```
fqutishat: we need timeout option and tlsconfig option similar to https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/transport/http/outbound.go#L39

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
...
+
+// HTTPDIDResolver DID http resolver
+type HTTPDIDResolver struct {
+	baseURL        string
+	requestCreator RequestExecutor
+	client         HTTPClient
+}
+
+// NewHTTPDIDResolver creates new generic NewHTTPDIDResolver
+func NewHTTPDIDResolver(client HTTPClient, baseURL string, requestExecutor RequestExecutor) *HTTPDIDResolver {
```
fqutishat: why NewHTTPDIDResolver need client as parameter.

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
...
+		params.Add(paramName, did)
+
+		baseURI.RawQuery = params.Encode()
+
+		return baseURI.String(), nil
+	}
+}
+
+// RequestExecutor is a factory of http request executor for DID resolving
+type RequestExecutor func(client HTTPClient, baseURL, did string) ([]byte, error)
```
llorllale: Can this be collapsed into `HTTPClient`?

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
...
+// NewHTTPDIDResolver creates new generic NewHTTPDIDResolver
+func NewHTTPDIDResolver(client HTTPClient, baseURL string, requestExecutor RequestExecutor) *HTTPDIDResolver {
+	return &HTTPDIDResolver{
+		client:         client,
+		requestCreator: requestExecutor,
+		baseURL:        baseURL}
+}
+
+// NewHTTPGetByQueryParamDIDResolver creates HTTP Get NewHTTPDIDResolver with passing DID as query parameter
+func NewHTTPGetByQueryParamDIDResolver(client HTTPClient, baseURL, didRequestParamName string) *HTTPDIDResolver {
```
llorllale: > "with passing DID as query parameter"

The DID is passed via `DidMethod.Read()`. Do we really need this constructor?

pkg/didmethod/httpuniversal/resolver.go
```diff
@@ -0,0 +1,116 @@
...
+// QueryParamURLBuilder provides a factory of URL builder using query parameters
+func QueryParamURLBuilder(paramName string) URLBuilderCreator {
+	return func(baseURL, did string) (string, error) {
+		baseURI, err := url.ParseRequestURI(baseURL)
+		if err != nil {
+			return "", errors.Errorf("Invalid base url: %w", err)
+		}
+
+		params := url.Values{}
+		params.Add(paramName, did)
```
troyronda: The DID should be supplied in the path rather than as a query parameter.
Example: https://uniresolver.io/1.0/identifiers/did:sov:WRfXPg8dantKVubE3HX8pw

Note: the host and base path need to be configurable via option. Examples:
- https://example.com/did:sov:WRfXPg8dantKVubE3HX8pw
- https://example.com/document/did:sidetree:WRfXPg8dantKVubE3HX8pw

are also valid server configurations.

pkg/didmethod/httpuniversal/resolver.go
kdimak: Agree, fixed.

pkg/didmethod/httpuniversal/resolver.go
kdimak: Agree. This is simplified.

pkg/didmethod/httpuniversal/resolver.go
kdimak: Agree, used the configuration approach from `outbound.go`

pkg/didmethod/httpuniversal/resolver.go
kdimak: Agree, this one is simplified.

pkg/didmethod/httpuniversal/resolver.go
kdimak: Agree, this one is simplified.

pkg/didmethod/httpuniversal/resolver.go
kdimak: True. My next version follows the spec more strictly.

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+func WithHTTPClientInstance(client *http.Client) HTTPOptSetter {
+	return func(opts *HTTPClientOpts) {
+		opts.client = client
+	}
+}
+
+// WithHTTPClientTimeout option is for creating an Outbound HTTP transport using a client timeout value
+func WithHTTPClientTimeout(timeout time.Duration) HTTPOptSetter {
+	return func(opts *HTTPClientOpts) {
+		opts.client.Timeout = timeout
```
llorllale: Panic if `WithHTTPClientInstance` is not called first.

I wonder if we should consider substituting this `WithXXX` options pattern with the one discussed [here](https://github.com/hyperledger/aries-framework-go/pull/120#discussion_r314558349)

@troyronda @Baha-sk

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+// WithHTTPClientTLSConfig option is for creating an Outbound HTTP transport using a tls.Config instance
+func WithHTTPClientTLSConfig(tlsConfig *tls.Config) HTTPOptSetter {
+	return func(opts *HTTPClientOpts) {
+		opts.client = &http.Client{
```
llorllale: This overrides `WithHTTPClientInstance`

pkg/didmethod/httpbinding/resolver.go
troyronda: We should use functional options for the New constructor. A struct is too brittle.

troyronda: see https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis for a longer explanation. also an example of this pattern is widely used in fabric-sdk-go and grpc.

pkg/didmethod/httpbinding/resolver.go
kdimak: Yes, that's not good...

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+
+// WithHTTPClientBasePath options is for setting optional base path, for example "document"
+func WithHTTPClientBasePath(basePath string) HTTPOptSetter {
+	return func(opts *HTTPClientOpts) {
+		opts.basePath = basePath
+	}
+}
+
+// resolveDid makes DID resolution via HTTP
+func (res *DIDResolver) resolveDid(url string) ([]byte, error) {
```
rolsonquadras: Just a nitpick - Currently, the naming for DID is not consistent (Did v DID). As part of the [issue](https://github.com/hyperledger/aries-framework-go/issues/87), we are planning to use DID for the naming. May be the function name can be changed to `resolveDID`.

Same applies to other places as well.

pkg/didmethod/httpbinding/resolver.go
llorllale: Opened #127

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+
+// HTTPClientOpts holds options for the HTTP transport implementation of DIDResolver
+// it has a required http.Client instance and optional basePath which is appended to the remove DID Resolver URL
+type HTTPClientOpts struct {
+	client   *http.Client
+	basePath string
+}
+
+// HTTPOptSetter is an outbound HTTP transport option
+type HTTPOptSetter func(opts *HTTPClientOpts)
```
fqutishat: Rename HTTPOptSetter to HTTPOpt

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+
+// WithHTTPClientInstance option is for creating an Outbound HTTP transport using an http.Client instance
+func WithHTTPClientInstance(client *http.Client) HTTPOptSetter {
```
fqutishat: we don't need WithHTTPClientInstance

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+		opts.client = &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: tlsConfig,
+			},
+		}
+	}
+}
+
+// WithHTTPClientBasePath options is for setting optional base path, for example "document"
+func WithHTTPClientBasePath(basePath string) HTTPOptSetter {
```
fqutishat: we don't need WithHTTPClientBasePath. You can move basePath to New

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// TODO support for service endpoint URL resolution
+	if containsDidDocument(resp) {
+		var gotBody []byte
+		gotBody, err = ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, errors.Errorf("Failed to read response body: %w", err)
+		}
+
+		err = resp.Body.Close()
```
fqutishat: body will not close if err return from ReadAll

pkg/didmethod/httpbinding/resolver.go
kdimak: OK!

pkg/didmethod/httpbinding/resolver.go
kdimak: Why? From [DID Resolving via HTTP(s)] (https://w3c-ccg.github.io/did-resolution/#bindings-https) I see that HTTPS is not mandatory and we should not close the possibility for not-secure HTTP clients.

pkg/didmethod/httpbinding/resolver.go
kdimak: This indicates that DIDs could be resolved using hostname like https://uniresolver.io/ directly or by appending extra relative path like "https://uniresolver.io/document/".
@troyronda what do you think about this?

pkg/didmethod/httpbinding/resolver.go
kdimak: Agree, good catch!

pkg/didmethod/httpbinding/resolver.go
kdimak: I left functional options approach in this PR.

pkg/didmethod/httpbinding/resolver.go
kdimak: I changed this to just change the `Transport` field of the client.

pkg/didmethod/httpbinding/resolver.go
kdimak: Corrected!

pkg/didmethod/httpbinding/resolver.go
kdimak: Done!

pkg/didmethod/httpbinding/resolver.go
kdimak: Also setting `DefaultClient` at the initialization before applying functional options for safety reasons.

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,139 @@
...
+
+// WithHTTPClientInstance option is for creating an Outbound HTTP transport using an http.Client instance
+func WithHTTPClientInstance(client *http.Client) HTTPOpt {
+	return func(opts *HTTPClientOpts) {
+		opts.client = client
+	}
+}
+
+// WithHTTPClientTimeout option is for creating an Outbound HTTP transport using a client timeout value
+func WithHTTPClientTimeout(timeout time.Duration) HTTPOpt {
```
troyronda: We could consider shortening the method names - I don't think "Client" is needed.

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	errors "golang.org/x/xerrors"
+)
+
+var logger = log.New("aries-framework/didmethod/httpbinding")
+
+// HTTPClientOpts holds options for the HTTP transport implementation of DIDResolver
+// it has a required http.Client instance and optional basePath which is appended to the remove DID Resolver URL
+type HTTPClientOpts struct {
```
troyronda: This struct should not be exported.

pkg/didmethod/httpbinding/resolver.go
troyronda: @fqutishat @kdimak I would likely not add a new basePath argument to New.

pkg/didmethod/httpbinding/resolver.go
kdimak: Yeah, this is a redundant word, agree!

pkg/didmethod/httpbinding/resolver.go
kdimak: Ah, I renamed it to make capital "HTTP" and did not notice the side effect. Thanks!

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,139 @@
...
+
+// containsDIDDocument checks weather reply from remote DID resolver contains DID document
+func containsDIDDocument(resp *http.Response) bool {
+	return resp.StatusCode == http.StatusOK && resp.Header.Get("Content-type") == "application/did+ld+json"
+}
+
+// New creates new DIDResolver
+func New(host string, opts ...HTTPOpt) (*DIDResolver, error) {
+	// Apply options
+	clOpts := &HTTPClientOpts{client: http.DefaultClient}
```
llorllale: Let's use `&http.Client{}` instead of `http.DefaultClient`. Any code can mutate the object assigned to `http.DefaultClient` and thus we should not trust this variable.

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,139 @@
...
+type HTTPClientOpts struct {
+	client   *http.Client
+	basePath string
+}
+
+// HTTPOpt is an outbound HTTP transport option
+type HTTPOpt func(opts *HTTPClientOpts)
+
+// WithHTTPClientInstance option is for creating an Outbound HTTP transport using an http.Client instance
+func WithHTTPClientInstance(client *http.Client) HTTPOpt {
```
llorllale: Let's remove this option for now - we don't anticipate usage of this function right now.

pkg/didmethod/httpbinding/resolver.go
troyronda: The team also thinks HTTP is not needed.

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,139 @@
...
+	return resp.StatusCode == http.StatusNotFound
+}
+
+// containsDIDDocument checks weather reply from remote DID resolver contains DID document
+func containsDIDDocument(resp *http.Response) bool {
+	return resp.StatusCode == http.StatusOK && resp.Header.Get("Content-type") == "application/did+ld+json"
+}
+
+// New creates new DIDResolver
+func New(host string, opts ...HTTPOpt) (*DIDResolver, error) {
```
llorllale: Let's rename `host` to `url` or `endpointUrl` or something similar - the spec just specifies the endpoint to which the algorithm appends the input DID: https://w3c-ccg.github.io/did-resolution/#bindings-https

The `basePath` option in `HTTPClientOpts` is not required.

pkg/didmethod/httpbinding/resolver.go
kdimak: I have the same feeling indeed!

pkg/didmethod/httpbinding/resolver.go
kdimak: Agree!

pkg/didmethod/httpbinding/resolver.go
kdimak: 👍

pkg/didmethod/httpbinding/resolver.go
kdimak: 👍

pkg/didmethod/httpbinding/resolver.go
kdimak: OK

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	errors "golang.org/x/xerrors"
+)
+
+var logger = log.New("aries-framework/didmethod/httpbinding")
+
+// optsOfHTTP holds options for the HTTP transport implementation of DIDResolver
+// it has a required http.Client instance and optional basePath which is appended to the remove DID Resolver URL
+type optsOfHTTP struct {
```
troyronda: These options are package level - not just for HTTP (so it is more accurately opts or resolverOpts).

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+	client *http.Client
+}
+
+// HTTPOpt is an outbound HTTP transport option
+type HTTPOpt func(opts *optsOfHTTP)
```
troyronda: HTTP  doesn't really seem needed - these are for the package constructor.

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+
+// WithHTTPTimeout option is for definition of HTTP client timeout value
+func WithHTTPTimeout(timeout time.Duration) HTTPOpt {
```
troyronda: HTTP doesn't really seem needed for the Option methods.

pkg/didmethod/httpbinding/resolver.go
kdimak: Sure!

pkg/didmethod/httpbinding/resolver.go
kdimak: Agree

pkg/didmethod/httpbinding/resolver.go
kdimak: Corrected

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+		var gotBody []byte
+		gotBody, err = ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, errors.Errorf("Failed to read response body: %w", err)
+		}
+
+		return gotBody, nil
+
+	} else if notExistentDID(resp) {
+		return nil, errors.Errorf("Input DID does not exist: %w", err)
```
troyronda: We will need to add the NotFound sentinel in a followup.

PR #120, opened 2019-08-16
refactor: #65 DidMethod.Read()

### LINKED ISSUE #65: Refactor `DidMethod.Read`
https://github.com/hyperledger/aries-framework-go/blob/695f0bf9ae71b8fee6a3b66397d34a5101bb35e6/pkg/framework/didresolver/api.go#L22

All args aside from the DID should be [optional](https://w3c-ccg.github.io/did-resolution/#resolving).

Let's also add a `bool` return arg that indicates whether the DID method found a result or not.

llorllale:
Closes #65

* Make optional all args to DidMethod.Read() (aside from the DID)
* Add didresolver.ErrNotFound sentinel error to indicate when DID was not found
* Add GoDocs to `didresolver.DidMethod`

### DIFF
--- a/pkg/didmethod/peer/resolver.go
+++ b/pkg/didmethod/peer/resolver.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package peer
 
 import (
+	"github.com/hyperledger/aries-framework-go/pkg/framework/didresolver"
 	errors "golang.org/x/xerrors"
 )
 
@@ -21,11 +22,15 @@ func NewDIDResolver(store *DIDStore) *DIDResolver {
 }
 
 // Read implements didresolver.DidMethod.Read interface (https://w3c-ccg.github.io/did-resolution/#resolving-input)
-func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+func (resl *DIDResolver) Read(did string, _ ...didresolver.ResolveOpt) ([]byte, error) {
 	// get the document from the store
 	doc, err := resl.store.Get(did)
 	if err != nil {
 		return nil, errors.Errorf("Fetching data from store failed: %w", err)
+	}
+
+	if doc == nil {
+		return nil, didresolver.ErrNotFound
 	}
 
 	// convert the doc to JSON as DID Resolver expects byte result.
--- a/pkg/framework/didresolver/api.go
+++ b/pkg/framework/didresolver/api.go
@@ -7,6 +7,8 @@ package didresolver
 
 import (
 	"time"
+
+	"golang.org/x/xerrors"
 )
 
 // ResultType input option can be used to request a certain type of result.
@@ -19,9 +21,16 @@ const (
 	ResolutionResult
 )
 
-// DidMethod operations
+// ErrNotFound is returned when a DID resolver does not find the DID.
+var ErrNotFound = xerrors.New("DID not found")
+
+// DidMethod resolves a DID into a result type (default: DidDocumentResult).
+// See the DID resolution spec: https://w3c-ccg.github.io/did-resolution.
 type DidMethod interface {
-	Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error)
+	// Read implements the 'DID Resolution' algorithm defined in
+	// https://w3c-ccg.github.io/did-resolution/#resolving.
+	Read(did string, opts ...ResolveOpt) ([]byte, error)
+	// Accept registers this DID method resolver with the given method.
 	Accept(method string) bool
 }
 
--- a/pkg/framework/didresolver/didresolver.go
+++ b/pkg/framework/didresolver/didresolver.go
@@ -34,7 +34,6 @@ func (r *DIDResolver) Resolve(did string, opts ...ResolveOpt) (*diddoc.Doc, erro
 	for _, opt := range opts {
 		opt(resolveOpts)
 	}
-
 	// TODO Validate that the input DID conforms to the did rule of the Generic DID Syntax (https://w3c-ccg.github.io/did-spec/#generic-did-syntax)
 	// For now we do simple validation
 	didParts := strings.SplitN(did, ":", 3)
@@ -51,14 +50,16 @@ func (r *DIDResolver) Resolve(did string, opts ...ResolveOpt) (*diddoc.Doc, erro
 	}
 
 	// Obtain the DID Document
-	didDocBytes, err := method.Read(did, resolveOpts.versionID, resolveOpts.versionTime, resolveOpts.noCache)
+	didDocBytes, err := method.Read(did, opts...)
 	if err != nil {
+		if err == ErrNotFound {
+			return nil, err
+		}
 		return nil, errors.Errorf("did method read failed failed: %w", err)
 	}
 
-	// If the input DID does not exist, return a nil
 	if len(didDocBytes) == 0 {
-		return nil, nil
+		return nil, ErrNotFound
 	}
 
 	// Validate that the output DID Document conforms to the serialization of the DID Document data model
--- a/pkg/didmethod/peer/resolver_test.go
+++ b/pkg/didmethod/peer/resolver_test.go
@@ -80,7 +80,7 @@ func TestPeerDIDResolver(t *testing.T) {
 	require.NoError(t, err)
 
 	resl := NewDIDResolver(store)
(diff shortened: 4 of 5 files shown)


pkg/framework/didresolver/api.go
```diff
@@ -19,63 +15,22 @@ const (
...
-
-// WithDidMethod to add did method
-// DID methods are checked in the order added
-func WithDidMethod(method DidMethod) Opt {
-	return func(opts *didResolverOpts) {
-		opts.didMethods = append(opts.didMethods, method)
-	}
+// Options holds the optional arguments to the DID resolution algorithm defined in
+// 'Resolving a DID': https://w3c-ccg.github.io/did-resolution/#resolving
+type Options struct {
```
troyronda: I do not understand this change. Why are you changing from functional options to a struct?

pkg/didmethod/peer/resolver.go
```diff
@@ -21,7 +22,7 @@ func NewDIDResolver(store *DIDStore) *DIDResolver {
 }
 
 // Read implements didresolver.DidMethod.Read interface (https://w3c-ccg.github.io/did-resolution/#resolving-input)
-func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+func (resl *DIDResolver) Read(did string, opts ...didresolver.Options) ([]byte, error) {
```
troyronda: I thought the purpose of the refactor was to able to return a doc. I still see []byte.

troyronda: I do like moving to an ...Option idiom rather than a long list or optional arguments in the signature. Later in the PR i am confused about using a struct rather than functional arguments,

pkg/framework/didresolver/didresolver.go
```diff
@@ -28,13 +44,13 @@ func New(opts ...Opt) *DIDResolver {
 }
 
 // Resolve did document
-func (r *DIDResolver) Resolve(did string, opts ...ResolveOpt) (*diddoc.Doc, error) {
-	resolveOpts := &resolveOpts{}
-	// Apply options
-	for _, opt := range opts {
-		opt(resolveOpts)
+func (r *DIDResolver) Resolve(did string, options ...Options) (*diddoc.Doc, error) {
```
troyronda: As I mentioned earlier, I am wondering why change this signature.

pkg/didmethod/peer/resolver.go
troyronda: I like collapsing the optional arguments. As I mention later, I am wondering why use an Option struct instead of functional options?

FYI: #65 also mentions a boolean return value.

pkg/framework/didresolver/api.go
llorllale: Two reasons:

* I think we can achieve the same goal (optional args) with less LoC (look at the relevant code that is no longer required with this change (ie. none of those With*** methods)):

```go
resolver.Read(did, didresolver.Options{
    NoCache: true,
    VersionID: 123,
    ResultType: DidDocumentResult,
    })
```
as opposed to:
```go
resolver.Read(did,
    WithNoCache(true),
    WithVersionId(123),
    WithResultType(DidDocumentResult))
```

* Slightly less confusion: there were two sets of "WithXXX" methods, several belonging to `ResolverOpt` and one belonging to`Opts`. The former applied to `didresolver.DIDResolver.Resolve()` while the latter applied to the `didresolver.New()` constructor. I found this a bit confusing and figured it made sense to more clearly distinguish between the two sets by way of their usage. If I hadn't made this change, I would've had to figure out how to make sense of adding `WithXXX` methods as options to `DidMethod.Read()`

pkg/didmethod/peer/resolver.go
llorllale: > FYI: #65 also mentions a boolean return value.

Added - I had overlooked that completely

pkg/didmethod/peer/resolver.go
```diff
@@ -21,20 +22,20 @@ func NewDIDResolver(store *DIDStore) *DIDResolver {
...
 
 	// convert the doc to JSON as DID Resolver expects byte result.
 	jsonDoc, err := doc.JSONBytes()
 	if err != nil {
-		return nil, errors.Errorf("Json marshalling of document failed: %w", err)
+		return nil, false, errors.Errorf("Json marshalling of document failed: %w", err)
 	}
 
-	return jsonDoc, nil
+	return jsonDoc, true, nil
```
Baha-sk: is a boolean really necessary? if jsonDoc is nil, then it means doc is not found. 
Also the boolean is set to false in case of errors only ( where jsonDoc returned is nil).
This boolean is redundant.

llorllale: I tend to agree. @troyronda thoughts?

troyronda: I think (nil, nil) is not a valid thing to return. If I get a non-nil err, I expect a valid value.

Baha-sk: (nil, nil) should never be returned
(!nil, !nil) should never be returned either,

the return should be either error != nil or jsonDoc != nil, the above cases should be treated as a bug

troyronda: So @Baha-sk Are you are saying that there is a NotFound error? How does the caller easily tell the error is NotFound?

llorllale: I'm not sure I'm following here.

(nil,nil) --> not found, no error
(!nil, nil) --> found, no error
(nil, !nil) --> not found, error

(!nil, !nil) should be illegal

troyronda: I dislike (nil, nil) as a signal for NotFound. I continue to think that a nil error means the value should be safe (and should exist). 

It would be helpful if someone has already documented the correct idiom to use in Go.

fqutishat: from spec:
https://w3c-ccg.github.io/did-resolution/#resolving-algorithm

If the input DID does not exist, return a null result.

troyronda: @fqutishat the spec should not be used as a substitute for correct idioms in the implementation language.

pkg/framework/didresolver/api.go
troyronda: - WithNoCache should not have taken an argument.
- I’m not sure that WithResultType should even exist externally. It seems that a typed return value method should exist for each type. i.e., different external methods.
- Is 0 the default/initial value for Version ID?

troyronda: Re: Confusion. Yup - there can be some confusion due to the two groups when reading source files.  (I would also suggest better code comments and groupings.)

Note: From an external developer PoV, interestingly GoDoc groups the With functions according to their type. E.g., https://godoc.org/github.com/hyperledger/fabric-sdk-go/pkg/client/resmgmt

See ClientOption vs RequestOption.

troyronda: I also think we could improve this situation by introducing multiple variations of WithVersionID that has an appropriate type as argument rather than interface{}.

(actually that is another advantage of functional options).

llorllale: Opened #127

llorllale:
Refactored to include *functional-style* optional parameters as per discussion.

llorllale:
The underlying `Store` API needs to indicate a "not found" outcome such that the peer DID resolver can appropriately return `didresolver.ErrNotFound`

pkg/framework/didresolver/api.go
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 package didresolver
 
 import (
+	"errors"
```
troyronda: Should import xerrors and alias to errors.

llorllale: Fixed

pkg/didmethod/peer/resolver.go
```diff
@@ -21,13 +22,17 @@ func NewDIDResolver(store *DIDStore) *DIDResolver {
 }
 
 // Read implements didresolver.DidMethod.Read interface (https://w3c-ccg.github.io/did-resolution/#resolving-input)
-func (resl *DIDResolver) Read(did string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+func (resl *DIDResolver) Read(did string, _ ...didresolver.ResolveOpt) ([]byte, error) {
```
troyronda: Looks like there was an already missing TODO to make use of arguments.

PR #147, opened 2019-08-22
fix: DidMethod interface impl by HTTP DID Resolver

### LINKED ISSUE #145: Adopt HTTP DID Resolver to the changed Resolver interface
Adopt HTTP DID Resolver method signature
https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didmethod/httpbinding/resolver.go#L109

to reflect the change of the Resolver interface:

https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/didresolver/api.go#L32

kdimak:

closes #145 

Adopts HTTP DID Resolver to implement the changed Resolver interface.

### DIFF
--- a/pkg/didmethod/httpbinding/resolver.go
+++ b/pkg/didmethod/httpbinding/resolver.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/didresolver"
 	errors "golang.org/x/xerrors"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/didmethod/httpbinding/resolver.go
```diff
@@ -106,14 +107,19 @@ func New(endpointURL string, opts ...ResolverOpt) (*DIDResolver, error) {
 }
 
 // Read implements didresolver.DidMethod.Read interface (https://w3c-ccg.github.io/did-resolution/#resolving-input)
-func (res *DIDResolver) Read(DID string, versionID interface{}, versionTime string, noCache bool) ([]byte, error) {
+func (res *DIDResolver) Read(DID string, _ ...didresolver.ResolveOpt) ([]byte, error) {
```
llorllale: Does the universal resolver support these options?

llorllale:
we now have an unfortunate similarity between the symbols `httpbinding.ResolverOpt` and `didresolver.ResolveOpt`

troyronda:
Should ResolveOpt actually be MethodOpt?

pkg/didmethod/httpbinding/resolver.go
fqutishat: we will use them in the future

kdimak: > Does the universal resolver support these options?

Currently, there is no instruction on how to use these parameters in HTTP(s) binding in the [spec](https://w3c-ccg.github.io/did-resolution/#bindings).

PR #159, opened 2019-08-23
feat: Create a state machine basic flow

talwinder50:
**Note :This is just initial draft not a final implementation.


pkg/didcomm/protocol/exchange/state/state.go
```diff
@@ -0,0 +1,200 @@
...
+	ack        = "ack"
+)
+
+//Context defines the DIDComm state context
+type Context struct {
+	Store storage.Store
+	State *State
+}
+
+// New creates the new context
```
llorllale: "creates the new did-exchange state machine" perhaps?

pkg/didcomm/protocol/exchange/state/state.go
```diff
@@ -0,0 +1,200 @@
...
+func New(store storage.Store, state *State) *Context {
+	return &Context{
+		Store: store,
+		State: state,
+	}
+}
+
+// DidCommState describes the internal State machine for Inviter and Invitee. Reference: https://github.com/hyperledger/aries-rfcs/blob/master/features/0023-did-exchange/chrome_2019-01-29_07-59-38.png
```
llorllale: Suggest we break these long lines into several with a max of ~100 chars per line

pkg/didcomm/protocol/exchange/state/state.go
```diff
@@ -0,0 +1,200 @@
...
+type DidCommState interface {
```
llorllale: Sounds like this should be renamed to some short variant of "DidExchangeStateMachine".

troyronda: @llorllale I would have assumed the StateMachine was part of the service subpackage.

(presumably the StateMachine object isn't accessed by the normal developer?)

pkg/didcomm/protocol/exchange/state/state.go
```diff
@@ -0,0 +1,200 @@
...
+	MsgType      string            `json:"@type"`
+	ConnectionID string            `json:"@id"`
+	ThreadID     map[string]string `json:"~thread"`
+	requestType  string
+	protocol     string
+	version      string
+	current      string
+}
+
+//Handle accepts the didComm protocol message and then set the state based on the message
```
llorllale: "accepts the did-exchange protocol message"

pkg/didcomm/protocol/exchange/state/state.go
talwinder50: shall i move this to service subpackage ?

llorllale: I would personally move it to an `exchange` package, whether it's the exported one or an `/internal/` one as @troyronda seems to suggest.

troyronda: My point is that the top level exchange package (and other top level protocol packages) should only contain developer facing APIs. Keep the GoDocs clean.

troyronda: Also - I don’t think StateMachine will need to be accessed outside the Exchange protocol package and sub packages, right?

pkg/didcomm/protocol/exchange/state/state.go
```diff
@@ -0,0 +1,200 @@
...
+
+//CheckState checks the current state of didComm state
+func (c *Context) CheckState() string {
+	s := []string{c.State.protocol, c.State.version, c.State.ConnectionID, c.State.ThreadID["thid"]}
+	compositeKey := strings.Join(s, "")
+	respBytes, _ := c.Store.Get(compositeKey)
+
+	if respBytes == nil {
+		c.State.current = NullState
+	} else {
```
fqutishat: you don't need else

pkg/didcomm/protocol/exchange/state/state.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+//Context defines the DIDComm state context
+type Context struct {
+	Store storage.Store
+	State *State
+}
+
+// New creates the new context
+func New(store storage.Store, state *State) *Context {
```
fqutishat: we you need to pass state to New method

pkg/didcomm/protocol/exchange/state/state.go
fqutishat: How exchange protocol will integrate with state machine how can we create the msg

PR #201, opened 2019-09-03
refactor: Load Protocol Services in framework.New()

### LINKED ISSUE #189: Load protocol services in framework.New()
The framework.Context() returns a [new instance](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/aries/framework.go#L100) of context each time its invoked. Same context should be return whenever Context() is called on same framework instance.

rolsonquadras:
Currently the protocol services are loaded in context.New() as protocols require ctx. This change moves that to framework.New() and creates a ctx object in framework.New() function.

Closes #189 

### DIFF
--- a/pkg/framework/aries/framework.go
+++ b/pkg/framework/aries/framework.go
@@ -9,6 +9,7 @@ package aries
 import (
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/context"
@@ -27,6 +28,7 @@ type Aries struct {
 	didResolver         DIDResolver
 	storeProvider       storage.Provider
 	protocolSvcCreators []api.ProtocolSvcCreator
+	services            []dispatcher.Service
 }
 
 // Option configures the framework.
@@ -50,6 +52,20 @@ func New(opts ...Option) (*Aries, error) {
 	err := defFrameworkOpts(frameworkOpts)
 	if err != nil {
 		return nil, fmt.Errorf("default option initialization failed: %w", err)
+	}
+
+	ctxProvider, err := frameworkOpts.Context()
+	if err != nil {
+		return nil, fmt.Errorf("context creation failed: %w", err)
+	}
+
+	//Load services
+	for _, v := range frameworkOpts.protocolSvcCreators {
+		svc, err := v(ctxProvider)
+		if err != nil {
+			return nil, fmt.Errorf("new protocol service failed: %w", err)
+		}
+		frameworkOpts.services = append(frameworkOpts.services, svc)
 	}
 
 	return frameworkOpts, nil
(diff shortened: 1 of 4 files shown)


pkg/framework/aries/framework.go
```diff
@@ -27,6 +27,8 @@ type Aries struct {
 	didResolver         DIDResolver
 	storeProvider       storage.Provider
 	protocolSvcCreators []api.ProtocolSvcCreator
+	ctx                 *context.Provider
+	services            []dispatcher.Service
```
llorllale: Seems odd to have both `dispatcher.Service` and `api.ProtocolSvcCreator` (the latter provide the former).

rolsonquadras: `protocolSvcCreators` are passed a functional options and `services` are derived from it. We need this while creating a new context as we pass `services` to `context.WithProtocolServices()`. [Previously](https://github.com/hyperledger/aries-framework-go/pull/201/files#diff-9988f5820bf05228e5657f2842fa071eL23) these 2 were there in `context.Provider` and now moved to `Aries`. Do you have anything specific in mind ? I can change the code.

llorllale: It seems like we no longer need `api.ProtocolSvcCreator`, do we?

rolsonquadras: we need it for the [option](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/aries/framework.go#L83). And we can create `services` once we have ctx.

llorllale: @troyronda @rolsonquadras I'm not sure what direction to head in this case. Yes - context needs to be available in order to instantiate the protocol services. However, the user does not (currently) have access to a "default" context that they can use to instantiate their own protocol services and pass in to `framework.New()`. Perhaps we need to expose a `defaults.Context()` somewhere?

rolsonquadras: @llorllale @troyronda this PR is just doing code refactor. The comments above looks more like a bigger design discussion. Can we this discussion moved to a new issue and merge this PR (if they are no blocking issues in this PR) ?

pkg/framework/aries/framework.go
```diff
@@ -9,6 +9,8 @@ package aries
 import (
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+
```
troyronda: Extra space

pkg/framework/aries/framework.go
troyronda: It’s okay to me. What is the issue number?

pkg/framework/context/context.go
```diff
@@ -91,10 +82,10 @@ func WithOutboundTransport(ot transport.OutboundTransport) ProviderOption {
 	}
 }
 
-// WithProtocols injects protocol svc into context
-func WithProtocols(protocolSvcCreator ...api.ProtocolSvcCreator) ProviderOption {
+// WithProtocolServices injects protocol services into the context.
```
troyronda: Why did we need to rename this method?

pkg/framework/aries/framework.go
rolsonquadras: @troyronda @llorllale created #212

pkg/framework/aries/framework.go
rolsonquadras: [Removed](https://github.com/hyperledger/aries-framework-go/compare/581fd542546b921481bf770071aa503f3cdcac10..69eb474db784896df411bcd385eca107c37518de)

pkg/framework/context/context.go
rolsonquadras: Before this change Context required `Protocol` to be passed in to create the `ProtocolServices`. Now services are created in framework.New(), we can directly pass `ProtocolServices` to context.

Issue #204, opened 2019-09-04
Update to Go 1.13


troyronda:
- CircleCI tag
- switch xerrors to errors
- crypto/ed25519 instead of golang.org/x/crypto/ed25519 (note for future changes)
- go.mod version
- go.mod tidying
- update linter and resolve issues (if needed).
  - pkg/framework/aries/api/protocol.go: error var SvcErrNotFound should have name of the form ErrFoo (golint)
  - should replace `errors.New(fmt.Sprintf(...))` with `fmt.Errorf(...)` (golint)
  - errors.Errorf => fmt.Errorf
  - errors strings should not be capitalized or end in punctuation (GoLand) - will be handled in #205
- update README
- Remove explicit GO111MODULE=on

PR #206, opened 2019-09-04
chore: update to Go 1.13

troyronda:
This change updates the project to Go 1.13
- CI is updated.
- README is updated.
- Built-in errors package is used.

Closes: #204

### DIFF
--- a/pkg/restapi/restapi_test.go
+++ b/pkg/restapi/restapi_test.go
@@ -20,7 +20,7 @@ import (
 func TestNew_Failure(t *testing.T) {
 	controller, err := New(&context.Provider{})
 	require.Error(t, err)
-	require.Equal(t, err, api.SvcErrNotFound)
+	require.Equal(t, err, api.ErrSvcNotFound)
 	require.Nil(t, controller)
 }
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,14 +14,14 @@ jobs:
       - run: golangci-lint run
   unit-test:
     docker:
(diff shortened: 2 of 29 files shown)


pkg/restapi/restapi_test.go
```diff
@@ -20,7 +20,7 @@ import (
 func TestNew_Failure(t *testing.T) {
 	controller, err := New(&context.Provider{})
 	require.Error(t, err)
-	require.Equal(t, err, api.SvcErrNotFound)
+	require.Equal(t, err, api.ErrSvcNotFound)
```
troyronda: Should I just shorten to ErrNotFound?

rolsonquadras: @troyronda in that case, question is how can we differentiate between different `NotFound` error cases (in same package)? e.g. 'Data not found' and 'Service not found'. Having a specific error message might help in future when we come across these type of scenario's imo.

llorllale: > Should I just shorten to ErrNotFound?

Yes.

sudeshrshetty: > @troyronda in that case, question is how can we going to differentiate between `NotFound` error cases (in same package)? e.g. 'Data not found' and 'Service not found'. Having a specific error message might help in future when we come across these type of scenario's imo.

`ErrSvcNotFound` is good for long run

troyronda:
Created followup regarding ErrNotFound: https://github.com/hyperledger/aries-framework-go/issues/207

Issue #171, opened 2019-08-28
Extensibility of Verifiable Credential data model


kdimak:
Verifiable Credentials Data Model 1.0 supports extensions through the definition of links to custom JSON-LD based definitions.
https://w3c.github.io/vc-data-model/#extensibility

We need to provide a way to apply for valid schema extensions and read custom fields of Verifiable Credential into a dictionary or any other appropriate data structure.

PR #211, opened 2019-09-05
feat: Extensibility of Verifiable Credential data model

kdimak:
- Provides a mechanism for extension of model by custom data structures which embedded verifiable.Credential

#171 

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -214,11 +214,20 @@ type embeddedCompositeIssuer struct {
 	CompositeIssuer compositeIssuer `json:"issuer,omitempty"`
 }
 
+// CredentialDecoder makes a custom decoding of Verifiable Credential in JSON form to existent
+// instance of Credential.
+type CredentialDecoder func(dataJSON []byte, credential *Credential) error
+
+// CredentialTemplate defines a factory method to create new Credential template.
+type CredentialTemplate func() *Credential
+
 // credentialOpts holds options for the Verifiable Credential decoding
 // it has a http.Client instance initialized with default parameters
 type credentialOpts struct {
-	schemaDownloadClient   *http.Client
-	disabledExternalSchema bool
+	schemaDownloadClient *http.Client
+	disabledCustomSchema bool
+	decoders             []CredentialDecoder
+	template             CredentialTemplate
 }
 
 // CredentialOpt is the Verifiable Credential decoding option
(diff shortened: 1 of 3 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -319,7 +365,7 @@ func validate(data []byte, schema *CredentialSchema, opts *credentialOpts) error
 	loader := gojsonschema.NewStringLoader(string(data))
 	result, err := gojsonschema.Validate(schemaLoader, loader)
 	if err != nil {
-		return fmt.Errorf("Validation of Verifiable Credential failed: %w", err)
```
troyronda: Why %v instead of %w? Is wrapping the error not appropriate here?
(and the other similar changes in this PR).

llorllale:
@kdimak please include a reference to the relevant issue in the PR's description

pkg/doc/verifiable/credential.go
kdimak: agree - I will revert it!

kdimak:
> @kdimak please include a reference to the relevant issue in the PR's description

done!

pkg/doc/verifiable/credential.go
kdimak: Fixed

pkg/doc/verifiable/credential.go
```diff
@@ -239,48 +244,89 @@ func WithDisabledCustomSchemaCheck() CredentialOpt {
...
+
+	credential.Issuer = Issuer{ID: issuerID, Name: issuerName}
+	return
+}
+
+// CredentialBuilder provides a way to extend Credential Data model.
+// It defines a factory method to create new Credential instance and
+// provides decoders needed to fill custom fields of Credential.
+type CredentialBuilder interface {
+	New() *Credential
```
troyronda: I think it is confusing to have a New() inside an interface.

Perhaps CreateCredential?

pkg/doc/verifiable/credential.go
```diff
@@ -239,48 +244,89 @@ func WithDisabledCustomSchemaCheck() CredentialOpt {
...
+	Decoders() []CredentialDecoder
```
troyronda: CredentialDecoders?

pkg/doc/verifiable/credential.go
```diff
@@ -239,48 +244,89 @@ func WithDisabledCustomSchemaCheck() CredentialOpt {
...
+	return func(opts *credentialOpts) {
+		opts.cb = cb
+	}
+}
+
+func issuerMapping(data []byte, credential *Credential) (err error) {
+	issuerID, issuerName, err := issuerFromBytes(data)
+	if err != nil {
+		err = fmt.Errorf("JSON unmarshalling of Verifiable Credential bytes failed: %w", err)
+		return
```
rolsonquadras: Better to have a explicit return than naked return -  [Named Result Parameters](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)

pkg/doc/verifiable/credential.go
```diff
@@ -239,48 +244,89 @@ func WithDisabledCustomSchemaCheck() CredentialOpt {
...
+// CredentialBuilder provides a way to extend Credential Data model.
+// It defines a factory method to create new Credential instance and
+// provides decoders needed to fill custom fields of Credential.
+type CredentialBuilder interface {
+	New() *Credential
+	Decoders() []CredentialDecoder
+}
+
+// New Factory method of base Credential.
+func (vc *Credential) New() *Credential {
```
troyronda: Credential.New returning a Credential seems pretty confusing.

pkg/doc/verifiable/credential.go
llorllale: I opened #216

pkg/doc/verifiable/example_credential.go
```diff
@@ -0,0 +1,74 @@
+/*
```
llorllale: This should be a `_test` file (ie. not in production code)

pkg/doc/verifiable/example_credential.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+//UniversityDegreeCredential University Degree credential, from examples of https://w3c.github.io/vc-data-model
+type UniversityDegreeCredential struct {
+	c Credential
+
+	Subject *UniversityDegreeSubject
+}
+
+type rawDegree struct {
```
llorllale: I think we would benefit from removing these `raw*` structs altogether from this test and have the non-raw structs hold the json tags. The reason is that I wouldn't want to give the wrong impression and have users think they necessarily have to follow this roundabout way of unmarshalling the credential when a simpler option is available.

pkg/doc/verifiable/credential.go
llorllale: And its method set seems disjointed. I suggest splitting them into two new options, perhaps with names `WithTemplate` and `WithDecoder`.

pkg/doc/verifiable/credential.go
kdimak: Useful link regarding code review comments! Thanks, I will update this.

pkg/doc/verifiable/example_credential.go
kdimak: OK!

pkg/doc/verifiable/example_credential.go
kdimak: OK!

pkg/doc/verifiable/credential.go
kdimak: I like the simplification @llorllale proposed!

pkg/doc/verifiable/credential.go
kdimak: agree!

pkg/doc/verifiable/example_credential_test.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (udc *UniversityDegreeCredential) decode(dataJSON []byte, credential *Credential) error {
+	return json.Unmarshal(dataJSON, udc)
+}
+
+func TestCredentialExtensibility(t *testing.T) {
+	udc := &UniversityDegreeCredential{}
+	cred, err := NewCredential(
+		[]byte(validCredential),
```
troyronda: We should add a utility function that would supply the credential type from the marshalled data. From there, someone could determine which type to instantiate.

troyronda: Created https://github.com/hyperledger/aries-framework-go/issues/218

PR #215, opened 2019-09-05
refactor: aries framework options should not return err

### LINKED ISSUE #214: aries framework options should not return err (defaults pkg)
The intention of options is that you can use directly inside aries.New
Having an error return prevents this usage.

Suggest using an anonymous function. Something like this:

```
return func(opts *Aries) error {
  [your code]
  return aries.WithStoreProvider(storeProv)(opts)
}
```

fqutishat:
closes #214

### DIFF
--- a/cmd/aries-agentd/main.go
+++ b/cmd/aries-agentd/main.go
@@ -60,9 +60,10 @@ func main() {
 		return
 	}
 
-	opts, err := createOpts()
-	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/aries-agentd/main.go
```diff
@@ -28,11 +28,12 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/defaults"
+
```
troyronda: extra line

fqutishat: done

PR #221, opened 2019-09-06
feat: DIDComm Inbound Transport Initiation in framework.New()

### LINKED ISSUE #188: Framework : Inbound transport instantiation
This issue is for the default instantiation of DIDComm Inbound transport based on question raised in #176. The code currently starts inbound transport outside the framework. ie, main starts the server. This needs to be moved inside framework.New(). 

DIDComm Inbound is transport agonistic. Framework should have a abstraction to support different implementations.

rolsonquadras:
- Create interface for InboundTransport
- Provide HTTP InboundTransport based on the new interface
- Add HTTPInboundTransport as the default inbound transport provider
- Start the inbound transport in while creating a new framework
- Remove code from reference agent to build and start the HTTP inbound transport

Closes #188 

### DIFF
--- a/pkg/didcomm/transport/http/inbound.go
+++ b/pkg/didcomm/transport/http/inbound.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package http
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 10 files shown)


pkg/didcomm/transport/http/inbound.go
```diff
@@ -88,3 +89,49 @@ func validateHTTPMethod(w http.ResponseWriter, r *http.Request) bool {
...
+			logger.Fatalf("HTTP server start with address [%s] failed, cause:  %s", i.server.Addr, err)
+		}
+	}()
+
+	return nil
+}
+
+// Stop the http server.
+func (i *Inbound) Stop() error {
+	if err := i.server.Shutdown(context.TODO()); err != nil {
```
troyronda: why context.TODO instead of context.Background?

(I assume you are planning to make this cancellable?)

pkg/framework/aries/framework.go
```diff
@@ -54,27 +56,51 @@ func New(opts ...Option) (*Aries, error) {
 		return nil, fmt.Errorf("default option initialization failed: %w", err)
 	}
 
+	// TODO -https://github.com/hyperledger/aries-framework-go/issues/212 Define clear relationship between framework
```
troyronda: The line width on this comment is a bit long for GitHub review panel.

pkg/didcomm/transport/http/inbound.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/6f6d39856a36112ab2f3116510c390f4c2855da9..1183264d25a8e484f015ea047b61c78c7a982cd5?diff=split#diff-18bf71ef9d572ae715312cab65be25eeR127)

pkg/framework/aries/framework.go
rolsonquadras: I think it depends on the view - unified or split view. In case of split, the lines wrap and no scrolling is required.

Issue #202, opened 2019-09-03
State Handler : Mock storage provider using In memory store in the service unit test 


talwinder50:
Mock storage provider using in memory store rather than levelDB storage

rolsonquadras:
@troyronda this has been addressed in #225. There was a typo with "closes", hence this is not closed automatically.

PR #225, opened 2019-09-06
test: Storage Provider/Store mock

rolsonquadras:
- New Storage provider and Store mock
- Moved outbound provider mock to /internal/mock folder

Cloes #202 


pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
-	mocktransport "github.com/hyperledger/aries-framework-go/pkg/internal/didcomm/transport/mock"
```
troyronda: do we really want all mocks in a single folder?

fqutishat: it's easier to find in one folder

rolsonquadras: Well, we can argue on both sides though. Personally, i feel folder structure becomes too complicated if we have a folder for each mock. anyways, the mocks are inside internal folder.

troyronda: even if you don’t want one folder per package, are we sure we shouldn’t separate mocks at the top level … e.g., didcomm

fqutishat: I think we should go with  pkg/internal/mocks/didcomm

rolsonquadras: @troyronda @fqutishat may be going down one level makes sense, i guess.

reg name - i feel putting mock at higher level than feature will give a good readability. People can go to mocks folder and see variable feature level mocks available.

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/257b15506fd69db7afc2e44467e044d9f3af5a18..9e430efccae39bf51273eafe671bffd5120dec70)

troyronda:
Should be rebased on master @rolsonquadras

Issue #218, opened 2019-09-06
Utility function to determine verifiable credential type from marshalled data


troyronda:
From https://github.com/hyperledger/aries-framework-go/pull/211#discussion_r321688986

Summary: When receiving a verifiable credential, the developer doesn't know which struct type to instantiate. We need to help them determine the type and provide a nice pattern to allow instantiation of the correct type.

PR #227, opened 2019-09-09
test: example selecting VC template based on basic one

kdimak:
According to @troyronda:
we should add an additional test case to provide an illustration of how a developer could choose which template to use.

decode into a basic verifiable credential
based on type info in the basic verifiable credential, decode into the more specialized type.
i.e., keep this test function that you created in this PR and add another one to illustrate the above.

Story: #218

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -187,7 +187,7 @@ type Credential struct {
 	RefreshService *RefreshService
 }
 
-// rawCredential
+// rawCredential is a basic verifiable credential
 type rawCredential struct {
 	Context        []string          `json:"@context,omitempty"`
 	ID             string            `json:"id,omitempty"`
@@ -294,7 +294,7 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error) 
 		return nil, err
 	}
 
-	var cred = crOpts.template()
+	cred := crOpts.template()
 	cred.Context = raw.Context
 	cred.ID = raw.ID
 	cred.Type = raw.Type
--- a/pkg/doc/verifiable/credential_test.go
+++ b/pkg/doc/verifiable/credential_test.go
@@ -7,6 +7,7 @@ package verifiable
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -25,7 +26,7 @@ const validCredential = `
   "id": "http://example.edu/credentials/1872",
   "type": [
     "VerifiableCredential",
-    "AlumniCredential"
+    "UniversityDegreeCredential"
   ],
   "credentialSubject": {
     "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
@@ -95,7 +96,7 @@ const issuerAsObject = `
 }
 `
 
-func TestNew(t *testing.T) {
+func TestNewCredential(t *testing.T) {
 	t.Run("test creation of new Verifiable Credential from JSON with valid structure", func(t *testing.T) {
 		vc, err := NewCredential([]byte(validCredential))
 		require.NoError(t, err)
@@ -112,7 +113,7 @@ func TestNew(t *testing.T) {
 		// validate type
 		require.Equal(t, vc.Type, []string{
 			"VerifiableCredential",
-			"AlumniCredential"})
+			"UniversityDegreeCredential"})
 
 		// validate not null credential subject
 		require.NotNil(t, vc.Subject)
@@ -156,6 +157,18 @@ func TestNew(t *testing.T) {
 		_, err := NewCredential([]byte("non json"))
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "JSON unmarshalling of verifiable credential failed")
+	})
+
+	t.Run("test a try to create a new Verifiable Credential with failing custom decoder", func(t *testing.T) {
+		_, err := NewCredential(
+			[]byte(validCredential),
+			WithDecoders([]CredentialDecoder{
+				func(dataJSON []byte, credential *Credential) error {
+					return errors.New("test decoding error")
+				},
+			}))
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "test decoding error")
 	})
 }
 
--- a/pkg/doc/verifiable/credential_extension_switch_test.go
+++ b/pkg/doc/verifiable/credential_extension_switch_test.go
@@ -1,0 +1,220 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package verifiable
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	validCred1 = `
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://www.w3.org/2018/credentials/examples/ext/type1"
+  ],
+  "id": "http://example.edu/credentials/1872",
+  "type": [
+    "VerifiableCredential",
+    "CredType1"
+  ],
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "s1": "custom subject 1"
+  },
+
+  "c1": "custom field 1",
+
+  "issuer": {
+    "id": "did:example:76e12ec712ebc6f1c221ebfeb1f",
+    "name": "Example University"
+  },
+
+  "issuanceDate": "2010-01-01T19:23:24Z"
+}
+`
+
+	validCred2 = `
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://www.w3.org/2018/credentials/examples/ext/type2"
+  ],
+  "id": "http://example.edu/credentials/1872",
+  "type": [
+    "VerifiableCredential",
+    "CredType2"
+  ],
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "s2": "custom subject 2"
+  },
+
+  "c2": "custom field 2",
+
+  "issuer": {
+    "id": "did:example:76e12ec712ebc6f1c221ebfeb1f",
+    "name": "Example University"
+  },
+
+  "issuanceDate": "2010-01-01T19:23:24Z"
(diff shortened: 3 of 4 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -320,7 +339,7 @@ func defaultCredentialOpts() *credentialOpts {
 		schemaDownloadClient: &http.Client{},
 		disabledCustomSchema: false,
 		decoders:             []CredentialDecoder{decodeIssuer},
-		template:             func() *Credential { return &Credential{} },
+		template:             func(credentialTypes []string) (*Credential, error) { return &Credential{}, nil },
```
troyronda: should the second return type be a boolean or an error?

pkg/doc/verifiable/credential_test.go
```diff
@@ -157,6 +158,28 @@ func TestNew(t *testing.T) {
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "JSON unmarshalling of verifiable credential failed")
 	})
+
```
troyronda: we should add an additional test case to provide an illustration of how a developer could choose which template to use.

1) decode into a basic verifiable credential
2) based on type info in the basic verifiable credential, decode into the more specialized type.

pkg/doc/verifiable/credential.go
```diff
@@ -219,7 +219,21 @@ type embeddedCompositeIssuer struct {
 type CredentialDecoder func(dataJSON []byte, credential *Credential) error
 
 // CredentialTemplate defines a factory method to create new Credential template.
-type CredentialTemplate func() *Credential
+// It's possible that Verifiable Credential could be not relevant to the Template's data model.
+// It can be determined by Template checking the types of the Verifiable Credential.
+// In this case error is returned.
+type CredentialTemplate func(credentialTypes []string) (*Credential, error)
```
llorllale: It is incorrect to determine the credential based just on a token encountered in the `type` field. You also need the `@context`.

Types are defined within `@context`s. The `VerifiableCredential` type is defined in the "https://www.w3.org/2018/credentials/v1" context (you can fetch that URL to verify this yourself). Likewise, additional types need to be defined in a context somewhere, and that includes adding it to the `@context` property.

pkg/doc/verifiable/credential.go
```diff
@@ -219,7 +219,21 @@ type embeddedCompositeIssuer struct {
...
+
+// ContainsTemplateType is a helper function to be used by Templates which extend base data model of Verifiable
+// Credential to identify that the types of latter contains target type of the Template.
+func ContainsTemplateType(credentialTypes []string, targetType string) bool {
```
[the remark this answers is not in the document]
kdimak: I have a feeling it can be used by other extensions of the VC data model. Why not export it?

pkg/doc/verifiable/credential.go
kdimak: I think that returning an error is more explanative than just boolean value. We can say why the template can't be applied.

pkg/doc/verifiable/credential_test.go
kdimak: I think there is the one at `extension_credential_test.go`. Do you this it's enough?

pkg/doc/verifiable/credential.go
kdimak: Hey @llorllale ! I agree with you that we need to pass `@context` as well as `type` so that we can avoid potential collision of same types defined in different contexts. I will change this. Thank you!

pkg/doc/verifiable/credential.go
kdimak: Well, after consideration I decided to make it internal. No need for making it public really.

pkg/doc/verifiable/credential.go
```diff
@@ -187,8 +187,8 @@ type Credential struct {
 	RefreshService *RefreshService
 }
 
-// rawCredential
-type rawCredential struct {
+// RawCredential is a basic verifiable credential
```
troyronda: I'm not excited about exporting a type name that starts with Raw. Is there a better description? Is this a BaseCredential?

pkg/doc/verifiable/credential_test.go
troyronda: @kdimak Does it show the above steps in sequence in a single test?

kdimak: @troyronda No, just because it's designed in a different way.
`udc := &UniversityDegreeCredential{}`
`cred, err := NewCredential(`
`	[]byte(validCredential),`
`	WithDecoders([]CredentialDecoder{udc.decode}),`
`	WithTemplate(udc.credential),`
`)`
We will get an error if the concrete VC is not supported by the VC extension.
I see your point here: in case we have several VC variants, we get basic VC and then decide what variant to use depending on the e.g. type (and context) of the basic VC.
I agree that it makes sense. Should we file a separate issue for this (as this goes beyond the current issue scope)?

pkg/doc/verifiable/credential.go
kdimak: Makes sense, also thought about this afterwards my last commit...

pkg/doc/verifiable/credential.go
```diff
@@ -220,7 +220,10 @@ type embeddedCompositeIssuer struct {
 type CredentialDecoder func(dataJSON []byte, credential *Credential) error
 
 // CredentialTemplate defines a factory method to create new Credential template.
-type CredentialTemplate func() *Credential
+// It's possible that Verifiable Credential could be not relevant to the Template's data model.
+// It can be determined by Template checking the types of the Verifiable Credential.
+// In this case error is returned.
+type CredentialTemplate func(credential *RawCredential) (*Credential, error)
```
llorllale: I don't understand why this template function needs the whole "raw credential" - it would appear it only needs the `@context` and the `type` properties as inputs, as evidenced [here](https://github.com/hyperledger/aries-framework-go/pull/227/files#diff-3f2d2cc11d10cd0a96b1574e700ae6dfR46)

kdimak: In my opinion, this is a more flexible approach - provide the basic VC to make a decision if the template is capable of extending it. Anyway, this approach can be changed, see the thread with @troyronda above (https://github.com/hyperledger/aries-framework-go/pull/227#discussion_r324332470).

pkg/doc/verifiable/credential_test.go
kdimak: hey @troyronda 
I added this extra test case, see **extension_credential_switch_test.go**.
I did not add config of proof validations enablement as for now we are not supporting proofs at all.

pkg/doc/verifiable/credential_test.go
```diff
@@ -95,7 +96,7 @@ const issuerAsObject = `
 }
 `
 
-func TestNew(t *testing.T) {
+func TestNewCredential(t *testing.T) { //nolint:funlen
```
Baha-sk: this function name doesn't seem to be long, there's no need for `nolint` here

pkg/doc/verifiable/extension_credential_switch_test.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+	}
+}
+
+func DecodeCredentials(producers ...CustomCredentialProducer) (chan<- []byte, <-chan interface{}, <-chan error) {
+	input := make(chan []byte)
+	output := make(chan interface{})
+	errors := make(chan error)
+
+	go func() {
```
troyronda: Why a new goroutine?

pkg/doc/verifiable/extension_credential_switch_test.go
```diff
@@ -0,0 +1,232 @@
...
+			return nil, err
+		},
+		Accept: func(vc *Credential) bool {
+			return containsString(vc.Context, "https://www.w3.org/2018/credentials/examples/ext/type2") &&
+				containsString(vc.Type, "CredType2")
+		},
+	}
+}
+
+func DecodeCredentials(producers ...CustomCredentialProducer) (chan<- []byte, <-chan interface{}, <-chan error) {
```
troyronda: Why do you need channels?

pkg/doc/verifiable/extension_credential_switch_test.go
```diff
@@ -0,0 +1,232 @@
...
+  ],
+  "id": "http://example.edu/credentials/1872",
+  "type": [
+    "VerifiableCredential"
+  ]
+}`
+)
+
+type CustomCredentialProducer struct {
+	Accept func(cred *Credential) bool
```
troyronda: Is there anything about this Accept/Apply pattern that should exist outside of the example/test (as part of the implementation)?

troyronda:
I think the commit description is now about providing an example rather than a utility function, at this point.

pkg/doc/verifiable/extension_credential_switch_test.go
kdimak: Yes, I think so.

pkg/doc/verifiable/extension_credential_switch_test.go
kdimak: Might be just for fun... Seriously, we can use function call here.

pkg/doc/verifiable/credential_test.go
kdimak: Right, Baha! It was longer at the previos version, now it's no need to use that.

kdimak:
> I think the commit description is now about providing an example rather than a utility function, at this point.

Agree with Troy! I've updated the task description.

pkg/doc/verifiable/extension_credential_switch_test.go
troyronda: Ok - please create a GitHub issue with the bits that should be moved into the implementation.

troyronda:
Remember the commit title too :).

pkg/doc/verifiable/extension_credential_switch_test.go
kdimak: Done! #263

pkg/doc/verifiable/extension_credential_switch_test.go
troyronda: As an example, we should show the simplest version that helps the reader understand - in this case understand how to use a basic credential to create a customized one.

kdimak: Simplified! And now it actually looks much clear and compact - so thanks for the suggestion @troyronda !

kdimak:
> Remember the commit title too :).

Done

Issue #216, opened 2019-09-05
Suggest: disallow naked returns under any circumstances


llorllale:
From [this comment](https://github.com/hyperledger/aries-framework-go/pull/211/files#r321359446), suggest we set `max-func-lines` to 0:

https://github.com/hyperledger/aries-framework-go/blob/0890946c897d13975d33fe02a87d396675d83b44/.golangci.yml#L59-L60

Issue #205, opened 2019-09-04
Enable go linter suite


troyronda:
Use a wider suite of go linter checks. E.g., https://github.com/golangci/golangci-lint/blob/master/.golangci.yml

llorllale:
Is #216 in scope of this story?

PR #235, opened 2019-09-10
Enable go linter suite

troyronda:
- Upgrade to golangci-lint v1.18
- Enable gochecknoglobals, nolen and dupl
- Disallow naked returns
- Resolve warnings

Noted TODOs to review global usage.

Closes: #205
Closes: #216

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,7 +60,7 @@ linters-settings:
   unparam:
     check-exported: false
   nakedret:
-    max-func-lines: 30
+    max-func-lines: 0
   gocritic:
     enabled-tags:
       - diagnostic
@@ -69,15 +69,24 @@ linters-settings:
       - opinionated
     disabled-checks:
       - unnamedResult
+  funlen:
+    lines: 60
+    statements: 40
 
 linters:
   enable-all: true
   disable:
     - maligned
     - prealloc
-    - dupl
-    - gochecknoglobals
 
 issues:
   exclude-use-default: false
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - dupl
+        - funlen
 
+  exclude:
+    # Allow package logger variables (for now)
+    - \`logger\` is a global variable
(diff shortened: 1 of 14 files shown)


.golangci.yml
```diff
@@ -44,7 +44,7 @@ linters-settings:
   maligned:
     suggest-new: true
   dupl:
-    threshold: 100
+    threshold: 150
```
troyronda: I reset to the default of 150.

.golangci.yml
```diff
@@ -69,15 +69,21 @@ linters-settings:
       - opinionated
     disabled-checks:
       - unnamedResult
+  funlen:
+    lines: 60
```
troyronda: 60 is the default.

pkg/framework/aries/default.go
```diff
@@ -21,9 +21,13 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/storage/leveldb"
 )
 
-// DBPath Level DB Path.
-var dbPath = "/tmp/peerstore/"
-var defaultInboundPort = ":8090"
+// TODO handle the test scenario better (make dbPath constant).
+//nolint:gochecknoglobals
+var (
```
fqutishat: should be const

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
```diff
@@ -26,9 +25,6 @@ const C20P = ContentEncryption("C20P") // Chacha20 encryption + Poly1035 authent
 // XC20P XChacha20Poly1035 algorithm
 const XC20P = ContentEncryption("XC20P") // XChacha20 encryption + Poly1035 authenticator cipher (192 bits nonce)
 
-// randReader is a cryptographically secure random number generator.
-var randReader = rand.Reader
```
Baha-sk: although not currently manipulated in tests, 
but this variable could be used to force random repeatable values in tests (inspired from go-jose).

troyronda: These things should be documented / commented. Global variables raises flags.

troyronda: Also - we shouldn't do these things until we actually are using them :).

.golangci.yml
```diff
@@ -69,15 +69,24 @@ linters-settings:
...
   enable-all: true
   disable:
     - maligned
     - prealloc
-    - dupl
-    - gochecknoglobals
 
 issues:
   exclude-use-default: false
+  exclude-rules:
```
troyronda: Any else that should be excluded from tests?

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
Baha-sk: sure.. it can be removed then

PR #236, opened 2019-09-10
feat: Create defulat Implementation for Wallet CreateSigningKey

fqutishat:
- Create Default Implementation for wallet
- Implement CreateSigningKey, GetSigningKey  method

closes #234

Signed-off-by: Firas Qutishat <firas.qutishat@securekey.com>

Issue #198, opened 2019-09-03
DID Exchange Inbound Message Events (service)


rolsonquadras:
Created issue based on discussion in https://wiki.hyperledger.org/display/ARIES/2019-09-03+Framework+Go+Weekly+Planning.

Framework should provide the ability to register events on inbound protocol messages. 

When protocol service [receives](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/dispatcher/dispatcher.go#L11) the message, the service should invoke the events registered by the [clients](https://github.com/hyperledger/aries-framework-go/tree/master/pkg/client).

llorllale:
This sounds like either it's a duplicate of #140 or depends on it

fqutishat:
duplicate of #140

llorllale:
Rough proposal, as discussed:

* expose identifiers for the different states ("Invited", "Requested", "Responded", "Completed")
* add `AddListener(string, <-chan dispatcher.DIDCommMsg) chan<- bool` to `didexchange.Service`

Usage:

```go
package user

func foo () {
    svc := /* get handle to didexchange.Service */
    incoming := make(<-chan dispatcher.DIDCommMsg)
    outgoing := svc.AddListener(didexchange.Invited, incoming)
    select {
    case msg <- incoming:
        // verify incoming invitation
        outgoing <- true    // aprove
    case <- time.After(100 * time.Second):
        // abort
    }
}
```

sudeshrshetty:
instead of `incoming := make(<-chan dispatcher.DIDCommMsg)`
we can create channel of struct type which includes 

- ID : for correlation
- Error: in case of some failure
- dispatcher.DIDCommMsg: actual message

llorllale:
Sounds good.

Note: `ID` would be for correlating *events* to the callback.

Issue #195, opened 2019-08-30
storage.Store.Get() should return `ErrNotFound` when item is not found


llorllale:
Similar to https://github.com/hyperledger/aries-framework-go/blob/ba35c80a2158fe9f3b991016dff1fe063f7ba973/pkg/framework/didresolver/api.go#L25

This would unblock this TODO:
https://github.com/hyperledger/aries-framework-go/blob/ba35c80a2158fe9f3b991016dff1fe063f7ba973/pkg/didmethod/peer/resolver_test.go#L96-L98

And is affecting this PR: https://github.com/hyperledger/aries-framework-go/pull/183/files (service.go, lines 132-135)

PR #237, opened 2019-09-10
feat: #231 skeleton - auto execution of did-exchange

### LINKED ISSUE #231: DIDExchange state machine: automatic execution
https://github.com/hyperledger/aries-framework-go/blob/4dd258fdba02e5b94510f9ae6595c385ec637046/pkg/didcomm/protocol/didexchange/service.go#L68-L70

The did-exchange state machine must execute actions automatically

llorllale:
Closes #231 

This PR provides the basic skeleton for the automatic state transitions as per https://github.com/hyperledger/aries-framework-go/issues/231#issuecomment-529939840.

Followup issues should implement the actions needed for each step, as well as registering/calling listeners.


pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -122,182 +122,165 @@ func TestSendResponse(t *testing.T) {
...
+			ID:    thid,
 			Label: "Bob",
 			Connection: &Connection{
 				DID: "B.did@B:A",
 			},
 		})
 	require.NoError(t, err)
-
-	msg = dispatcher.DIDCommMsg{Type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0/request", Payload: payloadBytes} //nolint:lll
+	msg := dispatcher.DIDCommMsg{Type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0/request", Payload: payloadBytes} //nolint:lll
```
troyronda: In several cases I am seeing a constant used for the DIDComm type, but in a couple cases not?

llorllale: What level of consistency are we looking for? Across the whole test file, within a test function... ?

llorllale: I guess I'll use literals then

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -102,3 +162,23 @@ func (s *completed) Name() string {
...
 }
+
+func (s *completed) Execute(msg dispatcher.DIDCommMsg) (state, error) {
+	switch msg.Type {
+	case connectionResponse:
+		if msg.Outbound {
+			return nil, fmt.Errorf("outbound responses are not allowed for state %s", s.Name())
+		}
+		// send ACK
+		return nil, nil
```
troyronda: I am seeing return nil, nil several times in the state machine. I also see a null struct. Is this supposed to be return *null{}, nil?

(I'm a bit confused.)

llorllale: nope - `null` != `nil`. Want me to rename `null`?

There are implicit TODOs in this file - like the line above that says "send ACK". If that results in an error then the return would change to `return nil, err`

pkg/didcomm/protocol/didexchange/service_test.go
llorllale: Fixed

troyronda: I had thought constants were the right way to go :)

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -164,10 +181,12 @@ func (s *Service) SendExchangeResponse(exchangeResponse *Response, destination s
 }
 
 func (s *Service) marshalAndSend(data interface{}, errorMsg, destination string) (string, error) {
+	// TODO need access to the wallet in order to first pack() the msg before sending
```
troyronda: It would be good to have an outbound sender function injected, and that function takes care of calling pack.

troyronda: As we just discussed, we also need a way to map a URI to an outbound transport.

i.e., an outbound dispatcher that does both the mapping and the packing.

pkg/didcomm/protocol/didexchange/service_test.go
llorllale: Ok. I'll revert to constants everywhere except the `TestService_Accept` test cases - I believe those should stay exactly as they are.

pkg/didcomm/protocol/didexchange/states.go
llorllale: As per our discussion - I added a new special state `noOp` to aid clarity in states.go in cases where we previously had `return nil, nil`

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -43,6 +53,21 @@ func stateFromName(name string) (state, error) {
...
+func (s *noOp) Name() string {
+	return "noop"
+}
+
+func (s *noOp) CanTransitionTo(_ state) bool {
+	return false
+}
+
+func (s *noOp) Execute(_ dispatcher.DIDCommMsg) (state, error) {
+	return nil, nil
```
troyronda: i would suggest error should be returned in this case.

pkg/didcomm/protocol/didexchange/states_test.go
```diff
@@ -9,6 +9,8 @@ package didexchange
 import (
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+
```
troyronda: the framework imports have been grouped after third party in other files.

pkg/didcomm/protocol/didexchange/states.go
llorllale: done

pkg/didcomm/protocol/didexchange/states_test.go
llorllale: done (are there any linters for this?)

llorllale:
@talwinder50 

> Can you explain why we need specific outbound bool parameter in the DIDCommMsg.

It's because different actions are required to be performed at each state depending on your role (inviter vs. invitee). This PR determines the role by the direction of the message (outbound vs. inbound). It might be worthwhile to look and see if storing a "role" attribute would make things clearer...

rolsonquadras:
> Followup issues should implement the actions needed for each step, as well as registering/calling listeners.
@llorllale registering/calling listeners - Is this same as #198 ? Just want to confirm.

rolsonquadras:
> @talwinder50
> 
> > Can you explain why we need specific outbound bool parameter in the DIDCommMsg.
> 
> It's because different actions are required to be performed at each state depending on your role (inviter vs. invitee). This PR determines the role by the direction of the message (outbound vs. inbound). It might be worthwhile to look and see if storing a "role" attribute would make things clearer...

The outbound flag is inside the DIDCommMsg struct. Are we expecting clients to invoke Handle() function even for outbound requests ?

llorllale:
@rolsonquadras I expect the state machine to send messages, yes. It should take care of all this machinery for you. When your role is Inviter (aka. "protocol initiator") then you can use `SendExchangeRequest` - this should delegate to `Service.Handle`

rolsonquadras:
> @rolsonquadras I expect the state machine to send messages, yes. It should take care of all this machinery for you. When your role is Inviter (aka. "protocol initiator") then you can use `SendExchangeRequest` - this should delegate to `Service.Handle`

My understanding of the flow from this PR and other discussion on events/listeners:

1. Handle() function is triggered when agent receives the invitation.
2. Event is triggered on this.
3. A callback event triggers #198 and control comes to service executing the state transitions.
4. The service(without going to client pkg) calls SendRequest() which in-turn will call handle() and we exclude outbound messages from event. 
5. Invoke outbound.Send() after pack/encrypt.

@llorllale @talwinder50 correct me, if it's wrong.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -65,17 +65,33 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 		return fmt.Errorf("invalid state transition: %s -> %s", current.Name(), next.Name())
 	}
 	// TODO: call pre-transition listeners -  Issue: https://github.com/hyperledger/aries-framework-go/issues/140
-	// TODO execute actions for the current state *and the next (if any)*. Eg. if we receive an
-	//      invitation, execute actions for 'invited' state and then execute actions for
-	//      'requested' state. Implement the relevant actions for each state in state.go.
+	followup, err := next.Execute(msg)
```
fqutishat: increase code coverage to 85%

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -43,6 +53,21 @@ func stateFromName(name string) (state, error) {
 	return nil, fmt.Errorf("invalid state name %s", name)
 }
 
+type noOp struct {
+}
+
+func (s *noOp) Name() string {
```
fqutishat: increase code coverage

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -65,17 +65,33 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+		return err
+	}
 	err = s.update(thid, next)
 	if err != nil {
 		return err
 	}
+	for ; !isNoOp(followup); followup = next {
+		next, err = followup.Execute(msg)
+		if err != nil {
+			return err
```
fqutishat: return err with message

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -65,17 +65,33 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+		}
+		err = s.update(thid, followup)
+		if err != nil {
+			return err
```
fqutishat: https://github.com/hyperledger/aries-framework-go/pull/237/files#r323339449

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -65,17 +65,33 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
-	// TODO execute actions for the current state *and the next (if any)*. Eg. if we receive an
-	//      invitation, execute actions for 'invited' state and then execute actions for
-	//      'requested' state. Implement the relevant actions for each state in state.go.
+	followup, err := next.Execute(msg)
+	if err != nil {
+		return err
+	}
 	err = s.update(thid, next)
 	if err != nil {
 		return err
```
fqutishat: https://github.com/hyperledger/aries-framework-go/pull/237/files#r323339449

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -65,17 +65,33 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 		return fmt.Errorf("invalid state transition: %s -> %s", current.Name(), next.Name())
 	}
 	// TODO: call pre-transition listeners -  Issue: https://github.com/hyperledger/aries-framework-go/issues/140
-	// TODO execute actions for the current state *and the next (if any)*. Eg. if we receive an
-	//      invitation, execute actions for 'invited' state and then execute actions for
-	//      'requested' state. Implement the relevant actions for each state in state.go.
+	followup, err := next.Execute(msg)
+	if err != nil {
+		return err
```
fqutishat: https://github.com/hyperledger/aries-framework-go/pull/237/files#r323339449

llorllale:
@fqutishat I believe I addressed your concerns - please check

llorllale:
@rolsonquadras 

> My understanding of the flow from this PR and other discussion on events/listeners:
> 
> 1. Handle() function is triggered when agent receives the invitation.
> 2. Event is triggered on this.
> 3. A callback event triggers #198 and control comes to service executing the state transitions.
> 4. The service(without going to client pkg) calls SendRequest() which in-turn will call handle() and we exclude outbound messages from event.
> 5. Invoke outbound.Send() after pack/encrypt.

Yes, that is more or less the inbound flow.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -122,182 +122,228 @@ func TestSendResponse(t *testing.T) {
...
+		require.NoError(t, err)
+		err = s.Handle(dispatcher.DIDCommMsg{Type: connectionResponse, Outbound: false, Payload: response})
+		require.Error(t, err)
+	})
+	t.Run("error when updating store on first state transition", func(t *testing.T) {
+		s := &Service{
+			outboundTransport: newMockOutboundTransport(),
+			store: &mockStore{
+				get: func(string) ([]byte, error) {
+					return nil, errors.New("not found")
```
Baha-sk: we should probably use a preset Not Found error, like this one: https://github.com/hyperledger/aries-framework-go/pull/236/files#diff-f761c91e753eba1c42a68076df06ce12R97

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -122,182 +122,228 @@ func TestSendResponse(t *testing.T) {
...
-				ID:   randomString(),
-				ConnectionSignature: &ConnectionSignature{
-					Type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single",
+	t.Run("error when updating store on followup state transition", func(t *testing.T) {
+		counter := 0
+		s := &Service{
+			outboundTransport: newMockOutboundTransport(),
+			store: &mockStore{
+				get: func(string) ([]byte, error) {
+					return nil, errors.New("not found")
```
Baha-sk: idem

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -378,6 +435,27 @@ func TestService_update(t *testing.T) {
...
+	data := make(map[string][]byte)
+	return &mockStore{
+		put: func(k string, v []byte) error {
+			data[k] = v
+			return nil
+		},
+		get: func(k string) ([]byte, error) {
+			v, found := data[k]
+			if !found {
+				return nil, errors.New("not found")
```
Baha-sk: same here

pkg/didcomm/dispatcher/dispatcher.go
```diff
@@ -15,6 +15,7 @@ type Service interface {
 
 // DIDCommMsg did comm msg
 type DIDCommMsg struct {
-	Type    string
-	Payload []byte
+	Outbound bool
```
rolsonquadras: Godoc would be helpful as what value need to be set here.

pkg/didcomm/protocol/didexchange/service_test.go
llorllale: #195 was opened for that

pkg/didcomm/protocol/didexchange/service_test.go
llorllale: #195 was opened for that

pkg/didcomm/protocol/didexchange/service_test.go
llorllale: #195

pkg/didcomm/dispatcher/dispatcher.go
llorllale: done

fqutishat: I think we don't need Outbound bool flag as from the msgtype you would know the state and follow up state

llorllale: @fqutishat this comment illustrates why it's useful to know the direction the message is heading in: https://github.com/hyperledger/aries-framework-go/issues/231#issuecomment-529939840

To be fair - what we *actually* need to know is what role we're participating in in this protocol. This PR uses the message's direction as a proxy for that. But maybe it's worthwhile to look into how to just store our role (as well as state) instead of changing `dispatcher.DIDCommMsg`. I'll look into this in a couple of hours.

llorllale: @fqutishat I think it's *crucial* to know the message's direction.

Look again at this comment: https://github.com/hyperledger/aries-framework-go/issues/231#issuecomment-529939840

Now let's ask ourselves: `Service.Handle(msg)` just received a `Request` message. It doesn't know whether it's inbound or outbound. It doesn't have any context persisted to the store, so okay - it knows this is a new instance of the did-exchange protocol. What should it do now - should it just call the listeners, or should it send the `Request` message to the other agent?

PR #240, opened 2019-09-11
feat: DID Exchange Inbound Message Event

### LINKED ISSUE #198: DID Exchange Inbound Message Events (service)

rolsonquadras:
- Support to register action events(go channels) for inbound DIDExchange message types
- Support to register for auto execution of message processing
- Support to register stream events(go channels) for inbound DIDExchange message types
- Support for callback for action events

Closes #198 

### DIFF
--- a/cmd/aries-agentd/main_test.go
+++ b/cmd/aries-agentd/main_test.go
@@ -103,7 +103,7 @@ func validateRequests(t *testing.T) {
 								"@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0/invitation"
 							}`),
 				"application/didcomm-envelope-enc"),
-			expectedStatus:     http.StatusAccepted,
+			expectedStatus:     http.StatusInternalServerError,
 			expectResponseData: false,
 		},
 	}
--- a/pkg/didcomm/event/event.go
+++ b/pkg/didcomm/event/event.go
@@ -1,0 +1,26 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package event
+
+import "github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+
+// DIDCommEvent message type to pass events in go channels.
+type DIDCommEvent struct {
+	// DIDComm message
+	Message dispatcher.DIDCommMsg
+	// Callback function to be called by the consumer for further processing the message.
+	Callback Callback
+}
+
+// DIDCommCallback message type to pass service callback in go channels.
+type DIDCommCallback struct {
+	// Set the value in case of any error while processing the DIDComm message event by the consumer.
+	Err error
+}
+
+// Callback type to pass service callbacks.
+type Callback func(DIDCommCallback)
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -11,24 +11,49 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sync"
 
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/common/metadata"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/event"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
+// TODO https://github.com/hyperledger/aries-framework-go/issues/104
+var logger = log.New("aries-framework/didexchange")
+
+// didCommChMessage type to correlate actionEvent message(go channel) with callback message(internal go channel).
+type didCommChMessage struct {
+	ID              string
+	DIDCommCallback event.DIDCommCallback
+}
+
 const (
 	// DIDExchange did exchange protocol
 	DIDExchange = "didexchange"
 	// DIDExchangeSpec defines the did-exchange spec
-	DIDExchangeSpec    = metadata.AriesCommunityDID + ";spec/didexchange/1.0/"
-	connectionInvite   = DIDExchangeSpec + "invitation"
-	connectionRequest  = DIDExchangeSpec + "request"
-	connectionResponse = DIDExchangeSpec + "response"
-	connectionAck      = DIDExchangeSpec + "ack"
+	DIDExchangeSpec = metadata.AriesCommunityDID + ";spec/didexchange/1.0/"
+	// ConnectionInvite defines the did-exchange invite message type.
+	ConnectionInvite = DIDExchangeSpec + "invitation"
+	// ConnectionRequest defines the did-exchange request message type.
+	ConnectionRequest = DIDExchangeSpec + "request"
+	// ConnectionResponse defines the did-exchange response message type.
+	ConnectionResponse = DIDExchangeSpec + "response"
+	// ConnectionAck defines the did-exchange ack message type.
+	ConnectionAck = DIDExchangeSpec + "ack"
 )
+
+// message type to store data for eventing. This is retrieved during callback.
+type message struct {
+	Msg           dispatcher.DIDCommMsg
+	ThreadID      string
+	NextStateName string
+}
 
 // provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
 type provider interface {
@@ -39,15 +64,43 @@ type provider interface {
 type Service struct {
 	outboundTransport transport.OutboundTransport
 	store             storage.Store
+	callbackChannel   chan didCommChMessage
+	actionEvent       chan<- event.DIDCommEvent
+	statusEvents      []chan<- dispatcher.DIDCommMsg
+	lock              sync.RWMutex
+	statusEventLock   sync.RWMutex
+	execute           bool
 }
 
 // New return didexchange service
 func New(store storage.Store, prov provider) *Service {
-	return &Service{outboundTransport: prov.OutboundTransport(), store: store}
+	svc := &Service{
+		// TODO Outbound dispatcher - https://github.com/hyperledger/aries-framework-go/issues/259
+		outboundTransport: prov.OutboundTransport(),
+		store:             store,
+		// TODO channel size - https://github.com/hyperledger/aries-framework-go/issues/246
+		callbackChannel: make(chan didCommChMessage, 10),
+		// set execute to false. Consumers have to enable this by setting either RegisterEvent() or
+		// RegisterAutoExecute()
+		execute: false,
+	}
+
+	svc.startInternalListener()
+
+	return svc
 }
 
 // Handle didexchange msg
 func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
+	// throw error if there are no action events are registered or auto execute set (if it's not outbound)
+	s.lock.RLock()
+	execute := s.execute
+	s.lock.RUnlock()
+
+	if !msg.Outbound && !execute {
+		return errors.New("no clients are registered to handle the message")
+	}
+
 	thid, err := threadID(msg.Payload)
 	if err != nil {
 		return err
@@ -63,26 +116,215 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 	if !current.CanTransitionTo(next) {
 		return fmt.Errorf("invalid state transition: %s -> %s", current.Name(), next.Name())
 	}
-	// TODO: call pre-transition listeners -  Issue: https://github.com/hyperledger/aries-framework-go/issues/140
-	followup, err := next.Execute(msg)
+
+	eventsTriggered := false
+	if !msg.Outbound {
+		// trigger the actionEvent if the message type is registered (if it's not outbound)
+		eventsTriggered, err = s.sendEvent(msg, thid, next)
+		if err != nil {
+			return fmt.Errorf("send events failed: %w", err)
+		}
+	}
+
+	// if no events are trigger continue the execution
+	if !eventsTriggered {
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	}
+
+	return nil
+}
+
+// RegisterEvent on DID Exchange protocol messages. The events are triggered for incoming ConnectionRequest,
+// ConnectionResponse or ConnectionAck message types. The consumer need to invoke the callback to resume processing.
+// Only one channel can be registered for the action events. If called multiple times, the events will be sent to the
+// last channel. This works in conjunction with RegisterAutoExecute(). If this function is called after
+// RegisterAutoExecute(), the service will not execute the processing automatically and it'll wait for the callback. If
+// both are not set, then service will not handle the messages and throw an error.
+func (s *Service) RegisterEvent(ch chan<- event.DIDCommEvent) error {
+	s.lock.Lock()
+	s.actionEvent = ch
+	s.execute = true
+	s.lock.Unlock()
+
+	return nil
+}
+
+// UnregisterEvent on DID Exchange protocol messages. Refer RegisterEvent().
+func (s *Service) UnregisterEvent() error {
+	return s.disableExecute()
+}
+
+// RegisterMsg on DID Exchange protocol messages. The events are triggered for incoming ConnectionRequest,
+// ConnectionResponse or ConnectionAck message types. The Callback is set to nil in the actionEvent message and service
+// won't be expecting any response from the consumers.
+func (s *Service) RegisterMsg(ch chan<- dispatcher.DIDCommMsg) error {
+	s.statusEventLock.Lock()
+	s.statusEvents = append(s.statusEvents, ch)
+	s.statusEventLock.Unlock()
+
+	return nil
+}
+
+// UnregisterMsg on DID Exchange protocol messages. Refer RegisterMsg().
+func (s *Service) UnregisterMsg(ch chan<- dispatcher.DIDCommMsg) error {
+	s.statusEventLock.Lock()
+	for i := 0; i < len(s.statusEvents); i++ {
+		if s.statusEvents[i] == ch {
+			s.statusEvents = append(s.statusEvents[:i], s.statusEvents[i+1:]...)
+			i--
+		}
+	}
+	s.statusEventLock.Unlock()
+
+	return nil
+}
+
+// RegisterAutoExecute on DID Exchange protocol messages. When this function is called, the service will auto execute
+// the workflow. This works in conjunction with RegisterEvent(). If this function is called after
+// RegisterEvent(), the service will execute the processing automatically and no action events will be triggered.
+// If both are not set, then service will not handle the messages and throw an error.
+func (s *Service) RegisterAutoExecute() error {
+	s.lock.Lock()
+	s.execute = true
+	s.lock.Unlock()
+
+	return nil
+}
+
+// UnregisterAutoExecute on DID Exchange protocol messages. Refer RegisterAutoExecute().
+func (s *Service) UnregisterAutoExecute() error {
+	return s.disableExecute()
+}
+
+func (s *Service) disableExecute() error {
+	s.lock.Lock()
+	s.actionEvent = nil
+	s.execute = false
+	s.lock.Unlock()
+
+	return nil
+}
+
+func (s *Service) handle(msg *message) error {
+	next, err := stateFromName(msg.NextStateName)
+	if err != nil {
+		return fmt.Errorf("invalid state name: %w", err)
+	}
+
+	followup, err := next.Execute(msg.Msg)
 	if err != nil {
 		return fmt.Errorf("failed to execute state %s %w", next.Name(), err)
 	}
-	err = s.update(thid, next)
+	err = s.update(msg.ThreadID, next)
 	if err != nil {
 		return fmt.Errorf("failed to persist state %s %w", next.Name(), err)
 	}
 	for ; !isNoOp(followup); followup = next {
-		next, err = followup.Execute(msg)
+		next, err = followup.Execute(msg.Msg)
 		if err != nil {
 			return fmt.Errorf("failed to execute state %s %w", followup.Name(), err)
 		}
-		err = s.update(thid, followup)
+		err = s.update(msg.ThreadID, followup)
 		if err != nil {
 			return fmt.Errorf("failed to persist state %s %w", followup.Name(), err)
 		}
 	}
 	// TODO call post-transition listeners -  Issue: https://github.com/hyperledger/aries-framework-go/issues/140
+	return nil
+}
+
+// sendEvent triggeres the status events and action events. Returns true if action events are triggered. The events are
+// triggered for ConnectionRequest, ConnectionResponse or ConnectionAck message types.
+func (s *Service) sendEvent(msg dispatcher.DIDCommMsg, threadID string, nextState state) (bool, error) {
+	// trigger the status events
+	s.statusEventLock.RLock()
+	statusEvents := s.statusEvents
+	s.statusEventLock.RUnlock()
+
+	for _, handler := range statusEvents {
+		handler <- msg
+	}
+
+	// invoke events for ConnectionRequest, ConnectionResponse or ConnectionAck and if action events are registered
+	if s.actionEvent == nil || msg.Type != ConnectionRequest &&
+		msg.Type != ConnectionResponse && msg.Type != ConnectionAck {
+		return false, nil
+	}
+
+	jsonDoc, err := json.Marshal(&message{
+		Msg:           msg,
+		ThreadID:      threadID,
+		NextStateName: nextState.Name(),
+	})
+
+	if err != nil {
+		return false, fmt.Errorf("JSON marshalling of document failed: %w", err)
+	}
+
+	// save the incoming message in the store (to retrieve later when callback events are fired)
+	id := generateRandomID()
+	err = s.store.Put(id, jsonDoc)
+	if err != nil {
+		return false, fmt.Errorf("JSON marshalling of document failed: %w", err)
+	}
+
+	// create the message for the channel
+	didCommEvent := event.DIDCommEvent{
+		Message: msg,
+		Callback: func(didCommCallback event.DIDCommCallback) {
+			s.processCallback(id, didCommCallback)
+		},
+	}
+
+	// trigger the registered action events
+	s.lock.RLock()
+	event := s.actionEvent
+	s.lock.RUnlock()
+	event <- didCommEvent
+
+	return true, nil
+}
+
+// startInternalListener listens to messages in gochannel for callback messages from clients.
+func (s *Service) startInternalListener() {
+	go func() {
+		for msg := range s.callbackChannel {
+			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
+			s.process(msg.ID, msg.DIDCommCallback)
+		}
+	}()
+}
+
+func (s *Service) processCallback(id string, didCommCallback event.DIDCommCallback) {
+	// pass the callback data to internal channel. This is created to unblock consumer go routine and wrap the callback
+	// channel internally.
+	s.callbackChannel <- didCommChMessage{ID: id, DIDCommCallback: didCommCallback}
+}
+
+// processCallback processes the callback events.
+func (s *Service) process(id string, didCommCallback event.DIDCommCallback) error {
+	if didCommCallback.Err != nil {
+		// TODO https://github.com/hyperledger/aries-framework-go/issues/242 Service callback processing error handling
+		return nil
+	}
+
+	// fetch the record
+	jsonDoc, err := s.store.Get(id)
+	if err != nil {
+		return fmt.Errorf("document for the id doesn't exists in the database: %w", didCommCallback.Err)
+	}
+
+	document := &message{}
+	err = json.Unmarshal(jsonDoc, document)
+	if err != nil {
+		return fmt.Errorf("JSON marshalling failed: %w", didCommCallback.Err)
+	}
+
+	// continue the processing
+	err = s.handle(document)
+	if err != nil {
+		return fmt.Errorf("processing of the message failed: %w", err)
+	}
 	return nil
 }
 
@@ -128,10 +370,10 @@ func (s *Service) update(thid string, state state) error {
 
 // Accept msg checks the msg type
 func (s *Service) Accept(msgType string) bool {
-	return msgType == connectionInvite ||
-		msgType == connectionRequest ||
-		msgType == connectionResponse ||
-		msgType == connectionAck
+	return msgType == ConnectionInvite ||
+		msgType == ConnectionRequest ||
+		msgType == ConnectionResponse ||
+		msgType == ConnectionAck
 }
 
 // Name return service name
@@ -162,7 +404,7 @@ func (s *Service) SendExchangeRequest(exchangeRequest *Request, destination stri
 		return errors.New("exchangeRequest cannot be nil")
 	}
 
-	exchangeRequest.Type = connectionRequest
+	exchangeRequest.Type = ConnectionRequest
 
 	// ignore response data as it is not used in this communication mode as defined in the spec
 	_, err := s.marshalAndSend(exchangeRequest, "Error Marshalling Exchange Request", destination)
@@ -175,7 +417,7 @@ func (s *Service) SendExchangeResponse(exchangeResponse *Response, destination s
 		return errors.New("exchangeResponse cannot be nil")
 	}
 
-	exchangeResponse.Type = connectionResponse
+	exchangeResponse.Type = ConnectionResponse
 
 	// ignore response data as it is not used in this communication mode as defined in the spec
 	_, err := s.marshalAndSend(exchangeResponse, "Error Marshalling Exchange Response", destination)
@@ -192,8 +434,12 @@ func (s *Service) marshalAndSend(data interface{}, errorMsg, destination string)
 	return s.outboundTransport.Send(string(jsonString), destination)
 }
 
+func generateRandomID() string {
+	return uuid.New().String()
+}
+
 func encodedExchangeInvitation(inviteMessage *Invitation) (string, error) {
-	inviteMessage.Type = connectionInvite
+	inviteMessage.Type = ConnectionInvite
 
 	invitationJSON, err := json.Marshal(inviteMessage)
 	if err != nil {
--- a/pkg/didcomm/protocol/didexchange/service_test.go
+++ b/pkg/didcomm/protocol/didexchange/service_test.go
@@ -10,24 +10,25 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/event"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
 	"github.com/hyperledger/aries-framework-go/pkg/internal/mock/didcomm"
+	mockstorage "github.com/hyperledger/aries-framework-go/pkg/internal/mock/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
-	"github.com/hyperledger/aries-framework-go/pkg/storage/leveldb"
 )
 
 const (
 	destinationURL  = "https://localhost:8090"
 	successResponse = "success"
+	invalidThreadID = "invalidThreadID"
 )
 
 func TestGenerateInviteWithPublicDID(t *testing.T) {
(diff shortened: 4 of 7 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -64,18 +96,138 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+			case msg := <-s.callback[connectionAck]:
+				s.processCallback(msg)
+			}
+		}
+	}()
+}
+
+// processCallback processes the callback events.
+func (s *Service) processCallback(didCommCallback event.DIDCommCallback) {
+	// TODO what to do in case of any failures with callback processing ? - currently, logging the error messages and
```
llorllale: "What do do?"

Probably communicate the error via `DIDCommEvent` (adding an "Err" attribute maybe?)?

llorllale: Or perhaps we should have a separate stream of errors encountered.

pkg/didcomm/event/event.go
```diff
@@ -0,0 +1,21 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package event
+
+import "github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+
+// DIDCommEvent message type to pass events in go channels.
+type DIDCommEvent struct {
```
llorllale: I think both these new structs need more extensive godocs, including their attributes.

Eg. clearly specify what `DIDCommCallback.Err` does.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -64,18 +96,138 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	err = s.update(msg.ThreadID, nextState)
 	if err != nil {
 		return err
 	}
 	// TODO call post-transition listeners -  Issue: https://github.com/hyperledger/aries-framework-go/issues/140
 	return nil
 }
 
+// Register events on DID Exchange protocol.
+func (s *Service) Register(msgType string, ch chan<- event.DIDCommEvent) (chan<- event.DIDCommCallback, error) {
```
llorllale: Looks like we need to export the msgTypes as well.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Created a issue for the TODO https://github.com/hyperledger/aries-framework-go/issues/242

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Valid point. [Updated](https://github.com/hyperledger/aries-framework-go/compare/35ffd11efb21de8bceb7b468e53a833a2b1a8f82..e20d32d8200881f1bfdc2f36aad1988fa5a78e93#diff-f761c91e753eba1c42a68076df06ce12R34).

pkg/didcomm/event/event.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/35ffd11efb21de8bceb7b468e53a833a2b1a8f82..e20d32d8200881f1bfdc2f36aad1988fa5a78e93#diff-d3e84998ff09ae2ecd4128d1239a97e7R24).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -40,11 +57,31 @@ type provider interface {
...
 	store             storage.Store
+	events            map[string][]chan<- event.DIDCommEvent
+	callback          map[string]chan event.DIDCommCallback
 }
 
 // New return didexchange service
 func New(store storage.Store, prov provider) *Service {
-	return &Service{outboundTransport: prov.OutboundTransport(), store: store}
+	// create channels for callback
+	// TODO configure callback channel size
```
rolsonquadras: Issue #246

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +85,60 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+
+// Register events on DID Exchange protocol.
+func (s *Service) Register(msgType string, ch chan<- event.DIDCommEvent) error {
+	if msgType != ConnectionInvite && msgType != ConnectionRequest &&
+		msgType != ConnectionResponse && msgType != ConnectionAck {
+		return fmt.Errorf("event registration not supported for the message type %s", msgType)
+	}
+
+	if _, ok := s.events[msgType]; ok {
+		s.events[msgType] = append(s.events[msgType], ch)
```
bstasyszyn: I think that you'll need a mutex around s.events.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -478,30 +486,354 @@ func (p *mockProvider) OutboundTransport() transport.OutboundTransport {
...
+	go func() {
+		for {
+			select {
+			// receive the event on ConnectionRequest message type
+			case e := <-invitationChannel:
+				require.NotEmpty(t, e.Message)
+				require.Equal(t, ConnectionInvite, e.Message.Type)
+
+				// invoke callback
+				err = e.Callback(event.DIDCommCallback{Err: nil})
```
bstasyszyn: When Callback is invoked, processCallback will be invoked and handle(...) will now run in this Go routine (i.e. the client's Go routine). This Go routine will be blocked until the flow has completed which may result in deadlock if the client needs to handle more events.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +85,60 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+}
+
+// Register events on DID Exchange protocol.
+func (s *Service) Register(msgType string, ch chan<- event.DIDCommEvent) error {
+	if msgType != ConnectionInvite && msgType != ConnectionRequest &&
+		msgType != ConnectionResponse && msgType != ConnectionAck {
+		return fmt.Errorf("event registration not supported for the message type %s", msgType)
+	}
+
+	if _, ok := s.events[msgType]; ok {
```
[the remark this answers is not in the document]
llorllale: I'm suggesting the statement on line 114 is sufficient for both cases - when the key `msgType` already exists or when it does not.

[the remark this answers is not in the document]
rolsonquadras: Got it, Thanks. Updated in the new [Patch](https://github.com/hyperledger/aries-framework-go/compare/2fc8ca5f292de18ad5657f1f1317e24252476a3c..9e40655340de2e0f5c4fc9f2074c4f8cd3e1126d#diff-f761c91e753eba1c42a68076df06ce12R116).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/2fc8ca5f292de18ad5657f1f1317e24252476a3c..9e40655340de2e0f5c4fc9f2074c4f8cd3e1126d#diff-f761c91e753eba1c42a68076df06ce12R61) mutex.

pkg/didcomm/protocol/didexchange/service_test.go
rolsonquadras: @bstasyszyn Very valid point. I was thinking to reduce the work for consumers to call the go-channel as they had to map the correlation keys.

For this, may be we can start a go routine or invoke channel inside call back function. So that it can be returned the control back immediately. If this is not a blocker for now, can we tackle this issue in next PRs/issue ? I can create the issue for this.

rolsonquadras:
1. Remove go routine - (https://github.com/hyperledger/aries-framework-go/pull/240/files#diff-f761c91e753eba1c42a68076df06ce12R179)
2. Non-block processCallback
3. Have a  errorEvent
4. Client deals with one channel (event). 
5. Hide all other async details from clients. eg - callback appears to be blocking to the client even though its non-blocking.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +87,58 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	// if no events are trigger continue the execution
+	if !eventsTriggered {
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	}
+
+	return nil
+}
+
+// Register events on DID Exchange protocol.
+func (s *Service) Register(msgType string, ch chan<- event.DIDCommEvent) error {
```
troyronda: This function is currently conflating five activities:

1. Register a Go channel to listen to the event stream.
2. Register a Go channel to listen to workflow steps that require an action.
3. Set which events require an explicit callback action to progress the state machine.
4. Allow a different Go channel for each event type.
5. Filter the events that are returned to the client.

I argue that:
- (3) should have been done prior to the protocol starting.
- (4) seems unnecessary.
- (5) I could see some arguments for filtering... but I think the caller code looks the same even with filtering... and this is all local (not filtering a remote stream)...

troyronda: Also what is hidden behind the pattern of passing a channel as an argument is that:
- the caller can control the channel buffer size.
- the caller can use the same channel against multiple protocols.

pkg/didcomm/protocol/didexchange/service_test.go
rolsonquadras: [Wrapped](https://github.com/hyperledger/aries-framework-go/compare/9e40655340de2e0f5c4fc9f2074c4f8cd3e1126d..686bbecbf5c870012cfbab30d759addb432f52e9#diff-f761c91e753eba1c42a68076df06ce12R207) internal go channel inside the callback function for non-blocking callback from the consumer.

rolsonquadras:
> 1. Remove go routine - (https://github.com/hyperledger/aries-framework-go/pull/240/files#diff-f761c91e753eba1c42a68076df06ce12R179)
> 2. Non-block processCallback
> 3. Have a  errorEvent
> 4. Client deals with one channel (event).
> 5. Hide all other async details from clients. eg - callback appears to be blocking to the client even though its non-blocking.

Addressed all the above comments except point#3 (TODO #242 )

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -86,6 +147,75 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 	return nil
 }
 
+func (s *Service) sendEvent(msg dispatcher.DIDCommMsg, threadID string, nextState state) (bool, error) {
+	invoked := false
+	if _, ok := s.events[msg.Type]; ok {
```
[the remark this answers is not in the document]
rolsonquadras: noted.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -39,11 +61,22 @@ type provider interface {
...
 	store             storage.Store
+	ch                chan didCommChMessage
+	event             []chan<- event.DIDCommEvent
 }
 
 // New return didexchange service
 func New(store storage.Store, prov provider) *Service {
-	return &Service{outboundTransport: prov.OutboundTransport(), store: store}
+	svc := &Service{
+		outboundTransport: prov.OutboundTransport(),
```
troyronda: There should be a TODO that we actually need an outbound dispatcher rather than the transport directly.

The dispatcher needs to call Pack prior to using the outbound transport.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -39,11 +61,22 @@ type provider interface {
 type Service struct {
 	outboundTransport transport.OutboundTransport
 	store             storage.Store
+	ch                chan didCommChMessage
```
troyronda: Suggest a more descriptive name.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -86,6 +150,92 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	}
+
+	return invoked, nil
+}
+
+// startInternalListener listens to messages in gochannel for callback messages from clients.
+func (s *Service) startInternalListener() {
+	go func() {
+		for msg := range s.ch {
+			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
```
troyronda: Also need to ensure we have tests for:

- no listeners call the callback.
- multiple listeners call the callback.
- some listeners call the callback and some do not.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +96,52 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	}
+
+	return nil
+}
+
+// Register events on DID Exchange protocol. The events are triggered for incoming ConnectionRequest, ConnectionResponse
+// or ConnectionAck message types.
+func (s *Service) Register(ch chan<- event.DIDCommEvent) error {
+	s.event = append(s.event, ch)
```
troyronda: Is this thread safe?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +96,52 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	if !eventsTriggered {
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	}
+
+	return nil
+}
+
+// Register events on DID Exchange protocol. The events are triggered for incoming ConnectionRequest, ConnectionResponse
+// or ConnectionAck message types.
+func (s *Service) Register(ch chan<- event.DIDCommEvent) error {
```
troyronda: How does a listener unregister?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -12,24 +12,46 @@ import (
...
+	ConnectionInvite = DIDExchangeSpec + "invitation"
+	// ConnectionRequest defines the did-exchange request message type.
+	ConnectionRequest = DIDExchangeSpec + "request"
+	// ConnectionResponse defines the did-exchange response message type.
+	ConnectionResponse = DIDExchangeSpec + "response"
+	// ConnectionAck defines the did-exchange ack message type.
+	ConnectionAck = DIDExchangeSpec + "ack"
 )
 
+type message struct {
```
troyronda: It would be good to have a comment on how message is being used.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +96,52 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 	if !current.CanTransitionTo(next) {
 		return fmt.Errorf("invalid state transition: %s -> %s", current.Name(), next.Name())
 	}
-	// TODO: call pre-transition listeners -  Issue: https://github.com/hyperledger/aries-framework-go/issues/140
-	followup, err := next.Execute(msg)
+
+	// trigger the event if the message type is registered (if it's not outbound)
+	eventsTriggered := false
+	if !msg.Outbound {
+		eventsTriggered, err = s.sendEvent(msg, thid, next)
```
troyronda: I assume eventsTriggered is only set to true for events that have callback actions (e.g., not error events).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +96,52 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+
+	// if no events are trigger continue the execution
+	if !eventsTriggered {
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	}
+
+	return nil
+}
+
+// Register events on DID Exchange protocol. The events are triggered for incoming ConnectionRequest, ConnectionResponse
```
troyronda: There is an undocumented side-effect: auto execution when you don't register; auto execution if you do.

Also as a TODO: you might only be registering for events not for callbacks. Related to the thought above: https://github.com/hyperledger/aries-framework-go/pull/240/files#r324795675

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +96,52 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	if !msg.Outbound {
+		eventsTriggered, err = s.sendEvent(msg, thid, next)
+		if err != nil {
+			return fmt.Errorf("send events failed: %w", err)
+		}
+	}
+
+	// if no events are trigger continue the execution
+	if !eventsTriggered {
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
```
troyronda: We have to be careful here.

There is a race between starting the protocol and disabling auto execution (currently via register). We don't want to accidentally auto execute state and we don't want to accidentally drop messages either :).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -192,8 +342,12 @@ func (s *Service) marshalAndSend(data interface{}, errorMsg, destination string)
 	return s.outboundTransport.Send(string(jsonString), destination)
 }
 
+func (s *Service) generateRandomID() string {
+	return uuid.New().String()
```
Baha-sk: not sure why, but the uuid docs says "creates new UUID or panics": https://godoc.org/github.com/google/uuid#New

it's maybe rare, but we should probably not allow panics here.. or at least catch them.

generally speaking, for random bytes, I prefer using crypto.Rand
have you tried  uuid.NewRandom() ?

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -478,30 +487,352 @@ func (p *mockProvider) OutboundTransport() transport.OutboundTransport {
...
 	require.NoError(t, err)
-	return dbstore, cleanup
+	return dbstore, func() {
+		err := prov.Close()
+		require.NoError(t, err)
+	}
+}
+
+func randomString() string {
+	u := uuid.New()
```
Baha-sk: same here

pkg/didcomm/protocol/didexchange/service.go
troyronda: Note uuid.New and uuid.NewRandom are [equivalent](https://github.com/google/uuid/blob/master/version4.go#L14) except for the panic vs error return.


So the following blurb from NewRandom also applies to New.
> The strength of the UUIDs is based on the strength of the crypto/rand

rolsonquadras: @Baha-sk just curious about the error logic for such cases. Whats the error/retry logic processing in such cases (error, panic)? Should we throw error back to the consumers or retry till we succeed?

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/6ca38f34c66c2017dee287d5ecd2734ca169e7b7..5839b0f509b0932a67abd562fb6a05dc181f2e08#diff-f761c91e753eba1c42a68076df06ce12R50)

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/6ca38f34c66c2017dee287d5ecd2734ca169e7b7..5839b0f509b0932a67abd562fb6a05dc181f2e08#diff-f761c91e753eba1c42a68076df06ce12R73) a TODO. Issue #259.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/6ca38f34c66c2017dee287d5ecd2734ca169e7b7..5839b0f509b0932a67abd562fb6a05dc181f2e08#diff-f761c91e753eba1c42a68076df06ce12R66).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Updated the Godoc and created a TODO. #261

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: valid point. Created #261 .

pkg/didcomm/protocol/didexchange/service.go
troyronda: The panic happens if the application [cannot read](https://github.com/google/uuid/blob/master/version4.go#L37) from crypto/rand. This seems a bit unrecoverable...

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: At the moment, events are triggered only for connectionRequest, response and ack. [Ref](https://github.com/hyperledger/aries-framework-go/pull/240/files#diff-f761c91e753eba1c42a68076df06ce12R162)

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: I was thinking about adding this in second cut as this PR is already more than 600 lines. For now added an [empty function](https://github.com/hyperledger/aries-framework-go/compare/5839b0f509b0932a67abd562fb6a05dc181f2e08..5dfa12128f1ba2867b1a8c7c47464e6f9b8c5b7a#diff-f761c91e753eba1c42a68076df06ce12R134) with a TODO.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: No. As per I remember during the last discussion, there were comments that registration happens when protocol is initiated. Hence locks are not necessary as such and having it will be an overhead during fetching the registered event to trigger it. Since we don't have that code now, I have added the mutex now.

Note - I had the locks earlier and removed it in previous patch.

pkg/didcomm/protocol/didexchange/service_test.go
rolsonquadras: Discussion is tracked [here](https://github.com/hyperledger/aries-framework-go/pull/240#discussion_r324806191).

pkg/didcomm/protocol/didexchange/service.go
troyronda: I don't know how registrations can happen when the protocol is initiated.

A client package is created "on the fly".

rolsonquadras:
Updated the code based on discussion with @troyronda. 

Configs:
1. Action Events - Consumer can register one event for message processing. The event message will have a callback function. The consumer need to call this to continue processing the message.
2. Stream Events -  Consume can register multiple event for stream listening. This is not related to message handling and just a mean for the consumers to listen to the messages.
3. Auto Execution - Consumers can set Auto execution to execute the message processing. 

Behaviour:
1. If Action events or auto execution not set - the service will throw error while handling the messages.
2. If action event set after auto execution - the service will trigger action event and wait for callback
3. If action event set before auto execution - the service will auto execute
3. If multiple action event set - the events will triggered on last event channel.

cmd/aries-agentd/main_test.go
```diff
@@ -103,7 +103,7 @@ func validateRequests(t *testing.T) {
 								"@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0/invitation"
 							}`),
 				"application/didcomm-envelope-enc"),
-			expectedStatus:     http.StatusAccepted,
+			expectedStatus:     http.StatusInternalServerError,
```
troyronda: Is this the status for the protocol doesn't yet have action callback listeners?

There seems to be a couple of relevant error codes to choose from: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors

e.g., 503

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: This has been addressed in this PR. Godoc - https://github.com/hyperledger/aries-framework-go/pull/240/files#diff-f761c91e753eba1c42a68076df06ce12R133.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +112,115 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+// RegisterActionEvent on DID Exchange protocol messages. The events are triggered for incoming ConnectionRequest,
+// ConnectionResponse or ConnectionAck message types. The consumer need to invoke the callback to resume processing.
+// Only one channel can be registered for the action events. If called multiple times, the events will be sent to the
+// last channel. This works in conjunction with RegisterAutoExecute(). If this function is called after
+// RegisterAutoExecute(), the service will not execute the processing automatically and it'll wait for the callback. If
+// both are not set, then service will not handle the messages and throw an error.
+func (s *Service) RegisterActionEvent(ch chan<- event.DIDCommEvent) error {
+	s.lock.Lock()
+	s.actionEvent = ch
+	s.execute = true
```
troyronda: is execute meaning service is allowed to respond?

cmd/aries-agentd/main_test.go
llorllale: why this change?

rolsonquadras: @troyronda yes. The error codes are set in [inbound.go](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/transport/http/inbound.go#L62).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +112,115 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+
+// UnregisterActionEvent on DID Exchange protocol messages. Refer RegisterActionEvent().
+func (s *Service) UnregisterActionEvent() error {
+	return s.disableExecute()
+}
+
+// RegisterStreamEvent on DID Exchange protocol messages. The events are triggered for incoming ConnectionRequest,
+// ConnectionResponse or ConnectionAck message types. The Callback is set to nil in the actionEvent message and service
+// won't be expecting any response from the consumers.
+func (s *Service) RegisterStreamEvent(ch chan<- dispatcher.DIDCommMsg) error {
```
troyronda: RegisterStatusEvent (or similar) might be the better corollary to RegisterActionEvent

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: yes.

pkg/didcomm/protocol/didexchange/service.go
troyronda: You may also want to match up the type names.

If you are calling the type DIDCommMsg then I would expect this method to be RegisterMsg[Foo].
Similarly with RegisterActionEvent returning DIDCommEvent then I would expect that method to be RegisterEvent[Foo].

cmd/aries-agentd/main_test.go
rolsonquadras: [Here](https://github.com/hyperledger/aries-framework-go/pull/240#discussion_r325327735).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +112,115 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	s.lock.Lock()
+	s.actionEvent = ch
+	s.execute = true
+	s.lock.Unlock()
+
+	return nil
+}
+
+// UnregisterActionEvent on DID Exchange protocol messages. Refer RegisterActionEvent().
+func (s *Service) UnregisterActionEvent() error {
```
troyronda: I'd expect the channel to be passed into such a method.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +112,115 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	if !eventsTriggered {
+		return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	}
+
+	return nil
+}
+
+// RegisterActionEvent on DID Exchange protocol messages. The events are triggered for incoming ConnectionRequest,
+// ConnectionResponse or ConnectionAck message types. The consumer need to invoke the callback to resume processing.
+// Only one channel can be registered for the action events. If called multiple times, the events will be sent to the
```
troyronda: The alternative, of course, is to return an error if currently registered.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -192,8 +430,12 @@ func (s *Service) marshalAndSend(data interface{}, errorMsg, destination string)
 	return s.outboundTransport.Send(string(jsonString), destination)
 }
 
+func (s *Service) generateRandomID() string {
```
troyronda: I don't see a reference to s in this method.

pkg/didcomm/protocol/didexchange/service.go
llorllale: `dispatcher.DIDCommMsg` is not an event...

cmd/aries-agentd/main_test.go
troyronda: Ah. We should consider a TODO where we have a more specific code like 503.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +112,115 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+func (s *Service) RegisterAutoExecute() error {
+	s.lock.Lock()
+	s.execute = true
+	s.lock.Unlock()
+
+	return nil
+}
+
+// UnregisterAutoExecute on DID Exchange protocol messages. Refer RegisterAutoExecute().
+func (s *Service) UnregisterAutoExecute() error {
```
llorllale: This should not disable the events stream.

troyronda: I assume you mean DIDCommMsg as a result of RegisterStreamEvent :). Agree those should not be disabled.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -39,15 +64,39 @@ type provider interface {
...
+
+	svc.startInternalListener()
+
+	return svc
 }
 
 // Handle didexchange msg
 func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
+	// throw error if there are no action events are registered or auto execute set (if it's not outbound)
+	if !msg.Outbound && !s.execute {
```
bstasyszyn: Not sure if this is related to the comment by Troy below, but s.execute needs an RLock in order to read.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: @llorllale @troyronda Is the code missing something here? Does this need any change ?

pkg/didcomm/protocol/didexchange/service.go
llorllale: This is not checking whether "action events" have been registered. If there are, then the msg should be processed, even if auto-execute is disabled.

cmd/aries-agentd/main_test.go
rolsonquadras: #271

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -39,15 +64,39 @@ type provider interface {
 type Service struct {
 	outboundTransport transport.OutboundTransport
 	store             storage.Store
+	callbackChannel   chan didCommChMessage
```
llorllale: Suggest use of "inbound/outbound" or "incoming/outgoing" prefixes to add clarity:

* callbackChannel -> inboundCallbacks
* actionEvent -> outboundActions
* streamEvents -> outboundEvents

`execute` could use a rename to `autoExecute`

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -39,15 +64,39 @@ type provider interface {
...
+		outboundTransport: prov.OutboundTransport(),
+		store:             store,
+		// TODO channel size - https://github.com/hyperledger/aries-framework-go/issues/246
+		callbackChannel: make(chan didCommChMessage, 10),
+		// set execute to false. Consumers have to enable this by setting either RegisterActionEvent() or
+		// RegisterAutoExecute()
+		execute: false,
+	}
+
+	svc.startInternalListener()
```
llorllale: Suggest rename to "startListeningForIncomingCallbacks". At least it makes the intent more clear.

pkg/didcomm/protocol/didexchange/service.go
llorllale: Also naming is everything. This method's purpose would be a lot clearer with a name such as "Register**For**EventsStream" or something similar. Keyword: for.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -63,21 +112,115 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
...
+	return nil
+}
+
+// RegisterActionEvent on DID Exchange protocol messages. The events are triggered for incoming ConnectionRequest,
+// ConnectionResponse or ConnectionAck message types. The consumer need to invoke the callback to resume processing.
+// Only one channel can be registered for the action events. If called multiple times, the events will be sent to the
+// last channel. This works in conjunction with RegisterAutoExecute(). If this function is called after
+// RegisterAutoExecute(), the service will not execute the processing automatically and it'll wait for the callback. If
+// both are not set, then service will not handle the messages and throw an error.
+func (s *Service) RegisterActionEvent(ch chan<- event.DIDCommEvent) error {
```
llorllale: Same here w.r.t. naming - suggest "RegisterForEventsStream" (the other one should stay with "RegisterForMsgStream")

troyronda: I would drop "For"

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/6dc9c7de378434ea27520364556f649ccf434c5f..36d8a92bfaa82b830e44a047521b376920c82c72#diff-f761c91e753eba1c42a68076df06ce12R437).

pkg/didcomm/protocol/didexchange/service.go
troyronda: execute doesn't mean autoExecute (which also means it could use a rename to something more explicit.)

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Updated.
1. RegisterActionEvent to [RegisterEvent](https://github.com/hyperledger/aries-framework-go/compare/6dc9c7de378434ea27520364556f649ccf434c5f..36d8a92bfaa82b830e44a047521b376920c82c72#diff-f761c91e753eba1c42a68076df06ce12R143)
2. RegisterStreamEvent to [RegisterMsg](https://github.com/hyperledger/aries-framework-go/compare/6dc9c7de378434ea27520364556f649ccf434c5f..36d8a92bfaa82b830e44a047521b376920c82c72#diff-f761c91e753eba1c42a68076df06ce12R160)

rolsonquadras: Discussion is tracker [here](https://github.com/hyperledger/aries-framework-go/pull/240#discussion_r325329926).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Discussion is tracker [here](https://github.com/hyperledger/aries-framework-go/pull/240#discussion_r325329926).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/6dc9c7de378434ea27520364556f649ccf434c5f..36d8a92bfaa82b830e44a047521b376920c82c72#diff-f761c91e753eba1c42a68076df06ce12R97).

rolsonquadras: s.execute is derived based on actionEvent or AutoExecute registration.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Since it's one channel(event), what are we expecting to do with that arg ?

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: @troyronda thoughts on this?

troyronda: I think we need to do further changes as a new PR. It's getting hard to follow.

pkg/didcomm/protocol/didexchange/service.go
llorllale: @troyronda the intent is to register a channel via which events would be received by the caller of this function.

`RegisterEventStream(chan event.DIDCommEvent)` reads as though whoever is calling this function is going to produce the events.

`RegisterForEventStream(chan event.DIDCommEvent)` removes that ambiguity.

Also, "Stream" is probably redundant. Instead, just say "Events" - the "stream" is implicit.

Just my 2c.

Issue #247, opened 2019-09-11
Controller API : add missing endpoints and validate spec


sudeshrshetty:
Related to epic #95 

- add validate spec makefile target in aries-framework-go
- add missing endpoints like remove connection etc

llorllale:
@sudeshrshetty what is a "spec target"?

sudeshrshetty:
All restapi pkg functions and models are now following open API annotation specifications, this target will generate spec and validate to find any discrepancies.
Once this make file target is in place, anyone is free to update any restapi models or functions, this target will make sure that open API specification is followed. In future, we are going to use this spec for docs and tests.

PR #265, opened 2019-09-16
feat:added missing endpoints in controller API

sudeshrshetty:
- added connection.remove() endpoint in client and rest API
 - all except connection query endpoints are changed to HTTP POST
 - improved test coverage for `pkg/restapi/operation/didexchange`
 - related to #247

### DIFF
--- a/pkg/restapi/operation/didexchange/didexchange.go
+++ b/pkg/restapi/operation/didexchange/didexchange.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/google/uuid"
@@ -36,6 +37,7 @@ const (
 	connections           = operationID
 	connectionsByID       = operationID + "/{id}"
 	acceptExchangeRequest = operationID + "/{id}/accept-request"
+	removeConnection      = operationID + "/{id}/remove"
 )
 
 // provider contains dependencies for the Exchange protocol and is typically created by using aries.Context()
@@ -77,7 +79,7 @@ type Operation struct {
 	handlers []operation.Handler
 }
 
-// CreateInvitation swagger:route GET /connections/create-invitation did-exchange createInvitation
+// CreateInvitation swagger:route POST /connections/create-invitation did-exchange createInvitation
 //
 // Creates a new connection invitation....
 //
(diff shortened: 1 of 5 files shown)


pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -217,10 +210,26 @@ func (c *Operation) QueryConnections(rw http.ResponseWriter, req *http.Request)
 		},
 	}
 
-	err = json.NewEncoder(rw).Encode(response)
+	c.writeResponse(rw, response)
+}
+
+// getQueryParams unmarshals query strings values to interface value
+func (c *Operation) getQueryParams(v interface{}, vals url.Values) error {
```
troyronda: I don't see any usage of c within this function.

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -217,10 +210,26 @@ func (c *Operation) QueryConnections(rw http.ResponseWriter, req *http.Request)
 		},
 	}
 
-	err = json.NewEncoder(rw).Encode(response)
+	c.writeResponse(rw, response)
+}
+
+// getQueryParams unmarshals query strings values to interface value
```
troyronda: It would be helpful to describe the v argument. 
You are expecting an annotated struct.

This function appears to populate an annotated struct from the query params.

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -217,10 +210,26 @@ func (c *Operation) QueryConnections(rw http.ResponseWriter, req *http.Request)
...
+		}
+	}
+
+	bytes, err := json.Marshal(args)
 	if err != nil {
-		logger.Errorf("Unable to write response, %s", err)
+		return err
 	}
+
+	return json.Unmarshal(bytes, v)
```
troyronda: Interesting pattern. I wonder if the Go stdlib has a more efficient method (rather than marshal and unmarshal json).

pkg/restapi/operation/didexchange/didexchange.go
sudeshrshetty: made it private static

pkg/restapi/operation/didexchange/didexchange.go
sudeshrshetty: This is simple util function to convert `map[string][]string` to given type v. It doesn't have to be annoted, simple json Unmarshal rule applies here.

pkg/restapi/operation/didexchange/didexchange.go
sudeshrshetty: Using Go standard lib supports, 
can convert 'map[string]string' to struct with string properties.
can convert 'map[string]string[]' to struct with string[] properties.

In this case, net.http `request.URL.Query()` was returning `map[string]string[]` and we needed struct with string properties out of it.

I didn't want to manually lookup for `key,value` pair inside `request.URL.Query()` to set them one by one to target struct (hard to maintain whenever queryString/model changes and developers may forgot update this logic whenever they change models).

So went for dynamic solution.

troyronda: Right - I guess I was surprised that the only way to accomplish the above (dynamically) is via JSON marshal/unmarshal.

PR #266, opened 2019-09-16
feat: Create Pack() for Authcrypt logic

### LINKED ISSUE #38: Create Pack() for Authcrypt logic
Once Go-Jose has Chacha20/Poly1035 support, create the Pack interface and implementation for **Authcrypt** using this new addition as per the new standardized JWE format in this Aries-rfc issue:
https://github.com/hyperledger/aries-rfcs/issues/133

and following the Pack spec for **Authcrypt** (disregard format from this RFC):
https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope

ref #36

fqutishat:
- imlpement pack method in wallet
- implement unpack method in wallet

closes #38

### DIFF
--- a/pkg/didcomm/crypto/crypter.go
+++ b/pkg/didcomm/crypto/crypter.go
@@ -8,20 +8,21 @@ package crypto
 
 // Crypter is an Aries envelope encrypter to support
 // secure DIDComm exchange of envelopes between Aries agents
+// TODO create a higher-level crypto that switches implementations based on the algorithm - Issue #273
 type Crypter interface {
 	// Encrypt a payload in an Aries compliant format using the sender keypair
 	// and a list of recipients public keys
 	// returns:
 	// 		[]byte containing the encrypted envelope
 	//		error if encryption failed
-	// TODO add key type of recipients and sender keys to be validated by the implementation
+	// TODO add key type of recipients and sender keys to be validated by the implementation - Issue #272
 	Encrypt(payload []byte, sender KeyPair, recipients [][]byte) ([]byte, error)
 	// Decrypt an envelope in an Aries compliant format with the recipient's private key
 	// and the recipient's public key both set in recipientKeyPair
 	// returns:
 	// 		[]byte containing the decrypted payload
 	//		error if decryption failed
-	// TODO add key type of recipients keys to be validated by the implementation
+	// TODO add key type of recipients keys to be validated by the implementation - Issue #272
 	Decrypt(envelope []byte, recipientKeyPair KeyPair) ([]byte, error)
 }
 
--- a/pkg/wallet/api.go
+++ b/pkg/wallet/api.go
@@ -90,7 +90,8 @@ type Pack interface {
 type Envelope struct {
 	Message    []byte
 	FromVerKey string
-	ToVerKeys  string
+	// TODO add key type - issue #272
+	ToVerKeys []string
 }
 
 // ErrKeyNotFound is returned when key not found
--- a/pkg/wallet/wallet.go
+++ b/pkg/wallet/wallet.go
@@ -9,32 +9,31 @@ package wallet
 import (
 	"crypto/rand"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/btcsuite/btcutil/base58"
 	"golang.org/x/crypto/nacl/box"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/crypto"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/crypto/jwe/authcrypt"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
 // BaseWallet wallet implementation
 type BaseWallet struct {
-	store storage.Store
+	store   storage.Store
+	crypter crypto.Crypter
 }
 
-// key json structure include public,private and metadata
-type key struct {
-	Public  string    `json:"public"`
-	Private *[32]byte `json:"private"`
-}
-
+// TODO use the context / provider mechanism
 // New return new instance of wallet implementation
-func New(storeProvider storage.Provider) (*BaseWallet, error) {
+func New(storeProvider storage.Provider, crypter crypto.Crypter) (*BaseWallet, error) {
 	store, err := storeProvider.GetStoreHandle()
 	if err != nil {
 		return nil, fmt.Errorf("failed to GetStoreHandle: %w", err)
 	}
-	return &BaseWallet{store: store}, nil
+	return &BaseWallet{store: store, crypter: crypter}, nil
 }
 
 // CreateKey create a new public/private signing keypair.
@@ -45,7 +44,7 @@ func (w *BaseWallet) CreateKey() (string, error) {
 	}
 	base58Pub := base58.Encode(pub[:])
 	// TODO - need to encrypt the priv before putting them in the store.
-	if err := w.persistKey(&key{Public: base58Pub, Private: priv}); err != nil {
+	if err := w.persistKey(base58Pub, &crypto.KeyPair{Pub: pub[:], Priv: priv[:]}); err != nil {
 		return "", err
 	}
 	return base58Pub, nil
@@ -63,12 +62,57 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
 
 // PackMessage Pack a message for one or more recipients.
 func (w *BaseWallet) PackMessage(envelope *Envelope) ([]byte, error) {
-	return nil, fmt.Errorf("not implemented")
+	if envelope == nil {
+		return nil, errors.New("envelope argument is nil")
+	}
+	// get keypair from db
+	senderKeyPair, err := w.getKey(envelope.FromVerKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed from getKey: %w", err)
+	}
+
+	var recipients [][]byte
+	for _, verKey := range envelope.ToVerKeys {
+		// TODO It is possible to have different key schemes in an interop situation
+		// there is no guarantee that each recipient is using the same key types
+		// decode base58 ver key
+		verKeyBytes := base58.Decode(verKey)
+		// create 32 byte key
+		recipients = append(recipients, verKeyBytes)
+	}
+	// encrypt message
+	bytes, err := w.crypter.Encrypt(envelope.Message, *senderKeyPair, recipients)
+	if err != nil {
+		return nil, fmt.Errorf("failed from encrypt: %w", err)
+	}
+	return bytes, nil
 }
 
 // UnpackMessage Unpack a message.
 func (w *BaseWallet) UnpackMessage(encMessage []byte) (*Envelope, error) {
-	return nil, fmt.Errorf("not implemented")
+	var e authcrypt.Envelope
+	if err := json.Unmarshal(encMessage, &e); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal encMessage: %w", err)
+	}
+	var keysNotFound []string
+	for _, v := range e.Recipients {
+		recipVKeyB58 := v.Header.KID
+		// get keypair from db
+		recipientKeyPair, err := w.getKey(recipVKeyB58)
+		if err != nil {
+			if errors.Is(err, ErrKeyNotFound) {
+				keysNotFound = append(keysNotFound, recipVKeyB58)
+				continue
+			}
+			return nil, fmt.Errorf("failed from getKey: %w", err)
+		}
+		bytes, err := w.crypter.Decrypt(encMessage, *recipientKeyPair)
+		if err != nil {
+			return nil, fmt.Errorf("failed from decrypt: %w", err)
+		}
+		return &Envelope{Message: bytes}, nil
+	}
+	return nil, fmt.Errorf("no corresponding recipient key found in {%s}", keysNotFound)
 }
 
 // Close wallet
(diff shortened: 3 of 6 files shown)


pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+	// create 32 byte key
+	var pubKey [chacha.KeySize]byte
+	copy(pubKey[:], pubBytes)
+
+	var recipients []*[chacha.KeySize]byte
+	for _, verKey := range envelope.ToVerKeys {
+		// decode base58 ver key
+		verKeyBytes := base58.Decode(verKey)
+		// create 32 byte key
+		var recipient [chacha.KeySize]byte
```
troyronda: I don’t think we should reference chacha directly - should come from crypter.

pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
 
 // PackMessage Pack a message for one or more recipients.
 func (w *BaseWallet) PackMessage(envelope *Envelope) ([]byte, error) {
-	return nil, fmt.Errorf("not implemented")
+	if envelope == nil {
+		return nil, errors.New("envelope argument is nil")
+	}
+	// get keypair from db
+	key, err := w.getKey(envelope.FromVerKey)
```
Baha-sk: probably rename `key` to `senderKeyPair` or `fromKeyPair` as there are many keys involved in this function, for the sake of clarity

pkg/wallet/wallet.go
```diff
@@ -29,12 +34,12 @@ type key struct {
 }
 
 // New return new instance of wallet implementation
-func New(storeProvider storage.Provider) (*BaseWallet, error) {
+func New(storeProvider storage.Provider, crypter crypto.Crypter) (*BaseWallet, error) {
```
troyronda: We aren’t using the context / provider mechanism here?

pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+	// get keypair from db
+	key, err := w.getKey(envelope.FromVerKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed from getKey: %w", err)
+	}
+	// decode base58 public key
+	pubBytes := base58.Decode(key.Public)
+
+	// create 32 byte key
+	var pubKey [chacha.KeySize]byte
```
Baha-sk: same thing with `pubKey`, call it `senderPubKey` or `fromPubKey`

pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+		verKeyBytes := base58.Decode(verKey)
+		// create 32 byte key
+		var recipient [chacha.KeySize]byte
+		copy(recipient[:], verKeyBytes)
+		recipients = append(recipients, &recipient)
+	}
+
+	// encrypt message
+	bytes, err := w.crypter.Encrypt(envelope.Message,
+		crypto.KeyPair{Priv: key.Private, Pub: &pubKey}, recipients)
```
Baha-sk: I don't understand why private key is used as is while public key had to be base58 decoded first. Shouldn't private key be base58 decoded too? Consistency is key.

what format do we expect to store keys in a wallet?

pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
-	return nil, fmt.Errorf("not implemented")
+	var e authcrypt.Envelope
+	if err := json.Unmarshal(encMessage, &e); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal encMessage: %w", err)
+	}
+	var keysNotFound []string
+	for _, v := range e.Recipients {
+		recipVKeyB58 := v.Header.KID
+		// get keypair from db
+		key, err := w.getKey(recipVKeyB58)
```
Baha-sk: same comment about naming key as above, in this case, this is a `recipientKeyPair`

pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+			}
+			return nil, fmt.Errorf("failed from getKey: %w", err)
+		}
+		// decode base58 public key
+		pubBytes := base58.Decode(key.Public)
+
+		// create 32 byte key
+		var pubKey [chacha.KeySize]byte
+		copy(pubKey[:], pubBytes)
+		bytes, err := w.crypter.Decrypt(encMessage, crypto.KeyPair{Priv: key.Private, Pub: &pubKey})
```
Baha-sk: Same comment about private vs public key formats above, why base58 decode public key and not the private one?

pkg/wallet/wallet.go
```diff
@@ -63,12 +68,71 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+		if err != nil {
+			return nil, fmt.Errorf("failed from decrypt: %w", err)
+		}
+		return &Envelope{Message: bytes}, nil
+	}
+	return nil, fmt.Errorf("no corresponding recipient key found in {%s}", keysNotFound)
```
Baha-sk: to be discussed, if we should be listing keys in error messages..

pkg/wallet/wallet.go
```diff
@@ -88,3 +152,19 @@ func (w *BaseWallet) persistKey(key *key) error {
 	}
 	return nil
 }
+
+// getKey get key
+func (w *BaseWallet) getKey(verkey string) (*key, error) {
```
Baha-sk: we should probably use the same key structure between our crypto and wallet packages (and remove the duplicate one)

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: I add TODO

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: I think it's ok to expose public keys now

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
```diff
@@ -63,12 +62,59 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+		// get keypair from db
+		recipientKeyPair, err := w.getKey(recipVKeyB58)
+		if err != nil {
+			if errors.Is(err, ErrKeyNotFound) {
+				keysNotFound = append(keysNotFound, recipVKeyB58)
+				continue
+			}
+			return nil, fmt.Errorf("failed from getKey: %w", err)
+		}
+		bytes, err := w.crypter.Decrypt(encMessage, crypto.KeyPair{Priv: recipientKeyPair.Priv, Pub: recipientKeyPair.Pub})
```
Baha-sk: the advantage of using the same struct is you can send it as is here too...

second argument can simply be `recipientKeyPair`, no need to create a new KeyPair instance

pkg/wallet/wallet.go
```diff
@@ -63,12 +62,59 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+		verKeyBytes := base58.Decode(verKey)
+		// create 32 byte key
+		var recipient [crypto.KeySize]byte
+		copy(recipient[:], verKeyBytes)
+		recipients = append(recipients, &recipient)
+	}
+
+	// encrypt message
+	bytes, err := w.crypter.Encrypt(envelope.Message,
+		crypto.KeyPair{Priv: senderKeyPair.Priv, Pub: senderKeyPair.Pub}, recipients)
```
Baha-sk: second argument can simply be senderKeyPair, no need to create a new KeyPair instance

pkg/didcomm/crypto/crypter.go
```diff
@@ -10,6 +10,11 @@ import (
 	chacha "golang.org/x/crypto/chacha20poly1305"
 )
 
+const (
+	// KeySize is the size of the key used by this AEAD, in bytes.
+	KeySize = 32
```
troyronda: FYI: You can assign a const from another const (e.g., chacha.KeySize)

pkg/didcomm/crypto/crypter.go
```diff
@@ -30,9 +35,9 @@ type Crypter interface {
 // KeyPair represents a private/public key pair each with 32 bytes in size
 type KeyPair struct {
 	// Priv is a private key
-	Priv *[chacha.KeySize]byte
+	Priv *[chacha.KeySize]byte `json:"private"`
```
troyronda: It's a bit odd to define a KeySize const in this file and not use it?

pkg/wallet/wallet.go
fqutishat: done

pkg/wallet/wallet.go
fqutishat: done

pkg/didcomm/crypto/crypter.go
troyronda: i.e., you are still importing chacha into this interface package.

pkg/wallet/wallet.go
```diff
@@ -63,12 +62,58 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
 
 // PackMessage Pack a message for one or more recipients.
 func (w *BaseWallet) PackMessage(envelope *Envelope) ([]byte, error) {
```
troyronda: As discussed in the planning meeting, the abstraction of Pack/Unpack and underlying crypto primitives in the wallet should be removed.

More specifically, does the wallet need to know about envelopes?

pkg/wallet/api.go
```diff
@@ -90,7 +90,7 @@ type Pack interface {
 type Envelope struct {
 	Message    []byte
 	FromVerKey string
-	ToVerKeys  string
+	ToVerKeys  []string
```
troyronda: Are these encoded keys?

(also wondering why string is being used for keys, instead of using []byte)

pkg/wallet/wallet_test.go
```diff
@@ -48,33 +56,188 @@ func TestBaseWallet_CreateKey(t *testing.T) {
...
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "decrypt error")
+
 	})
 }
 
 func TestBaseWallet_PackMessage(t *testing.T) {
-	t.Run("test error not implemented", func(t *testing.T) {
-		w, err := New(&mockstorage.MockStoreProvider{})
+	t.Run("test success", func(t *testing.T) {
```
troyronda: Do we have an expected test test case for Pack and Unpack (e.g., from RFC) that we are comparing against?

pkg/didcomm/crypto/crypter.go
fqutishat: done remove all chacha.KeySize

pkg/wallet/api.go
fqutishat: they are encoded keys with base58

pkg/wallet/wallet.go
```diff
@@ -63,12 +62,58 @@ func (w *BaseWallet) DecryptMessage(encMessage []byte, toVerKey string) ([]byte,
...
+	// get keypair from db
+	senderKeyPair, err := w.getKey(envelope.FromVerKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed from getKey: %w", err)
+	}
+
+	var recipients []*[crypto.KeySize]byte
+	for _, verKey := range envelope.ToVerKeys {
+		// decode base58 ver key
+		verKeyBytes := base58.Decode(verKey)
```
troyronda: I'm still hesitant about this logic.

It is possible to have different key schemes in an interop situation. I wouldn't expect guarantee that each recipient is using the same key types. Yet another TODO...

fqutishat: ok

pkg/didcomm/crypto/crypter.go
```diff
@@ -10,6 +10,11 @@ import (
 	chacha "golang.org/x/crypto/chacha20poly1305"
 )
 
+const (
+	// KeySize is the size of the key used by this AEAD, in bytes.
+	KeySize = chacha.KeySize
```
troyronda: We'll need to revisit this logic when we have more than chacha.

pkg/didcomm/crypto/crypter.go
```diff
@@ -18,7 +23,7 @@ type Crypter interface {
 	// returns:
 	// 		[]byte containing the encrypted envelope
 	//		error if encryption failed
-	Encrypt(payload []byte, sender KeyPair, recipients []*[chacha.KeySize]byte) ([]byte, error)
+	Encrypt(payload []byte, sender KeyPair, recipients []*[KeySize]byte) ([]byte, error)
```
troyronda: Why doesn't Encrypt take a slice rather than an array?

fqutishat: @Baha-sk ^^^

pkg/wallet/api.go
troyronda: Issue #272 again

pkg/wallet/wallet.go
troyronda: I assume #272 applies here too.

pkg/wallet/wallet_test.go
fqutishat: to test success we just pack msg and unpack it. I'm not sure how else can I test success

pkg/wallet/api.go
fqutishat: I think all public keys should encoded with base58

pkg/wallet/wallet.go
fqutishat: I added TODO

pkg/wallet/wallet_test.go
troyronda: We would need a test fixture from the RFC and ensure we produce the same. We may have issues when we try to interop with another agent if we only test against our own implementation.

Issue #94, opened 2019-08-13
Demo : Controller API for DID exchange


sudeshrshetty:
Should be done after #93 

- [x] - start `aries-agentd` as command line command (COBRA). Provide port etc as command line arguments (covered in #250 )
- [x] - build docker image for demo agent 
- [x] - run bddtests using agents as separate docker containers (#491)
- [x] - run demo using  swagger UI based on controller API spec

epic #67

PR #274, opened 2019-09-17
feat: generate and validate controller OpenAPI specs

### LINKED ISSUE #247: Controller API : add missing endpoints and validate spec

sudeshrshetty:
- generate build target added to generate rest api spec against
'openAPI/swagger specification 2.0'
- closes #247


Makefile
```diff
@@ -29,3 +31,10 @@ bdd-test: clean
 .PHONY: clean
 clean:
 	rm -f coverage.txt
+	rm -Rf ./build
+
+.PHONY: validate-api-spec
```
troyronda: I think we should be more explicit that this is an openapi/swagger target

validate-openapi-spec or validate-openapi

troyronda: Also I see below that the openapi is generated.

Are you sure we shouldn't call the target generate-foo?

Makefile
```diff
@@ -3,9 +3,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 GO_CMD ?= go
+# Controller API entry point to be used for generating REST API specifications
+API_SPEC_META=cmd/aries-agentd/main.go
```
troyronda: Same comment here. Be explicit on the type of API.

scripts/validate-spec.sh
```diff
@@ -0,0 +1,22 @@
...
+SPEC_LOC="${SPEC_LOC}"
+SPEC_META="${SPEC_META:-cmd/aries-agentd/main.go}"
+OUTPUT="$SPEC_LOC/openAPI.yml"
+
+# generate and validate commands
+GENERATE_COMMAND="generate spec $SPEC_META -o $OUTPUT"
+VALIDATE_COMMAND="validate $OUTPUT"
+
+echo "Generating Open API spec"
+docker run --rm -it -e GOPATH=$HOME/go:/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger $GENERATE_COMMAND
```
troyronda: they still called it swagger?

sudeshrshetty: command lines and annotations are still called `swagger` since we are using Open API 2.0.
Created #275 for upgrade, where these commands may get replaced.

sudeshrshetty: command lines and annotations are still called `swagger` since we are using Open API 2.0.
Created #275 for upgrade, where these commands may get replaced.

Makefile
sudeshrshetty: > I think we should be more explicit that this is an rest/openapi/swagger target
> 
> e.g., validate-openapi-spec, validate-openapi, ...

done

sudeshrshetty: > Also I see below that the openapi is generated.
> 
> Are you sure we shouldn't call the target generate-foo?

we generate the spec only to feed it into validate command, we are not yet including spec in codebase. Overall objective of this target it to validate the rest api code against open API specifications.

Makefile
sudeshrshetty: done

Makefile
troyronda: Would someone want to consume the generated specs separate from validation?

README.md
```diff
@@ -46,6 +46,14 @@ If you wish to regenerate it, you can delete this folder and:
  or
 2. cd into `pkg/didcomm/transport/http/` and run `go generate`
 
+##### Controller REST API specifications generation
+Controller REST API specifications are generated according to Open API 2.0 standards.
+
+It can be generated by running command `make generate-openapi-test`
+
+Generated spec can be found under 
+`build/rest/openapi/spec`
```
troyronda: It would be useful to also give instructions on how to run the openapi client with the generated specs as input.

scripts/generate-spec.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
troyronda: filename should be more explicit.

README.md
sudeshrshetty: Will be added as part of #94

scripts/generate-spec.sh
sudeshrshetty: done

Makefile
```diff
@@ -29,3 +31,10 @@ bdd-test: clean
 .PHONY: clean
 clean:
 	rm -f coverage.txt
+	rm -Rf ./build
+
+.PHONY: generate-openapi-spec
+generate-openapi-spec: clean
+	@echo "Generating and validating controller API specifications using Open API"
```
llorllale: I wonder why the command that generates the spec file needs to validate it as well.

sudeshrshetty: swagger doesn't provide two-in-one command which generates & validates in one go.
`swagger generate` can fail if there is a syntax or missing fields in annotations.
`swagger validate` fails if any request/response model is missing or if there are any orphan models or if any misconfiguration under 'Open API Specification 2.0' standards.

llorllale: I was imprecise - by "command" I meant the `make` recipe "generate-openapi-spec".

My assumption is that if `swagger generate` succeeds then `swagger validate` will succeed as well. If `swagger generate` fails then `make generate-openapi-spec` fails and `swagger validate` won't be executed at all.

Makefile
llorllale: Good question - and it's what I understood from our last conversation.

Makefile
sudeshrshetty: I understand what you are saying, but your assumption is wrong here.

`swagger generate` passes 90% of the time since it fills spec with defaults if any information is missing. It fails if there any compilation error and if there any annotation syntax issue.

`swagger validate` fails most of the time since it never scans code, it scans generated spec and make sure all necessary info and all mappings are intact (between path, models etc), and there aren't any orphan request/response/models.

llorllale: Oh ok - I think I see now. So, for example, there is the possibility that `swagger generate` will output orphaned models and then `swagger validate` would detect this error. Did I get it right?

sudeshrshetty: Yes, and generate scans code, validate scans spec.

PR #278, opened 2019-09-18
feat: Outbound dispatcher

fqutishat:
- create outbound dispatcher interface
- allow user to inject custom outbound dispatcher
- access outbound dispatcher from context


pkg/didcomm/dispatcher/api.go
```diff
@@ -0,0 +1,42 @@
...
+// Destination provides the recipientKeys, routingKeys, and serviceEndpoint populated from Invitation
+type Destination struct {
+	RecipientKeys   []string
+	ServiceEndpoint string
+	RoutingKeys     []string
+}
+
+// Outbound interface
+type Outbound interface {
+	Send([]byte, *Destination) error
```
llorllale: nitpick - I'd appreciate if instead of `[]byte` this was an `interface{}`, thus relieving all callers from having to marshal the messages themselves (there will be several callers that will have to do this)

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"errors"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+)
+
+// OutboundDispatcher dispatch msgs to destination
+type OutboundDispatcher struct {
+	outboundTransport transport.OutboundTransport
```
llorllale: outbound transport cannot be set before inspecting the endpointURL contained in the `*Destination` passed to the `Send()` method

pkg/framework/aries/framework.go
```diff
@@ -25,14 +25,16 @@ type DIDResolver interface {
...
+	transport                 api.TransportProviderFactory
+	didResolver               DIDResolver
+	storeProvider             storage.Provider
+	protocolSvcCreators       []api.ProtocolSvcCreator
+	services                  []dispatcher.Service
+	inboundTransport          transport.InboundTransport
+	walletCreator             api.WalletCreator
+	wallet                    api.CloseableWallet
+	outboundDispatcherCreator dispatcher.OutboundCreator
+	outboundDispatcher        dispatcher.Outbound
```
llorllale: Second instance of this pattern inside this `Aries` struct: we're adding an attribute with a value provided by another one. In this case, this `outboundDispatcher` was created by `outboundDispatcherCreator`. Previously we added `services` being created by `procotolSvcCreators`. Are we *certain* that `Aries` needs to hold all these?

fqutishat: I removed outboundDispatcher not needed aries to hold it

pkg/didcomm/dispatcher/outbound.go
fqutishat: Done I add array of outboundTransport and add type method

pkg/didcomm/dispatcher/api.go
fqutishat: @troyronda @rolsonquadras what do you think?

rolsonquadras: @fqutishat i feel interface{} would be more appropriate.

talwinder50: yea i think interface would be more appropriate

fqutishat: ok I will change it

fqutishat: done

pkg/didcomm/transport/transport_interface.go
```diff
@@ -11,6 +11,8 @@ package transport
 type OutboundTransport interface {
 	// Send send a2a exchange data
 	Send(data string, destination string) (string, error)
+	// Type return outbound type
+	Type() string
```
llorllale: Consider renaming this to `Scheme()`

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"errors"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+)
+
+// OutboundDispatcher dispatch msgs to destination
+type OutboundDispatcher struct {
+	outboundTransport []transport.OutboundTransport
```
llorllale: Consider instead just accepting a `func(scheme string) (transport.OutboundTransport, error)`. That way this OutboundDispatcher doesn't have to search through them itself - a nice separation of concerns.

pkg/didcomm/transport/transport_interface.go
fqutishat: done

PR #279, opened 2019-09-18
refactoring: Prevent deadlock when writing to a nil channel

soluchok:
* double check actionEvent to nil


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -278,10 +281,13 @@ func (s *Service) sendEvent(msg dispatcher.DIDCommMsg, threadID string, nextStat
 
 	// trigger the registered action events
 	s.lock.RLock()
-	event := s.actionEvent
+	aEvent = s.actionEvent
```
bstasyszyn: We already have aEvent set on line 248. Do we need to set it again?

soluchok: I did that on purpose, `RegisterEvent` function has a comment `If called multiple times, the events will be sent to the last channel.` So, it is possible that `aEvent` of the beginning can be `chan1` and by the end of the function `aEvent` can be `chan2` according to the comment I mentioned above looks like we need to double-check that bc `aEvent` may not be equal to `s.actionEvent` at the end of the function.

Solutions:
1. we can use `aEvent` which we got at the beginning
2. double-check: we will check `aEvent` at the beginning to nill, and at the end, we will check it one more time (because `aEvent` may not be equal to `s.actionEvent`)
3. we can ignore `s.actionEvent` at the beginning (the function will be executed even if `s.actionEvent` is nill) and at the end we can decide whether we can send a message to the channel or not

bstasyszyn: 1. is the best solution. The reason we lock member variables is so that we get a consistent view of the variables during the execution of the function. It doesn't matter if the variable changes immediately after we release the lock - we should still use the one we retrieved at the beginning.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -277,11 +281,8 @@ func (s *Service) sendEvent(msg dispatcher.DIDCommMsg, threadID string, nextStat
 	}
 
 	// trigger the registered action events
-	s.lock.RLock()
-	event := s.actionEvent
-	s.lock.RUnlock()
-	event <- didCommEvent
-
+	// TODO: it can be panic if the channel was closed
```
bstasyszyn: Is there an action item (issue) associated with this TODO?

soluchok: no, it is more for general understanding, need to decide whether it is our or a client responsibility

bstasyszyn: Then I would remove the TODO.

PR #285, opened 2019-09-19
protocol: Added states for introduce protocol

soluchok:
* defined states (introduce protocol)

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -1,0 +1,24 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package introduce
+
+import "github.com/hyperledger/aries-framework-go/pkg/common/metadata"
+
+const (
+	// Introduce protocol name
+	Introduce = "introduce"
+	// Spec defines the introduce spec
+	Spec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// ConnectionProposal defines the introduce proposal message type.
+	ConnectionProposal = Spec + "proposal"
+	// ConnectionResponse defines the introduce response message type.
+	ConnectionResponse = Spec + "response"
+	// ConnectionInvitation defines the introduce invitation message type.
+	ConnectionInvitation = Spec + "invitation"
+	// ConnectionConfirm defines the introduce confirm message type.
+	ConnectionConfirm = Spec + "confirm"
+)
--- a/pkg/didcomm/protocol/introduce/states.go
+++ b/pkg/didcomm/protocol/introduce/states.go
@@ -64,7 +64,9 @@ func (s *start) Name() string {
 }
 
 func (s *start) CanTransitionTo(next state) bool {
-	return false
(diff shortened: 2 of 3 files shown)


llorllale:
@soluchok please create an issue for this task and reference it in this PR's description

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -0,0 +1,24 @@
...
+	// Spec defines the introduce spec
+	Spec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// ConnectionProposal defines the introduce proposal message type.
+	ConnectionProposal = Spec + "proposal"
+	// ConnectionResponse defines the introduce response message type.
+	ConnectionResponse = Spec + "response"
+	// ConnectionInvitation defines the introduce invitation message type.
+	ConnectionInvitation = Spec + "invitation"
+	// ConnectionConfirm defines the introduce confirm message type.
+	ConnectionConfirm = Spec + "confirm"
```
llorllale: There is no "confirm" message in the `introduce` protocol - what we should use instead is an [`ack`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0015-acks/README.md) (but using the `spec/introduce/1.0` namespace).

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -0,0 +1,187 @@
...
+// start state
+type start struct {
+}
+
+func (s *start) Name() string {
+	return stateNameStart
+}
+
+func (s *start) CanTransitionTo(next state) bool {
+	// Introducer can go to arranging or done state
```
llorllale: I've opened #288 to track hyperledger/aries-rfcs#227.

It doesn't make sense to have a direct transition start -> done.

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -0,0 +1,187 @@
...
+	// TODO: need clarification: Introducee can go to stateNameDeciding
```
llorllale: This is correct. There is no TODO required here, or is there?

pkg/didcomm/protocol/introduce/states_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+)
+
+// nolint:gochecknoglobals
```
llorllale: @troyronda

pkg/didcomm/protocol/introduce/states.go
llorllale: Btw - I wouldn't want this issue to block this PR's merge. #288 can refactor if required later on.

pkg/didcomm/protocol/introduce/states.go
soluchok: I added todo bc in State Machine it was `transition to "invited"` I suppose it was a mistake it should be `transition to "deciding"` instead of `transition to "invited"`.

llorllale: We should just remove the TODO and let #288 sort it out.

Issue #287, opened 2019-09-19
RFC 0028: Determine states for introduce protocol


soluchok:
Need to define states for introduce protocol according to the state machine follow the link below for more information

https://github.com/hyperledger/aries-rfcs/tree/master/features/0028-introduce

PR #289, opened 2019-09-19
protocol: Added states for introduce protocol

soluchok:
This branch name was renamed to see old comments visit this [link](https://github.com/hyperledger/aries-framework-go/pull/285)

Issue: https://github.com/hyperledger/aries-framework-go/issues/287

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -1,0 +1,24 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package introduce
+
+import "github.com/hyperledger/aries-framework-go/pkg/common/metadata"
+
+const (
+	// Introduce protocol name
+	Introduce = "introduce"
+	// IntroduceSpec defines the introduce spec
+	IntroduceSpec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// ProposalMsgType defines the introduce proposal message type.
+	ProposalMsgType = IntroduceSpec + "proposal"
+	// RequestMsgType defines the introduce request message type.
+	RequestMsgType = IntroduceSpec + "request"
+	// ResponseMsgType defines the introduce response message type.
+	ResponseMsgType = IntroduceSpec + "response"
+	// AckMsgType defines the introduce ack message type.
+	AckMsgType = IntroduceSpec + "ack"
+)
--- a/pkg/didcomm/protocol/introduce/states.go
+++ b/pkg/didcomm/protocol/introduce/states.go
@@ -1,0 +1,188 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package introduce
+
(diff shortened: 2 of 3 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -0,0 +1,24 @@
...
+	// Spec defines the introduce spec
+	Spec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// ConnectionProposal defines the introduce proposal message type.
+	ConnectionProposal = Spec + "proposal"
+	// ConnectionResponse defines the introduce response message type.
+	ConnectionResponse = Spec + "response"
+	// ConnectionInvitation defines the introduce invitation message type.
+	ConnectionInvitation = Spec + "invitation"
+	// ConnectionConfirm defines the introduce confirm message type.
+	ConnectionConfirm = Spec + "confirm"
```
llorllale: This is the only blocker for me - like I said [here](https://github.com/hyperledger/aries-framework-go/pull/285#discussion_r326180812), "confirm" is not a msg defined in the `introduce` RFC. We should use `ack`.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -0,0 +1,24 @@
...
+	// Introduce protocol name
+	Introduce = "introduce"
+	// Spec defines the introduce spec
+	Spec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// ConnectionProposal defines the introduce proposal message type.
+	ConnectionProposal = Spec + "proposal"
+	// ConnectionResponse defines the introduce response message type.
+	ConnectionResponse = Spec + "response"
+	// ConnectionInvitation defines the introduce invitation message type.
+	ConnectionInvitation = Spec + "invitation"
```
llorllale: `invitation` is also not an `introduce` message type. This is actually a `didexchange` msgType. What the Introducer does is it forwards the `Invitation`s to each introducee.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "github.com/hyperledger/aries-framework-go/pkg/common/metadata"
+
+const (
+	// Introduce protocol name
+	Introduce = "introduce"
+	// Spec defines the introduce spec
+	Spec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// ConnectionProposal defines the introduce proposal message type.
+	ConnectionProposal = Spec + "proposal"
```
rolsonquadras: May be we can use "Introduce" prefix here as this is an introduce protocol.

rolsonquadras:
@soluchok could you please link the issue# to this PR ?

pkg/didcomm/protocol/introduce/service.go
soluchok: I did that on purpose, my IDE (GoLand) says `Name starts with package name`

llorllale: @soluchok I don't understand your comment? This is package `introduce`. Also, we're missing the [`request`](https://github.com/hyperledger/aries-rfcs/tree/master/features/0028-introduce#request) message.

Issue #250, opened 2019-09-12
aries-agentd CLI


DRK3:
Created from #94 

start aries-agentd as command line command (COBRA). Provide port etc as command line arguments

PR #290, opened 2019-09-19
feat: aries-agentd CLI

DRK3:
https://github.com/hyperledger/aries-framework-go/issues/250

Start aries-agentd with command line arguments via Cobra. What used to be environment variables are now command line arguments.

Signed-off-by: Derek Trider <derektrider@gmail.com>


cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,304 @@
...
+		args := []string{"--" + AgentInboundHostFlagName, testInboundHostURL, "--" + AgentDBPathFlagName, ""}
+		startCmd.SetArgs(args)
+		startCmd.Execute()
+		done <- true
+	}()
+
+	select {
+	case res := <-done:
+		require.True(t, res)
+	case <-time.After(5 * time.Second):
```
DRK3: Talked to Bob and he suggested that I can mock the server instead of letting it actually launch here.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,102 @@
...
+const AgentDBPathFlagUsage = "Path to database"
+
+func Cmd() *cobra.Command {
+	startCmd := &cobra.Command{
+		Use:   "start",
+		Short: "Start an agent",
+		Long:  `Start an Aries agent controller`,
+
+		Run: func(cmd *cobra.Command, args []string) {
+			host, _ := cmd.Flags().GetString(AgentHostFlagName)
```
rolsonquadras: Need to handle error here. The startAgent will fail with "Unable to start aries agentd, host not provided" error when host is passed as empty, without giving the actual error message to the user.

Same for other cases.

cmd/aries-agentd/startcmd/start_test.go
DRK3: This suggestion came up because I was asking how I can test a successful server start without causing it to block. Running the server in another goroutine causes issues too since the goroutine lives on for other tests and causes them to fail.

llorllale:
@DRK3 please link to the issue in this PR's description

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,102 @@
...
+	startCmd.MarkFlagRequired(AgentHostFlagName)
+	startCmd.Flags().StringP(AgentInboundHostFlagName, AgentInboundHostFlagShorthand, "", AgentInboundHostFlagUsage)
+	startCmd.MarkFlagRequired(AgentInboundHostFlagName)
+	startCmd.Flags().StringP(AgentDBPathFlagName, AgentDBPathFlagShorthand, "", AgentDBPathFlagUsage)
+	startCmd.MarkFlagRequired(AgentDBPathFlagName)
+
+	return startCmd
+}
+
+func startAgent(host string, inboundHost string, dbPath string) {
```
llorllale: This function should be returning all those errors instead of logging them.

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const testHostURL = "localhost:8080"
+const testInboundHostURL = "localhost:8081"
+
+func TestStartCmdContents(t *testing.T) {
+	startCmd := Cmd()
+
+	if startCmd.Use != "start" {
```
llorllale: use `require.Equal(t, "start", startCmd.Use)` to decrease verbosity

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,304 @@
...
+		t.Error()
+	}
+	if startCmd.Short != "Start an agent" {
+		t.Error()
+	}
+	if startCmd.Long != `Start an Aries agent controller` {
+		t.Error()
+	}
+
+	if !areFlagPropertiesCorrect(startCmd, AgentHostFlagName, AgentHostFlagShorthand, AgentHostFlagUsage) {
```
llorllale: We won't know which of them are incorrect when this returns `false`. Rather just use `require.Equals()` within the body of `areFlagPropertiesCorrect()` - this way we can trace the failed assertion to an exact line of test code

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,304 @@
...
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+const testHostURL = "localhost:8080"
+const testInboundHostURL = "localhost:8081"
```
soluchok: We need to rewrite this, please see what I did  https://github.com/hyperledger/aries-framework-go/pull/317
Probably it will not conflict because the file name was changed but be careful

DRK3:
> @DRK3 please link to the issue in this PR's description

Done. @llorllale

DRK3:
@troyronda @fqutishat @llorllale 

Note that I believe I am unable to increase code coverage any further - I made comments on the lines of code in files changed that I don't think I can reasonably test. Let me know your thoughts.

go.mod
```diff
@@ -7,10 +7,11 @@ module github.com/hyperledger/aries-framework-go
 require (
 	github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
-	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/golang/protobuf v1.3.1 // indirect
```
fqutishat: Why we need golang/protobuf

cmd/aries-agentd/main.go
```diff
@@ -25,74 +25,24 @@ SPDX-License-Identifier: Apache-2.0
...
-	logger.Infof("Starting aries agentd on host [%s]", host)
+	rootCmd.AddCommand(startcmd.Cmd(&startcmd.RealServer{}))
 
-	// start server on given port and serve using given handlers
-	err = http.ListenAndServe(host, router)
-	if err != nil {
-		logger.Fatalf("Failed to start aries agentd on port [%s], cause:  %s", host, err)
+	if err := rootCmd.Execute(); err != nil {
+		logger := log.New("aries-framework/agentd")
+		logger.Fatalf(err.Error())
```
DRK3: I'm unable to get unit test coverage on lines 45 and 46 since logger.Fatalf(err.Error()) executes os.exit(1) and that seems to prevent the unit test coverage framework from detecting it.
I tried a workaround I found online which involves calling another process and checking the return status on that process. While this would work for doing the test, the code coverage tool is unable to pick it up since it happens in another process.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,135 @@
...
+
+			dbPath, err := cmd.Flags().GetString(AgentDBPathFlagName)
+			if err != nil {
+				logger.Errorf(err.Error())
+			}
+
+			err = startAgent(server, host, inboundHost, dbPath)
+			if err != nil {
+				logger.Errorf(err.Error())
+			}
```
DRK3: I can't seem to test these logger.Errorf(err.Error()) lines because the Cobra framework actually prevents this "run" func from executing if the incorrect arguments are passed in. So lines 59, 64, 69 and 74 will never execute.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	framework, err := aries.New(opts...)
+	if err != nil {
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to initialize framework :  %w", host, err)
+	}
+
+	ctx, err := framework.Context()
+	if err != nil {
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get aries context : %w", host, err)
```
DRK3: Unsure how to unit test line 110. Aries framework is created through controlled inputs. Note also that the existing test cases didn't test this (was originally in main.go)

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// get all HTTP REST API handlers available for controller A PI
+	restService, err := restapi.New(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get rest service api :  %w", host, err)
```
DRK3: Also not sure how to test this.

cmd/aries-agentd/main_test.go
```diff
@@ -2,280 +2,23 @@
...
-		if err != nil {
-			t.Fatalf("Failed to clear leveldb directory: %s", err)
-		}
-	}
-}
+// Strips out the extra args that the unit test framework adds
+// This allows main() to execute as if it was called directly from the command line
+func setUpArgs() {
+	args := []string{os.Args[0]}
+	os.Args = args
```
soluchok: can be done in one line `os.Args = os.Args[:1]`

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,135 @@
...
+func startAgent(server server, host string, inboundHost string, dbPath string) error {
+	if host == "" {
+		return errors.New(strings.ToLower(MissingHostErrorMessage))
+	}
+
+	var opts []aries.Option
+	if inboundHost == "" {
+		return errors.New(strings.ToLower(MissingInboundHostErrorMessage))
+	}
+	opts = append(opts, defaults.WithInboundHTTPAddr(inboundHost))
```
soluchok: `var opts []aries.Option` should be before line 97

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	for _, tt := range tests {
+
+		resp, err := http.DefaultClient.Do(tt.r)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		defer func() {
```
soluchok: Do not use defer in a loop it can be a potential memory leak

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,344 @@
...
+			}
+			require.NotEmpty(t, response)
+			require.True(t, isJSON(response))
+		}
+	}
+
+}
+
+// isJSON checks if response is json
+func isJSON(res []byte) bool {
```
soluchok: use empty struct
```
func isJSON(res []byte) bool {
	var js struct{}
	return json.Unmarshal(res, &js) == nil
}
```

DRK3:
I should have mentioned more clearly: I moved the tests that used to be in main_test.go to start_test.go, so some of the tests in start_test.go weren't actually written by me but will appear to be in this PR

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -0,0 +1,344 @@
+/*
```
DRK3: Heads up: Some of the tests here I had moved from main_test.go but I didn't originally write.

cmd/aries-agentd/startcmd/start_test.go
DRK3: This wasn't actually code that I wrote - I just moved it from the previous file to here. Can you make a separate issue for this? Or if you know how to refactor the method then let me know and I can implement it if it doesn't take too long.

cmd/aries-agentd/startcmd/start_test.go
DRK3: This was code that I didn't originally write - I just moved it from another file to here. It was an easy change to make though, so I did it.

cmd/aries-agentd/startcmd/start.go
DRK3: This was code that I didn't originally write - I just moved it from another file to here. It was an easy change to make though, so I did it.

cmd/aries-agentd/main_test.go
DRK3: Done!

go.mod
DRK3: Just reran 'go mod tidy' and it disappeared. Not sure why it was there in the first place, but it's gone now.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,166 @@
...
+	err = startCmd.MarkFlagRequired(AgentInboundHostFlagName)
+	if err != nil {
+		logger.Errorf(err.Error())
+	}
+	startCmd.Flags().StringP(AgentDBPathFlagName, AgentDBPathFlagShorthand, "", AgentDBPathFlagUsage)
+	err = startCmd.MarkFlagRequired(AgentDBPathFlagName)
+	if err != nil {
+		logger.Errorf(err.Error())
+	}
+
```
DRK3: The "MarkFlagRequired" method throws an error if the flag name passed into it doesn't exist in the command. But it always will, so I'm not sure how I can actually test that. The linter will complain if I don't handle the error.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/defaults"
+	"github.com/hyperledger/aries-framework-go/pkg/restapi"
+)
+
+// AgentHostFlagName is the flag name for the agent host command line argument.
+const AgentHostFlagName = "api-host"
```
rolsonquadras: We can group all the constant into one block. [Reference](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/service.go#L31).

cmd/aries-agentd/startcmd/start_test.go
sudeshrshetty: thats function supposed to find if given bytes is in JSON format, not sure empty struct can help here. `map[string]interface{}` makes sure if given bytes are in JSON format @soluchok

cmd/aries-agentd/startcmd/start.go
DRK3: Done!

cmd/aries-agentd/main.go
```diff
@@ -25,74 +25,25 @@ SPDX-License-Identifier: Apache-2.0
...
-	logger.Infof("Starting aries agentd on host [%s]", host)
+	rootCmd.AddCommand(startcmd.Cmd(&startcmd.HTTPServer{}))
 
-	// start server on given port and serve using given handlers
-	err = http.ListenAndServe(host, router)
-	if err != nil {
-		logger.Fatalf("Failed to start aries agentd on port [%s], cause:  %s", host, err)
+	if err := rootCmd.Execute(); err != nil {
+		logger := log.New("aries-framework/agentd")
+		logger.Fatalf(err.Error())
```
rolsonquadras: Better to log this with a message. [Reference](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/transport/http/support_test.go#L49).

cmd/aries-agentd/startcmd/start_test.go
DRK3: For now I'll revert back to the original way. Since I didn't originally write this code I think it's probably safer to leave as is.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -0,0 +1,168 @@
...
+func Cmd(server server) *cobra.Command {
+	startCmd := &cobra.Command{
+		Use:   "start",
+		Short: "Start an agent",
+		Long:  `Start an Aries agent controller`,
+
+		Run: func(cmd *cobra.Command, args []string) {
+			host, err := cmd.Flags().GetString(AgentHostFlagName)
+			if err != nil {
+				logger.Errorf(err.Error())
```
rolsonquadras: log error with a message. [Reference](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/transport/http/outbound.go#L88).

cmd/aries-agentd/main_test.go
```diff
@@ -2,280 +2,22 @@
...
-
-	select {
-	case res := <-done:
-		require.True(t, res)
-	case <-time.After(5 * time.Second):
-		t.Fatal("agent should fail to start when inbound host address not provided")
-	}
+// Correct behaviour is for main to finish with exit code 0.
+// This test fails otherwise. However, this can't be checked by the unit test framework, so t goes unused.
+func TestCallingWithNoUserArgs(t *testing.T) { //nolint - see above
```
sudeshrshetty: typo in comment: so t goes unused.
fn name: TestWithOutUserArgs??

cmd/aries-agentd/startcmd/start.go
sudeshrshetty: return error, use `RunE` for command

DRK3: Done

cmd/aries-agentd/main_test.go
DRK3: It actually wasn't a typo - the "t" is referring to t *testing.T. But it does look confusing. I reworded it to: // This test fails otherwise. However, this can't be checked by the unit test framework. The *testing.T argument is
// only there so that this test gets picked up by the framework but otherwise we don't need it.

DRK3: Done - renamed the test.

cmd/aries-agentd/startcmd/start.go
DRK3: Done.

cmd/aries-agentd/main.go
DRK3: Done!

PR #292, opened 2019-09-19
feat: Legacy encrypt (current aries RFC 19), compatible with ACA-Py

### LINKED ISSUE #139: Create package to support legacy envelope encryption format
As discussed in [aries-rfcs/issues/133](https://github.com/hyperledger/aries-rfcs/issues/133), implement the 'legacy' crypter from [RFC #0019](https://github.com/hyperledger/aries-rfcs/blob/master/features/0019-encryption-envelope/README.md) in a sibling package to the JWE implementation (#37).

Part of #36 

@Baha-sk

Moopli:
Implements legacy encrypter for compatibility with ACA-Py.

Closes #139

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@
 module github.com/hyperledger/aries-framework-go
 
 require (
+	github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/google/uuid v1.1.1
--- a/pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
+++ b/pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
@@ -1,0 +1,183 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package authcrypt
+
+import (
+	"crypto/rand"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/agl/ed25519/extra25519"
+	"golang.org/x/crypto/blake2b"
+	"golang.org/x/crypto/ed25519"
+	"golang.org/x/crypto/nacl/box"
+)
+
+type privateEd25519 [ed25519.PrivateKeySize]byte
+type publicEd25519 [ed25519.PublicKeySize]byte
+
+type keyPairEd25519 struct {
+	priv *privateEd25519
+	pub  *publicEd25519
+}
+
+// CurveKeySize is the size of public and private Curve25519 keys in bytes
+const CurveKeySize int = 32
+
+type privateCurve25519 [CurveKeySize]byte
+type publicCurve25519 [CurveKeySize]byte
+
+type keyPairCurve25519 struct {
+	priv *privateCurve25519
+	pub  *publicCurve25519
+}
+
+// Crypter represents an Authcrypt Encrypter (Decrypter) that outputs/reads legacy Aries envelopes
+type Crypter struct {
+	sender     keyPairEd25519
+	recipients []*publicEd25519
+	randSource io.Reader
+}
+
+// New will create a Crypter that encrypts messages using the legacy Aries format
+// Note: legacy crypter does not support XChacha20Poly1035 (XC20P), only Chacha20Poly1035 (C20P)
+func New(sender keyPairEd25519, recipients []*publicEd25519) (*Crypter, error) { // nolint: lll
+
+	if len(recipients) == 0 {
+		return nil, errors.New("empty recipients keys, must have at least one recipient")
+	}
+
+	c := &Crypter{
+		sender:     sender,
+		recipients: recipients,
+		randSource: rand.Reader,
+	}
+
+	if !isKeyPairValid(sender) {
+		return nil, fmt.Errorf(
+			"sender keyPair not supported, it must have a %d byte private key and %d byte public key",
+			ed25519.PrivateKeySize, ed25519.PublicKeySize)
+	}
+
+	return c, nil
+}
+
+func (c *Crypter) setRandSource(source io.Reader) {
+	c.randSource = source
+}
+
+func isKeyPairValid(kp keyPairEd25519) bool {
+	if kp.priv == nil || kp.pub == nil {
+		return false
+	}
+
+	return true
+}
+
+// envelope is the full payload envelope for the JSON message
+type envelope struct {
+	Protected  string `json:"protected,omitempty"`
+	IV         string `json:"iv,omitempty"`
+	CipherText string `json:"ciphertext,omitempty"`
+	Tag        string `json:"tag,omitempty"`
+}
+
+// protected is the protected header of the JSON envelope
+type protected struct {
+	Enc        string      `json:"enc,omitempty"`
+	Typ        string      `json:"typ,omitempty"`
+	Alg        string      `json:"alg,omitempty"`
+	Recipients []recipient `json:"recipients,omitempty"`
+}
+
+// recipient holds the data for a recipient in the envelope header
+type recipient struct {
+	EncryptedKey string          `json:"encrypted_key,omitempty"`
+	Header       recipientHeader `json:"header,omitempty"`
+}
+
+// recipientHeader holds the header data for a recipient
+type recipientHeader struct {
+	KID    string `json:"kid,omitempty"`
+	Sender string `json:"sender,omitempty"`
+	IV     string `json:"iv,omitempty"`
+}
+
+// publicEd25519toCurve25519 takes an Ed25519 public key and provides the corresponding Curve25519 public key
+//  This function wraps PublicKeyToCurve25519 from Adam Langley's ed25519 repo: https://github.com/agl/ed25519
+func publicEd25519toCurve25519(pub *publicEd25519) (*publicCurve25519, error) {
+	if pub == nil {
+		return nil, errors.New("key is nil")
+	}
+	pkOut := new([CurveKeySize]byte)
+	success := extra25519.PublicKeyToCurve25519(pkOut, (*[ed25519.PublicKeySize]byte)(pub))
+	if !success {
+		return nil, errors.New("failed to convert public key")
+	}
+	return (*publicCurve25519)(pkOut), nil
+}
+
+// secretEd25519toCurve25519 converts a secret key from Ed25519 to curve25519 format
+//  This function wraps PrivateKeyToCurve25519 from Adam Langley's ed25519 repo: https://github.com/agl/ed25519
+func secretEd25519toCurve25519(priv *privateEd25519) (*privateCurve25519, error) {
+	if priv == nil {
+		return nil, errors.New("key is nil")
+	}
+	sKOut := new([CurveKeySize]byte)
+	extra25519.PrivateKeyToCurve25519(sKOut, (*[ed25519.PrivateKeySize]byte)(priv))
+	return (*privateCurve25519)(sKOut), nil
+}
+
+func makeNonce(pub1, pub2 []byte) ([]byte, error) {
+	var nonce [24]byte
+	// generate an equivalent nonce to libsodium's (see link above)
+	nonceWriter, err := blake2b.New(24, nil)
+	if err != nil {
+		return nil, err
+	}
+	_, err = nonceWriter.Write(pub1)
+	if err != nil {
+		return nil, err
+	}
+	_, err = nonceWriter.Write(pub2)
+	if err != nil {
+		return nil, err
+	}
+
+	nonceOut := nonceWriter.Sum(nil)
+	copy(nonce[:], nonceOut)
+
+	return nonce[:], nil
+}
+
+// sodiumBoxSeal will encrypt a msg (in the case of this package, it will be
+// an ephemeral key concatenated to the sender's public key) using the
+// recipient's pubKey, this is equivalent to libsodium's C function: crypto_box_seal()
+// https://libsodium.gitbook.io/doc/public-key_cryptography/sealed_boxes#usage
+func sodiumBoxSeal(msg []byte, recPub *publicCurve25519, randSource io.Reader) ([]byte, error) {
+	var nonce [24]byte
+	// generate ephemeral curve25519 asymmetric keys
+	epk, esk, err := box.GenerateKey(randSource)
+	if err != nil {
+		return nil, err
+	}
+	// generate an equivalent nonce to libsodium's (see link above)
+	nonceSlice, err := makeNonce(epk[:], recPub[:])
+	if err != nil {
+		return nil, err
+	}
+	copy(nonce[:], nonceSlice)
+
+	var out = make([]byte, len(epk))
+	copy(out, epk[:])
+
+	// now seal the msg with the ephemeral key, nonce and recPub (which is recipient's publicKey)
+	ret := box.Seal(out, msg, &nonce, (*[32]byte)(recPub), esk)
+
+	return ret, nil
+}
--- a/pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
+++ b/pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
@@ -1,0 +1,573 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package authcrypt
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io"
+	insecurerand "math/rand"
+	"testing"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/ed25519"
+	"golang.org/x/crypto/nacl/box"
+	"golang.org/x/crypto/nacl/sign"
+)
+
+// failReader wraps a Reader, used for testing different failure checks for encryption tests.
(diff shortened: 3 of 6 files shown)


pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -0,0 +1,219 @@
...
+import (
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/json"
+
+	"github.com/btcsuite/btcutil/base58"
+	"golang.org/x/crypto/nacl/box"
+	"golang.org/x/crypto/poly1305"
+
```
troyronda: Please cleanup imports

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -0,0 +1,219 @@
...
+func (c *Crypter) Encrypt(payload []byte) ([]byte, error) {
+	var _, err error
+
+	nonce := make([]byte, c.nonceSize)
+	_, err = rand.Reader.Read(nonce)
+	if err != nil {
+		return nil, err
+	}
+
+	var cek *[chacha20poly1305.KeySize]byte
```
troyronda: I would suggest aliasing this import to chacha.

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -0,0 +1,219 @@
...
+			return nil, err
+		}
+
+		encodedRecipients = append(encodedRecipients, *recipient)
+	}
+
+	return encodedRecipients, nil
+}
+
+// Note: the spec is incorrect with the parameters for encrypted_key: the CEK is the message to encrypt
```
troyronda: Did you raise an issue on aries-rfcs?

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -0,0 +1,219 @@
...
+	encCEK := box.Seal(nil, cek[:], &nonce, (*[32]byte)(recPKCurve), (*[32]byte)(senderSKCurve))
+
+	var encSender []byte
+	encSender, err = sodiumBoxSeal([]byte(base58.Encode(c.sender.pub[:])), recPKCurve)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Recipient{
+		EncryptedKey: base64.URLEncoding.EncodeToString(encCEK),
```
troyronda: Why not RawURLEncoding like the jwe package is doing?

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
```diff
@@ -0,0 +1,201 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
```
troyronda: Do we have a test where the inputs are generated by the python codebase (and pasted into the test?)

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
```diff
@@ -0,0 +1,43 @@
...
+	"golang.org/x/crypto/chacha20poly1305"
+	"golang.org/x/crypto/nacl/box"
+	errors "golang.org/x/xerrors"
+)
+
+// Decrypt will decode the envelope using the legacy format
+// Using (X)Chacha20 encryption algorithm and Poly1035 authenticator
+func (c *Crypter) Decrypt(envelope []byte, recipientPrivKey *[chacha20poly1305.KeySize]byte) ([]byte, error) {
+
+	return nil, nil
```
troyronda: Is there a missing TODO?

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
Moopli: The spec in question is being amended to a JWE format (that @Baha-sk is working on) anyway - perhaps we should request it be split to keep a spec for the legacy format for backwards-compatibility, and amend that.

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
Moopli: Encoding in raw base64 URL encoding, ie, without padding, is safe when the recipient knows exactly how long the data should be. I don't expect it would work when I'm dealing with a generic decoder function (in python) that isn't being told the length of the output data. I can test it.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
Moopli: Yes, they'll come with decrypt.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
```diff
@@ -0,0 +1,224 @@
...
+		nonceSize = chacha.NonceSize
+	default:
+		return nil, fmt.Errorf("encryption algorithm '%s' not supported", alg)
+	}
+
+	if len(recipients) == 0 {
+		return nil, errors.New("empty recipients keys, must have at least one recipient")
+	}
+	var recipientsKey []*publicEd25519
+	recipientsKey = append(recipientsKey, recipients...)
```
llorllale: is there a reason for assigning `recipients` into `recipientsKey`?

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
Moopli: Added, not marking as outdated for some reason

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
```diff
@@ -0,0 +1,224 @@
...
+func isKeyPairValid(kp keyPairEd25519) bool {
+	if kp.priv == nil || kp.pub == nil {
+		return false
+	}
+
+	return true
+}
+
+// Envelope is the full payload envelope for the JSON message
+type Envelope struct {
```
llorllale: Consider un-exporting these structs

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
Moopli: #294

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
```diff
@@ -0,0 +1,224 @@
...
+func publicEd25519toCurve25519(pub *publicEd25519) (*publicCurve25519, error) {
+	pkOut := new([CurveKeySize]byte)
+	success := extra25519.PublicKeyToCurve25519(pkOut, (*[CurveKeySize]byte)(pub))
+	if !success {
+		return nil, errors.New("failed to convert public key")
+	}
+	return (*publicCurve25519)(pkOut), nil
+}
+
+// TODO: consider wrapping agl's secret key conversion instead of reimplementing
```
llorllale: is this TODO current? Can it be removed?

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -0,0 +1,218 @@
...
+	protectedBytes, err := json.Marshal(protected)
+	if err != nil {
+		return nil, err
+	}
+
+	m, err := prettyPrint(protectedBytes)
+	if err != nil {
+		return nil, err
+	}
+	print(m, "\n")
```
llorllale: this needs to be removed

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
```diff
@@ -0,0 +1,198 @@
...
+	})
+}
+
+func TestSodiumBoxSeal(t *testing.T) {
+	var err error
+
+	recipient1Key, err := randCurveKeyPair(rand.Reader)
+	require.NoError(t, err)
+
+	t.Run("Generate a box_seal message to compare to ACA-Py:", func(t *testing.T) {
```
llorllale: where is the comparison to ACA-Py's output?

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	/*
+			Note: box.GenerateKey() generates a curve25519 keypair, not an Ed25519 keypair.
+			I think we should be generating Ed25519 keypairs since they are converted into
+				curve25519 keypairs for processing.
+
+		TODO: use Ed25519 keys, convert to curve25519 internally, encode the Ed25519 form for message packing
+	*/
+
+	t.Run("Success test case: given keys, generate envelope", func(t *testing.T) {
```
llorllale: it would be good to compare the actual result to an expected value

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -0,0 +1,218 @@
...
+
+// TODO: refactor Crypter into legacy/authcrypt/authcrypt.go, encrypt into legacy/authcrypt/encrypt.go,
+//  shared code into crypto/internal, and also add decrypt.go
+
+// TODO generate testcases in ACA-Py, make tests here
+
+// Encrypt will encode the payload argument
+// Using the protocol defined by Aries RFC 0019
+func (c *Crypter) Encrypt(payload []byte) ([]byte, error) {
+	var _, err error
```
llorllale: `_,` not needed

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
Moopli: That's something I want reviewed - I reimplemented a function with reference to agl's secret key conversion and libsodium's secret key conversion, since I didn't want the dependency. But I bit the bullet and imported agl's curve25519 package rather than implementing the _public_ key conversion, so maybe I should just turn this function into a wrapper or call the wrapped function directly as well.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
Moopli: Randomness (for nonces, CEKs) keeps me from generating encrypted envelopes in ACA-Py to compare against my own output. Right now I generate test data which can be pasted into a python test script to verify using ACA-Py decrypt, but once I have decrypt implemented I can create test data in ACA-Py that can be part of test cases here, testing decrypt, and I can also have test cases for my encrypt against my decrypt.

In essence, verification of correctness needs to happen on the decrypt side. I could work on making encrypt take a Reader as parameter for the randomness source, with the wallet being relied on to submit a CSPRNG.

Moopli:
> I'm concerned that most of the tests just consist of a) checking that there is no error, and b) the output is not empty.
> 
> It would be better if we tested actual results against expected, verifiable values.

I just pushed fixes to most of what's been brought up, some improvements to test coverage, and I made the crypter deterministic, taking an RNG as a parameter - `crypto/rand.Reader` under production conditions, but can take a seeded RNG for tests. I verified with a test using an empty payload, and I'll add more tests that I can now do, with a deterministic crypter.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
Moopli: My above concern should be solved now. You should soon see more tests like what you want to see.

go.mod
```diff
@@ -21,6 +22,7 @@ require (
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58 // indirect
 	golang.org/x/sys v0.0.0-20190616124812-15dcb6c0061f // indirect
 	golang.org/x/text v0.3.2 // indirect
+	golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7
```
troyronda: This needs to be tidied- xerrors should not be here.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
Baha-sk: yes, we should call/wrap agl's library, not reimplement it.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, err
+	}
+
+	nonceOut := nonceWriter.Sum(nil)
+	copy(nonce[:], nonceOut)
+
+	return nonce[:], nil
+}
+
+// TODO dupe of jwe/authcrypt encryptOID, refactor out
```
Baha-sk: most probably this function will stay here as jwe/authcrypt encryptOID will eventually be replaced with a JWK field (and this function removed from there).

Moopli: Will remove TODO

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
```diff
@@ -0,0 +1,420 @@
...
+			"3JVueJUSpj68N2HnRwzywiupgVhqeZBpQSztBD8YEtUqDUmWdrAd6K51dZyuoSTNTPCmPEDgkikuFGwWuh63dz8T")
+		rec3 := getB58EdKey("7snUUwA23DVBmafz9ibmBgwFFCUwzgTzmvcJGepuzjmK",
+			"3VaBzg9FgCq87etzLXktPWRFDJyUJrkxU9v5SwXVJE2WFMuaeffAjr3QTWWwjf3U8byJCwusZQqeuBZgAZphH16y")
+		rec4 := getB58EdKey("GSRovbnQy8HRjVjvzGbbfN387EX9NFfLj89C1ScXYfrF",
+			"jM6QdBopsuAz95PdFwwSg3ARSVQohArPKaZWNcvgLBz5NkqkZNCMMjftsXjHbGk7QLq7nabBV5FdHPnmXdhhdCB")
+
+		source := insecurerand.NewSource(6572692) // just a random const
+		constRand := insecurerand.New(source)
+
+		crypter, err := New(*sender, []*publicEd25519{rec1.pub, rec2.pub, rec3.pub, rec4.pub}, constRand)
```
Baha-sk: rather than passing in a Rand instance to New (it's supposed to be hidden from the caller), use a variable `randReader` that you can update in the test then revert at the end of the test

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
Baha-sk: need a test with ref payload built in Python and test decryption  here

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
```diff
@@ -0,0 +1,200 @@
...
+// Crypter represents an Authcrypt Encrypter (Decrypter) that outputs/reads legacy Aries envelopes
+type Crypter struct {
+	sender     keyPairEd25519
+	recipients []*publicEd25519
+	randSource io.Reader
+}
+
+// New will create a Crypter that encrypts messages using the legacy Aries format
+// Note: legacy crypter does not support XChacha20Poly1035 (XC20P), only Chacha20Poly1035 (C20P)
+func New(sender keyPairEd25519, recipients []*publicEd25519, randSource io.Reader) (*Crypter, error) { // nolint: lll
```
Baha-sk: rather than passing in a randSource instance to New (it's supposed to be hidden from the caller), use a variable randReader that you can update/mock in the tests then revert it back to crypt.Rand at the end of the test

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
Moopli: I made a private function that will set the rand source, and removed the parameter from New, now users cannot set an insecure rand source.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
Moopli: Ref payload from python decrypted in Go will come with Decrypter #294 
Rand source comment has been addressed in latest push

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
Moopli: Now that I can run the crypter using a deterministic rand source, I have a few tests which do this. Got some ideas on more I can do though.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
Moopli: Done

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
Moopli: These tests will come with #294 when the decrypter exists

Moopli:
@fqutishat every live code path is now exercised - only future regressions or cosmic rays will make errors happen in the few error cases that aren't tested.

pkg/didcomm/crypto/legacy/authcrypt/authcrypt.go
```diff
@@ -0,0 +1,183 @@
...
+	if !isKeyPairValid(sender) {
+		return nil, fmt.Errorf(
+			"sender keyPair not supported, it must have a %d byte private key and %d byte public key",
+			ed25519.PrivateKeySize, ed25519.PublicKeySize)
+	}
+
+	return c, nil
+}
+
+func (c *Crypter) setRandSource(source io.Reader) {
```
llorllale: This method is unused

Moopli: It's used in tests, but I can replace it with direct access to the Crypter.

PR #305, opened 2019-09-20
fix: Random port for tests

soluchok:
* added getRandomPort function

### DIFF
--- a/cmd/aries-agentd/main_test.go
+++ b/cmd/aries-agentd/main_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


rolsonquadras:
@DRK3 FYI. This file has changes to main_test.go which has been removed in the [PR](https://github.com/hyperledger/aries-framework-go/pull/290/files#diff-2bf1adf59ea642b335e892465b4aca8fR49). You might need to rebase again.

cmd/aries-agentd/main_test.go
```diff
@@ -19,8 +20,25 @@ import (
...
 
-const testURL = "localhost:8080"
-const testInboundURL = "localhost:8081"
+// nolint:gochecknoglobals
+var (
+	testURL        = fmt.Sprintf("localhost:%d", getRandomPort(false))
+	testInboundURL = fmt.Sprintf("localhost:%d", getRandomPort(false))
+)
+
+func getRandomPort(exit bool) int {
```
rolsonquadras: I see we are recursing here. Can [this](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/transport/http/support_test.go#L107) be used here to generate the dynamic port number ?

soluchok: Do you want me to make getServerPort public and use it instead of `listener.Addr().(*net.TCPAddr).Port`?

rolsonquadras: Now that I think about it again, I see this is different than the link from first post. Looks like we are creating a port and closing it, assuming it won't be taken by other processes before we start the server in main_test.go. There would be still some gap, but surely better than having static port.

troyronda: I assume this could have been structured as a loop with a number of attempts counter; rather than recursion.

cmd/aries-agentd/main_test.go
```diff
@@ -19,8 +20,25 @@ import (
...
+	const network = "tcp"
+	addr, _ := net.ResolveTCPAddr(network, "localhost:0")
+	listener, err := net.ListenTCP(network, addr)
+	if err != nil && !exit {
+		return getRandomPort(true)
+	}
+	if listener == nil {
```
troyronda: Should be checking err not listener

cmd/aries-agentd/main_test.go
```diff
@@ -19,8 +20,28 @@ import (
...
+func getRandomPort(n int) int {
+	const network = "tcp"
+	addr, err := net.ResolveTCPAddr(network, "localhost:0")
+	if err != nil {
+		panic(err)
+	}
+	listener, err := net.ListenTCP(network, addr)
+	if err != nil {
+		if n > 0 {
+			return getRandomPort(n - 1)
```
troyronda: I still prefer a loop to recursion :).

PR #308, opened 2019-09-20
ci: switch to GitHub Actions

troyronda:
Closes: #307

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,19 +3,19 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-name: Go
+name: build
 on: [push, pull_request]
 jobs:
-
   unitTest:
-    name: Unit Test
+    name: Unit test
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
     strategy:
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-18.04, macOS-10.14]
     steps:
 
-    - name: Set up Go 1.13
+    - name: Setup Go 1.13
       uses: actions/setup-go@v1
       with:
         go-version: 1.13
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -3,29 +3,32 @@
...
 
     - uses: actions/checkout@v1
 
-    - name: Unit Test
-      timeout-minutes: 5
+    - name: Run unit test
       run: make unit-test
+    
+    - name: Upload coverage to Codecov  
+      uses: codecov/codecov-action@v1.0.2
```
soluchok: Add `timeout-minutes: 5` per job
and move it to `Run checks` job bc here we have matrix strategy and it (Codecov) doesn't work for mac

.github/workflows/build.yml
```diff
@@ -3,29 +3,32 @@
...
       uses: actions/setup-go@v1
       with:
         go-version: 1.13
       id: go
 
     - uses: actions/checkout@v1
 
-    - name: Unit Test
-      timeout-minutes: 5
+    - name: Run unit test
```
soluchok: Add `timeout-minutes: 5` as well

troyronda: timeout goes with the job not the step AFAIK.
https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

(this was the change I made).

.github/workflows/build.yml
troyronda: codecov uses the result of unit test run. I had to duplicate the unit test section.

troyronda: (I left the matrix so the OS is filled into the check name).

.github/workflows/build.yml
troyronda: The documentation is unclear on what the default step timeout is...
https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes

.github/workflows/build.yml
troyronda: replaced with a conditional.

.github/workflows/build.yml
```diff
@@ -3,54 +3,64 @@
...
 jobs:
-
   unitTest:
-    name: Unit Test
+    name: Unit test
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
```
llorllale: Suggest we pin the OS versions

.github/workflows/build.yml
troyronda: We will need to revisit the timeouts. I just copied 10mins into each long step and job.

.github/workflows/build.yml
```diff
@@ -3,54 +3,64 @@
...
 
     - uses: actions/checkout@v1
 
-    - name: Unit Test
-      timeout-minutes: 5
+    - name: Run unit test
+      timeout-minutes: 10
       run: make unit-test
 
+    # TODO: need secret for Codecov from Actions
```
llorllale: Doesn't look like we have the necessary option under this project's Settings tab: https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables

.github/workflows/build.yml
troyronda: done

.github/workflows/build.yml
troyronda: Yup - that's why I left CircleCI unit test step enabled.

PR #309, opened 2019-09-20
feat: Add unpack method to inbound handler

### LINKED ISSUE #301: Add unpack method to inbound handler
- Add unpack method to inbound handler

fqutishat:
- Add unpack in inbound handler

closes #301

### DIFF
--- a/pkg/didcomm/transport/http/inbound.go
+++ b/pkg/didcomm/transport/http/inbound.go
@@ -13,6 +13,8 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	"github.com/hyperledger/aries-framework-go/pkg/wallet"
+
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
 )
@@ -22,6 +24,7 @@ var logger = log.New("aries-framework/transport")
 // provider contains dependencies for the HTTP Handler creation and is typically created by using aries.Context()
 type provider interface {
 	InboundMessageHandler() transport.InboundMessageHandler
+	PackWallet() wallet.Pack
 }
 
 // NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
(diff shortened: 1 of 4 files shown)


pkg/didcomm/transport/transport_interface.go
```diff
@@ -23,6 +25,7 @@ type InboundMessageHandler func(payload []byte) error
 // It is typically created by using aries.Context().
 type InboundProvider interface {
 	InboundMessageHandler() InboundMessageHandler
+	PackWallet() wallet.Pack
```
rolsonquadras: Just wondering if we can name this `Pack()` instead of `PackWallet()`. From DIDComm perspective Pack/Unpack terms are well defined and gives context by itself.

@fqutishat @troyronda @llorllale thoughts ?

Baha-sk: Since it's a Provider interface... maybe rename to `Packager()`?

pkg/didcomm/transport/http/inbound.go
```diff
@@ -19,29 +19,24 @@ import (
...
-}
-
 // NewInboundHandler will create a new handler to enforce Did-Comm HTTP transport specs
 // then routes processing to the mandatory 'msgHandler' argument.
 //
 // Arguments:
 // * 'msgHandler' is the handler function that will be executed with the inbound request payload.
 //    Users of this library must manage the handling of all inbound payloads in this function.
-func NewInboundHandler(prov provider) (http.Handler, error) {
+func NewInboundHandler(prov transport.InboundProvider) (http.Handler, error) {
```
troyronda: Why this change? Note: if you need another provider in the future, you will need to break your API signature

pkg/didcomm/transport/transport_interface.go
troyronda: I’m not a fan of PackWallet either.

pkg/didcomm/transport/http/inbound.go
fqutishat: done

Issue #173, opened 2019-08-28
JSON Web Token proof of Verifiable Credential


kdimak:
Support [JSON Web Token](https://w3c.github.io/vc-data-model/#json-web-token) proof of Verifiable Credential.

troyronda:
> In the DID Auth WG we reconfirmed that ES256K, Ed25519 and RS256 are the algorithms we want to support.

troyronda:
Also see https://github.com/square/go-jose

PR #312, opened 2019-09-20
feat: decoding of Verifiable Credential proved by JWT signature

kdimak:

#173

Implement decoding of Credential verified by JWT signature.
https://w3c.github.io/vc-data-model/#jwt-decoding

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -13,10 +13,14 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/square/go-jose/v3"
+	"github.com/square/go-jose/v3/jwt"
 	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 )
+
+//go:generate testdata/scripts/openssl_env.sh testdata/scripts/generate_test_keys.sh
 
 var logger = log.New("aries-framework/doc/verifiable")
 
@@ -177,7 +181,7 @@ type Credential struct {
 	Context        []string
 	ID             string
 	Type           []string
-	Subject        *Subject
+	Subject        Subject
 	Issuer         Issuer
 	Issued         *time.Time
 	Expired        *time.Time
@@ -187,12 +191,39 @@ type Credential struct {
 	RefreshService *RefreshService
 }
 
+// JWTAlgorithm defines JWT signature algorithms of Verifiable Credential
+type JWTAlgorithm int
+
+const (
+	// ES256K JWT Algorithm
+	ES256K JWTAlgorithm = iota
+	// RS256 JWT Algorithm
+	RS256
+	// EdDSA JWT Algorithm
+	EdDSA
+)
+
+// Jose converts JWTAlgorithm to JOSE one.
+func (ja JWTAlgorithm) Jose() jose.SignatureAlgorithm {
+	switch ja {
+	case ES256K:
+		return jose.ES256
+	case RS256:
+		return jose.RS256
+	case EdDSA:
+		return jose.EdDSA
+	default:
+		logger.Errorf("Unsupported algorithm: %v, fallback to RS256", ja)
+		return jose.ES256
+	}
+}
+
 // rawCredential is a basic verifiable credential
 type rawCredential struct {
 	Context        []string          `json:"@context,omitempty"`
 	ID             string            `json:"id,omitempty"`
 	Type           []string          `json:"type,omitempty"`
-	Subject        *Subject          `json:"credentialSubject,omitempty"`
+	Subject        Subject           `json:"credentialSubject,omitempty"`
 	Issued         *time.Time        `json:"issuanceDate,omitempty"`
 	Expired        *time.Time        `json:"expirationDate,omitempty"`
 	Proof          *Proof            `json:"proof,omitempty"`
@@ -222,13 +253,19 @@ type CredentialDecoder func(dataJSON []byte, credential *Credential) error
 // CredentialTemplate defines a factory method to create new Credential template.
 type CredentialTemplate func() *Credential
 
+// PublicKeyFetcher fetches public key for JWT signing verification based on Issuer ID (possibly DID)
+// and Key ID.
+// If not defined, JWT encoding is not tested.
+type PublicKeyFetcher func(issuerID, keyID string) (interface{}, error)
+
 // credentialOpts holds options for the Verifiable Credential decoding
 // it has a http.Client instance initialized with default parameters
 type credentialOpts struct {
-	schemaDownloadClient *http.Client
-	disabledCustomSchema bool
-	decoders             []CredentialDecoder
-	template             CredentialTemplate
+	schemaDownloadClient   *http.Client
+	disabledCustomSchema   bool
+	decoders               []CredentialDecoder
+	template               CredentialTemplate
+	issuerPublicKeyFetcher PublicKeyFetcher
 }
 
 // CredentialOpt is the Verifiable Credential decoding option
@@ -264,6 +301,13 @@ func WithTemplate(template CredentialTemplate) CredentialOpt {
 	}
 }
 
+// WithJWTPublicKeyFetcher defines a fetcher of public key required for verification of JWT signature
+func WithJWTPublicKeyFetcher(fetcher PublicKeyFetcher) CredentialOpt {
+	return func(opts *credentialOpts) {
+		opts.issuerPublicKeyFetcher = fetcher
+	}
+}
+
 func decodeIssuer(data []byte, credential *Credential) error {
 	issuerID, issuerName, err := issuerFromBytes(data)
 	if err != nil {
@@ -276,20 +320,38 @@ func decodeIssuer(data []byte, credential *Credential) error {
 
 // NewCredential creates an instance of Verifiable Credential by reading a JSON document from bytes.
 // It also applies miscellaneous options like custom decoders or settings of schema validation.
-func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error) {
+func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, error) {
 	// Apply options
 	crOpts := defaultCredentialOpts()
 	for _, opt := range opts {
 		opt(crOpts)
 	}
 
-	raw := &rawCredential{}
-	err := json.Unmarshal(dataJSON, &raw)
-	if err != nil {
-		return nil, fmt.Errorf("JSON unmarshalling of verifiable credential failed: %w", err)
+	var err error
+	var raw *rawCredential
+	var vcDataDecoded []byte
+	rawDecoding := true
+
+	if crOpts.issuerPublicKeyFetcher != nil {
+		var vcDataFromJwt []byte
+		if vcDataFromJwt, raw, err = decodeJWT(vcData, crOpts.issuerPublicKeyFetcher); err != nil {
+			return nil, fmt.Errorf("JWT decoding failed: %w", err)
+		}
+
+		rawDecoding = false
+		vcDataDecoded = vcDataFromJwt
 	}
 
-	err = validate(dataJSON, raw.Schema, crOpts)
+	if rawDecoding {
+		raw = &rawCredential{}
+		err = json.Unmarshal(vcData, raw)
+		if err != nil {
+			return nil, fmt.Errorf("JSON unmarshalling of verifiable credential failed: %w", err)
+		}
+		vcDataDecoded = vcData
+	}
+
+	err = validate(vcDataDecoded, raw.Schema, crOpts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -316,12 +381,133 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
 	return cred, nil
 }
 
+// CredentialClaims is JWT Claims extension by Credential.
+type CredentialClaims struct {
+	*jwt.Claims
+
+	Credential *rawCredential `json:"vc,omitempty"`
```
llorllale: This is leaking the un-exported `rawCredential` type

pkg/doc/verifiable/credential.go
```diff
@@ -222,13 +253,20 @@ type CredentialDecoder func(dataJSON []byte, credential *Credential) error
 // CredentialTemplate defines a factory method to create new Credential template.
 type CredentialTemplate func() *Credential
 
+// PublicKeyFetcher fetches public key for JWT signing verification based on Issuer ID (possibly DID)
+// and Key ID.
+// If not defined, JWT encoding is not tested.
```
llorllale: It looks like we're always fetching the key if `verifyJwtSignature` is `true`. Maybe remove this line?

pkg/doc/verifiable/credential.go
```diff
@@ -316,12 +381,133 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+		return nil, nil, fmt.Errorf("VC is not JWS: %w", err)
+	}
+
+	var jwtPayload CredentialClaims
+	err = parsedJwt.UnsafeClaimsWithoutVerification(&jwtPayload)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse JWT claims: %w", err)
+	}
+
+	if toVerifyJwtSignature {
```
llorllale: Why would we **not** want to verify the signature?

pkg/doc/verifiable/credential.go
```diff
@@ -316,12 +381,133 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+}
+
+func decodeJws(rawJwt []byte, toVerifyJwtSignature bool, fetcher PublicKeyFetcher) ([]byte, *rawCredential, error) {
+	parsedJwt, err := jwt.ParseSigned(string(rawJwt))
+	if err != nil {
+		return nil, nil, fmt.Errorf("VC is not JWS: %w", err)
+	}
+
+	var jwtPayload CredentialClaims
+	err = parsedJwt.UnsafeClaimsWithoutVerification(&jwtPayload)
```
llorllale: This seems fishy to me (contingent on your answer to my [question below](https://github.com/hyperledger/aries-framework-go/pull/312/files#r327225969)).

pkg/doc/verifiable/credential.go
```diff
@@ -316,12 +381,133 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
 
+// CredentialClaims is JWT Claims extension by Credential.
+type CredentialClaims struct {
+	*jwt.Claims
+
+	Credential *rawCredential `json:"vc,omitempty"`
+}
+
+// RawClaims is used to get raw content of "vc" claim of JWT.
+type RawClaims struct {
```
llorllale: I think this doesn't need to be exported

pkg/doc/verifiable/credential.go
```diff
@@ -276,20 +328,33 @@ func decodeIssuer(data []byte, credential *Credential) error {
...
-	if err != nil {
-		return nil, fmt.Errorf("JSON unmarshalling of verifiable credential failed: %w", err)
+	var raw *rawCredential
+	var err error
+	if crOpts.issuerPublicKeyFetcher != nil {
+		var vcDataFromJwt []byte
+		if vcDataFromJwt, raw, err = decodeJws(vcData, crOpts.verifyJwtSignature, crOpts.issuerPublicKeyFetcher); err != nil {
+			logger.Infof("cannot decode JWS: %w", err)
+		} else {
+			vcData = vcDataFromJwt
```
llorllale: As a best practice, it's best not to reassign input parameters.

pkg/doc/verifiable/credential.go
kdimak: The reason is to not to do it twice when using VC extension. We first encode into the base data model (with signature verification) and check what's the extension is it (e.g. by `@context` and `type`) and then make another encoding but into the concrete data model extension. At the latter stage, we do not need to verify the signature once again.

pkg/doc/verifiable/credential.go
kdimak: We deserialize JWT claims without verification in order to get `issuer` claim. It could contain a DID of the issuer which should direct us where to look for the public key needed to make signature verification. We can also use `kid` JOSE header to pick appropriate public key if there are several available in the resolved DID Document.
Look at how it's implemented at **decentralized-identity/interop-project**:
https://github.com/decentralized-identity/interop-project/blob/1e64dcbd3470c703c063a14465c3ac8d2a4585ba/cli-demo/src/index.js#L59

pkg/doc/verifiable/credential.go
kdimak: Agree! My negligence...

pkg/doc/verifiable/credential.go
kdimak: Got it! I will change it, thanks!

pkg/doc/verifiable/credential.go
```diff
@@ -187,12 +191,39 @@ type Credential struct {
 	RefreshService *RefreshService
 }
 
+// JwtAlgorithm JWT signature algorithms of Verifiable Credential
+type JwtAlgorithm int
```
troyronda: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/doc/verifiable/credential.go
kdimak: @llorllale @troyronda After talking to Georgy, we've decided for simplification to always verify the signature.

pkg/doc/verifiable/credential.go
```diff
@@ -276,20 +320,38 @@ func decodeIssuer(data []byte, credential *Credential) error {
...
-		return nil, fmt.Errorf("JSON unmarshalling of verifiable credential failed: %w", err)
+	var err error
+	var raw *rawCredential
+	var vcDataDecoded []byte
+	rawDecoding := true
+
+	if crOpts.issuerPublicKeyFetcher != nil {
+		var vcDataFromJwt []byte
+		if vcDataFromJwt, raw, err = decodeJws(vcData, crOpts.issuerPublicKeyFetcher); err != nil {
+			logger.Infof("cannot decode JWS: %w", err)
```
llorllale: I think we should simply abort and return the error at this point. One of the possible errors of `decodeJws()` results from validation of the signature. If that fails, we wouldn't want to continue.

pkg/doc/verifiable/credential.go
```diff
@@ -421,9 +603,80 @@ func loadCredentialSchema(url string, client *http.Client) ([]byte, error) {
...
+		IssuedAt:  jwt.NewNumericDate(*vc.Issued), // iat (not in spec, follow the interop project approach)
+	}
+	if vc.Expired != nil {
+		jwtClaims.Expiry = jwt.NewNumericDate(*vc.Expired) // exp
+	}
+
+	var raw *rawCredential
+	if minimizeVc {
+		vcCopy := *vc
+		vcCopy.Expired = nil
```
llorllale: This claim is missing from `jwtClaims` above

pkg/doc/verifiable/credential.go
```diff
@@ -421,9 +603,80 @@ func loadCredentialSchema(url string, client *http.Client) ([]byte, error) {
 	return gotBody, nil
 }
 
-// JSONBytes converts Verifiable Credential to JSON bytes
-func (vc *Credential) JSONBytes() ([]byte, error) {
-	rawCred := &rawCredential{
+// JWT serializes Credential to signed JWT.
+func (vc *Credential) JWT(
+	signatureAlg JWTAlgorithm,
+	privateKey interface{},
```
llorllale: TODO - need to refactor this. Private keys will never leave the wallet, therefore it cannot be passed in as a parameter to anything outside of the wallet itself.

pkg/doc/verifiable/credential.go
```diff
@@ -316,6 +378,126 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+}
+
+func decodeJws(rawJwt []byte, fetcher PublicKeyFetcher) ([]byte, *rawCredential, error) {
+	parsedJwt, err := jwt.ParseSigned(string(rawJwt))
+	if err != nil {
+		return nil, nil, fmt.Errorf("VC is not JWS: %w", err)
+	}
+
+	var credClaims credentialClaims
+	err = parsedJwt.UnsafeClaimsWithoutVerification(&credClaims)
```
llorllale: Now that we agreed to always verify the signature, I don't think we should be using any of these "UnsafeXXX" functions (unless I'm mistaken)

kdimak: That's because of an indirect algorithm: we need to read **Issuer** JWT claim first, find the proper public key (by **Issuer (DID)** and **Key ID**) and only then verify the signature.

pkg/doc/verifiable/credential.go
kdimak: OK, I need more input about this, let's discuss this tomorrow!

pkg/doc/verifiable/credential.go
kdimak: It's set between, as `Expired` field of Verifiable Credential is optional and could be nil. See lines 621-623.

pkg/doc/verifiable/credential.go
kdimak: Also thought about it, but considered it as use case when e.g. a Holder can get Verifiable Credential with proof embedded into Verifiable Credential (**proof** property). From the other hand, we don't support embedded proofs for now.

@llorllale do you think we should abort and return the error at this point, right?

pkg/doc/verifiable/credential.go
llorllale: Ok, got it.

This is kinda funky: we shouldn't be touching these claims at all without verification. But, on the other hand, including the sender's identifier in the `kid` will reveal it to eavesdroppers... not sure what can be done about it.

pkg/doc/verifiable/credential.go
llorllale: Don't worry about it for now - create an issue and add a TODO here that references it.

pkg/doc/verifiable/credential.go
llorllale: I think we should, yes.

There are plenty of other possible error conditions in `decodeJws()` - I don't see why we need to distinguish between them on this line and attempt to decide whether it's worth continuing or not. Aborting is the simplest path forward.

pkg/doc/verifiable/credential.go
kdimak: #339

pkg/doc/verifiable/credential.go
kdimak: Deal!

pkg/doc/verifiable/credential.go
```diff
@@ -316,6 +378,126 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+		iatTime := iat.Time().UTC()
+		raw.Issued = &iatTime
+	}
+
+	if exp := credClaims.Expiry; exp != nil {
+		expTime := exp.Time().UTC()
+		raw.Expired = &expTime
+	}
+}
+
```
troyronda: JWT

pkg/doc/verifiable/credential.go
```diff
@@ -316,6 +378,126 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+	// Issuer of Verifiable Credential could be either string (id) or struct (with "id" field).
+	switch issuer := raw.Issuer.(type) {
+	case string:
+		raw.Issuer = iss
+	case map[string]interface{}:
+		issuer["id"] = iss
+	}
+}
+
+func mergeRefinedVcIntoRawVcData(rawVcClaim map[string]interface{}, raw *rawCredential) {
```
troyronda: VC

pkg/doc/verifiable/credential.go
kdimak: Fixed!

pkg/doc/verifiable/credential.go
kdimak: Fixed!

pkg/doc/verifiable/credential.go
```diff
@@ -316,6 +378,128 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+	case map[string]interface{}:
+		issuer["id"] = iss
+	}
+}
+
+func (rawClaims *rawCredentialClaims) mergeRefinedVC(raw *rawCredential) {
+	rawVCClaims := rawClaims.Raw
+
+	// these operations are safe
+	rawData, _ := json.Marshal(raw) // nolint:errcheck
```
rolsonquadras: @troyronda @llorllale @fqutishat are we going to ignore marshalling errors ? We had this discussion during daily Aries call today.

troyronda: I'm not understanding why marshal errors are being ignored.

pkg/doc/verifiable/credential.go
```diff
@@ -187,12 +191,39 @@ type Credential struct {
...
+	RS256
+	// EdDSA JWT Algorithm
+	EdDSA
+)
+
+// Jose converts JWTAlgorithm to JOSE one.
+func (ja JWTAlgorithm) Jose() jose.SignatureAlgorithm {
+	switch ja {
+	case ES256K:
+		return jose.ES256
```
Baha-sk: to be verified, I don't think `jose.ES256` is ES256k

pkg/doc/verifiable/credential.go
Baha-sk: agree

pkg/doc/verifiable/credential.go
```diff
@@ -316,6 +378,128 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+
+func (rawClaims *rawCredentialClaims) mergeRefinedVC(raw *rawCredential) {
+	rawVCClaims := rawClaims.Raw
+
+	// these operations are safe
+	rawData, _ := json.Marshal(raw) // nolint:errcheck
+
+	var rawMap map[string]interface{}
+
+	_ = json.Unmarshal(rawData, &rawMap) // nolint:errcheck
```
Baha-sk: idem

pkg/doc/verifiable/credential.go
```diff
@@ -316,6 +378,128 @@ func NewCredential(dataJSON []byte, opts ...CredentialOpt) (*Credential, error)
...
+func (credClaims *credentialClaims) refineVCFromJWTClaims() {
+	raw := credClaims.Credential
+
+	if iss := credClaims.Issuer; iss != "" {
+		refineVCIssuerFromJWTClaims(raw, iss)
+	}
+
+	if nbf := credClaims.NotBefore; nbf != nil {
+		nbfTime := nbf.Time().UTC()
+		raw.Issued = &nbfTime
```
Baha-sk: is this intentional? `raw.Issued` instead of `raw.NotBefore`

pkg/doc/verifiable/credential.go
```diff
@@ -421,9 +605,75 @@ func loadCredentialSchema(url string, client *http.Client) ([]byte, error) {
...
+
+	// create an instance of Builder that uses the rsa signer
+	builder := jwt.Signed(rsaSigner)
+
+	builder = builder.Claims(credClaims)
+
+	// validate all ok, sign with the RSA key, and return a compact JWT
+	rawJWT, err := builder.CompactSerialize()
+	if err != nil {
+		return "", fmt.Errorf("failed to sign JWT")
```
Baha-sk: to be discussed, wrap err or not here ..

pkg/doc/verifiable/credential_jwt_proof_test.go
```diff
@@ -0,0 +1,196 @@
...
+			WithJWTPublicKeyFetcher(func(issuerID, keyID string) (interface{}, error) {
+				publicKey, err := readPublicKey(filepath.Join(certPrefix, "holder_public.pem"))
+				require.NoError(t, err)
+				require.NotNil(t, publicKey)
+
+				return publicKey, nil
+			}))
+
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "JWT decoding failed")
```
Baha-sk: use `require.EqualError(t, e, ...)` instead

pkg/doc/verifiable/credential.go
Moopli: The takeaway was to forgive the drop to test coverage rather than ignore the errors altogether (or panic), and we aim for at least 85% to give ourselves a margin for cases like this. If some error should never happen, you don't want that error to be silent if the impossible ever happens because then you might be processing spurious data.

kdimak: I fixed that in #382

pkg/doc/verifiable/credential.go
kdimak: Yes, true! Filed #381

pkg/doc/verifiable/credential.go
kdimak: Yes, as `rawCredential.Issued` is written to both `iat` and `nbf` headers.

pkg/doc/verifiable/credential.go
kdimak: Fixed it by error wrapping at #382

pkg/doc/verifiable/credential_jwt_proof_test.go
kdimak: This is intentional - I'm checking only a part of the error message here...

Issue #325, opened 2019-09-24
RFC 0028: Define models for introduce protocol


soluchok:
Define models for the [introduce protocol ](https://github.com/hyperledger/aries-rfcs/tree/master/features/0028-introduce) according to the documentation

PR #328, opened 2019-09-24
protocol: Added models for the introduce protocol

soluchok:
* models for the protocol

Issue: https://github.com/hyperledger/aries-framework-go/issues/325

### DIFF
--- a/pkg/didcomm/common/model/ack.go
+++ b/pkg/didcomm/common/model/ack.go
@@ -1,0 +1,17 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package model
+
+import "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+
+// Ack acknowledgement struct
+type Ack struct {
+	Type   string            `json:"@type,omitempty"`
+	ID     string            `json:"@id,omitempty"`
+	Status string            `json:"status,omitempty"`
+	Thread *decorator.Thread `json:"~thread,omitempty"`
+}
--- a/pkg/didcomm/protocol/introduce/models.go
+++ b/pkg/didcomm/protocol/introduce/models.go
@@ -1,0 +1,99 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package introduce
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
+)
+
+// Proposal defines proposal request
+type Proposal struct {
+	Type   string            `json:"@type,omitempty"`
+	ID     string            `json:"@id,omitempty"`
+	To     To                `json:"to,omitempty"`
+	NWise  bool              `json:"nwise,omitempty"`
+	Timing *decorator.Timing `json:"~timing,omitempty"`
+}
+
+// To introducee descriptor keeps information about the introduction
+// e.g introducer wants to introduce Bot to introducee { "name": "Bob" }
+type To struct {
+	Name            string          `json:"name,omitempty"`
+	Description     string          `json:"description,omitempty"`
+	DescriptionL10N DescriptionL10N `json:"description~l10n,omitempty"`
+	Where           string          `json:"where,omitempty"`
+	ImgAttach       ImgAttach       `json:"img~attach,omitempty"`
+	Proposed        bool            `json:"proposed,omitempty"`
+}
+
+// DescriptionL10N may contain locale field and key->val pair for translation
+// e.g { "locale": "en", "es": "Donde se toma el MRI; no en el centro"},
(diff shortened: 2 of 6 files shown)


pkg/didcomm/protocol/introduce/models.go
```diff
@@ -0,0 +1,97 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package introduce
+
+import (
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
+
```
troyronda: extra newline

soluchok:
@llorllale please take a look

pkg/didcomm/protocol/introduce/models.go
```diff
@@ -0,0 +1,106 @@
...
+// Content keeps image data
+type Content struct {
+	Link      string `json:"link,omitempty"`
+	ByteCount int    `json:"byte_count,omitempty"`
+	Sha256    string `json:"sha256,omitempty"`
+}
+
+// Timing keeps expiration time for Proposal
+type Timing struct {
```
llorllale: Please move this to the `decorator` package.

pkg/didcomm/protocol/introduce/models.go
```diff
@@ -0,0 +1,106 @@
...
+	ID         string                  `json:"@id,omitempty"`
+	Thread     *decorator.Thread       `json:"~thread,omitempty"`
+	Approve    bool                    `json:"approve,omitempty"`
+	Invitation *didexchange.Invitation `json:"invitation,omitempty"`
+}
+
+// Ack confirming to an introducee which let them know that their invitation was forwarded
+// TODO: Ack was copy-pasted from didexchange clarify what status does.
+//       Do we need status for introduce protocol? What value it can be e.g "success", "error" ?
+type Ack struct {
```
llorllale: We should move this to something like a "common" package under the existing `didcomm` one. This same structure is used by the did-exchange protocol as well (and is reused in several other RFCs).

soluchok: I moved common model to  `pkg/didcomm/common/model`

pkg/didcomm/protocol/introduce/models.go
```diff
@@ -0,0 +1,99 @@
...
+type Content struct {
+	Link      string `json:"link,omitempty"`
+	ByteCount int    `json:"byte_count,omitempty"`
+	Sha256    string `json:"sha256,omitempty"`
+}
+
+// PleaseIntroduceTo includes all field from To structure
+// also it has Discovered the field which should be provided by help-me-discover protocol
+type PleaseIntroduceTo struct {
+	To
```
troyronda: missing a TODO?

pkg/didcomm/protocol/introduce/models.go
```diff
@@ -0,0 +1,99 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package introduce
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
```
troyronda: as an aside, is decorator a protocol or should it be under protocol/common?

@llorllale

soluchok: @troyronda  Do you mean to move `decorator` to pkg/didcomm/common/decorator ?

troyronda: that is the question i am posing.

pkg/didcomm/protocol/introduce/models.go
soluchok: I am going to add TODO, we need this field but it should be provided by help-me-discover protocol which is not implemented yet

pkg/didcomm/protocol/introduce/models.go
llorllale: A decorator is not a protocol. From the [decorator RFC](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0011-decorators/README.md#tutorial) (emphasis mine):

> **A decorator is an optional chunk of JSON that conveys metadata.** Decorators are not declared in a core schema but rather supplementary to it. Decorators add semantic content broadly relevant to messaging in general, and not so much tied to the problem domain of a specific type of interaction.

Decorators don't have any of the key [ingredients](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/README.md#ingredients) of protocols.

However, since they are [adopted](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/template.md#adopted-messages) by other protocols (and some, like `~thread`, are integral to every protocol), I don't have a problem placing them inside the `protocol` directory.

pkg/didcomm/common/model/ack.go
```diff
@@ -0,0 +1,17 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package model
```
llorllale: I dislike having a namespace named "model"; however, an `ack` is neither a protocol nor a decorator, so I'm not sure where else to put it. I opened hyperledger/aries-rfc#234 to see if I get an answer.

PR #330, opened 2019-09-24
feature: Add support for multiple proofs to DID Document

### LINKED ISSUE #324: Add support for multiple proofs to DID Document
Currently DID Document supports only one proof. Add support for multiple proofs to DID Document. Also, change signatureValue to proofValue.

sandrask:
Currently DID Document supports only one proof. Add support for multiple proofs to DID Document. Also, change signatureValue to proofValue.

Closes #324

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 package did
 
 import (
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"encoding/pem"
@@ -24,6 +25,12 @@ const (
 	jsonldID           = "id"
 	jsonldServicePoint = "serviceEndpoint"
 	jsonldController   = "controller"
+
+	jsonldCreator    = "creator"
+	jsonldCreated    = "created"
+	jsonldProofValue = "proofValue"
+	jsonldDomain     = "domain"
+	jsonldNonce      = "nonce"
 
 	// various public key encodings
 	jsonldPublicKeyBase58 = "publicKeyBase58"
@@ -77,16 +84,22 @@ const (
       }
     },
     "created": {
-      "type": "string",
-      "pattern": "\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\dZ"
+      "type": "string"
     },
     "updated": {
-      "type": "string",
-      "pattern": "\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\dZ"
+      "type": "string"
     },
     "proof": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/proof"
+      }
+    }
+  },
+  "definitions": {
+	"proof": {
       "type": "object",
-      "required": [ "type", "creator", "created", "signatureValue"],
+      "required": [ "type", "creator", "created", "proofValue"],
       "properties": {
         "type": {
           "type": "string",
@@ -97,10 +110,9 @@ const (
           "format": "uri-reference"
         },
         "created": {
-          "type": "string",
-          "pattern": "\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\dZ"
+          "type": "string"
         },
-        "signatureValue": {
+        "proofValue": {
           "type": "string"
         },
         "domain": {
@@ -109,10 +121,8 @@ const (
         "nonce": {
           "type": "string"
         }
-  }
-    }
-  },
-  "definitions": {
+	  }
+    },
     "publicKey": {
       "required": [
         "id",
@@ -169,7 +179,7 @@ type Doc struct {
 	Authentication []VerificationMethod
 	Created        *time.Time
 	Updated        *time.Time
-	Proof          *Proof
+	Proof          []Proof
 }
 
 // PublicKey DID doc public key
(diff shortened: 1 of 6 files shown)


pkg/doc/did/doc.go
```diff
@@ -201,17 +210,17 @@ type rawDoc struct {
 	Authentication []interface{}            `json:"authentication,omitempty"`
 	Created        *time.Time               `json:"created,omitempty"`
 	Updated        *time.Time               `json:"updated,omitempty"`
-	Proof          *Proof                   `json:"proof,omitempty"`
+	Proof          []interface{}            `json:"proof,omitempty"`
```
soluchok: I would rename it to Proofs

pkg/doc/did/doc.go
```diff
@@ -236,17 +245,50 @@ func FromBytes(data []byte) (*Doc, error) {
...
 		Authentication: authPKs,
 		Created:        raw.Created,
 		Updated:        raw.Updated,
-		Proof:          raw.Proof,
+		Proof:          proofs,
 	}, nil
 }
 
+func populateProofs(rawProofs []interface{}) ([]Proof, error) {
+	var proofs []Proof
```
soluchok: Please specify capacity  `proofs := make([]Proof,0,len(rawProofs))`

pkg/doc/did/doc.go
```diff
@@ -236,17 +245,50 @@ func FromBytes(data []byte) (*Doc, error) {
...
+		}
+
+		proofs = append(proofs, proof)
+	}
+
+	return proofs, nil
+}
+
 func populateServices(rawServices []map[string]interface{}) []Service {
 	var services []Service
```
soluchok: same, specify capacity

pkg/doc/did/doc.go
```diff
@@ -428,3 +470,25 @@ func populateRawAuthentications(vms []VerificationMethod) []interface{} {
...
+	var rawProofs []interface{}
+	var local Proof
+	for _, p := range proofs {
+		local = p
+		rawProofs = append(rawProofs, populateRawProof(&local))
+	}
+	return rawProofs
+}
+
+func populateRawProof(proof *Proof) map[string]interface{} {
```
soluchok: no need to use pointer `proof *Proof`  we have only read here

pkg/doc/did/doc.go
```diff
@@ -428,3 +470,25 @@ func populateRawAuthentications(vms []VerificationMethod) []interface{} {
 
 	return rawAuthentications
 }
+
+func populateRawProofs(proofs []Proof) []interface{} {
+	var rawProofs []interface{}
+	var local Proof
+	for _, p := range proofs {
+		local = p
```
soluchok: Why? remove line `var local Proof` and `local = p` and pass data to populateRawProof without a pointer, and set capacity to `var rawProofs []interface{}` -> rawProofs := make([]interface{},0,len(proofs))

pkg/doc/did/doc.go
sandrask: Per spec it is proof, just like authentication etc.

pkg/doc/did/doc.go
sandrask: Same as above, artificial code to avoid linter warning. I guess I can add nolint.

pkg/doc/did/doc.go
sandrask: Linter complained with error along the line: "struct is too big, pass by pointer" , that was the only reason I added it. I can revert the change and put nolint.

sandrask: Exact message: pkg/doc/did/doc.go:482:23: hugeParam: proof is heavy (88 bytes); consider passing it by pointer (gocritic)

pkg/doc/did/doc.go
sandrask: Done

pkg/doc/did/doc.go
sandrask: Done

pkg/doc/did/doc.go
sandrask: Added nolint to populateRawProof which resolved this issue too.

pkg/doc/did/doc.go
```diff
@@ -236,19 +245,52 @@ func FromBytes(data []byte) (*Doc, error) {
...
+		Proof:          proofs,
 	}, nil
 }
 
+func populateProofs(rawProofs []interface{}) ([]Proof, error) {
+	proofs := make([]Proof, 0, len(rawProofs))
+	for _, rawProof := range rawProofs {
+		emap, ok := rawProof.(map[string]interface{})
+		if !ok {
+			return nil, errors.New("rawProofs is not map[string]interface{}")
```
Baha-sk: no `s` in rawProof

pkg/doc/did/doc.go
```diff
@@ -201,17 +210,17 @@ type rawDoc struct {
...
-	Type           string     `json:"type,omitempty"`
-	Created        *time.Time `json:"created,omitempty"`
-	Creator        string     `json:"creator,omitempty"`
-	SignatureValue string     `json:"signatureValue,omitempty"`
-	Domain         string     `json:"domain,omitempty"`
-	Nonce          string     `json:"nonce,omitempty"`
+	Type       string     `json:"type,omitempty"`
+	Created    *time.Time `json:"created,omitempty"`
+	Creator    string     `json:"creator,omitempty"`
+	ProofValue string     `json:"proofValue,omitempty"`
```
troyronda: Should ProofValue and Nonce be []byte?
(Seems odd to see a string.)

pkg/doc/did/doc.go
```diff
@@ -424,3 +466,23 @@ func populateRawAuthentications(vms []VerificationMethod) []interface{} {
 
 	return rawAuthentications
 }
+
+func populateRawProofs(proofs []Proof) []interface{} {
```
soluchok: in that case, it will be better to create only one function 
*  we can avoid copying a big structure to the function stack (if arg without a pointer)
*  we can prevent moving it to the heap (if arg is a pointer)
```
func populateRawProofs(proofs []Proof) []interface{} {
	rawProofs := make([]interface{},0,len(proofs))
	for _, p := range proofs {
		rawProofs = append(rawProofs, map[string]interface{}{
			jsonldType : p.Type,
			jsonldCreated : p.Created,
			jsonldCreator : p.Creator,
			jsonldProofValue : p.ProofValue,
			jsonldDomain : p.Domain,
			jsonldNonce : p.Nonce,
		})
	}
	return rawProofs
}
```

sandrask: Done

pkg/doc/did/doc.go
sandrask: This was existing struct. I changed ProofValue and Nonce to []byte. Done.

Issue #294, opened 2019-09-19
Legacy Authcrypt Decrypter 


Moopli:
Following on from #139, create a decrypter that will decrypt authcrypt envelopes encrypted with ACA-Py and with the legacy encrypter.

Part of #36

PR #340, opened 2019-09-24
feat: Legacy Authcrypt Decrypter

Moopli:
Implements the legacy authcrypt decrypter, following on from #139, the encrypter.

Fixes: #294

### DIFF
--- a/pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
+++ b/pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
@@ -7,16 +7,158 @@ SPDX-License-Identifier: Apache-2.0
 package authcrypt
 
 import (
+	"bytes"
+	"crypto/ed25519"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
+	"fmt"
 
+	"github.com/btcsuite/btcutil/base58"
+	chacha "golang.org/x/crypto/chacha20poly1305"
 	"golang.org/x/crypto/nacl/box"
 )
 
 // Decrypt will decode the envelope using the legacy format
 // Using (X)Chacha20 encryption algorithm and Poly1035 authenticator
-func (c *Crypter) Decrypt(envelope []byte, recipientPrivKey *privateEd25519) ([]byte, error) {
-	// TODO implement legacy auth decrypt https://github.com/hyperledger/aries-framework-go/issues/294
-	return nil, errors.New("not implemented")
+func (c *Crypter) Decrypt(envelope []byte, recipientKey *keyPairEd25519) ([]byte, error) {
+	var envelopeData legacyEnvelope
+	err := json.Unmarshal(envelope, &envelopeData)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


Moopli:
Note: refactoring work is left for the next PR, including:
- Updating legacy authcrypter (encrypt and decrypt) to fit Crypter interface
- Refactoring shared code with JWE authcrypt
- Refactoring out the secret-using operations (for recipient headers) to be under wallet

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
```diff
@@ -7,16 +7,165 @@ SPDX-License-Identifier: Apache-2.0
...
+	if err != nil {
+		return nil, err
+	}
+
+	var protectedData protected
+	err = json.Unmarshal(protectedBytes, &protectedData)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: do we keep this algorithm sanity check? Only commented out right now to accommodate a bug in ACA-Py
```
Moopli: Welcoming input here - we want to interop with ACA-Py, but we also want to meet the spec and thus interop with other hypothetical aries agents that decide not to support the incoming JWE format.

troyronda: What is ACA-PY using here?

(yes - we want to be able to interop with ACA-Py).

Baha-sk: the aim here to support backward compatibility with ACA-Py implementation.. the (new) jwe implementation takes care of properly handling the format required by the spec

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
```diff
@@ -7,16 +7,165 @@ SPDX-License-Identifier: Apache-2.0
...
 import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
+	"fmt"
 
+	"github.com/btcsuite/btcutil/base58"
+	chacha "golang.org/x/crypto/chacha20poly1305"
+	"golang.org/x/crypto/ed25519"
```
Baha-sk: now that the linter supports Go 1.13, let's try to use the internal crypto package instead..

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
Moopli: ACA-Py labels their messages as using xchacha20, while the algorithm used is always chacha20. They'll be updating to the spec once it's finalized, so I won't be keeping this check.

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
Moopli: I've switched `crypto/curve25519` to import from the standard library instead of `golang.org/x/crypto`. The others I use have not been moved over yet.

Moopli:
h/o merging for a minute, gotta fix my history - rebased onto origin master instead of upstream master

Moopli:
(lastest update just rebasing while I wait for more reviews...)

PR #343, opened 2019-09-25
feature: Implement Ed25519Signature2018 signature suite

### LINKED ISSUE #248: Implement Ed25519VerificationKey2018 Linked Data Signature Suite
In #241 we create an interface for a generic pluggable Linked Data Signature Suite. Here we create an implementation of the [Ed25519 2018](https://w3c-dvcg.github.io/lds-ed25519-2018/) signature suite, so Sign and Verify can be done using `Ed25519VerificationKey2018`.

Part of #96

sandrask:
Closes #248


pkg/doc/signature/ed25519signature2018/suite_test.go
```diff
@@ -0,0 +1,110 @@
...
+	require.NoError(t, err)
+	require.NotEmpty(t, privKey)
+
+	ss := New()
+	signature, err := ss.Sign(privKey, []byte("test doc"))
+	require.NoError(t, err)
+	require.NotEmpty(t, signature)
+}
+
+func TestSignatureSuite_Verify(t *testing.T) {
```
Moopli: Do you have any example documents with signatures that you can use to test Sign & Verify?
If there is test data elsewhere (provided by the spec? In the tests of reference implementations? Generated by reference implementations?) then you can use such data to test that Sign produces the right signature, and to test that Verify will accept data with correct signatures and reject data with incorrect signatures.

Moopli: You can also test that canonicalization will produce the same result even if you change whitespace, element order etc in a document.

sandrask: We are using standard go library "crypto/ed25519" to produce signature and verify. I don't think that we need additional unit tests to verify this.

sandrask: Canonicalization will aways produce the same result (same order of elements) and my unit test already tests that. I chose Test 28 from https://json-ld.org/test-suite/reports/#test_30bc80ba056257df8a196e8f65c097fc. 

json-gold (library used) passed all 119 tests

PR #360, opened 2019-09-27
fix: Prevent panic when writing to a closed channel

soluchok:
* added recovery

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service_test.go
+++ b/pkg/didcomm/protocol/didexchange/service_test.go
@@ -107,6 +107,30 @@ func TestService_Name(t *testing.T) {
 	require.Equal(t, DIDExchange, prov.Name())
 }
 
+func TestService_Handle_Panic(t *testing.T) {
+	dbstore, cleanup := store(t)
+	defer cleanup()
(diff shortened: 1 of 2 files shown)


pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -107,6 +107,30 @@ func TestService_Name(t *testing.T) {
 	require.Equal(t, DIDExchange, prov.Name())
 }
 
+func TestService_Handle_Panic(t *testing.T) {
+	dbstore, cleanup := store(t)
+	defer cleanup()
+
+	s := New(dbstore, &mockdid.MockDIDCreator{Doc: getMockDID()}, &mockProvider{})
+	actionCh := make(chan dispatcher.DIDCommAction)
+	close(actionCh)
```
troyronda: This doesn’t seem like a valid scenario. The client should call unregister first. If they don’t gracefully close things then panic will occur.

soluchok: ok, if it is expected behavior I do not mind

PR #362, opened 2019-09-27
bug: Fix TestServiceEvents client test

sandrask:
We have to feed proper did documents into service (has to have at least one public key)

### DIFF
--- a/pkg/client/didexchange/client_test.go
+++ b/pkg/client/didexchange/client_test.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package didexchange
 
 import (
+	"crypto/ed25519"
+	"crypto/rand"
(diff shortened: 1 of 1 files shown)


pkg/client/didexchange/client_test.go
```diff
@@ -221,5 +223,30 @@ type mockDIDCreator struct {
 }
 
 func (m *mockDIDCreator) CreateDID(opts ...wallet.DocOpts) (*did.Doc, error) {
-	return &did.Doc{}, nil
+	const didContext = "https://w3id.org/did/v1"
```
sudeshrshetty: May be for next patch, use this
https://github.com/hyperledger/aries-framework-go/blob/master/pkg/internal/mock/common/did/mock_creator.go

Issue #336, opened 2019-09-24
introduce protocol: fix some state transitions


llorllale:
The outcome of #288 were several fixes to the `introduce` protocol's RFC - these need to be reflected in our codebase:

* Introducer: disallowed start->done transition. From start they can transition either to the delivering state or the arranging state
* Introducer: now supports sending an ack msg to each introducee to confirm that their invitations were forwarded
* Introducee: disallowed start->invited. Instead they can transition start->deciding

PR #373, opened 2019-09-30
protocol: Change state transitions for introduce protocol

soluchok:
* added transitions to delivering and arranging

Issue: https://github.com/hyperledger/aries-framework-go/issues/336


soluchok:
@llorllale I have not seen `invited` state, could you please explain what does this comment mean?
`Introducee: disallowed start->invited. Instead they can transition start->deciding`

If we missed `invited` state probably we need to add it, let me know.

llorllale:
@soluchok the RFC previously included the transition `start`->`invited`. That was an error and was fixed.

troyronda:
@soluchok Should use semantic prefix in PR and commit titles.

PR #380, opened 2019-09-30
feat: handleInboundInvitation creating and passing the thread id

### LINKED ISSUE #378: handleInboundInvitation is creating new id but never persisting and state is changed to noop instead of requested
This is bug found in the integration test

talwinder50:
closes #378

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -99,26 +99,33 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 	aEvent := s.actionEvent
 	s.lock.RUnlock()
 
+	logger.Infof("entered into Handle exchange message : %s", msg.Payload)
+
 	if !msg.Outbound && aEvent == nil {
 		return errors.New("no clients are registered to handle the message")
 	}
 
-	thid, err := threadID(msg.Payload)
+	thid, err := threadID(msg)
 	if err != nil {
 		return err
 	}
+	logger.Infof("thread id value for the did exchange msg : %s", thid)
+
 	current, err := s.currentState(thid)
 	if err != nil {
 		return err
 	}
+	logger.Infof("current state : %s", current.Name())
+
 	next, err := stateFromMsgType(msg.Type)
 	if err != nil {
 		return err
 	}
+	logger.Infof("state will transition to -> %s if the msgType is processed", next.Name())
+
 	if !current.CanTransitionTo(next) {
 		return fmt.Errorf("invalid state transition: %s -> %s", current.Name(), next.Name())
 	}
-
 	// trigger message events
 	logger.Infof("send pre event for state %s", next.Name())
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -74,6 +74,7 @@ type Service struct {
 type context struct {
 	outboundDispatcher dispatcher.Outbound
 	didCreator         did.Creator
+	thid               string
```
troyronda: this does not belong here.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -276,7 +276,7 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation) error {
 	//  using this request's ID. issue-280
 	request := &Request{
 		Type:  ConnectionRequest,
-		ID:    uuid.New().String(),
+		ID:    ctx.thid,
```
fqutishat: You need to pass thid as argument and remove thid from context

pkg/didcomm/protocol/didexchange/service.go
talwinder50: we basically need to find calculate threadID/ID out of state machine handle and the pass it as a parameter to state machine for persistence purposes. Context is internal struct. Other place is to add in DIDCommMsg struct, but that doesn't seem ok. 

Any suggestions ?

pkg/didcomm/protocol/didexchange/states.go
talwinder50: ok i will do that way

pkg/didcomm/protocol/didexchange/states_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/google/uuid"
+
```
troyronda: extra newline

talwinder50: removed

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -39,7 +39,7 @@ type state interface {
 	CanTransitionTo(next state) bool
 	// Executes this state, returning a followup state to be immediately executed as well.
 	// The 'noOp' state should be returned if the state has no followup.
-	Execute(msg dispatcher.DIDCommMsg, ctx context) (followup state, err error)
+	Execute(msg dispatcher.DIDCommMsg, thid string, ctx context) (followup state, err error)
```
troyronda: @rolsonquadras @fqutishat @talwinder50 @llorllale I'm wondering if thid could make sense in DIDCommMsg?

(I see these arguments paired together in the PR).

rolsonquadras: It would surely be a good candidate for ID inside DIDCommMsg. The one problem i see is that it is not consistent in different DIDComm messages (DIDCommMsg is constructed at inbound transport layer). `~thread` decorator is not present in exchangeRequest. The DIDExchange Service derives from `id` attribute. Apart from this, I believe all DIDComm message types will have `~thread` decorator.

Issue #381, opened 2019-10-01
Support JSON Web Token (JWT) with ES256K (secp256k1) signature


kdimak:
Currently, the following curves are supported in [go-jose](https://github.com/square/go-jose):
- P-256
- P-384
- P-521
We need to add support of P-521-K elliptic curve which is used at [VC Test Suite](https://github.com/w3c/vc-test-suite) and is mentioned among required JWT signatures at [Interoperability Project](https://github.com/decentralized-identity/interoperability) (see [comment](https://github.com/decentralized-identity/interoperability/issues/5#issuecomment-516794494))

kdimak:
`go-jose` looks for the secp256k1 curve impl library other than `btcsuite/btcd` (https://github.com/square/go-jose/pull/278#issuecomment-555338996).

Baha-sk:
> `go-jose` looks for the secp256k1 curve impl library other than `btcsuite/btcd` ([square/go-jose#278 (comment)](https://github.com/square/go-jose/pull/278#issuecomment-555338996)).

I think they're looking to move this change in a separate package.. I don't know how though since the PR is changing the core of Go Jose..

Issue #372, opened 2019-09-29
Generation of JWT from Verifiable Credential without the signature


kdimak:
Encode Verifiable Credential into plain JWT (JSON document) without making JWS.
The result can be verified through `--jwt-no-jws` option in the [VC Test Suite](https://github.com/w3c/vc-test-suite)

kdimak:
To pass the tests:
1) If no JWS is present, a proof property MUST be provided.
2) If only the proof attribute is used, the alg header MUST be set to none.

Issue #371, opened 2019-09-29
Support Verifiable Presentation data model 


kdimak:
https://w3c.github.io/vc-data-model/#presentations-0

Issue #180, opened 2019-08-28
Zero Knowledge Proof of Verifiable Credential (with Verifiable Presentation)


troyronda:
https://w3c.github.io/vc-data-model/#zero-knowledge-proofs
using Camenisch-Lysyanskaya (CL) Signatures and presumably Hyperledger Ursa library.

Note: If Hyperledger Ursa is not installed on the system, the Framework should still build and operate with support for ZKP.

kdimak:
To verify, remove `zkp` from `sectionsNotSupported` (at **pkg/doc/verifiable/test-suite/config.json**)

PR #382, opened 2019-10-01
fix: VC data model and JWT to pass the test suite

### LINKED ISSUE #358: Fix Verifiable Credential data model and JWT to pass the Verifiable Claims WG Test Suite

kdimak:
closes #358

VC Test Suite statistics:
- 57 passing
- 19 pending
- 18 failing

The reasons for failing:
- JWT suite errors, ES256K support required (#381 )
- Cannot generate JWT from Verifiable Credential without the signature (#372 )
- Support of Verifiable Presentation (#371 )
- Several inconsistencies in VC Test Suite: [#96](https://github.com/w3c/vc-test-suite/issues/96), [#97](https://github.com/w3c/vc-test-suite/issues/97), [#98](https://github.com/w3c/vc-test-suite/issues/98)

The reasons for pending:
- Not implemented support of Zero Knowledge Proof (#180 ), `zkp` test suite is disabled for now.

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -24,8 +24,7 @@ import (
 
 var logger = log.New("aries-framework/doc/verifiable")
 
-const defaultSchema = `
-{
+const defaultSchema = `{
   "required": [
     "@context",
     "type",
@@ -42,8 +41,16 @@ const defaultSchema = `
           "pattern": "^https://www.w3.org/2018/credentials/v1$"
         }
       ],
+      "uniqueItems": true,
       "additionalItems": {
-        "type": "string"
+        "oneOf": [
+          {
+            "type": "object"
+          },
+          {
+            "type": "string"
+          }
+        ]
       }
     },
     "id": {
@@ -51,8 +58,16 @@ const defaultSchema = `
       "format": "uri"
     },
     "type": {
-      "type": "array",
-      "items": [
+      "oneOf": [
+        {
+          "type": "array",
+          "items": [
+            {
+              "type": "string",
+              "pattern": "^VerifiableCredential$"
+            }
+          ]
+        },
         {
           "type": "string",
           "pattern": "^VerifiableCredential$"
@@ -76,7 +91,8 @@ const defaultSchema = `
     "issuer": {
       "anyOf": [
         {
-          "type": "string"
+          "type": "string",
+          "format": "uri"
         },
         {
           "type": "object",
@@ -112,7 +128,10 @@ const defaultSchema = `
       "$ref": "#/definitions/typedID"
     },
     "credentialSchema": {
-      "$ref": "#/definitions/typedID"
+      "$ref": "#/definitions/typedIDs"
+    },
+    "evidence": {
+      "$ref": "#/definitions/typedIDs"
     },
     "refreshService": {
       "$ref": "#/definitions/typedID"
@@ -135,9 +154,32 @@ const defaultSchema = `
           "format": "uri"
         },
         "type": {
-          "type": "string"
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          ]
         }
       }
+    },
+    "typedIDs": {
+      "anyOf": [
+        {
+          "$ref": "#/definitions/typedID"
+        },
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/typedID"
+          }
+        }
+      ]
     }
   }
 }
@@ -149,9 +191,10 @@ const jsonSchema2018Type = "JsonSchemaValidator2018"
 var defaultSchemaLoader = gojsonschema.NewStringLoader(defaultSchema)
 
 // Proof defines embedded proof of Verifiable Credential
-type Proof struct {
-	Type string `json:"type,omitempty"`
-}
+type Proof interface{}
+
+// Evidence defines evidence of Verifiable Credential
+type Evidence interface{}
 
 type typedID struct {
 	ID   string `json:"id,omitempty"`
@@ -176,18 +219,23 @@ type CredentialSchema typedID
 // RefreshService provides a way to automatic refresh of expired Verifiable Credential
 type RefreshService typedID
 
+// TermsOfUse represents terms of use of Verifiable Credential by Issuer or Verifiable Presentation by Holder.
+type TermsOfUse typedID
+
 // Credential Verifiable Credential definition
 type Credential struct {
-	Context        []string
+	Context        []interface{}
 	ID             string
-	Type           []string
+	Type           interface{}
 	Subject        Subject
 	Issuer         Issuer
 	Issued         *time.Time
 	Expired        *time.Time
 	Proof          *Proof
 	Status         *CredentialStatus
-	Schema         *CredentialSchema
+	Schemas        []CredentialSchema
+	Evidence       *Evidence
+	TermsOfUse     []TermsOfUse
 	RefreshService *RefreshService
 }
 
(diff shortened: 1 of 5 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -562,20 +699,20 @@ func describeSchemaValidationError(result *gojsonschema.Result) string {
...
 func getCredentialSchema(schema *CredentialSchema, opts *credentialOpts) (gojsonschema.JSONLoader, error) {
-	schemaLoader := defaultSchemaLoader
 	if schema != nil && !opts.disabledCustomSchema {
 		switch schema.Type {
 		case jsonSchema2018Type:
-			if customSchemaData, err := loadCredentialSchema(schema.ID, opts.schemaDownloadClient); err == nil {
-				schemaLoader = gojsonschema.NewBytesLoader(customSchemaData)
-			} else {
-				return nil, fmt.Errorf("loading custom credential schema from %s failed: %w", schema.ID, err)
+			var err error
```
troyronda: this line can (and should be eliminated if you change L707 to not embed the if condition (and instead create a line after).

pkg/doc/verifiable/credential.go
```diff
@@ -562,20 +699,20 @@ func describeSchemaValidationError(result *gojsonschema.Result) string {
...
+			var customSchemaData []byte
+			if customSchemaData, err = loadCredentialSchema(schema.ID, opts.schemaDownloadClient); err == nil {
```
troyronda: if err != nil is usually first.

pkg/doc/verifiable/credential.go
kdimak: Yeah, my stupid miss.

pkg/doc/verifiable/credential.go
kdimak: OK!

pkg/doc/verifiable/credential.go
```diff
@@ -562,20 +699,19 @@ func describeSchemaValidationError(result *gojsonschema.Result) string {
...
+			if err != nil {
 				return nil, fmt.Errorf("loading custom credential schema from %s failed: %w", schema.ID, err)
 			}
+			return gojsonschema.NewBytesLoader(customSchemaData), nil
 		default:
 			logger.Warnf("unsupported credential schema: %s. Using default schema for validation", schema.Type)
 		}
 	}
-	return schemaLoader, nil
+	return nil, nil
```
troyronda: We shouldn't see return nil, nil

pkg/doc/verifiable/credential.go
```diff
@@ -220,19 +268,37 @@ func (ja JWTAlgorithm) Jose() jose.SignatureAlgorithm {
...
 
+type typeSingle struct {
+	Type string `json:"type,omitempty"`
+}
+
+type typeMultiple struct {
+	Types []string `json:"type,omitempty"`
+}
+
+type credentialSchemaSingle struct {
```
troyronda: Could you explain a bit more why you needed an extra `credentialSchemaSingle` type?

pkg/doc/verifiable/test-suite/verifiable_suite_test.go
```diff
@@ -142,7 +150,7 @@ func parseKeys(packedKeys string) (private, public interface{}) {
 }
 
 func encodeVCToJSON(vcBytes []byte) {
-	credential, err := verifiable.NewCredential(vcBytes)
+	credential, err := verifiable.NewCredential(vcBytes, verifiable.WithNoCustomSchemaCheck())
```
troyronda: Is there an issue?

pkg/doc/verifiable/credential.go
kdimak: Agree!

pkg/doc/verifiable/credential.go
kdimak: You are right.

pkg/doc/verifiable/test-suite/verifiable_suite_test.go
kdimak: In the Test Suite, there is a suite that checks Credential Schemas of Credential. They use fake URLs and `NewCredential()` because of that. Using `WithNoCustomSchemaCheck()`, we disable loading of custom schemas and thus get rid of the issue.

pkg/doc/verifiable/credential.go
kdimak: Refactored.

pkg/doc/verifiable/credential.go
```diff
@@ -561,21 +687,26 @@ func describeSchemaValidationError(result *gojsonschema.Result) string {
 	return errMsg
 }
 
-func getCredentialSchema(schema *CredentialSchema, opts *credentialOpts) (gojsonschema.JSONLoader, error) {
-	schemaLoader := defaultSchemaLoader
-	if schema != nil && !opts.disabledCustomSchema {
+func getSchemaLoader(schemas []CredentialSchema, opts *credentialOpts) (gojsonschema.JSONLoader, error) {
```
troyronda: Just double checking that it was intentional to remove the pointer type.
i.e., you have []CredentialSchema rather than []*CredentialSchema.

kdimak: Sure, Troy. Previously, `getCredentialSchema()` got Schema Loader for single credential schema, now it gets the same for the first found appropriate schema. So it's correct here.

troyronda: Sure - agree that a slice is needed. I was more curious about the pointer type inside.

llorllale:
@kdimak @troyronda can we add this check to the CI?

pkg/doc/verifiable/credential.go
```diff
@@ -655,21 +786,46 @@ func (vc *Credential) JWT(signatureAlg JWTAlgorithm, privateKey interface{}, key
...
-		return "", fmt.Errorf("failed to sign JWT")
+		return "", fmt.Errorf("failed to sign JWT: %w", err)
 	}
 
 	return rawJWT, nil
 }
 
-func (vc *Credential) getFirstSubjectID() string {
+// FirstSubjectID gets ID of first subject if present or return empty string otherwise
+func (vc *Credential) FirstSubjectID() string {
```
llorllale: This function shouldn't exist at all. If the JSON-LD VC contains multiple subjects, the `JWT()` method should return an error indicating as such.

troyronda:
> @kdimak @troyronda can we add this check to the CI?

This test doesn't fully pass yet. I think we should create a separate test suites job (and suppress the failure flag on it for now.)

pkg/doc/verifiable/test-suite/config.json
```diff
@@ -1,8 +1,8 @@
 {
   "generator": "aries-framework-go-gen",
-  "presentationGenerator": "aries-framework-go-gen",
+  "presentationGenerator": "aries-framework-go-gen --presentation",
```
llorllale: `generatorOptions` is there precisely for passing parameters to the `generator` and `presentationGenerator`

troyronda:
> > @kdimak @troyronda can we add this check to the CI?
> 
> @llorllale This test doesn't fully pass yet. I think we should create a separate test suites job (and suppress the failure flag on it for now.)

Done: https://github.com/hyperledger/aries-framework-go/pull/410

Rebase this PR to see results in the CI.

pkg/doc/verifiable/credential.go
kdimak: Agree! I will fix that.

pkg/doc/verifiable/test-suite/config.json
kdimak: On the one hand, we would like to use the same generator for both Verifiable Credential and Verifiable Presentation testing. One the other hand, we need to distinguish the processing of VC from VP in the generator. That's why I put `--presentation` there for VP.

Issue #212, opened 2019-09-05
question - Expose a `defaults.Context()` somewhere?


rolsonquadras:
@troyronda @rolsonquadras I'm not sure what direction to head in this case. Yes - context needs to be available in order to instantiate the protocol services. However, the user does not (currently) have access to a "default" context that they can use to instantiate their own protocol services and pass in to `framework.New()`. Perhaps we need to expose a `defaults.Context()` somewhere?

_Originally posted by @llorllale in https://github.com/hyperledger/aries-framework-go/pull/201_

llorllale:
@rolsonquadras @troyronda allow me to explain:

What triggered my comment above is that `framework.Aries` now has two attributes for the same thing - `services` and `protocolSvcCreators`:

https://github.com/hyperledger/aries-framework-go/blob/0890946c897d13975d33fe02a87d396675d83b44/pkg/framework/aries/framework.go#L26-L32

Why is this? Just so that `Aries.Context()` can keep creating new instances of `context.Provider` without instantiating new services?

llorllale:
It's not clear to me if instantiating new `dispatcher.Service`s would present a problem. They certainly are not responsible for managing resources (eg. inbound transports, storage), hence perhaps why `Aries.Close()` doesn't do anything with them?

rolsonquadras:
https://github.com/hyperledger/aries-framework-go/pull/1103#discussion_r370353265

PR #384, opened 2019-10-01
fix: Fix initialize services in aries farmework

### LINKED ISSUE #383: Fix initialize services in aries farmework

fqutishat:
closes #383

### DIFF
--- a/pkg/framework/aries/framework.go
+++ b/pkg/framework/aries/framework.go
@@ -65,50 +65,33 @@ func New(opts ...Option) (*Aries, error) {
 	//  on the context. The inbound transports require ctx.InboundMessageHandler(), which in-turn depends on
 	//  protocolServices. At the moment, there is a looping issue among these.
 	//  This needs to be resolved and should define a clear relationship between these.
-	ctxProvider, err := frameworkOpts.Context()
-	if err != nil {
-		return nil, fmt.Errorf("context creation failed: %w", err)
-	}
+
+	// Order of initializing service is important
 
 	// Create wallet
-	if e := createWallet(frameworkOpts, ctxProvider); e != nil {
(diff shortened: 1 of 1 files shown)


pkg/framework/aries/framework.go
```diff
@@ -220,8 +224,11 @@ func createWallet(frameworkOpts *Aries, ctxProvider api.Provider) error {
 	return nil
 }
 
-func createOutboundDispatcher(frameworkOpts *Aries, ctxProvider dispatcher.Provider) error {
-	var err error
+func createOutboundDispatcher(frameworkOpts *Aries) error {
+	ctxProvider, err := frameworkOpts.Context()
```
troyronda: why create the context twice instead of passing in from parent?

rolsonquadras:
@fqutishat I think the semantic prefix should be "fix" instead of "bug".

pkg/framework/aries/framework.go
fqutishat: 1- Wallet need context 
2- OutboundDispatcher need context that has wallet
3- Services need context that has OutboundDispatcher

fqutishat: @troyronda we create issue to address this issue

troyronda: It seems like we shouldn't be reusing the same Context() method for everything. The first two seem like special bootstrap steps.

(Re-using Context() for bootstrap is confusing - rather there should be specialized private methods that return only the subset type).

fqutishat: @troyronda we will fix in  https://github.com/hyperledger/aries-framework-go/issues/212

fqutishat: @troyronda I change it

PR #388, opened 2019-10-01
feat: DID Exchange Client Events

### LINKED ISSUE #199: DID Exchange Message Events (Client)
Created issue based on discussion in https://wiki.hyperledger.org/display/ARIES/2019-09-03+Framework+Go+Weekly+Planning.

Framework should provide the ability for the consumers of the framework to register events on protocol clients. This issue is related to #198 (clients registering events on the service).

rolsonquadras:
- Support to register/unregister action events
- Support to register/unregister message events

Closes #199 

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -10,17 +10,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
 
-	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/wallet"
 )
-
-var logger = log.New("aries-framework/didexchange-client")
 
 // provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
 type provider interface {
@@ -38,6 +36,10 @@ type Client struct {
 	inboundTransportEndpoint string
 	actionCh                 chan dispatcher.DIDCommAction
 	msgCh                    chan dispatcher.StateMsg
+	actionEvent              chan<- dispatcher.DIDCommAction
+	actionEventlock          sync.RWMutex
+	msgEvents                []chan<- dispatcher.StateMsg
+	msgEventsLock            sync.RWMutex
 }
 
 // New return new instance of didexchange client
@@ -59,6 +61,7 @@ func New(ctx provider) (*Client, error) {
 		msgCh:    make(chan dispatcher.StateMsg, 10),
 	}
 
+	// start listening for action/message events
 	err = c.startServiceEventListener()
 	if err != nil {
 		return nil, fmt.Errorf("service event listener startup failed: %w", err)
@@ -120,7 +123,6 @@ func (c *Client) RemoveConnection(id string) error {
 
 // startServiceEventListener listens to action and message events from DID Exchange service.
 func (c *Client) startServiceEventListener() error {
-	// register the action event channel
 	err := c.didexchangeSvc.RegisterActionEvent(c.actionCh)
 	if err != nil {
 		return fmt.Errorf("didexchange action event registration failed: %w", err)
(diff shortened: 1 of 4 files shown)


pkg/client/didexchange/client.go
```diff
@@ -119,35 +119,136 @@ func (c *Client) RemoveConnection(id string) error {
 }
 
 // startServiceEventListener listens to action and message events from DID Exchange service.
-func (c *Client) startServiceEventListener() error {
-	// register the action event channel
+func (c *Client) startServiceEventListener() {
+	// listen for action events
+	go func() {
```
troyronda: Did you desire to have two goroutines?

(note, you can alternatively use a for { select {} } within a single goroutine).

rolsonquadras: No, missed to refactor that. Thanks for pointing out.

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/e52e576769b0be364b0393262e689ca6f6dc369f..d58b6228a75eddb0dd13d89c07ec443bf5b490ec).

pkg/client/didexchange/client.go
```diff
@@ -119,35 +119,134 @@ func (c *Client) RemoveConnection(id string) error {
...
+	}
+
+	return nil
+}
+
+// RegisterMsgEvent on DID Exchange protocol messages. The message events are triggered for state transitions. Client
+// will not expect any callback on these events unlike Action events.
+func (c *Client) RegisterMsgEvent(ch chan<- dispatcher.StateMsg) error {
+	// register the service message event channel
+	err := c.didexchangeSvc.RegisterMsgEvent(c.msgCh)
```
bstasyszyn: If this is called multiple times then you'll register the same channel (c.msgCh) multiple times.

rolsonquadras: Valid point. [Added](https://github.com/hyperledger/aries-framework-go/compare/d58b6228a75eddb0dd13d89c07ec443bf5b490ec..7f2cd61c8bc3526f13b02f38f39e9946c1881f10#diff-427522b353a28b1ec350c29c79d387b0R192) a condition to register only once.

pkg/client/didexchange/client.go
```diff
@@ -119,35 +119,134 @@ func (c *Client) RemoveConnection(id string) error {
...
+// UnregisterActionEvent on DID Exchange protocol messages. Refer RegisterActionEvent().
+func (c *Client) UnregisterActionEvent(ch chan<- dispatcher.DIDCommAction) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.actionEvent != ch {
+		return errors.New("invalid channel passed to unregister the action event")
+	}
+
+	c.actionEvent = nil
```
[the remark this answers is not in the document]
rolsonquadras: Moved.

pkg/client/didexchange/client.go
```diff
@@ -119,35 +119,144 @@ func (c *Client) RemoveConnection(id string) error {
...
+
+// RegisterMsgEvent on DID Exchange protocol messages. The message events are triggered for state transitions. Client
+// will not expect any callback on these events unlike Action events.
+func (c *Client) RegisterMsgEvent(ch chan<- dispatcher.StateMsg) error {
+	c.msgEventLock.Lock()
+	count := len(c.msgEvents)
+	c.msgEventLock.Unlock()
+
+	// register the service message event channel (only once)
+	if count == 0 {
```
bstasyszyn: I just realized that there could be a race condition here. If two concurrent calls come here then both of them could get a count of 0 and then both would register the event.

troyronda:
@rolsonquadras rebase is needed.

pkg/client/didexchange/client.go
```diff
@@ -119,35 +119,144 @@ func (c *Client) RemoveConnection(id string) error {
...
+
+	// register the service action event channel
 	err := c.didexchangeSvc.RegisterActionEvent(c.actionCh)
 	if err != nil {
 		return fmt.Errorf("didexchange action event registration failed: %w", err)
 	}
 
-	// register the message event channel
-	err = c.didexchangeSvc.RegisterMsgEvent(c.msgCh)
+	c.actionEvent = ch
```
soluchok: Probably we shod use the same approach `if count == 0 {`

pkg/client/didexchange/client.go
```diff
@@ -38,6 +36,10 @@ type Client struct {
 	inboundTransportEndpoint string
 	actionCh                 chan dispatcher.DIDCommAction
 	msgCh                    chan dispatcher.StateMsg
+	actionEvent              chan<- dispatcher.DIDCommAction
+	msgEvents                []chan<- dispatcher.StateMsg
+	lock                     sync.RWMutex
+	msgEventLock             sync.RWMutex
```
soluchok: better to keep the mutex close to the data it is destined to protect
`msgEventLock` field should be above `msgEvents` the point is we can see what this mutex for
e.g
```
lock                          sync.RWMutex	
actionEvent              chan<- dispatcher.DIDCommAction
msgEventLock         sync.RWMutex
msgEvents               []chan<- dispatcher.StateMsg
```
in this structure, we can see that `lock` is responsible for `actionEvent` and `msgEventLock` for `msgEvents`

soluchok: and probably better to rename `lock` to `actionEventLock`

rolsonquadras: [Done](https://github.com/hyperledger/aries-framework-go/compare/7f2cd61c8bc3526f13b02f38f39e9946c1881f10..7e659b43cc8ebdfacb141bb9a0805dcddc74b450#diff-427522b353a28b1ec350c29c79d387b0R42).

pkg/client/didexchange/client.go
rolsonquadras: The original code had creation at startup. Reverted this logic.

rolsonquadras:
> @rolsonquadras rebase is needed.

Done.

pkg/client/didexchange/client.go
```diff
@@ -132,22 +134,109 @@ func (c *Client) startServiceEventListener() error {
...
+	c.actionEvent = nil
+
+	return nil
+}
+
+// RegisterMsgEvent on DID Exchange protocol messages. The message events are triggered for state transitions. Client
+// will not expect any callback on these events unlike Action events.
+func (c *Client) RegisterMsgEvent(ch chan<- dispatcher.StateMsg) error {
+	c.msgEventsLock.Lock()
+	c.msgEvents = append(c.msgEvents, ch)
```
troyronda: I wonder if multiple channels will become registered to a particular client.
(I understand why protocol services have multiple listeners - it is less clear that Clients do).

rolsonquadras: Since these are just stream listeners and client doesn't expect anything back. So, providing support for multiple listeners won't harm. You have anything specific in mind ?

troyronda: @rolsonquadras Just wondering if there is unnecessary complexity :).

rolsonquadras: @troyronda Created an issue https://github.com/hyperledger/aries-framework-go/issues/395 to discuss this further.

Issue #242, opened 2019-09-11
Service callback processing error handling


rolsonquadras:
Based on TODO in the PR - https://github.com/hyperledger/aries-framework-go/pull/240/files#diff-f761c91e753eba1c42a68076df06ce12R135

```
func (s *Service) processCallback(didCommCallback event.DIDCommCallback) {
	// TODO what to do in case of any failures with callback processing ? - currently, logging the error messages and
      //  ignoring the flow.
```

soluchok:
I think we can add a config to our service if the client wants to receive errors he should pass config with `ConsumerErrors` param equal to `true` e.g  Config{ConsumerErrors:true} , also we need to expose one more method `ConsumerErrors() <-chan error` which will return errors which happens during the execution. Under the hood, we will have internal error chan where we will send en errors. It should be unbuffered chan and the client must listen to ConsumerErrors() function in a different goroutine and handle those errors. If the client will not handle it it will lead to deadlock.
Everything described above applies only when `Config.ConsumerErrors=true`

rolsonquadras:
Tagging discussion on the design:
https://github.com/hyperledger/aries-framework-go/pull/393#issuecomment-539107508
https://github.com/hyperledger/aries-framework-go/pull/393#issuecomment-539177203

rolsonquadras:
Need more grooming on this issue, specifically retry.

PR #321, opened 2019-09-23
refactor: Service Event - Update to Register/Unregister APIs

rolsonquadras:
- Remove AutoExecute Register/Unregister API
- Consumer need to register a channel to process all the messages
- Utility function to AutoExecute Action Events

Design is tracked in #276. This PR also closes #276.

Signed-off-by: Rolson Quadras <rolson.quadras@securekey.com>

PR #393, opened 2019-10-02
refactor: Added protocol callback error handling

soluchok:
* ability to consume the result of the callback function

Closes: https://github.com/hyperledger/aries-framework-go/issues/242

### DIFF
--- a/pkg/didcomm/dispatcher/api.go
+++ b/pkg/didcomm/dispatcher/api.go
@@ -77,7 +77,7 @@ type DIDCommAction struct {
 	// DIDComm message
 	Message DIDCommMsg
 	// Continue function to be called by the consumer for further processing the message.
-	Continue func()
+	Continue func(args ...interface{})
 	// Stop invocation notifies the service that the consumer action event processing has failed or the consumer wants
 	// to stop the processing.
 	Stop func(err error)
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -27,8 +27,34 @@ var logger = log.New("aries-framework/did-exchange/service")
 
 // didCommChMessage type to correlate actionEvent message(go channel) with callback message(internal go channel).
 type didCommChMessage struct {
-	ID  string
-	Err error
+	id      string
+	err     error
+	action  *dispatcher.DIDCommAction
+	errChan chan<- ActionError
+}
+
+func (msg *didCommChMessage) applyOptions(args ...interface{}) {
+	for _, arg := range args {
+		if opt, ok := arg.(Opt); ok {
+			opt(msg)
+		}
+	}
+}
+
+// ActionError structure for error handling
+type ActionError struct {
+	Err    error
+	Action *dispatcher.DIDCommAction
+}
+
+// Opt options for didCommChMessage
+type Opt func(m *didCommChMessage)
+
+// WithError provides the possibility to receive errors
+func WithError(eChan chan<- ActionError) Opt {
+	return func(m *didCommChMessage) {
+		m.errChan = eChan
+	}
 }
 
 const (
@@ -281,13 +307,22 @@ func (s *Service) sendActionEvent(msg dispatcher.DIDCommMsg, aEvent chan<- dispa
 	// TODO pass invitation id #397
 	didCommAction := dispatcher.DIDCommAction{
 		Message: msg,
-		Continue: func() {
-			s.processCallback(id, nil)
-		},
 		Stop: func(err error) {
-			s.processCallback(id, err)
+			s.processCallback(didCommChMessage{
+				id:  id,
+				err: err,
+			})
 		},
 		Properties: s.createEventProperties(threadID, ""),
+	}
+
+	didCommAction.Continue = func(args ...interface{}) {
+		msg := didCommChMessage{
+			id:     id,
+			action: &didCommAction,
+		}
+		msg.applyOptions(args...)
+		s.processCallback(msg)
 	}
 
 	// trigger the registered action event
@@ -312,31 +347,36 @@ func (s *Service) sendMsgEvents(msg *dispatcher.StateMsg) {
 func (s *Service) startInternalListener() {
 	go func() {
 		for msg := range s.callbackChannel {
-			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
 			if err := s.process(msg); err != nil {
-				// TODO handle error
-				logger.Errorf(err.Error())
+				if msg.errChan == nil {
+					logger.Errorf(err.Error())
+					continue
+				}
+				msg.errChan <- ActionError{
+					Err:    err,
+					Action: msg.action,
+				}
 			}
 		}
 	}()
 }
 
-func (s *Service) processCallback(id string, err error) {
+func (s *Service) processCallback(msg didCommChMessage) {
 	// pass the callback data to internal channel. This is created to unblock consumer go routine and wrap the callback
 	// channel internally.
-	s.callbackChannel <- didCommChMessage{ID: id, Err: err}
+	s.callbackChannel <- msg
 }
 
 // processCallback processes the callback events.
 func (s *Service) process(msg didCommChMessage) error {
-	if msg.Err != nil {
+	if msg.err != nil {
 		// TODO https://github.com/hyperledger/aries-framework-go/issues/438 Cleanup/Update data on Stop Event Action
-		logger.Errorf("client action event processing failed - msgID:%s error:%s", msg.ID, msg.Err)
+		logger.Errorf("client action event processing failed - msgID:%s error:%s", msg.id, msg.err)
 		return nil
 	}
 
 	// fetch the record
-	jsonDoc, err := s.store.Get(msg.ID)
+	jsonDoc, err := s.store.Get(msg.id)
 	if err != nil {
 		return fmt.Errorf("document for the id doesn't exists in the database: %w", err)
 	}
--- a/pkg/didcomm/protocol/didexchange/service_test.go
+++ b/pkg/didcomm/protocol/didexchange/service_test.go
@@ -765,10 +765,14 @@ func handleConnectionResponseEvents(t *testing.T, svc *Service, e *dispatcher.DI
 	err := json.Unmarshal(e.Message.Payload, &pl)
 	require.NoError(t, err)
 
+	var invalid bool
 	if pl.ID == changeID {
-		e.Continue = func() {
-			svc.processCallback("invalid-id", nil)
+		e.Continue = func(args ...interface{}) {
+			msg := didCommChMessage{id: "invalid-id", action: e}
+			msg.applyOptions(args...)
+			svc.processCallback(msg)
 		}
+		invalid = true
 	}
 
 	if pl.ID == handleError {
@@ -781,13 +785,26 @@ func handleConnectionResponseEvents(t *testing.T, svc *Service, e *dispatcher.DI
 		err = svc.store.Put(id, jsonDoc)
 		require.NoError(t, err)
 
-		e.Continue = func() {
-			svc.processCallback(id, nil)
+		e.Continue = func(...interface{}) {
+			svc.processCallback(didCommChMessage{id: id})
 		}
 	}
 
+	if invalid {
+		eChan := make(chan ActionError)
+		e.Continue(WithError(eChan))
+		select {
+		case res := <-eChan:
+			require.EqualError(t, res.Err, "document for the id doesn't exists in the database: data not found")
+			require.Equal(t, e, res.Action)
+		case <-time.After(time.Millisecond * 50):
+			t.Error("timeout: did not receive a result")
+		}
+	} else {
+		e.Continue()
+	}
+
 	// invoke callback
-	e.Continue()
 	switch pl.ID {
 	case changeID:
 		// no state change since there was a error with processing
@@ -817,8 +834,8 @@ func handleConnectionAckEvents(t *testing.T, svc *Service, e *dispatcher.DIDComm
 		err := svc.store.Put(id, []byte("invalid json"))
 		require.NoError(t, err)
 
-		e.Continue = func() {
-			svc.processCallback(id, nil)
+		e.Continue = func(...interface{}) {
+			svc.processCallback(didCommChMessage{id: id})
 		}
 	}
 


troyronda:
@soluchok normal semantic PR prefixes: https://github.com/commitizen/conventional-commit-types/blob/v2.2.0/index.json (from https://github.com/probot/semantic-pull-requests)

ie refactor not refactoring

rolsonquadras:
@soluchok issue #242 is a feature issue. The PR title says "refactor". Should the title be updated with "feat" ?

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -739,7 +739,10 @@ func handleConnectionRequestEvents(t *testing.T, svc *Service, e dispatcher.DIDC
 	}(e)
 
 	// invoke callback
-	e.Callback(dispatcher.DIDCommCallback{Err: err})
+	msg := &dispatcher.DIDCommCallback{}
+	err = msg.Done(dispatcher.Result{Err: err})
+	require.NoError(t, err)
+	e.Callback(msg)
```
rolsonquadras: Can you please add a test case to validate the error from callback processing ? i.e, consumer calls the callback and validates the error from the callback.

soluchok: Done, I've modified the existing one

troyronda:
@rolsonquadras @soluchok "feat: Added error handling" will be a meaningless entry in the change log.

pkg/didcomm/dispatcher/api_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		const errMsg = "callback was already done use IsValid() function to prevent this error"
+		// double done
+		err = c.Done(Result{})
+		require.EqualError(t, err, errMsg)
+	})
+
+	t.Run("WithResult", func(t *testing.T) {
+		c := NewDIDCommCallback().WithResult()
+		done := make(chan bool)
```
troyronda: See https://github.com/hyperledger/aries-framework-go/pull/321#discussion_r327313409

FYI: done channel signatures are usually this:
`chan struct{}`

rolsonquadras:
> @soluchok issue #242 is a feature issue. The PR title says "refactor". Should the title be updated with "feat" ?

Any specific reason to revert back to "refactor". Also the commit message doesn't have semantic prefix.

soluchok:
> > @soluchok issue #242 is a feature issue. The PR title says "refactor". Should the title be updated with "feat" ?
> 
> Any specific reason to revert back to "refactor". Also the commit message doesn't have semantic prefix.

Ok, let's decide is it `refactor` or `feat`? @troyronda @rolsonquadras 
and I will change comment and title

rolsonquadras:
> > > @soluchok issue #242 is a feature issue. The PR title says "refactor". Should the title be updated with "feat" ?
> > 
> > 
> > Any specific reason to revert back to "refactor". Also the commit message doesn't have semantic prefix.
> 
> Ok, let's decide is it `refactor` or `feat`? @troyronda @rolsonquadras
> and I will change comment and title

My bad, missed @troyronda 's comment.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -775,13 +780,32 @@ func handleConnectionResponseEvents(t *testing.T, svc *Service, e dispatcher.DID
...
 	}
 
 	// invoke callback
-	e.Callback(dispatcher.DIDCommCallback{Err: nil})
+	msg := dispatcher.NewDIDCommCallback()
+	if invalid {
+		done := make(chan struct{})
+		result, err := msg.WithResult().Result()
+		go func() {
+			require.NoError(t, err)
```
fqutishat: move it up

pkg/didcomm/dispatcher/api.go
```diff
@@ -78,14 +80,63 @@ type DIDCommAction struct {
...
+	if c.res != nil {
+		c.res <- r
+		close(c.res)
+	}
+
+	return nil
+}
+
+// IsValid is helper function to check if callback was already executed
+func (c *DIDCommCallback) IsValid() bool {
```
llorllale: Should rename to `IsDone`

pkg/didcomm/dispatcher/api_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	t.Run("WithResult", func(t *testing.T) {
+		c := NewDIDCommCallback().WithResult()
+		done := make(chan struct{})
+		go func() {
+			ch, err := c.Result()
+			require.NoError(t, err)
+			select {
+			case <-ch:
+			case <-time.After(time.Millisecond * 50):
```
llorllale: this timeout is unnecessary

pkg/didcomm/dispatcher/api.go
```diff
@@ -78,14 +80,63 @@ type DIDCommAction struct {
...
+func NewDIDCommCallback() *DIDCommCallback {
+	return &DIDCommCallback{}
+}
+
+// WithResult creates result chan which provides the possibility to be notified when the callback is finished
+func (c *DIDCommCallback) WithResult() *DIDCommCallback {
+	if c.res != nil {
+		return c
+	}
+	c.res = make(chan Result)
```
troyronda: It seems we are forcing each Callback to have its own channel.

There is a trade-off here. (when channels are provided by the calling code, they can re-use them across many calls).

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -775,13 +780,32 @@ func handleConnectionResponseEvents(t *testing.T, svc *Service, e dispatcher.DID
...
 			svc.processCallback(id, didCommCallback)
 		}
 	}
 
 	// invoke callback
-	e.Callback(dispatcher.DIDCommCallback{Err: nil})
+	msg := dispatcher.NewDIDCommCallback()
+	if invalid {
+		done := make(chan struct{})
+		result, err := msg.WithResult().Result()
```
troyronda: I would suggest that the extra WithResult() is unnecessary from a caller's Point of View.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -294,38 +294,40 @@ func (s *Service) sendMsgEvents(msg *dispatcher.StateMsg) {
 func (s *Service) startInternalListener() {
 	go func() {
 		for msg := range s.callbackChannel {
-			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
-			if err := s.process(msg.ID, msg.DIDCommCallback); err != nil {
-				// TODO handle error
-				logger.Errorf(err.Error())
+			if !msg.DIDCommCallback.IsValid() {
+				continue
```
troyronda: This logic confuses me a bit. When will this scenario occur?

llorllale: For starters, it should be named `IsDone` or something similar.

troyronda: Why is there a msg in the queue that isn't "Done"?

pkg/didcomm/dispatcher/api.go
```diff
@@ -78,14 +80,63 @@ type DIDCommAction struct {
...
+	done bool
+}
+
+// NewDIDCommCallback creates DIDCommCallback message
+func NewDIDCommCallback() *DIDCommCallback {
+	return &DIDCommCallback{}
+}
+
+// WithResult creates result chan which provides the possibility to be notified when the callback is finished
+func (c *DIDCommCallback) WithResult() *DIDCommCallback {
```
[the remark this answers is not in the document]
rolsonquadras: Adding to the above point, I feel rather than giving the option to the consumer to receive the error result, service should always give back the error. It should be up-to the consumer to either process or ignore the error.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -775,13 +780,32 @@ func handleConnectionResponseEvents(t *testing.T, svc *Service, e dispatcher.DID
...
+		go func() {
+			select {
+			case res := <-result:
+				require.EqualError(t, res.Err, "document for the id doesn't exists in the database: data not found")
+			case <-time.After(time.Millisecond * 50):
+				t.Error("timeout: did not receive a result")
+			}
+			close(done)
+		}()
+		e.Callback(msg)
```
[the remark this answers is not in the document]
rolsonquadras: @troyronda There was a preliminary discussion of using MsgEvents for throwing the callback errors as well. Would it be worth while to explore this option ?

pkg/didcomm/protocol/didexchange/service.go
soluchok: `why is the logic to simply skip it` it is according to previous logic (unit tests depend on it) it can happen when a client execute callback inappropriate way (should never happen) 
I would suggest here panic instead of `continue` because there is no another way to notify the client about the error or we can just log this error

pkg/didcomm/dispatcher/api.go
soluchok: Done!

pkg/didcomm/dispatcher/api.go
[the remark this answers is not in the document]
soluchok: Done!

pkg/didcomm/dispatcher/api.go
```diff
@@ -78,14 +80,63 @@ type DIDCommAction struct {
...
+		c.res <- r
+		close(c.res)
+	}
+
+	return nil
+}
+
+// IsValid is helper function to check if callback was already executed
+func (c *DIDCommCallback) IsValid() bool {
+	return !c.done
```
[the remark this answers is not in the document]
soluchok: Done!

pkg/didcomm/dispatcher/api.go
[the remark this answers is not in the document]
soluchok: Done!

soluchok: I was thinking about reusing, but I did not like because when we receive an error we will never know for which callback it was.

pkg/didcomm/protocol/didexchange/service_test.go
soluchok: Done

troyronda:
This PR may need a discussion. I chatted with @rolsonquadras about the approach and he plans to write something up.

rolsonquadras:
> This PR may need a discussion. I chatted with @rolsonquadras about the approach and he plans to write something up.

The current consumer code looks something like follows without error handling (master branch).
```
	actionCh := make(chan dispatcher.DIDCommAction)
	err := svc.RegisterActionEvent(actionCh)
	for {
		select {
			case e := <-c.actionCh:
				e.Continue() 	// e.Callback() is changed to e.Continue  #437
				// TODO handle error from callback
		}
	}

```

With the error handling code added in this PR, the consumer code will look like below.
```
	actionCh := make(chan dispatcher.DIDCommAction)
	err := svc.RegisterActionEvent(actionCh)
	for {
		select {
			case e := <-c.actionCh:
				msg := dispatcher.NewDIDCommCallback()
				for {
					select {
					case res := <-msg.Result():
						// process error
				}()
				e.Continue(msg)
		}
	}

```
This would block the main select(actionCh) and other actionEvents will buffer. This would be similar to sync call err := e.Continue().

> 
> Briefly my thoughts were:
> 
> * Callback is currently handling two scenarios (Client wants to signal 'Continue with Protocol' vs 'Fail/Stop Protocol'). Suggest these be separated (e.g., Fail doesn't need an error channel).

This has been handled through #436 and #437.

> * For a client to have handle (multiple) async scenarios, it would have been logical for that client to pass-in a channel for receiving error events. In this way they could setup a for { select {} } for all event handling.

```
	actionCh := make(chan dispatcher.DIDCommAction)
	err := svc.RegisterActionEvent(actionCh)
	errorCh := make(chan <some_type)
	
	for {
		select {
			case e := <-c.actionCh:
				e.Continue(errorCh)
			case e := <-errorCh:
				// process the error
				// correlate based on ConnectionID - https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/service.go#L254
		}
	}

```

@troyronda @soluchok

soluchok:
@rolsonquadras 
> With the error handling code added in this PR, the consumer code will look like below.

```
	actionCh := make(chan dispatcher.DIDCommAction)
	err := svc.RegisterActionEvent(actionCh)
	for {
		select {
		case e := <-c.actionCh:
			msg := dispatcher.NewDIDCommCallback()
			res := <-msg.Result()
			if res.Err {
				e.Stop(res.Err)
			} else {
				e.Continue(msg)
			}
		}
	}
```

rolsonquadras:
> @rolsonquadras
> 
> > With the error handling code added in this PR, the consumer code will look like below.
> 
> ```
> 	actionCh := make(chan dispatcher.DIDCommAction)
> 	err := svc.RegisterActionEvent(actionCh)
> 	for {
> 		select {
> 		case e := <-c.actionCh:
> 			msg := dispatcher.NewDIDCommCallback()
>                         e.Continue(msg)
>                         go func(m *dispatcher.DIDCommCallback){
> 			    if  res := <-m.Result(); res.Err != nil {
> 				e.Stop(res.Err)
> 			    }
>                         }(msg)
> 		}
> 	}
> ```

Consumer will either invoke either Continue() or Stop(). For Stop(), consumer won't send back the error received from the callback. It's the error from action processing.

pkg/didcomm/dispatcher/api.go
```diff
@@ -72,12 +72,18 @@ type Provider interface {
...
+	Action *DIDCommAction
+}
+
 // DIDCommAction message type to pass events in go channels.
 type DIDCommAction struct {
 	// DIDComm message
 	Message DIDCommMsg
 	// Continue function to be called by the consumer for further processing the message.
-	Continue func()
+	Continue func(chan<- ActionError)
```
troyronda: It would be nice if we could avoid having to pas a nil argument when the caller doesn’t care about the error chan. I wonder if we should use a variadic pattern?

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -781,13 +783,26 @@ func handleConnectionResponseEvents(t *testing.T, svc *Service, e *dispatcher.DI
...
+			svc.processCallback(didCommChMessage{id: id})
 		}
 	}
 
+	if invalid {
+		eChan := make(chan dispatcher.ActionError)
+		e.Continue(eChan)
+		select {
+		case res := <-eChan:
+			require.EqualError(t, res.Err, "document for the id doesn't exists in the database: data not found")
```
troyronda: it would be good to have a test case showing correlation.

pkg/didcomm/dispatcher/api.go
soluchok: Done!

soluchok: In my opinion, better to pass a clear argument, it's much easier to work with it. But options are not a bad approach, especially for callback (probably it will require one more option in the future)

llorllale: I like Rolson's snippet towards the end of [this comment](https://github.com/hyperledger/aries-framework-go/pull/393#issuecomment-539177203) better. It would be simpler from a user perspective if the error message has a `Retry()` function or something similar.

soluchok: It has `res.Action.Continue()` if you need retry it

PR #406, opened 2019-10-03
feat: Persist Invitation during Create Invitation

### LINKED ISSUE #396: Store Invitation Details (Inviter - CreateInvitation)
- Store invitation details (object) to retriever it at later stages to map the exRequest to invitation. 
- Have a mapping of recipient pubKey to the invitation.

sudeshrshetty:
- closes #396

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -17,6 +17,8 @@ import (
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange/persistence"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/wallet"
 )
 
@@ -25,6 +27,7 @@ type provider interface {
 	Service(id string) (interface{}, error)
 	CryptoWallet() wallet.Crypto
 	InboundTransportEndpoint() string
+	StorageProvider() storage.Provider
 }
 
 // Client enable access to didexchange api
@@ -40,6 +43,7 @@ type Client struct {
 	actionEventlock          sync.RWMutex
 	msgEvents                []chan<- dispatcher.StateMsg
 	msgEventsLock            sync.RWMutex
+	recorder                 *persistence.ConnectionRecorder
 }
 
 // New return new instance of didexchange client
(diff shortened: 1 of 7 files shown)


pkg/internal/common/persistence/connection.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package persistence
```
troyronda: i am wary of a common package holding persistence structures. These structures should be in their associated packages.

sudeshrshetty: this package doesn't maintain any structures, it is just like a util package. Isolating read, write, hashing to separate package, so that it will be easier to maintain in future.
As per the flow: service, restapi and client may need to access persistence (sometimes duplicate logic), so thought of moving it to separate package and using it as util from all layers.

may be package name is confusing?

troyronda: // ConnectionRecord takes care of connection related persistence
 type ConnectionRecord struct {
 	store storage.Store
 }

sudeshrshetty: gotcha, any package suggestion where it can be moved?

sudeshrshetty: fixed

pkg/client/didexchange/client.go
```diff
@@ -48,17 +52,25 @@ func New(ctx provider) (*Client, error) {
...
+	}
+
 	c := &Client{
 		didexchangeSvc:           didexchangeSvc,
 		wallet:                   ctx.CryptoWallet(),
 		inboundTransportEndpoint: ctx.InboundTransportEndpoint(),
 		// TODO channel size - https://github.com/hyperledger/aries-framework-go/issues/246
 		actionCh: make(chan dispatcher.DIDCommAction, 10),
 		msgCh:    make(chan dispatcher.StateMsg, 10),
+		record:   persistence.NewConnectionRecord(store),
```
troyronda: "Record" makes it sound like this is an individual record rather than a handle to storage for a particular type.

sudeshrshetty: Will rename it to `Recorder`

sudeshrshetty: done

pkg/client/didexchange/client.go
```diff
@@ -40,6 +43,7 @@ type Client struct {
 	actionEventlock          sync.RWMutex
 	msgEvents                []chan<- dispatcher.StateMsg
 	msgEventsLock            sync.RWMutex
+	record                   *persistence.ConnectionRecord
```
troyronda: This still has "record"

sudeshrshetty: done

PR #407, opened 2019-10-03
refactor: Legacy Authcrypter implements Crypter interface

Moopli:
This is a prelude to the wallet-crypto refactor; where the legacy Authcrypter has been refactored to implement the Crypter interface and simplify some code (for example, in generating the protected header and using it as the AAD).


Baha-sk:
code coverage of `pkg/didcomm/crypto/legacy/authcrypt/encrypt.go` decreased a bit..

Moopli:
> code coverage of `pkg/didcomm/crypto/legacy/authcrypt/encrypt.go` decreased a bit..

Fixed that up, coverage now increased instead.

PR #417, opened 2019-10-03
refactor: Get public key in didexchange state

### LINKED ISSUE #416: Refactor public key calculation in didexchange states
This change to remove the directly assessing first public key from the did doc. Instead it will loop through and fetch the key which is supported by the didcomm message protocol.

talwinder50:
closes #416

This change to remove the directly assessing first public key from the did doc. Instead it will loop through and fetch the key which is supported by the didcomm message protocol.

### DIFF
--- a/pkg/didcomm/protocol/didexchange/states.go
+++ b/pkg/didcomm/protocol/didexchange/states.go
@@ -32,6 +32,8 @@ const (
 	stateNameResponded = "responded"
 	stateNameCompleted = "completed"
 	ackStatusOK        = "ok"
+	// Todo:How to find the key type -Issue-439
+	supportedPublicKeyType = "Ed25519VerificationKey2018"
 )
 
 //TODO: This is temporary to move forward with bdd test will be fixed in Issue-353
@@ -280,8 +282,11 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation, thid string)
 	if err != nil {
 		return nil, err
 	}
-	// choose the first public key
-	sendVerKey := string(newDidDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKeys(newDidDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, fmt.Errorf("error while getting public key %s", err)
+	}
+	sendVerKey := string(pubKey[0].Value)
 	temp = sendVerKey
 	// prepare the request :
 	// TODO Service.Handle() is using the ID from the Invitation as the threadID when instead it should be
@@ -327,7 +332,11 @@ func (ctx *context) handleInboundRequest(request *Request) (stateAction, error) 
 
 	destination := prepareDestination(request.Connection.DIDDoc)
 
-	sendVerKey := string(newDidDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKeys(newDidDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, err
+	}
+	sendVerKey := string(pubKey[0].Value)
 	// send exchange response
 	return func() error {
 		return ctx.outboundDispatcher.Send(response, sendVerKey, destination)
@@ -348,7 +357,11 @@ func (ctx *context) sendOutboundRequest(msg dispatcher.DIDCommMsg) (stateAction,
 		return nil, err
 	}
 	// choose the first public key
-	sendVerKey := string(request.Connection.DIDDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKeys(request.Connection.DIDDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, err
+	}
+	sendVerKey := string(pubKey[0].Value)
 	// send the exchange request
 	return func() error {
 		return ctx.outboundDispatcher.Send(request, sendVerKey, destination)
@@ -385,8 +398,13 @@ func (ctx *context) sendOutboundResponse(msg dispatcher.DIDCommMsg) (stateAction
 	if err != nil {
 		return nil, err
 	}
+
+	pubKey, err := getPublicKeys(connection.DIDDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, err
+	}
 	// choose the first public key
-	sendVerKey := string(connection.DIDDoc.PublicKey[0].Value)
+	sendVerKey := string(pubKey[0].Value)
 
 	return func() error {
 		return ctx.outboundDispatcher.Send(response, sendVerKey, destination)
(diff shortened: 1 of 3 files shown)


troyronda:
@talwinder50 please rebase and update PR & commit title.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -276,7 +278,11 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation, thid string)
 		return nil, err
 	}
 	// choose the first public key
-	sendVerKey := string(newDidDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKey(newDidDoc, supportedPublicKeyType)
```
sandrask: Coverage decreased in all files due to this new error condition. You should add test for no key found scenario. You can do it by passing in mock did creator that returns did without public keys or with wrong key types.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -323,7 +329,11 @@ func (ctx *context) handleInboundRequest(request *Request) (stateAction, error)
 
 	destination := prepareDestination(request.Connection.DIDDoc)
 
-	sendVerKey := string(newDidDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKey(newDidDoc, supportedPublicKeyType)
```
sandrask: Same comment as above regarding coverage.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -344,7 +354,11 @@ func (ctx *context) sendOutboundRequest(msg dispatcher.DIDCommMsg) (stateAction,
 		return nil, err
 	}
 	// choose the first public key
-	sendVerKey := string(request.Connection.DIDDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKey(request.Connection.DIDDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, err
```
sandrask: You should test error condition.

Moopli:
I don't think the coverage report is accurate when it says that coverage changed in the legacy authcrypter - reading through the code I don't see any connection with the crypter.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -281,7 +283,11 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation, thid string)
 		return nil, err
 	}
 	// choose the first public key
```
troyronda: is this comment still valid?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -281,7 +283,11 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation, thid string)
...
-	sendVerKey := string(newDidDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKey(newDidDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, err
+	}
+	sendVerKey := string(pubKey[0].Value)
```
troyronda: why pubKey[0]?

pkg/didcomm/protocol/didexchange/states.go
talwinder50: Removed

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -492,3 +511,15 @@ func (ctx *context) handleInboundResponse(response *Response) (stateAction, erro
 func getEpochTime() int64 {
 	return time.Now().Unix()
 }
+func getPublicKey(didDoc *did.Doc, pubKeyType string) ([]did.PublicKey, error) {
+	var publicKeys []did.PublicKey
+	for k := range didDoc.PublicKey {
+		if didDoc.PublicKey[k].Type == pubKeyType {
```
troyronda: why not have range return the pubKey rather than looking up again?

pkg/didcomm/protocol/didexchange/states.go
talwinder50: so did document can have a multiple keys for same type. Therefore we will pick the first one.

pkg/didcomm/protocol/didexchange/states.go
talwinder50: done

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -32,6 +32,8 @@ const (
 	stateNameResponded = "responded"
 	stateNameCompleted = "completed"
 	ackStatusOK        = "ok"
+	// Todo : How to find the key type
+	supportedPublicKeyType = "Ed25519VerificationKey2018"
```
rolsonquadras: Is there an issue # for this ?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -280,8 +282,11 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation, thid string)
 	if err != nil {
 		return nil, err
 	}
-	// choose the first public key
-	sendVerKey := string(newDidDoc.PublicKey[0].Value)
+	pubKey, err := getPublicKey(newDidDoc, supportedPublicKeyType)
+	if err != nil {
+		return nil, err
```
rolsonquadras: Better to return error with a message.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -492,3 +510,15 @@ func (ctx *context) handleInboundResponse(response *Response) (stateAction, erro
 func getEpochTime() int64 {
 	return time.Now().Unix()
 }
+func getPublicKey(didDoc *did.Doc, pubKeyType string) ([]did.PublicKey, error) {
```
rolsonquadras: imo the function name should end with` s` ie, `getPublicKeys` as this is returning a slice.

talwinder50: done

pkg/didcomm/protocol/didexchange/states.go
talwinder50: done

pkg/didcomm/protocol/didexchange/states.go
talwinder50: added

pkg/didcomm/protocol/didexchange/states.go
troyronda: Is this the documented behaviour in the protocol spec?

PR #419, opened 2019-10-04
fix: Return stateAction from sendOutboundAck

### LINKED ISSUE #418: Return stateAction from sendOutboundAck

fqutishat:
- return send functionality in func

closes #418


pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -255,14 +255,16 @@ func (s *completed) Execute(msg dispatcher.DIDCommMsg, thid string, ctx context)
 		}
 		return &noOp{}, action, nil
 	case ConnectionAck:
+		var action = func() error { return nil }
```
troyronda: action :=

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -445,11 +447,13 @@ func (ctx *context) sendOutboundAck(msg dispatcher.DIDCommMsg) error {
 
 	err := json.Unmarshal(msg.Payload, ack)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	// TODO : Issue-353
 	sendVerKey := temp
-	return ctx.outboundDispatcher.Send(ack, sendVerKey, destination)
+	return func() error {
```
troyronda: Suggest assigning the func to a variable as a new line. (and then returning that variable.)

troyronda:
@fqutishat a bit more description would be helpful :).

pkg/didcomm/protocol/didexchange/states.go
fqutishat: done

pkg/didcomm/protocol/didexchange/states.go
fqutishat: done

Issue #397, opened 2019-10-02
Update the service to save and retrieve connection record using connection recorder storage 


rolsonquadras:
- Create Own Unique ConnectionID (status - Pending) [Use - UUID] : 
- Create unique threadID (requestID) and store the mapping to connectionID

key : connectionID, val : {endpoint, recKey, inviationID}
key : protocolID=(myDID+threadID), val : connectionID

protocolID should be the hash.

talwinder50:
InvitationID here is the request ID

PR #422, opened 2019-10-04
feat: Pass connectionID and invitationID in events

### LINKED ISSUE #403: Pass connectionID and  invitationID in events
Event need to pass connectionID and eventID to consumers.

fqutishat:
- Pass connectionID and invitationID from did exchange service

closes #403

### DIFF
--- a/pkg/didcomm/dispatcher/api.go
+++ b/pkg/didcomm/dispatcher/api.go
@@ -47,6 +47,8 @@ type StateMsg struct {
 	Type    StateMsgType
 	StateID string
 	Msg     DIDCommMsg
+	// Properties contains different values from different protocols
+	Properties map[string]interface{}
 }
 
 // Destination provides the recipientKeys, routingKeys, and serviceEndpoint populated from Invitation
@@ -76,6 +78,8 @@ type DIDCommAction struct {
 	Message DIDCommMsg
 	// Callback function to be called by the consumer for further processing the message.
 	Callback Callback
+	// Properties contains different values from different protocols
+	Properties map[string]interface{}
 }
 
 // DIDCommCallback message type to pass service callback in go channels.
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -46,6 +46,10 @@ const (
 	ConnectionAck = DIDExchangeSpec + "ack"
 	// DIDExchangeServiceType is the service type to be used in DID document
 	DIDExchangeServiceType = "did-communication"
+	// ConnectionID connection id is created to retriever connection record from db
+	ConnectionID = "connectionID"
+	// InvitationID invitation id is created in invitation request
+	InvitationID = "invitationID"
 )
 
 // message type to store data for eventing. This is retrieved during callback.
@@ -127,8 +131,11 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 		return fmt.Errorf("invalid state transition: %s -> %s", current.Name(), next.Name())
 	}
 	// trigger message events
-	logger.Infof("send pre event for state %s", next.Name())
-	s.sendMsgEvents(&dispatcher.StateMsg{Type: dispatcher.PreState, Msg: msg, StateID: next.Name()})
+	// TODO change from thread id to connection id #397
+	// TODO pass invitation id #397
+	s.sendMsgEvents(&dispatcher.StateMsg{
+		Type: dispatcher.PreState, Msg: msg, StateID: next.Name(), Properties: s.createEventProperties(thid, "")})
+	logger.Infof("sent pre event for state %s", next.Name())
 
 	// trigger action event based on message type for inbound messages
 	if !msg.Outbound && canTriggerActionEvents(msg.Type) {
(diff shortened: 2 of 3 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -46,6 +46,10 @@ const (
 	ConnectionAck = DIDExchangeSpec + "ack"
 	// DIDExchangeServiceType is the service type to be used in DID document
 	DIDExchangeServiceType = "did-communication"
+	// ConnectionID connection id
```
troyronda: suggest describing the field. Similar for L51

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -204,58 +211,52 @@ func (s *Service) handle(msg *message) error {
...
-			return fmt.Errorf("failed to persist state %s %w", followup.Name(), err)
+		logger.Infof("finish execute next state: %s", next.Name())
+		if err = s.update(msg.ThreadID, next); err != nil {
+			return fmt.Errorf("failed to persist state %s %w", next.Name(), err)
 		}
-		logger.Infof("persisted the connection using %s and updating the state to %s", msg.ThreadID, followup.Name())
-		err = action()
-		if err != nil {
-			return fmt.Errorf("failed to execute state action %s %w", followup.Name(), err)
+		logger.Infof("persisted the connection using %s and updating the state to %s",
```
troyronda: seems the state is already updated at this point?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -204,58 +211,52 @@ func (s *Service) handle(msg *message) error {
...
+			msg.ThreadID, next.Name())
+		if err := action(); err != nil {
+			return fmt.Errorf("failed to execute state action %s %w", next.Name(), err)
 		}
-		logger.Infof("finish execute state action: %s", followup.Name())
+		logger.Infof("finish execute state action: %s", next.Name())
```
troyronda: there are too many individual log messages.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -204,58 +211,52 @@ func (s *Service) handle(msg *message) error {
...
 
-		logger.Infof("send post event for state %s", followup.Name())
-		s.sendMsgEvents(&dispatcher.StateMsg{Type: dispatcher.PostState, Msg: msg.Msg, StateID: followup.Name()})
+		logger.Infof("send post event for state %s", next.Name())
+		// TODO change from thread id to connection id
+		// TODO pass invitation id
```
troyronda: I'm confused by the TODOs. Are the invitation and connection IDs not being passed?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -128,7 +132,10 @@ func (s *Service) Handle(msg dispatcher.DIDCommMsg) error {
 	}
 	// trigger message events
 	logger.Infof("send pre event for state %s", next.Name())
-	s.sendMsgEvents(&dispatcher.StateMsg{Type: dispatcher.PreState, Msg: msg, StateID: next.Name()})
+	// TODO change from thread id to connection id
```
troyronda: Issue number?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -204,58 +211,52 @@ func (s *Service) handle(msg *message) error {
...
+		s.sendMsgEvents(&dispatcher.StateMsg{
+			Type: dispatcher.PostState, Msg: msg.Msg, StateID: next.Name(),
+			Properties: s.createEventProperties(msg.ThreadID, "")})
+		next = followup
 	}
-	logger.Infof("--exited internal handle function--")
 	return nil
 }
 
+func (s *Service) createEventProperties(connectionID, invitationID string) map[string]interface{} { //nolint: unparam
```
llorllale: this method's parameter is named `connectionID` yet I see the `threadID` passed in in all invocations.

pkg/didcomm/dispatcher/api.go
```diff
@@ -47,6 +47,8 @@ type StateMsg struct {
 	Type    StateMsgType
 	StateID string
 	Msg     DIDCommMsg
+	// Properties contains specific keys from different protocols
+	Properties map[string]interface{}
```
llorllale: Suggest we rename this to "Context". The godoc is misleading - in the context of Aries, "keys" usually means "cryptographic keys". Suggest we make this a bit more clear.

pkg/didcomm/protocol/didexchange/service.go
llorllale: do these fields need to be exported?

fqutishat: @llorllale https://github.com/hyperledger/aries-framework-go/pull/422/files#diff-427522b353a28b1ec350c29c79d387b0R27

fqutishat: @troyronda done

pkg/didcomm/protocol/didexchange/service.go
fqutishat: done

pkg/didcomm/dispatcher/api.go
troyronda: +1 on using a term other than keys for the reason given by @llorllale .

I'm not sure, though, that we should use the term context in yet another way :).

pkg/didcomm/protocol/didexchange/service.go
fqutishat: We will change all them to debug after we complete the 0.1.0 demo

pkg/didcomm/protocol/didexchange/service.go
fqutishat: we suppose to pass the connection id and invitation id but issue #397 not read. So to unblock myself I pass thread id as connection id

pkg/didcomm/protocol/didexchange/service.go
fqutishat: done

pkg/didcomm/protocol/didexchange/service.go
fqutishat: we suppose to pass the connection id and invitation id but issue #397 not read. So to unblock myself I pass thread id as connection id

pkg/didcomm/protocol/didexchange/service.go
troyronda: Even when you change to debug, you should have two logger statements in a row ;).

pkg/didcomm/dispatcher/api.go
fqutishat: I change the comments

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -204,58 +211,51 @@ func (s *Service) handle(msg *message) error {
...
+		logger.Infof("finish execute state action: %s", next.Name())
 
-		logger.Infof("send post event for state %s", followup.Name())
-		s.sendMsgEvents(&dispatcher.StateMsg{Type: dispatcher.PostState, Msg: msg.Msg, StateID: followup.Name()})
+		// TODO change from thread id to connection id #397
+		// TODO pass invitation id #397
+		s.sendMsgEvents(&dispatcher.StateMsg{
+			Type: dispatcher.PostState, Msg: msg.Msg, StateID: next.Name(),
+			Properties: s.createEventProperties(msg.ThreadID, "")})
+		logger.Infof("send post event for state %s", next.Name())
```
troyronda: Sent?

pkg/didcomm/protocol/didexchange/service.go
fqutishat: @troyronda I change it

pkg/didcomm/protocol/didexchange/service.go
fqutishat: done

pkg/didcomm/dispatcher/api.go
llorllale: > I'm not sure, though, that we should use the term context in yet another way :).

I was expecting this counterpoint, but this is really what it is. :)

As a matter of fact, we might want to consider using the actual `context.Context` in the future. The timeout/deadline functionality would align nicely with the `~timing` decorator, and with protocol-specific K-Vs stuffed into it, just like in this case.

PR #423, opened 2019-10-04
refactor: split wallet CreateKey into CreateSigningKey & CreateEncryptionKey

Moopli:
Replaced wallet.CreateKey with CreateEncryptionKey, since it's an encryption key (curve25519) and introduced a new CreateSigningKey method that actually creates signing keys (ed25519).

Updated references/usages to use correct keys.

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -84,7 +84,7 @@ func New(ctx provider) (*Client, error) {
 
 // CreateInvitation create invitation
 func (c *Client) CreateInvitation(label string) (*didexchange.Invitation, error) {
-	verKey, err := c.wallet.CreateKey()
+	verKey, err := c.wallet.CreateEncryptionKey()
 	if err != nil {
 		return nil, fmt.Errorf("failed CreateSigningKey: %w", err)
 	}
--- a/pkg/internal/mock/wallet/mock_wallet.go
+++ b/pkg/internal/mock/wallet/mock_wallet.go
@@ -13,15 +13,17 @@ import (
 
 // CloseableWallet mock wallet
 type CloseableWallet struct {
-	CreateSigningKeyValue string
-	CreateSigningKeyErr   error
-	SignMessageValue      []byte
-	SignMessageErr        error
-	PackValue             []byte
-	PackErr               error
-	UnpackValue           *wallet.Envelope
-	UnpackErr             error
-	MockDID               *did.Doc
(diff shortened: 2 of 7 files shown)


Moopli:
@sandrask @Baha-sk

pkg/client/didexchange/client.go
```diff
@@ -84,20 +84,20 @@ func New(ctx provider) (*Client, error) {
...
+	verKey, err := c.wallet.CreateEncryptionKey()
 	if err != nil {
 		return nil, fmt.Errorf("failed CreateSigningKey: %w", err)
 	}
 
 	invitation := &didexchange.Invitation{
 		ID:              uuid.New().String(),
 		Label:           label,
-		RecipientKeys:   []string{verKey},
+		RecipientKeys:   []string{string(verKey)},
```
llorllale: `verkey` is already a string...

pkg/internal/mock/wallet/mock_wallet.go
```diff
@@ -13,24 +13,31 @@ import (
...
 }
 
-// CreateKey create a new public/private signing keypair.
-func (m *CloseableWallet) CreateKey() (string, error) {
+// CreateEncryptionKey create a new public/private encryption keypair.
+func (m *CloseableWallet) CreateEncryptionKey() (wallet.EncryptionPubKey, error) {
+	return m.CreateEncryptionKeyValue, m.CreateEncryptionKeyErr
+}
+
+// CreateSigningKey create a new public/private encryption keypair.
```
sandrask: signing instead of encryption key

pkg/wallet/wallet.go
```diff
@@ -13,6 +13,8 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/crypto/nacl/sign"
```
sandrask: Should we use "crypto/ed25519" to generate key?

pkg/client/didexchange/client.go
troyronda: This does look ... odd :).

Moopli: It is.

Moopli: Ok so - this will need changing as we push more crypto refactoring, yes.

Moopli: We now use strings as identifiers instead of a wrapping type, this is fixed.

Issue #264, opened 2019-09-16
Protocol Service - Unregister option


rolsonquadras:
Based on TODO https://github.com/hyperledger/aries-framework-go/compare/5839b0f509b0932a67abd562fb6a05dc181f2e08..5dfa12128f1ba2867b1a8c7c47464e6f9b8c5b7a#diff-f761c91e753eba1c42a68076df06ce12R135

rolsonquadras:
This has been implemented in the same PR - https://github.com/hyperledger/aries-framework-go/pull/240/files#diff-f761c91e753eba1c42a68076df06ce12R153

PR #425, opened 2019-10-04
feat: Unsecured JWT from Verifiable Credential

### LINKED ISSUE #372: Generation of JWT from Verifiable Credential without the signature

kdimak:
closes #372

As a result, 2 VC Test Suite tests are passing.

As go-jose does not support unsecured JWT serialization, a workaround solution is used. [#264](https://github.com/square/go-jose/issues/264)is filed but was rejected by the authors of the go-jose.

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -332,6 +332,7 @@ type credentialOpts struct {
 	decoders               []CredentialDecoder
 	template               CredentialTemplate
 	issuerPublicKeyFetcher PublicKeyFetcher
+	unsecuredJWT           bool
 }
 
 // CredentialOpt is the Verifiable Credential decoding option
@@ -371,6 +372,13 @@ func WithTemplate(template CredentialTemplate) CredentialOpt {
 func WithJWTPublicKeyFetcher(fetcher PublicKeyFetcher) CredentialOpt {
 	return func(opts *credentialOpts) {
 		opts.issuerPublicKeyFetcher = fetcher
+	}
+}
+
+// WithUnsecuredJWT indicates the need to decode Verifiable Credential from unsecured JWT
+func WithUnsecuredJWT() CredentialOpt {
+	return func(opts *credentialOpts) {
+		opts.unsecuredJWT = true
 	}
 }
 
@@ -465,18 +473,25 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, error) {
 	return cred, nil
 }
 
-func decodeRaw(vcData []byte, crOpts *credentialOpts) (vcDataDecoded []byte, raw *rawCredential, err error) {
+func decodeRaw(vcData []byte, crOpts *credentialOpts) ([]byte, *rawCredential, error) {
 	if crOpts.issuerPublicKeyFetcher != nil {
-		var vcDataFromJwt []byte
-		if vcDataFromJwt, raw, err = decodeJWT(vcData, crOpts.issuerPublicKeyFetcher); err != nil {
-			return nil, nil, fmt.Errorf("JWT decoding failed: %w", err)
+		vcDataFromJwt, rawCred, err := decodeCredJWS(vcData, crOpts.issuerPublicKeyFetcher)
+		if err != nil {
+			return nil, nil, fmt.Errorf("JWS decoding failed: %w", err)
 		}
-
-		return vcDataFromJwt, raw, nil
+		return vcDataFromJwt, rawCred, nil
 	}
 
-	raw = &rawCredential{}
-	err = json.Unmarshal(vcData, raw)
+	if crOpts.unsecuredJWT {
+		rawBytes, rawCred, err := decodeCredJWTUnsecured(vcData)
+		if err != nil {
+			return nil, nil, fmt.Errorf("unsecured JWT decoding failed: %w", err)
+		}
+		return rawBytes, rawCred, nil
+	}
+
+	raw := &rawCredential{}
+	err := json.Unmarshal(vcData, raw)
 	if err != nil {
 		return nil, nil, fmt.Errorf("JSON unmarshalling of verifiable credential failed: %w", err)
 	}
@@ -492,137 +507,6 @@ func loadCredentialSchemas(raw *rawCredential, vcDataDecoded []byte) ([]Credenti
 		return schemas, nil
 	}
 	return []CredentialSchema{}, nil
-}
-
-// credentialClaims is JWT Claims extension by Credential.
-type credentialClaims struct {
-	*jwt.Claims
-
-	Credential *rawCredential `json:"vc,omitempty"`
-}
-
-// rawCredentialClaims is used to get raw content of "vc" claim of JWT.
-type rawCredentialClaims struct {
-	*jwt.Claims
-
-	Raw map[string]interface{} `json:"vc,omitempty"`
-}
-
-func decodeJWT(rawJwt []byte, fetcher PublicKeyFetcher) ([]byte, *rawCredential, error) {
-	parsedJwt, err := jwt.ParseSigned(string(rawJwt))
-	if err != nil {
-		return nil, nil, fmt.Errorf("VC is not JWS: %w", err)
-	}
-
-	credClaims := new(credentialClaims)
-	err = parsedJwt.UnsafeClaimsWithoutVerification(credClaims)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to parse JWT claims: %w", err)
-	}
-
-	if verifyErr := verifyJWTSignature(parsedJwt, fetcher, credClaims); verifyErr != nil {
-		return nil, nil, verifyErr
-	}
-
-	// Apply VC-related claims from JWT.
-	credClaims.refineVCFromJWTClaims()
-
-	// Decode again to get raw content of "vc" claim.
-	rawClaims := new(rawCredentialClaims)
-	err = parsedJwt.UnsafeClaimsWithoutVerification(rawClaims)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to parse JWT claims: %w", err)
-	}
-
-	// Complement original "vc" JSON claim with data refined from JWT claims.
-	if err = rawClaims.mergeRefinedVC(credClaims.Credential); err != nil {
-		return nil, nil, fmt.Errorf("failed to merge refined VC: %w", err)
-	}
-
-	var vcData []byte
-	if vcData, err = json.Marshal(rawClaims.Raw); err != nil {
-		return nil, nil, errors.New("failed to marshal 'vc' claim of JWT")
-	}
-
-	return vcData, credClaims.Credential, nil
-}
-
-func (credClaims *credentialClaims) refineVCFromJWTClaims() {
-	raw := credClaims.Credential
-
-	if iss := credClaims.Issuer; iss != "" {
-		refineVCIssuerFromJWTClaims(raw, iss)
-	}
-
-	if nbf := credClaims.NotBefore; nbf != nil {
-		nbfTime := nbf.Time().UTC()
-		raw.Issued = &nbfTime
-	}
-
-	if jti := credClaims.ID; jti != "" {
-		raw.ID = credClaims.ID
-	}
-
-	if iat := credClaims.IssuedAt; iat != nil {
-		iatTime := iat.Time().UTC()
-		raw.Issued = &iatTime
-	}
-
-	if exp := credClaims.Expiry; exp != nil {
-		expTime := exp.Time().UTC()
-		raw.Expired = &expTime
-	}
-}
-
-func refineVCIssuerFromJWTClaims(raw *rawCredential, iss string) {
-	// Issuer of Verifiable Credential could be either string (id) or struct (with "id" field).
-	switch issuer := raw.Issuer.(type) {
-	case string:
-		raw.Issuer = iss
-	case map[string]interface{}:
-		issuer["id"] = iss
-	}
-}
-
-func (rawClaims *rawCredentialClaims) mergeRefinedVC(raw *rawCredential) error {
-	rawVCClaims := rawClaims.Raw
-
-	rawData, err := json.Marshal(raw)
-	if err != nil {
-		return err
-	}
-
-	var rawMap map[string]interface{}
-
-	err = json.Unmarshal(rawData, &rawMap)
-	if err != nil {
-		return err
-	}
-
-	// make the merge
-	for k, v := range rawMap {
-		rawVCClaims[k] = v
-	}
-
-	return nil
-}
-
-func verifyJWTSignature(parsedJwt *jwt.JSONWebToken, fetcher PublicKeyFetcher, credClaims *credentialClaims) error {
-	var keyID string
-	for _, h := range parsedJwt.Headers {
-		if h.KeyID != "" {
-			keyID = h.KeyID
-			break
-		}
-	}
-	publicKey, err := fetcher(credClaims.Issuer, keyID)
-	if err != nil {
-		return fmt.Errorf("failed to get public key for JWT signature verification: %w", err)
-	}
-	if err = parsedJwt.Claims(publicKey, credClaims); err != nil {
-		return fmt.Errorf("JWT signature verification failed: %w", err)
-	}
-	return nil
 }
 
 func defaultCredentialOpts() *credentialOpts {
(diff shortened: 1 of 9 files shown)


pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	return rawJWT, nil
+}
+
+// Plain builds unsecured JWT.
+// Currently go-jose does not have the plaintext builder. A workaround solution used here is to build JWS
+// and then clean the signature part of the serialized JWT. That's why a private key is needed.
+// Also, typ and alg are set to none as VC spec required.
+// TODO use plain (unsecured) JWT builder when it will be implemented (https://github.com/square/go-jose/issues/264)
```
troyronda: Sounds like we will need a new solution.

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,205 @@
...
+
+// CredentialJWT is JWT Claims extension by Verifiable Credential.
+type CredentialJWT struct {
+	*jwt.Claims
+
+	Credential *rawCredential `json:"vc,omitempty"`
+}
+
+// Signed creates signed JWT (JWS)
+func (vcc *CredentialJWT) Signed(signatureAlg JWTAlgorithm, privateKey interface{}, keyID string) (string, error) {
```
troyronda: I’m not sure about this method name (or this particular abstraction). I see jwt has the same method name with a different signature.

I.e I think a Signed method would take a Signer.

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,205 @@
...
+
+	return rawJWT, nil
+}
+
+// Plain builds unsecured JWT.
+// Currently go-jose does not have the plaintext builder. A workaround solution used here is to build JWS
+// and then clean the signature part of the serialized JWT. That's why a private key is needed.
+// Also, typ and alg are set to none as VC spec required.
+// TODO use plain (unsecured) JWT builder when it will be implemented (https://github.com/square/go-jose/issues/264)
+func (vcc *CredentialJWT) Plain(privateKey interface{}) (string, error) {
```
troyronda: Why is there a privateKey argument? (Even if you are doing a workaround, we shouldn’t have arguments that wouldn’t make sense with the real solution).

pkg/doc/verifiable/credential_jwt.go
kdimak: Yeah, I've pushed a new version where we implement our own unsecured JWT serialization/deserialization. See `jwt_unsecured.go`.

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Signed creates signed JWT (JWS)
+func (vcc *CredentialJWT) Signed(signatureAlg JWTAlgorithm, privateKey interface{}, keyID string) (string, error) {
+	key := jose.SigningKey{Algorithm: signatureAlg.Jose(), Key: privateKey}
+
+	var signerOpts = &jose.SignerOptions{}
+	signerOpts.WithType("JWT")
+	signerOpts.WithHeader("kid", keyID)
+
+	rsaSigner, err := jose.NewSigner(key, signerOpts)
```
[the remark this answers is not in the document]
kdimak: Yes, it should be just `signer` of course.

pkg/doc/verifiable/credential_jwt.go
kdimak: Fixed in a new version where this workaround is not used anymore.

pkg/doc/verifiable/credential_jwt.go
kdimak: Fixed in a new version where this workaround is not used anymore.

pkg/doc/verifiable/credential_jwt.go
kdimak: Maybe `JWS()` would sound better here?

pkg/doc/verifiable/credential_jwt.go
troyronda: Ok - I assume we are waiting for this new version? @kdimak

kdimak: Yes, I've just pushed it.

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func decodeCredJWS(rawJwt []byte, fetcher PublicKeyFetcher) ([]byte, *rawCredential, error) {
+	decoder := &credJWSDecoder{
+		PKFetcher: fetcher,
+	}
+	return decodeCredJWT(rawJwt, decoder)
+}
+
+func decodeCredJWTPlain(rawJwt []byte) ([]byte, *rawCredential, error) {
```
troyronda: I think we should be consistent in naming these methods and structures.

This is an "Unsecured JWS" (rather than "Plain JWT" or "Unsecured JWT")?

kdimak: Agree! I will use "unsecured".

pkg/doc/verifiable/credential.go
```diff
@@ -374,6 +375,13 @@ func WithJWTPublicKeyFetcher(fetcher PublicKeyFetcher) CredentialOpt {
 	}
 }
 
+// WithUnsecuredJWT indicates the need to decode Verifiable Credential from unsecured JWT
```
troyronda: JWT or JWS?

(https://github.com/hyperledger/aries-framework-go/pull/425#discussion_r332511429)

pkg/doc/verifiable/credential.go
```diff
@@ -374,6 +375,13 @@ func WithJWTPublicKeyFetcher(fetcher PublicKeyFetcher) CredentialOpt {
...
+func WithUnsecuredJWT() CredentialOpt {
```
troyronda: I'm wondering if unsecured JWS should be detected in your `decodeRaw` method, and the option is rather about whether or not unsecured JWS are allowed (default to false).

pkg/doc/verifiable/credential.go
kdimak: I'm using JWT (JSON Web Token) term here. And JWS for a signed one. Here we have neither signing nor encryption (JWE), so it's JWT.

pkg/doc/verifiable/credential.go
kdimak: I also thought about this but decided to go the same way as for JWS - a user explicitly defines the way we need to decode VC. And if e.g. JWS decoding fails, we do not try to decode by just JSON unmarshalling (with built-in proof).

pkg/doc/verifiable/credential_jwt.go
kdimak: Corrected.

pkg/doc/verifiable/credential.go
troyronda: > Unsecured JWS
>       A JWS that provides no integrity protection.  Unsecured JWSs use
>       the "alg" value "none".
https://tools.ietf.org/html/rfc7515

Are you setting "alg: none"?

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+	rawJWT, err := builder.CompactSerialize()
+	if err != nil {
+		return "", fmt.Errorf("failed to sign JWT: %w", err)
+	}
+
+	return rawJWT, nil
+}
+
+// SerializeUnsecured serialized JWT into unsecured JWT.
+func (vcc *JWTCredClaims) SerializeUnsecured() (string, error) {
```
troyronda: I think the term Marshal has been used more frequently in this project.

pkg/doc/verifiable/credential.go
kdimak: I think we need to use the term "Unsecured JWT" solely.

pkg/doc/verifiable/credential_jwt.go
kdimak: Renamed.

pkg/doc/verifiable/credential.go
troyronda: Should also update the commit title to remove "(plain)".

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+)
+
+// JWTCredClaims is JWT Claims extension by Verifiable Credential.
+type JWTCredClaims struct {
+	*jwt.Claims
+
+	Credential *rawCredential `json:"vc,omitempty"`
+}
+
+// MarshalSigned marshals JWT into signed form (JWT)
```
troyronda: signed form meaning JWS?

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	return nil
+}
+
+// credJWTDecoder parses JWT claims from serialized token into rawCredential and byte array format.
+// The implementation depends on the type of serialization, e.g. signed (JWT), unsecured (plain JWT).
+type credJWTDecoder interface {
+	parseJWTClaims(rawJwt []byte) (*JWTCredClaims, error)
+	parseJWTRawClaims(rawJwt []byte) (*jwtRawCredClaims, error)
```
troyronda: The word "Raw" doesn't convey the difference between these two functions (to me).

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+
+// JWTCredClaims is JWT Claims extension by Verifiable Credential.
+type JWTCredClaims struct {
+	*jwt.Claims
+
+	Credential *rawCredential `json:"vc,omitempty"`
+}
+
+// MarshalSigned marshals JWT into signed form (JWT)
+func (vcc *JWTCredClaims) MarshalSigned(signatureAlg JWTAlgorithm, privateKey interface{}, keyID string) (string, error) { //nolint:lll
```
troyronda: Do I understand right that MarshalSigned produces a JWS and MarshalUnsecured produces an UnsecuredJWT?

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+		raw.Issued = &iatTime
+	}
+
+	if exp := vcc.Expiry; exp != nil {
+		expTime := exp.Time().UTC()
+		raw.Expired = &expTime
+	}
+}
+
+func decodeCredJWS(rawJwt []byte, fetcher PublicKeyFetcher) ([]byte, *rawCredential, error) {
```
troyronda: Sometimes in your above code, "Signed" is used rather than "JWS".

pkg/doc/verifiable/credential_jwt.go
kdimak: Yes.

pkg/doc/verifiable/credential_jwt.go
kdimak: OK, I will find a better naming.

pkg/doc/verifiable/credential_jwt.go
kdimak: Should be
> MarshalSigned marshals JWT into signed form **(JWS)**

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+
+	return rawJWT, nil
+}
+
+// MarshalUnsecured marshals JWT into unsecured JWT.
+func (vcc *JWTCredClaims) MarshalUnsecured() (string, error) {
+	headers := map[string]string{
+		"alg": "none",
+	}
+	return marshalUnsecuredJWT(headers, vcc)
```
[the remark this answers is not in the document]
kdimak: No, `marshalUnsecuredJWT()` is used only in one place.
As we used own impl of Unsigned JWT marshalling / unmarshalling, I decided to make it more abstract and not dependent on Verifiable Credential.

pkg/doc/verifiable/credential_jwt.go
kdimak: No exact pattern, it's mixed. I will try to improve it.

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+// credJWTDecoder parses JWT claims from serialized token into rawCredential and byte array format.
+// The implementation depends on the type of serialization, e.g. signed (JWT), unsecured (plain JWT).
+type credJWTDecoder interface {
+	parseJWTClaims(rawJwt []byte) (*JWTCredClaims, error)
+	parseJWTRawClaims(rawJwt []byte) (*jwtRawCredClaims, error)
+}
+
+// credJWSDecoder parses and verifies signature of serialized JWT. To verify the signature,
+// Public Key Fetcher is used.
+type credJWSDecoder struct {
```
Baha-sk: I would separate JWS logic from UnsecureJWT logic in separate files ( unless this logic will no grow beyond this PR)

pkg/doc/verifiable/credential.go
troyronda: @Baha-sk see https://github.com/hyperledger/aries-framework-go/pull/425#discussion_r332560468

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,260 @@
...
+	signer, err := jose.NewSigner(key, signerOpts)
+	if err != nil {
+		return "", fmt.Errorf("failed to create signer: %w", err)
+	}
+
+	// create an instance of Builder that uses the rsa signer
+	builder := jwt.Signed(signer).Claims(vcc)
+
+	// validate all ok, sign with the RSA key, and return a compact JWT
+	rawJWT, err := builder.CompactSerialize()
```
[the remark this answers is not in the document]
kdimak: Yes, bad naming. Will fix it.

pkg/doc/verifiable/credential.go
```diff
@@ -465,18 +473,25 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, error) {
 	return cred, nil
 }
 
-func decodeRaw(vcData []byte, crOpts *credentialOpts) (vcDataDecoded []byte, raw *rawCredential, err error) {
+func decodeRaw(vcData []byte, crOpts *credentialOpts) ([]byte, *rawCredential, error) {
 	if crOpts.issuerPublicKeyFetcher != nil {
```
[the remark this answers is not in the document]
kdimak: Agree. I also have doubts about clarity of configurations.
Now it works like this:
- if `issuerPublicKeyFetcher` is defined, try to deserialize JWS, if error, abort
- if `unsecuredJWT` is true, try to deserialize unsecured JWT, if error, abort
- otherwise, unmarshal from JSON.

I created a follow-up issue #445  for this, just because current PR is already too big.

pkg/doc/verifiable/credential_jwt.go
kdimak: Another follow-up issue #446 :)

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	return nil
+}
+
+// credJWTDecoder parses JWT claims from serialized token into JWTCredClaims struct and JSON object.
+// The implementation depends on the type of serialization, e.g. signed (JWT), unsecured (plain JWT).
+type credJWTDecoder interface {
+	CredClaims(rawJwt []byte) (*JWTCredClaims, error)
+	JSONObj(rawJwt []byte) (map[string]interface{}, error)
```
troyronda: I would rename JSONObj.

kdimak: Cannot find a better naming... Could you advise?

pkg/doc/verifiable/credential_jwt_test.go
```diff
@@ -0,0 +1,230 @@
...
+package verifiable
+
+import (
+	"errors"
+	"path/filepath"
+	"testing"
+
+	"github.com/square/go-jose/v3"
+	"github.com/square/go-jose/v3/jwt"
+
```
Moopli: extra line

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -0,0 +1,256 @@
...
+func (vcc *JWTCredClaims) MarshalUnsecuredJWT() (string, error) {
+	headers := map[string]string{
+		"alg": "none",
+	}
+	return marshalUnsecuredJWT(headers, vcc)
+}
+
+// jwtCredObjClaims is used to get generic JSON object content of "vc" claim of JWT.
+type jwtCredObjClaims struct {
+	CredObj map[string]interface{} `json:"vc,omitempty"`
```
troyronda: Not sure what the Obj suffix conveys.

pkg/doc/verifiable/credential_jwt.go
troyronda: I would convey the idea that you are unmarshaling (parsing) the VC field vs the JWT token that has a vc field.

Something like UnmarshalVC and the other is UnmarshalVCToken? Perhaps you have something better :).

kdimak: I have `UnmarshalClaims()` for the first and `UnmarshalVCClaim()` for the second.

pkg/doc/verifiable/credential_jwt.go
kdimak: Right, I'll fix it.

pkg/doc/verifiable/credential.go
kdimak: Yes, thanks for the reminder!

Issue #200, opened 2019-09-03
DID Exchange Inbound Message Events (REST API - webhooks)


rolsonquadras:
Created issue based on discussion in https://wiki.hyperledger.org/display/ARIES/2019-09-03+Framework+Go+Weekly+Planning.

Framework should provide the ability for the RestAPI controller to register events on protocol clients. This issue is related to #198 (clients registering events on the service).

llorllale:
Done in #427

PR #427, opened 2019-10-07
feat: Webhook for aries-agentd

DRK3:
#200 

Aries-agentd now has the ability to send notification via a webhook. Clients are subscribed through a command line argument. Currently the server just sends sample messages every two seconds.

### DIFF
--- a/cmd/aries-agentd/startcmd/start.go
+++ b/cmd/aries-agentd/startcmd/start.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/gorilla/mux"
 	"github.com/spf13/cobra"
@@ -48,14 +47,30 @@ const (
 	// AgentDBPathFlagUsage is the flag usage text for the database path command line argument.
 	AgentDBPathFlagUsage = "Path to database"
 
-	// MissingHostErrorMessage is the error message shown when the user provides a blank host argument.
-	MissingHostErrorMessage = "Unable to start aries agentd, host not provided"
+	// AgentWebhookFlagName is the flag name for the webhook command line argument.
+	AgentWebhookFlagName = "webhook-url"
 
-	// MissingInboundHostErrorMessage is the error message shown when the user provides a blank inbound host argument.
-	MissingInboundHostErrorMessage = "Unable to start aries agentd, HTTP Inbound transport host not provided"
+	// AgentWebhookFlagShorthand is the flag shorthand name for the webhook command line argument.
+	AgentWebhookFlagShorthand = "w"
+
+	// AgentWebhookFlagUsage is the flag usage text for the webhook command line argument.
+	AgentWebhookFlagUsage = "URL to send notifications to." +
+		" This flag can be repeated, allowing for multiple listeners."
 )
 
+// ErrMissingHost is the error when the user provides a blank host argument.
+var ErrMissingHost = errors.New("unable to start aries agentd, host not provided")
+
+// ErrMissingInboundHost is the error when the user provides a blank inbound host argument.
+var ErrMissingInboundHost = errors.New("unable to start aries agentd, HTTP Inbound transport host not provided")
+
 var logger = log.New("aries-framework/agentd")
+
+type agentParameters struct {
+	server                    server
+	host, inboundHost, dbPath string
+	webhookURLs               []string
+}
 
 type server interface {
 	ListenAndServe(host string, router *mux.Router) error
@@ -71,6 +86,7 @@ func (s *HTTPServer) ListenAndServe(host string, router *mux.Router) error {
 
 // Cmd returns the Cobra start command.
 func Cmd(server server) (*cobra.Command, error) {
+	var webhookURLs []string
 	startCmd := &cobra.Command{
 		Use:   "start",
 		Short: "Start an agent",
@@ -92,7 +108,8 @@ func Cmd(server server) (*cobra.Command, error) {
 				return fmt.Errorf("agent DB path flag not found: %s", err)
 			}
 
-			err = startAgent(server, host, inboundHost, dbPath)
+			parameters := &agentParameters{server, host, inboundHost, dbPath, webhookURLs}
+			err = startAgent(parameters)
 			if err != nil {
 				return fmt.Errorf("unable to start agent: %s", err)
 			}
@@ -117,38 +134,48 @@ func Cmd(server server) (*cobra.Command, error) {
 		return nil, fmt.Errorf("tried to mark DB path flag as required but it was not found: %s", err)
 	}
 
+	startCmd.Flags().StringSliceVarP(&webhookURLs, AgentWebhookFlagName, AgentWebhookFlagShorthand, []string{},
+		AgentWebhookFlagUsage)
+	err = startCmd.MarkFlagRequired(AgentWebhookFlagName)
+	if err != nil {
+		return nil, fmt.Errorf("tried to mark agent webhook host flag as required but it was not found: %s", err)
+	}
+
 	return startCmd, nil
 }
 
-func startAgent(server server, host, inboundHost, dbPath string) error {
-	if host == "" {
-		return errors.New(strings.ToLower(MissingHostErrorMessage))
+func startAgent(parameters *agentParameters) error {
+	if parameters.host == "" {
+		return ErrMissingHost
 	}
 
-	if inboundHost == "" {
-		return errors.New(strings.ToLower(MissingInboundHostErrorMessage))
+	if parameters.inboundHost == "" {
+		return ErrMissingInboundHost
 	}
 	var opts []aries.Option
-	opts = append(opts, defaults.WithInboundHTTPAddr(inboundHost))
+	opts = append(opts, defaults.WithInboundHTTPAddr(parameters.inboundHost))
 
-	if dbPath != "" {
-		opts = append(opts, defaults.WithStorePath(dbPath))
+	if parameters.dbPath != "" {
+		opts = append(opts, defaults.WithStorePath(parameters.dbPath))
 	}
 
 	framework, err := aries.New(opts...)
 	if err != nil {
-		return fmt.Errorf("failed to start aries agentd on port [%s], failed to initialize framework :  %w", host, err)
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to initialize framework :  %w",
+			parameters.host, err)
 	}
 
 	ctx, err := framework.Context()
 	if err != nil {
-		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get aries context : %w", host, err)
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get aries context : %w",
+			parameters.host, err)
 	}
 
-	// get all HTTP REST API handlers available for controller A PI
-	restService, err := restapi.New(ctx)
+	// get all HTTP REST API handlers available for controller API
+	restService, err := restapi.New(ctx, parameters.webhookURLs)
 	if err != nil {
-		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get rest service api :  %w", host, err)
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get rest service api :  %w",
+			parameters.host, err)
 	}
 	handlers := restService.GetOperations()
 
@@ -158,12 +185,12 @@ func startAgent(server server, host, inboundHost, dbPath string) error {
 		router.HandleFunc(handler.Path(), handler.Handle()).Methods(handler.Method())
 	}
 
-	logger.Infof("Starting aries agentd on host [%s]", host)
+	logger.Infof("Starting aries agentd on host [%s]", parameters.host)
 
 	// start server on given port and serve using given handlers
-	err = server.ListenAndServe(host, router)
+	err = parameters.server.ListenAndServe(parameters.host, router)
 	if err != nil {
-		return fmt.Errorf("failed to start aries agentd on port [%s], cause:  %w", host, err)
+		return fmt.Errorf("failed to start aries agentd on port [%s], cause:  %w", parameters.host, err)
 	}
 
 	return nil
--- a/cmd/aries-agentd/startcmd/start_test.go
+++ b/cmd/aries-agentd/startcmd/start_test.go
@@ -20,11 +20,12 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/spf13/cobra"
-
 	"github.com/stretchr/testify/require"
 )
 
 type mockServer struct{}
+
+const agentUnexpectedExitErrMsg = "agent server exited unexpectedly"
 
 func (s *mockServer) ListenAndServe(host string, router *mux.Router) error {
 	return nil
@@ -102,8 +103,9 @@ func TestStartAriesDRequests(t *testing.T) {
 	testInboundHostURL := randomURL()
 
 	go func() {
-		err := startAgent(&HTTPServer{}, testHostURL, testInboundHostURL, path)
-		require.NoError(t, err)
+		parameters := &agentParameters{&HTTPServer{}, testHostURL, testInboundHostURL, path, []string{}}
+		err := startAgent(parameters)
+		require.FailNow(t, agentUnexpectedExitErrMsg+": "+err.Error())
 	}()
 
 	waitForServerToStart(t, testHostURL, testInboundHostURL)
@@ -111,8 +113,8 @@ func TestStartAriesDRequests(t *testing.T) {
 	validateRequests(t, testHostURL, testInboundHostURL)
 }
 
-func listenFor(host string, d time.Duration) error {
-	timeout := time.After(d)
+func listenFor(host string) error {
+	timeout := time.After(2 * time.Second)
 	for {
 		select {
 		case <-timeout:
@@ -158,7 +160,7 @@ func validateRequests(t *testing.T, testHostURL, testInboundHostURL string) {
 		// DIDComm inbound API test
 		{
 			name: "200: testing didcomm inbound",
-			r: newreq("POST",
+			r: newreq(http.MethodPost,
 				fmt.Sprintf("http://%s", testInboundHostURL),
 				strings.NewReader(`
 							{
@@ -205,7 +207,8 @@ func isJSON(res []byte) bool {
 func TestStartCmdWithMissingHostArg(t *testing.T) {
 	startCmd, err := Cmd(&mockServer{})
 	require.NoError(t, err)
-	args := []string{"--" + AgentInboundHostFlagName, randomURL(), "--" + AgentDBPathFlagName, ""}
+	args := []string{"--" + AgentInboundHostFlagName, randomURL(), "--" + AgentDBPathFlagName, "",
+		"--" + AgentWebhookFlagName, ""}
 	startCmd.SetArgs(args)
 
 	err = startCmd.Execute()
@@ -215,16 +218,18 @@ func TestStartCmdWithMissingHostArg(t *testing.T) {
 }
 
 func TestStartAgentWithBlankHost(t *testing.T) {
-	err := startAgent(&mockServer{}, "", randomURL(), "")
+	parameters := &agentParameters{&mockServer{}, "", randomURL(), "", []string{}}
+	err := startAgent(parameters)
 
 	require.NotNil(t, err)
-	require.Equal(t, strings.ToLower(MissingHostErrorMessage), err.Error())
+	require.Equal(t, ErrMissingHost, err)
 }
 
 func TestStartCmdWithoutInboundHostArg(t *testing.T) {
 	startCmd, err := Cmd(&mockServer{})
 	require.NoError(t, err)
-	args := []string{"--" + AgentHostFlagName, randomURL(), "--" + AgentDBPathFlagName, ""}
+	args := []string{"--" + AgentHostFlagName, randomURL(), "--" + AgentDBPathFlagName, "",
+		"--" + AgentWebhookFlagName, ""}
 	startCmd.SetArgs(args)
 
 	err = startCmd.Execute()
@@ -234,10 +239,11 @@ func TestStartCmdWithoutInboundHostArg(t *testing.T) {
 }
 
 func TestStartAgentWithBlankInboundHost(t *testing.T) {
-	err := startAgent(&mockServer{}, randomURL(), "", "")
+	parameters := &agentParameters{&mockServer{}, randomURL(), "", "", []string{}}
+	err := startAgent(parameters)
 
 	require.NotNil(t, err)
-	require.Equal(t, strings.ToLower(MissingInboundHostErrorMessage), err.Error())
+	require.Equal(t, ErrMissingInboundHost, err)
 	if err == nil {
 		t.Fatal()
 	}
@@ -246,7 +252,8 @@ func TestStartAgentWithBlankInboundHost(t *testing.T) {
 func TestStartCmdWithoutDBPath(t *testing.T) {
 	startCmd, err := Cmd(&mockServer{})
 	require.NoError(t, err)
-	args := []string{"--" + AgentHostFlagName, randomURL(), "--" + AgentInboundHostFlagName, randomURL()}
+	args := []string{"--" + AgentHostFlagName, randomURL(), "--" + AgentInboundHostFlagName, randomURL(),
+		"--" + AgentWebhookFlagName, ""}
 	startCmd.SetArgs(args)
 
 	err = startCmd.Execute()
@@ -261,14 +268,15 @@ func TestStartCmdMissingAllArgs(t *testing.T) {
 
 	err = startCmd.Execute()
 	require.NotNil(t, err)
-	require.Equal(t, `required flag(s) "api-host", "db-path", "inbound-host" not set`, err.Error())
+	require.Equal(t,
+		`required flag(s) "api-host", "db-path", "inbound-host", "webhook-url" not set`, err.Error())
 }
 
 func TestStartCmdValidArgs(t *testing.T) {
 	startCmd, err := Cmd(&mockServer{})
 	require.NoError(t, err)
 	args := []string{"--" + AgentHostFlagName, randomURL(), "--" + AgentInboundHostFlagName,
-		randomURL(), "--" + AgentDBPathFlagName, ""}
+		randomURL(), "--" + AgentDBPathFlagName, "", "--" + AgentWebhookFlagName, ""}
 	startCmd.SetArgs(args)
 
 	err = startCmd.Execute()
@@ -284,15 +292,17 @@ func TestStartMultipleAgentsWithSameHost(t *testing.T) {
 	path1, cleanup1 := generateTempDir(t)
 	defer cleanup1()
 	go func() {
-		err := startAgent(&HTTPServer{}, host, inboundHost, path1)
-		require.NoError(t, err)
+		parameters := &agentParameters{&HTTPServer{}, host, inboundHost, path1, []string{}}
+		err := startAgent(parameters)
+		require.FailNow(t, agentUnexpectedExitErrMsg+": "+err.Error())
 	}()
 
 	waitForServerToStart(t, host, inboundHost)
 
 	path2, cleanup2 := generateTempDir(t)
 	defer cleanup2()
-	err := startAgent(&HTTPServer{}, host, inboundHost2, path2)
+	parameters := &agentParameters{&HTTPServer{}, host, inboundHost2, path2, []string{}}
+	err := startAgent(parameters)
 
 	require.NotNil(t, err)
 	addressAlreadyInUseErrorMessage := "failed to start aries agentd on port [" + host +
@@ -309,23 +319,25 @@ func TestStartMultipleAgentsWithSameDBPath(t *testing.T) {
 	defer cleanup()
 
 	go func() {
-		err := startAgent(&HTTPServer{}, host1, inboundHost1, path)
-		require.NoError(t, err)
+		parameters := &agentParameters{&HTTPServer{}, host1, inboundHost1, path, []string{}}
+		err := startAgent(parameters)
+		require.FailNow(t, agentUnexpectedExitErrMsg+": "+err.Error())
 	}()
 
 	waitForServerToStart(t, host1, inboundHost1)
 
-	err := startAgent(&HTTPServer{}, host2, inboundHost2, path)
+	parameters := &agentParameters{&HTTPServer{}, host2, inboundHost2, path, []string{}}
+	err := startAgent(parameters)
 
 	require.NotNil(t, err)
 	require.Contains(t, err.Error(), "storage initialization failed")
 }
 
 func waitForServerToStart(t *testing.T, host, inboundHost string) {
-	if err := listenFor(host, 2*time.Second); err != nil {
+	if err := listenFor(host); err != nil {
 		t.Fatal(err)
 	}
-	if err := listenFor(inboundHost, 2*time.Second); err != nil {
+	if err := listenFor(inboundHost); err != nil {
 		t.Fatal(err)
 	}
 }
--- a/pkg/restapi/restapi.go
+++ b/pkg/restapi/restapi.go
@@ -10,12 +10,14 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/framework/context"
 	"github.com/hyperledger/aries-framework-go/pkg/restapi/operation"
 	"github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/restapi/webhook"
 )
 
 // New returns new controller REST API instance.
 //
 // TODO: Allow customized operations.
-func New(ctx *context.Provider) (*Controller, error) {
+// TODO: Make webhookURLs an optional parameter (#472)
+func New(ctx *context.Provider, webhookURLs []string) (*Controller, error) {
 	var allHandlers []operation.Handler
 
 	// Add DID Exchange Rest Handlers
@@ -25,6 +27,8 @@ func New(ctx *context.Provider) (*Controller, error) {
 	}
 
 	allHandlers = append(allHandlers, exchange.GetRESTHandlers()...)
+
+	webhook.StartWebhookDispatcher(webhookURLs)
 
 	return &Controller{handlers: allHandlers}, nil
 }
--- a/pkg/restapi/webhook/webhook.go
+++ b/pkg/restapi/webhook/webhook.go
@@ -1,0 +1,81 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package webhook
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+const (
+	notificationInterval    = 2 * time.Second
+	notificationSendTimeout = 10 * time.Second
+)
+
+var logger = log.New("aries-framework/agentd")
+
+// StartWebhookDispatcher will launch the webhook dispatcher in a new goroutine.
+// The dispatcher is responsible for sending notifications to subscribers.
+func StartWebhookDispatcher(subscriberURLs []string) {
+	if len(subscriberURLs) == 0 {
+		logger.Warnf("No subscriber URLs provided. Webhook dispatcher will not start.")
+		return
+	}
+
+	go dispatch(subscriberURLs)
+}
+
+func dispatch(subscriberURLs []string) {
+	// TODO: Data should should be pushed whenever a record is created or its state property is updated  (see #472).
+	ticker := time.NewTicker(notificationInterval)
+	defer ticker.Stop()
+	for range ticker.C {
+		for _, subscriberURL := range subscriberURLs {
+			err := sendNotification(subscriberURL)
+			if err != nil {
+				logger.Errorf(err.Error())
+			}
+		}
+	}
+}
+
+func sendNotification(destination string) error {
+	//TODO: This HTTP post request needs to supply real data, not this sample text (see #472).
+	ctx, cancel := context.WithTimeout(context.Background(), notificationSendTimeout)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, destination,
+		bytes.NewBufferString("Sample notification from aries-agentd."))
+	if err != nil {
+		return fmt.Errorf("failed to create new http post request for %s: %s", destination, err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to post notification to %s: %s", destination, err)
+	}
+
+	defer closeResponse(resp.Body)
+
+	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusCreated {
+		logger.Infof("Notification sent to %s successfully. \n", destination)
+		return nil
+	}
+	return fmt.Errorf("notification was sent to %s, but %s was received",
+		destination, resp.Status)
+}
+
+func closeResponse(c io.Closer) {
+	err := c.Close()
+	if err != nil {
+		logger.Errorf("Failed to close response body")
+	}
+}
--- a/pkg/restapi/webhook/webhook_test.go
+++ b/pkg/restapi/webhook/webhook_test.go
@@ -1,0 +1,239 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package webhook
+
+import (
+	"context"
+	"errors"
+	"fmt"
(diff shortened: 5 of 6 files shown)


DRK3:
Not sure how to meaningfully increase code coverage further. The uncovered areas are in goroutines so while I know I can make them execute to satisfy the code coverage checker, I don't have any value to check against, so it wouldn't really be a genuine test.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -116,11 +132,16 @@ func Cmd(server server) (*cobra.Command, error) {
...
+	err = startCmd.MarkFlagRequired(AgentWebhookHostFlagName)
+	if err != nil {
+		return nil, fmt.Errorf("tried to mark agent webhook host flag as required but it was not found: %s", err)
+	}
 
 	return startCmd, nil
 }
 
-func startAgent(server server, host, inboundHost, dbPath string) error {
+func startAgent(server server, host, inboundHost, dbPath, webhookHost string) error {
```
troyronda: @sudeshrshetty @DRK3 suggest creating a struct - the argument count is getting high.

cmd/aries-agentd/startcmd/webhook/webhook.go
```diff
@@ -0,0 +1,119 @@
+/*
```
troyronda: @sudeshrshetty @DRK3 Is this package intended to be not re-usable? (it's in the cmd/ path rather than the pkg/ path).

cmd/aries-agentd/startcmd/testhelper/testhelper.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"errors"
+	"net"
+	"time"
+)
+
+// WaitForServerToStart stalls until the specified host is successfully contacted.
+// An error is returned if no response is received after the given duration is exceeded.
+func WaitForServerToStart(host string, maxTimeToWait time.Duration) error {
```
llorllale: This function is erroneously included in the coverage report because the filename doesn't end with "test"

DRK3:
Note to self: try using the randomURL() function in start_test.go instead of hardcoding my host names. I believe it may be possible for my hardcoded names to collide with the randomly generated URLs in the cases where I launch a server in a new goroutine, as the goroutines do not get killed when the test ends (which is the original reason why I have so many different hardcoded URLs)

cmd/aries-agentd/startcmd/webhook/webhook.go
```diff
@@ -0,0 +1,119 @@
...
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package webhook
```
llorllale: @DRK3 did you evaluate a 3rd party library for websockets? Eg. gorilla/websocket

cmd/aries-agentd/startcmd/start.go
```diff
@@ -168,3 +193,12 @@ func startAgent(server server, host, inboundHost, dbPath string) error {
 
 	return nil
 }
+
+func startWebhookServer(webhookHost string) {
+	go func() {
+		webhookServerErr := webhook.StartWebhookServer(webhookHost, "/webhook")
```
llorllale: ACA-Py's websocket handler is registered on the `/ws` endpoint

cmd/aries-agentd/startcmd/start.go
```diff
@@ -168,3 +193,12 @@ func startAgent(server server, host, inboundHost, dbPath string) error {
 
 	return nil
 }
+
+func startWebhookServer(webhookHost string) {
```
rolsonquadras: Should we call this as webhook server or controller server ? @troyronda @llorllale

cmd/aries-agentd/startcmd/start.go
DRK3: Done.

cmd/aries-agentd/startcmd/webhook/webhook.go
DRK3: Moved it to pfk/restapi/webhook.

cmd/aries-agentd/startcmd/testhelper/testhelper.go
DRK3: @llorllale Some background info: That WaitForServerToStart method used to be in start_test, but I wanted to use it in another package as well so I moved it to a common place. Doing that caused it to be a part of code coverage, so I added a unit test to it to help mitigate that. Unfortunately it seems like the go compiler won't let me import a file that ends in "test". Is there another solution to this problem? The only other thing I can think of would be to duplicate the WaitForServerToStart code in each package where it's used, but that's ugly as well.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -11,6 +11,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/hyperledger/aries-framework-go/pkg/restapi/webhook"
```
sudeshrshetty: fix import order

cmd/aries-agentd/startcmd/testhelper/testhelper.go
sudeshrshetty: move testhelper to internal package,
if testhelper is used in unit-tests then it should have been covered in test coverage.

Refer: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/internal/common/logging/modlog/logtestutils.go

cmd/aries-agentd/startcmd/testhelper/testhelper_test.go
```diff
@@ -0,0 +1,18 @@
...
+package testhelper
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestTimeout(t *testing.T) {
```
sudeshrshetty: may be bit more explicit test name?

pkg/restapi/webhook/samplewebhookclient.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"net/http"
+)
+
+type sampleWebhookClient struct {
+	addr           string
+	handlerPattern string
+}
+
+func (sampleWebhookClient *sampleWebhookClient) listen() error {
```
sudeshrshetty: names may conflict --> so use, func (s *sampleWebhookClient)

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,119 @@
...
+
+type webhookRequest struct {
+	Name        string
+	Destination string
+}
+
+var logger = log.New("aries-framework/agentd")
+
+// UnexpectedExitErrMsg is the string to be displayed when the webhook server exits unexpectedly.
+const UnexpectedExitErrMsg = "webhook server exited unexpectedly"
```
sudeshrshetty: error constant?

var UnexpectedExitErr = errors.New("webhook server exited unexpectedly")

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (dispatcher *dispatcher) start() {
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+	for range ticker.C {
+		dispatcher.dispatch()
+	}
+}
+func (dispatcher *dispatcher) addSubscriber(name, subscriberURL string) {
```
sudeshrshetty: to improve readability -->use  func (d *dispatcher

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,119 @@
...
+		logger.Errorf("Error adding new subscriber: %s", err)
+		return
+	}
+	dispatcher.addSubscriber(request.Name, request.Destination)
+}
+
+type dispatcher struct {
+	client         *http.Client
+	subscriberURLs map[string]string
+	mu             *sync.Mutex
```
sudeshrshetty: Use sync.RWMutex,

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,119 @@
...
+		dispatcher.dispatch()
+	}
+}
+func (dispatcher *dispatcher) addSubscriber(name, subscriberURL string) {
+	dispatcher.mu.Lock()
+	dispatcher.subscriberURLs[name] = subscriberURL
+	dispatcher.mu.Unlock()
+}
+func (dispatcher *dispatcher) dispatch() {
+	dispatcher.mu.Lock()
```
sudeshrshetty: use read lock here with rwmutex here

cmd/aries-agentd/startcmd/webhook/webhook.go
llorllale: @DRK3 and I investigated this further yesterday:

ACA-Py...
* Can notify events through webhooks and websockets
   * If both webhooks and websockets are registered then all events are sent to both channels
* Only registers webhooks via the command line parameter `--webhook-url`
* Registers websockets when the client does HTTP GET of `/ws`
* Uses the websocket channel for one purpose only - to send events to the client. ACA-Py does not listen for incoming commands via this channel.

cmd/aries-agentd/startcmd/start.go
DRK3: Will create a separate PR for websocket support. I'll be sure to register it on the /ws endpoint.

cmd/aries-agentd/startcmd/start.go
DRK3: just spoke to @llorllale to confirm. This isn't the controller, just the actual webhook dispatcher. I'm going to be reworking this code soon, perhaps I can give it a clearer name.

pkg/restapi/webhook/webhook.go
troyronda: note error variables are prefixed with Err

sudeshrshetty: ErrUnexpectedExit

cmd/aries-agentd/startcmd/testhelper/testhelper.go
DRK3: Talked to @sudeshrshetty, can't make it an internal package since one of the packages using it is not within the same tree rooted at the parent of the internal directory. So instead I am moving it to pkg/restapi/helper

cmd/aries-agentd/startcmd/start.go
DRK3: Done.

pkg/restapi/webhook/samplewebhookclient.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: I made it a var and changed it to ErrUnexpectedExitMsg, but I can't change it to an error type since I need it to be a string.

cmd/aries-agentd/startcmd/testhelper/testhelper_test.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: Done!

cmd/aries-agentd/startcmd/webhook/webhook.go
llorllale: Documentation has begun: https://hackmd.io/4MYrV-f3TYiwMKvIVgMn6A?view

I suggest watching that document and also for any PRs to the ACA-Py repo that add this docu.

pkg/restapi/webhook/webhook.go
DRK3: With my new changes, this var isn't used anymore, and so I have deleted it.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -48,6 +49,15 @@ const (
...
+	AgentWebhookHostFlagName = "webhook-url"
+
+	// AgentWebhookHostFlagShorthand is the flag shorthand name for the webhook host command line argument.
+	AgentWebhookHostFlagShorthand = "w"
+
+	// AgentWebhookHostFlagUsage is the flag usage text for the webhook host command line argument.
+	AgentWebhookHostFlagUsage = "Webhook Host Name:Port. Leave blank to prevent creation of a webhook server."
+
 	// MissingHostErrorMessage is the error message shown when the user provides a blank host argument.
 	MissingHostErrorMessage = "Unable to start aries agentd, host not provided"
```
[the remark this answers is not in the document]
DRK3: I think this comment is out of date now - the variable you were talking about has been removed. Let me know if this is about something else.

pkg/restapi/helper/helper.go
```diff
@@ -0,0 +1,43 @@
+/*
```
DRK3: Note that these were existing functions - I just moved them from start_test.go to a common location since I use them in multiple packages now.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -11,13 +11,14 @@ import (
...
-	"github.com/gorilla/mux"
-	"github.com/spf13/cobra"
-
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/defaults"
 	"github.com/hyperledger/aries-framework-go/pkg/restapi"
+	"github.com/hyperledger/aries-framework-go/pkg/restapi/webhook"
+
+	"github.com/gorilla/mux"
```
sudeshrshetty: these 2 imports should be before `github.com/hyperledger/` imports

cmd/aries-agentd/startcmd/start.go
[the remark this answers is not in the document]
sudeshrshetty: `MissingHostErrorMessage` may be unrelated to your current change.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -117,38 +136,48 @@ func Cmd(server server) (*cobra.Command, error) {
...
+		return nil, fmt.Errorf("tried to mark agent webhook host flag as required but it was not found: %s", err)
+	}
+
 	return startCmd, nil
 }
 
-func startAgent(server server, host, inboundHost, dbPath string) error {
-	if host == "" {
+func startAgent(parameters *agentParameters) error {
+	if parameters.host == "" {
```
sudeshrshetty: May be unrelated to your current change.
Nice to have var error for (MissingHostErrorMessage string), than doing errors.new + to lower everytime

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -18,49 +18,21 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/restapi/helper"
+
```
sudeshrshetty: import order

pkg/restapi/webhook/samplewebhookclient.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (s *sampleWebhookClient) listenAndStopAfterReceivingNotification() error {
+	srv := &http.Server{Addr: s.addr, Handler: http.DefaultServeMux}
+	http.HandleFunc(s.handlerPattern, func(resp http.ResponseWriter, req *http.Request) {
+		logger.Infof("Just received webhook notification")
+
+		err := srv.Shutdown(context.Background())
+		if err != nil {
+			logger.Fatalf("Failed to shutdown sample webhook client server: %s", err)
```
sudeshrshetty: Are you sure of using Fatal here? it calls os.Exit(1), be careful with it.

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+var logger = log.New("aries-framework/agentd")
+
+type dispatcher struct {
+	client         *http.Client
+	subscriberURLs []string
+	mu             *sync.RWMutex
```
sudeshrshetty: rwmutex pointers are only to share the mutex between goroutines/functions, so no need of pointer here.

use `mu             sync.RWMutex`

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// StartWebhookDispatcher will launch the webhook dispatcher in a new goroutine.
+// The dispatcher is responsible for sending notifications to subscribers.
+func StartWebhookDispatcher(subscriberURLs []string) {
+	dispatcher := &dispatcher{
+		client:         &http.Client{},
+		subscriberURLs: subscriberURLs,
+		mu:             &sync.RWMutex{},
```
sudeshrshetty: The mutexes are generally designed to work without any type of initialization. Remove line #32 once change it to non pointer type.

cmd/aries-agentd/startcmd/start.go
DRK3: Done!

cmd/aries-agentd/startcmd/start.go
DRK3: Done!

cmd/aries-agentd/startcmd/start_test.go
DRK3: I moved the hyperledger import below the third party libraries. Was that what you were expecting?

pkg/restapi/webhook/samplewebhookclient.go
DRK3: Oops, meant to use Errorf. Fixed!

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+var logger = log.New("aries-framework/agentd")
+
+type dispatcher struct {
+	client         *http.Client
+	subscriberURLs []string
+	mu             sync.RWMutex
```
sudeshrshetty: Looks like you don't need lock anymore

DRK3: Good catch! Fixed.

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+	for range ticker.C {
+		d.dispatch()
+	}
+}
+
+func (d *dispatcher) dispatch() {
+	for _, subscriberURL := range d.subscriberURLs {
+		go func(destination string) {
+			req, err := http.NewRequest("POST", destination,
+				bytes.NewBufferString(fmt.Sprintf("Sample notification from aries-agentd.")))
```
sudeshrshetty: You are not formatting anything here, use simple string

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+	dispatcher := &dispatcher{
+		client:         &http.Client{},
+		subscriberURLs: subscriberURLs,
+	}
+
+	go dispatcher.start()
+}
+
+func (d *dispatcher) start() {
+	ticker := time.NewTicker(2 * time.Second)
```
sudeshrshetty: constant may be, in long run may be we need to configure it.

pkg/restapi/webhook/webhook.go
DRK3: Fixed!

pkg/restapi/webhook/webhook.go
DRK3: Done

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -20,47 +20,19 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/spf13/cobra"
-
 	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/restapi/helper"
```
troyronda: helper is not a descriptive package name.

pkg/restapi/helper/helper.go
troyronda: This should be under a restapi/internal/<pkgname> path. Are these test utilities for the REST packages?

DRK3: Since I can't move it to an internal package due to it being used in different packages where the parent folders don't share a common path, @sudeshrshetty let me know that you suggested that I should instead duplicated those test utility functions in each test file. That's done.

cmd/aries-agentd/startcmd/start_test.go
DRK3: This file is gone now, see my other comment in the thread regarding internal package

llorllale:
@DRK3 please reference the relevant issue in this PR's description

cmd/aries-agentd/startcmd/webhook/webhook.go
llorllale: @DRK3 they've documented the webhook functionality in this PR: https://github.com/hyperledger/aries-cloudagent-python/pull/212/files

Interestingly, they do not mention the websocket functionality...

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,78 @@
...
+	ticker := time.NewTicker(notificationInterval)
+	defer ticker.Stop()
+	for range ticker.C {
+		d.dispatch()
+	}
+}
+
+func (d *dispatcher) dispatch() {
+	for _, subscriberURL := range d.subscriberURLs {
+		go func(subscriberURL string) {
```
llorllale: Rather than running a go routine per webhook url, it's better to have just one go routine to dispatch all messages to all the webhook urls. The ticker from `start()` should be moved into this go routine's loop.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -48,15 +48,31 @@ const (
...
+	// AgentWebhookFlagUsage is the flag usage text for the webhook command line argument.
+	AgentWebhookFlagUsage = "URL to send notifications to." +
+		" This flag can be repeated, allowing for multiple listeners."
 )
 
+// ErrMissingHost is the error when the user provides a blank host argument.
+var ErrMissingHost = errors.New("unable to start aries agentd, host not provided")
+
+// ErrMissingInboundHost is the error when the user provides a blank inbound host argument.
+var ErrMissingInboundHost = errors.New("unable to start aries agentd, HTTP Inbound transport host not provided")
```
llorllale: It would be great if we took this opportunity to fix this documentation: this "inbound host" is really the listen address for the DIDComm service served over the HTTP transport.

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,78 @@
...
+
+func sendNotification(destination string, c *http.Client) error {
+	//TODO: This HTTP post request needs to supply real data, not this sample text (see #472).
+	req, err := http.NewRequest("POST", destination,
+		bytes.NewBufferString("Sample notification from aries-agentd."))
+	if err != nil {
+		return fmt.Errorf("failed to create new http post request for %s: %s", destination, err)
+	}
+
+	resp, err := c.Do(req)
```
DRK3: Note to self: add a timeout. Golang's default is to just wait forever if no timeout is specified.

pkg/restapi/webhook/webhook.go
troyronda: We should not spinup a goroutine for each.

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,78 @@
...
+			if err != nil {
+				logger.Errorf(err.Error())
+			}
+		}(subscriberURL)
+	}
+}
+
+func sendNotification(destination string, c *http.Client) error {
+	//TODO: This HTTP post request needs to supply real data, not this sample text (see #472).
+	req, err := http.NewRequest("POST", destination,
```
troyronda: Use https://golang.org/pkg/net/http/#NewRequestWithContext and create a context with timeout.

cmd/aries-agentd/startcmd/start.go
DRK3: What should be changed exactly?

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: Done.

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,78 @@
...
+	dispatcher := &dispatcher{
+		client:         &http.Client{},
+		subscriberURLs: subscriberURLs,
+	}
+
+	dispatcher.start()
+}
+
+func (d *dispatcher) start() {
+	go func() {
```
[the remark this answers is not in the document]
DRK3: Done!

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,78 @@
...
+		ticker := time.NewTicker(notificationInterval)
```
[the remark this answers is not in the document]
DRK3: Can you elaborate on that? Not quite sure what you mean.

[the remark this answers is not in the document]
troyronda: I'm not sure what is the intended usage of the notificationInterval (and ticker) once you have "real" data. Are you queuing the "real" data and only sending on intervals? (or will you send "real" data as you get them?)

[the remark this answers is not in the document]
DRK3: Oh I see - that ticker is just temporary. It just allows me to demonstrate the webhook behaviour. According to the ACA-Py webhook documentation: "state-management records are sent to the provided URL via POST requests whenever a record is created or its state property is updated."

So in #472 I will get rid of the ticker and make it match ACA-Py with push notifications.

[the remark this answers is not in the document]
DRK3: I added a TODO above the ticker code that references #472.

pkg/restapi/webhook/samplewebhookclient.go
```diff
@@ -0,0 +1,47 @@
+/*
```
DRK3: Move this inside the webhook test

DRK3: Try adding _test after the name first

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+	//TODO: This HTTP post request needs to supply real data, not this sample text (see #472).
+	ctx, cancel := context.WithTimeout(context.Background(), notificationSendTimeout)
+	req, err := http.NewRequestWithContext(ctx, "POST", destination,
+		bytes.NewBufferString("Sample notification from aries-agentd."))
+	if err != nil {
+		cancel()
+		return fmt.Errorf("failed to create new http post request for %s: %s", destination, err)
+	}
+
+	resp, err := c.Do(req)
```
DRK3: Check http error codes

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+		err := sendNotification(subscriberURL, d.client)
+		if err != nil {
+			logger.Errorf(err.Error())
+		}
+	}
+}
+
+func sendNotification(destination string, c *http.Client) error {
+	//TODO: This HTTP post request needs to supply real data, not this sample text (see #472).
+	ctx, cancel := context.WithTimeout(context.Background(), notificationSendTimeout)
```
DRK3: defer cancel() instead of having it everywhere here

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+	resp, err := c.Do(req)
+	if err != nil {
+		cancel()
+		return fmt.Errorf("failed to post notification to %s: %s", destination, err)
+	}
+
+	defer func() {
+		e := resp.Body.Close()
+		if e != nil {
+			panic(err)
```
DRK3: change this to log error

pkg/restapi/webhook/webhook_test.go
```diff
@@ -0,0 +1,142 @@
...
+		} else {
+			close(subscriberReceivedNotification)
+		}
+	}()
+
+	if err := listenFor(clientHost, 2*time.Second); err != nil {
+		t.Fatal(err)
+	}
+
+	StartWebhookDispatcher([]string{"http://" + clientHost + clientHandlerPattern})
```
DRK3: use fmt.sprintf()

pkg/restapi/webhook/samplewebhookclient.go
DRK3: jsut made it a static function within testing file

cmd/aries-agentd/startcmd/start.go
```diff
@@ -158,12 +186,14 @@ func startAgent(server server, host, inboundHost, dbPath string) error {
 		router.HandleFunc(handler.Path(), handler.Handle()).Methods(handler.Method())
 	}
 
-	logger.Infof("Starting aries agentd on host [%s]", host)
+	logger.Infof("Starting aries agentd on host [%s]", parameters.host)
+
+	webhook.StartWebhookDispatcher(parameters.webhookURLs)
```
DRK3: move this inside restapi.New()

If no webhook urls provided, print warning and don't start dispatcher.

Ask George/Troy if it should be a mandatory argument. @llorllale @troyronda

pkg/restapi/webhook/webhook_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"fmt"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWebhookDispatcherSendNotification(t *testing.T) {
```
DRK3: Do a test with multiple URLs

pkg/restapi/webhook/webhook_test.go
```diff
@@ -0,0 +1,142 @@
...
+func TestWebhookNotificationMalformedURL(t *testing.T) {
+	err := sendNotification("%", &http.Client{})
+	require.Equal(t, `failed to create new http post request for %: parse %: invalid URL escape "%"`, err.Error())
+}
+
+func TestWebhookNotificationClientURLNoResponse(t *testing.T) {
+	expectedErrMessage := `failed to post notification to http://localhost:8080: ` +
+		`Post http://localhost:8080: dial tcp 127.0.0.1:8080: connect: connection refused`
+	// One of the test machine produces a slightly different error message
+	alternateExpectedErrMessage := `failed to post notification to http://localhost:8080: ` +
```
DRK3: do require.contains(the important part of the message)

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -333,6 +326,25 @@ func waitForServerToStart(t *testing.T, host, inboundHost string) {
...
+			return errors.New("timeout: server is not available")
+		default:
+			conn, err := net.Dial("tcp", host)
+			if err != nil {
+				continue
+			}
+			if err := conn.Close(); err != nil {
+				return err
+			}
+			return nil
```
soluchok: please simplify it `return conn.Close()`

cmd/aries-agentd/startcmd/start.go
DRK3: First two sentences: done.

cmd/aries-agentd/startcmd/start_test.go
DRK3: Done!

pkg/restapi/webhook/samplewebhookclient.go
DRK3: Done

pkg/restapi/webhook/webhook.go
DRK3: Done

pkg/restapi/webhook/webhook.go
DRK3: Done

pkg/restapi/webhook/webhook.go
DRK3: Done. @sudeshrshetty Let me know what you think. Note that the error is logged in dispatch(), which calls sendNotification.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -117,38 +134,48 @@ func Cmd(server server) (*cobra.Command, error) {
...
 	if err != nil {
-		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get aries context : %w", host, err)
+		return fmt.Errorf("failed to start aries agentd on port [%s], failed to get aries context : %w",
+			parameters.host, err)
 	}
 
-	// get all HTTP REST API handlers available for controller A PI
-	restService, err := restapi.New(ctx)
+	// get all HTTP REST API handlers available for controller API
+	restService, err := restapi.New(ctx, parameters.webhookURLs)
```
sudeshrshetty: Use optional parameter to pass webhooks, add a TODO and push it in next PR.

pkg/restapi/restapi.go
```diff
@@ -10,12 +10,13 @@ import (
...
 	"github.com/hyperledger/aries-framework-go/pkg/restapi/operation"
 	"github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/restapi/webhook"
 )
 
 // New returns new controller REST API instance.
 //
 // TODO: Allow customized operations.
-func New(ctx *context.Provider) (*Controller, error) {
+func New(ctx *context.Provider, webhookURLs []string) (*Controller, error) {
```
sudeshrshetty: Use variadic options...add a TODO for now to get this PR merged

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,94 @@
...
+
+type dispatcher struct {
+	client         *http.Client
+	subscriberURLs []string
+}
+
+// StartWebhookDispatcher will launch the webhook dispatcher in a new goroutine.
+// The dispatcher is responsible for sending notifications to subscribers.
+func StartWebhookDispatcher(subscriberURLs []string) {
+	if subscriberURLs != nil {
```
sudeshrshetty: check for empty... simply put len() check

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	defer func() {
+		e := resp.Body.Close()
+		if e != nil {
+			logger.Errorf("Failed to close response body")
+		}
+	}()
+
+	successStatusCodes := [2]int{200, 201}
```
sudeshrshetty: move to constant or variable.

pkg/restapi/webhook/webhook_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	select {
+	case <-subscriberReceivedNotification:
+		// Pass
+	case <-time.After(5 * time.Second):
+		require.FailNow(t, "samplewebhookclient did not receive a notification")
+	}
+}
+
+func TestWebhookDispatcherMultipleWebhooks(t *testing.T) {
```
sudeshrshetty: Use loop instead of repeating each logic 2 times.

pkg/restapi/webhook/webhook.go
soluchok: if those statuses are HTTP better to use `http.StatusCreated and http.StatusOK`

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,94 @@
...
+		return fmt.Errorf("failed to create new http post request for %s: %s", destination, err)
+	}
+
+	resp, err := c.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to post notification to %s: %s", destination, err)
+	}
+
+	defer func() {
+		e := resp.Body.Close()
```
soluchok: it is awful I mean `e` instead of `err` you can create a helper function
```
func closeOrLog(c io.Closer) {
	if err := c.Close(); err != nil {
		logger.Errorf("Failed to close response body")
	}
}
```

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,94 @@
...
+			logger.Errorf(err.Error())
+		}
+	}
+}
+
+func sendNotification(destination string, c *http.Client) error {
+	//TODO: This HTTP post request needs to supply real data, not this sample text (see #472).
+	ctx, cancel := context.WithTimeout(context.Background(), notificationSendTimeout)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, "POST", destination,
```
soluchok: "POST" to http.MethodPost

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,94 @@
...
+	client         *http.Client
+	subscriberURLs []string
+}
+
+// StartWebhookDispatcher will launch the webhook dispatcher in a new goroutine.
+// The dispatcher is responsible for sending notifications to subscribers.
+func StartWebhookDispatcher(subscriberURLs []string) {
+	if subscriberURLs != nil {
+		dispatcher := &dispatcher{
+			client:         &http.Client{},
```
soluchok: Do not use default client `&http.Client{}` at least timeout should be here

P.S function can be simplified
```
func StartWebhookDispatcher(subscriberURLs []string) {
	if subscriberURLs == nil {
		logger.Warnf("No subscriber URLs provided. Webhook dispatcher will not start.")
	}

	dispatcher := &dispatcher{
		client:         &http.Client{},
		subscriberURLs: subscriberURLs,
	}
	go dispatcher.start()
}
```

pkg/restapi/webhook/webhook.go
DRK3: Talked to Sudesh, the linter won't let me make it a global variable and the compiler won't let me make it a constant, so I'm just checking for both error codes instead of putting it in an array.

pkg/restapi/webhook/webhook.go
soluchok: and replace `subscriberURLs != nil` to `len(subscriberURLs) != 0`

pkg/restapi/webhook/webhook.go
sudeshrshetty: is it fixed?

pkg/restapi/webhook/webhook_test.go
DRK3: Done

pkg/restapi/webhook/webhook.go
DRK3: Yep

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
DRK3: I did the simplification and replaced subscriberURLs != nil with len(subscriberURLs) != 0

I sent you a message regarding the timeout.

pkg/restapi/webhook/webhook.go
DRK3: Done!

pkg/restapi/webhook/webhook.go
```diff
@@ -0,0 +1,80 @@
...
+	notificationInterval    = 2 * time.Second
+	notificationSendTimeout = 10 * time.Second
+)
+
+var logger = log.New("aries-framework/agentd")
+
+// StartWebhookDispatcher will launch the webhook dispatcher in a new goroutine.
+// The dispatcher is responsible for sending notifications to subscribers.
+func StartWebhookDispatcher(subscriberURLs []string) {
+	if len(subscriberURLs) == 0 {
```
sudeshrshetty: You should return when you find it empty.

pkg/restapi/webhook/webhook_test.go
```diff
@@ -0,0 +1,235 @@
...
+		if err := listenFor(testClientData.clientHost); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	StartWebhookDispatcher([]string{fmt.Sprintf("http://%s%s", allTestClientData[0].clientHost,
+		allTestClientData[0].clientHandlerPattern),
+		fmt.Sprintf("http://%s%s", allTestClientData[1].clientHost, allTestClientData[1].clientHandlerPattern)})
+
+	select {
```
sudeshrshetty: something like in below example?

for i := 0; i < 2; i++ {
	select {
	case n := <-ch1:
		fmt.Printf("ch1: %d\n", n)

	case n := <-ch2:
		fmt.Printf("ch2: %d\n", n)
	}
}

Issue #420, opened 2019-10-04
Update BDD test to validate Invitation to DIDExchange Complete status


rolsonquadras:
Based on https://github.com/hyperledger/aries-framework-go/issues/387#issuecomment-537660141.

BDD Test - Inviter generates and shares Invitation. At the end of the flow, test need to validate successful completion for this invitation (1 connection).

rolsonquadras:
This has been implemented in https://github.com/hyperledger/aries-framework-go/pull/428/#issue-325128422. Closing the issue.

PR #428, opened 2019-10-07
feat: Add ability to query connection from did exchange persistence 

### LINKED ISSUE #424: Add ability to query connection from did exchange persistence

fqutishat:
- Add query connection in did exchange persistence

closes #424 #420

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -11,7 +11,6 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	"time"
 
 	"github.com/google/uuid"
 
@@ -27,6 +26,9 @@ const (
 	// InvitationID invitation id is created in invitation request
 	InvitationID = didexchange.InvitationID
 )
+
+// ErrConnectionNotFound is returned when connection not found
+var ErrConnectionNotFound = errors.New("connection not found")
 
 // provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
 type provider interface {
@@ -49,7 +51,7 @@ type Client struct {
 	actionEventlock          sync.RWMutex
 	msgEvents                []chan<- dispatcher.StateMsg
 	msgEventsLock            sync.RWMutex
-	recorder                 *didexchange.ConnectionRecorder
+	connectionStore          *didexchange.ConnectionRecorder
 }
 
 // New return new instance of didexchange client
@@ -74,9 +76,9 @@ func New(ctx provider) (*Client, error) {
 		wallet:                   ctx.CryptoWallet(),
 		inboundTransportEndpoint: ctx.InboundTransportEndpoint(),
 		// TODO channel size - https://github.com/hyperledger/aries-framework-go/issues/246
-		actionCh: make(chan dispatcher.DIDCommAction, 10),
-		msgCh:    make(chan dispatcher.StateMsg, 10),
-		recorder: didexchange.NewConnectionRecorder(store),
+		actionCh:        make(chan dispatcher.DIDCommAction, 10),
+		msgCh:           make(chan dispatcher.StateMsg, 10),
+		connectionStore: didexchange.NewConnectionRecorder(store),
 	}
 
 	// start listening for action/message events
@@ -103,7 +105,7 @@ func (c *Client) CreateInvitation(label string) (*didexchange.Invitation, error)
 		Type:            didexchange.ConnectionInvite,
 	}
 
-	err = c.recorder.SaveInvitation(verKey, invitation)
+	err = c.connectionStore.SaveInvitation(verKey, invitation)
 	if err != nil {
 		return nil, fmt.Errorf("failed to save invitation: %w", err)
 	}
@@ -124,19 +126,25 @@ func (c *Client) HandleInvitation(invitation *didexchange.Invitation) error {
 }
 
 // QueryConnections queries connections matching given parameters
-func (c *Client) QueryConnections(request *QueryConnectionsParams) ([]*QueryConnectionResult, error) {
+func (c *Client) QueryConnections(request *QueryConnectionsParams) ([]*ConnectionResult, error) {
 	// TODO sample response, to be implemented as part of #226
-	return []*QueryConnectionResult{
-		{ConnectionID: uuid.New().String(), CreatedTime: time.Now()},
-		{ConnectionID: uuid.New().String(), CreatedTime: time.Now()},
+	return []*ConnectionResult{
+		{didexchange.ConnectionRecord{ConnectionID: uuid.New().String()}},
+		{didexchange.ConnectionRecord{ConnectionID: uuid.New().String()}},
 	}, nil
 }
 
-// QueryConnectionByID fetches single connection record for given id
-func (c *Client) QueryConnectionByID(id string) (*QueryConnectionResult, error) {
-	// TODO sample response, to be implemented as part of #226
-	return &QueryConnectionResult{
-		ConnectionID: uuid.New().String(), CreatedTime: time.Now(),
+// GetConnection fetches single connection record for given id
+func (c *Client) GetConnection(connectionID string) (*ConnectionResult, error) {
+	conn, err := c.connectionStore.GetConnection(connectionID)
+	if err != nil {
+		if errors.Is(err, storage.ErrDataNotFound) {
+			return nil, ErrConnectionNotFound
+		}
+		return nil, fmt.Errorf("cannot fetch state from store: connectionid=%s err=%s", connectionID, err)
+	}
+	return &ConnectionResult{
+		didexchange.ConnectionRecord{ConnectionID: connectionID, State: conn.State},
 	}, nil
 }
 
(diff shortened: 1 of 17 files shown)


test/bdd/features/didexchange_e2e.feature
```diff
@@ -18,4 +18,5 @@ Feature: Decentralized Identifier(DID) exchange between the agents
     And   "Bob" receives invitation from "Alice"
     And   "Alice" waits for post state event "completed"
     And   "Bob" waits for post state event "completed"
-
+    And   "Alice" retrieve connection record and validate state is "completed"
```
sandrask: retrieves connection record and validates that connection state is "completed"

sandrask: same as Alice

fqutishat: done

pkg/client/didexchange/client.go
```diff
@@ -134,10 +144,16 @@ func (c *Client) QueryConnections(request *QueryConnectionsParams) ([]*QueryConn
 }
 
 // QueryConnectionByID fetches single connection record for given id
-func (c *Client) QueryConnectionByID(id string) (*QueryConnectionResult, error) {
-	// TODO sample response, to be implemented as part of #226
+func (c *Client) QueryConnectionByID(connectionID string) (*QueryConnectionResult, error) {
```
troyronda: Out of curiosity, why is the Client method "QueryConnectionByID" and the underlying service method the more succinct "Connection"? Will the client have more methods for executing connection lookups?

fqutishat: Yes I think the client will more methods for connection lookups

troyronda: (what about the underlying service?)

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -409,9 +414,12 @@ func (s *Service) Name() string {
 }
 
 // Connection return connection
-func (s *Service) Connection(id string) {
-	// TODO add Connection logic
-
+func (s *Service) Connection(connectionID string) (*ConnectionRecord, error) {
```
troyronda: I wonder if we should prefix this method to indicate that this is a query/lookup rather than just an accessor?

fqutishat: I will rename it to QueryConnectionByID like the client

fqutishat: done

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -52,6 +46,17 @@ const (
...
 )
 
+// didCommChMessage type to correlate actionEvent message(go channel) with callback message(internal go channel).
+type didCommChMessage struct {
+	ID              string
+	DIDCommCallback dispatcher.DIDCommCallback
+}
+
+// ConnectionRecord contain info about did exchange connection
+type ConnectionRecord struct {
```
talwinder50: Sudesh has created separate file connections, do you think this struct belong there or here .. ?

troyronda: @sudeshrshetty @fqutishat ^^^

fqutishat: Yes I will move it

troyronda: A couple thoughts:
- As store is provided to Client, it would make more sense (to me) to have a set of protocol methods that can marshal/unmarshal using a store and protocol record types. I don't like the idea of unnecessary methods in the service package. These methods could live in a persistence sub-package as Sudesh started.
- "QueryFooByID" seems a bit verbose.
  - Retrieving a record by ID is the default mechanism (ByID is not really needed).
  - If we have a bunch of way to do queries, it might make sense to have a criteria object passed to a "QueryFoo" method. As such, we probably should use a different prefix for the default mechanism (other than Query).

llorllale: I think this applies here: https://golang.org/doc/effective_go.html#Getters

pkg/client/didexchange/client.go
llorllale: Strange that a method called `QueryConnectionByID` returns `QueryConnectionResult` objects instead of `Connection` objects.

Also, has `QueryConnectionResult` been formally adopted into the `client` package?

pkg/didcomm/protocol/didexchange/service.go
llorllale: One other thought: the "Record" suffix is unnecessary

fqutishat: done

pkg/client/didexchange/client.go
fqutishat: done

pkg/didcomm/protocol/didexchange/persistence/connection.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
+// ConnectionData contain info about did exchange connection
+type ConnectionData struct {
```
troyronda: I might have expected `ConnectionRecord` with the other naming in this package like recorder.

p.s., I don't think the Data suffix is really needed (as it's in a persistence package).

pkg/client/didexchange/client.go
```diff
@@ -133,11 +136,17 @@ func (c *Client) QueryConnections(request *QueryConnectionsParams) ([]*QueryConn
 	}, nil
 }
 
-// QueryConnectionByID fetches single connection record for given id
-func (c *Client) QueryConnectionByID(id string) (*QueryConnectionResult, error) {
-	// TODO sample response, to be implemented as part of #226
+// GetConnection fetches single connection record for given id
+func (c *Client) GetConnection(connectionID string) (*QueryConnectionResult, error) {
```
rolsonquadras: GetConnection returns QueryConnectionResult. I feel it would make more readable if it returns Connection/ConnectionResult.

pkg/client/didexchange/client.go
```diff
@@ -133,11 +136,17 @@ func (c *Client) QueryConnections(request *QueryConnectionsParams) ([]*QueryConn
...
+	conn, err := c.recorder.GetConnection(connectionID)
+	if err != nil {
+		if errors.Is(err, storage.ErrDataNotFound) {
+			return nil, ErrConnectionNotFound
+		}
+		return nil, fmt.Errorf("cannot fetch state from store: connectionid=%s err=%s", connectionID, err)
+	}
 	return &QueryConnectionResult{
```
troyronda: `QueryConnectionResult` is a verbose name for a connection record :).

Is this struct embedding persistence.ConnectionData?

pkg/client/didexchange/client.go
```diff
@@ -133,11 +136,17 @@ func (c *Client) QueryConnections(request *QueryConnectionsParams) ([]*QueryConn
 	}, nil
 }
 
-// QueryConnectionByID fetches single connection record for given id
-func (c *Client) QueryConnectionByID(id string) (*QueryConnectionResult, error) {
-	// TODO sample response, to be implemented as part of #226
+// GetConnection fetches single connection record for given id
+func (c *Client) GetConnection(connectionID string) (*QueryConnectionResult, error) {
+	conn, err := c.recorder.GetConnection(connectionID)
```
troyronda: `recorder` seems to be used as a `connectionStore`.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -64,15 +65,20 @@ type provider interface {
 	OutboundDispatcher() dispatcher.Outbound
 }
 
+type connectionRecorder interface {
```
troyronda: Record retrieval doesn't seem like a "Recorder"

troyronda:
This PR likely has merge conflicts with #441.

pkg/didcomm/protocol/didexchange/persistence/connection.go
fqutishat: Rename to ConnectionRecord. Connection name is already used in did exchange pkg https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/models.go#L72

pkg/client/didexchange/client.go
fqutishat: Rename to ConnectionResult

pkg/client/didexchange/client.go
troyronda: agree.

pkg/client/didexchange/client.go
fqutishat: Rename to connectionStore

pkg/didcomm/protocol/didexchange/service.go
fqutishat: Rename to connectionStore

pkg/client/didexchange/client.go
fqutishat: Rename to ConnectionResult

PR #437, opened 2019-10-07
refactor: DIDExchange Event - Separate API for Continue and Stop

### LINKED ISSUE #436: DIDExchange Event - Provide separate API for Continue and Fail instead of Callback
Based on comments in https://github.com/hyperledger/aries-framework-go/pull/393#issuecomment-539107508

`Callback is currently handling two scenarios (Client wants to signal 'Continue with Protocol' vs 'Fail/Stop Protocol'). Suggest these be separated (e.g., Fail doesn't need an error channel).`

rolsonquadras:
- Currently the callback handles 2 scenario's : Continue with Processing and Fail processing. Both these are handled by Callback using Err field (nil to continue and err to fail).
- This Change will separate these 2 behaviours and provide separate APIs for the scenarios.

Closes #436 

### DIFF
--- a/pkg/didcomm/dispatcher/api.go
+++ b/pkg/didcomm/dispatcher/api.go
@@ -76,20 +76,14 @@ type OutboundCreator func(prov Provider) (Outbound, error)
 type DIDCommAction struct {
 	// DIDComm message
 	Message DIDCommMsg
-	// Callback function to be called by the consumer for further processing the message.
-	Callback Callback
+	// Continue function to be called by the consumer for further processing the message.
+	Continue func()
+	// Stop invocation notifies the service that the consumer action event processing has failed or the consumer wants
+	// to stop the processing.
+	Stop func(err error)
 	// Properties contains different values from different protocols
 	Properties map[string]interface{}
 }
-
-// DIDCommCallback message type to pass service callback in go channels.
-type DIDCommCallback struct {
-	// Set the value in case of any error while processing the DIDComm message event by the consumer.
-	Err error
-}
-
-// Callback type to pass service callbacks.
-type Callback func(DIDCommCallback)
 
 // Event related apis.
 type Event interface {
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -27,8 +27,8 @@ var logger = log.New("aries-framework/did-exchange/service")
 
 // didCommChMessage type to correlate actionEvent message(go channel) with callback message(internal go channel).
 type didCommChMessage struct {
-	ID              string
-	DIDCommCallback dispatcher.DIDCommCallback
+	ID  string
+	Err error
 }
 
 const (
@@ -281,8 +281,11 @@ func (s *Service) sendActionEvent(msg dispatcher.DIDCommMsg, aEvent chan<- dispa
 	// TODO pass invitation id #397
 	didCommAction := dispatcher.DIDCommAction{
 		Message: msg,
-		Callback: func(didCommCallback dispatcher.DIDCommCallback) {
-			s.processCallback(id, didCommCallback)
+		Continue: func() {
+			s.processCallback(id, nil)
+		},
+		Stop: func(err error) {
+			s.processCallback(id, err)
 		},
 		Properties: s.createEventProperties(threadID, ""),
 	}
@@ -310,7 +313,7 @@ func (s *Service) startInternalListener() {
 	go func() {
 		for msg := range s.callbackChannel {
 			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
-			if err := s.process(msg.ID, msg.DIDCommCallback); err != nil {
+			if err := s.process(msg); err != nil {
 				// TODO handle error
 				logger.Errorf(err.Error())
 			}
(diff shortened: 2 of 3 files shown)


pkg/didcomm/dispatcher/api.go
```diff
@@ -76,20 +76,19 @@ type OutboundCreator func(prov Provider) (Outbound, error)
 type DIDCommAction struct {
 	// DIDComm message
 	Message DIDCommMsg
-	// Callback function to be called by the consumer for further processing the message.
-	Callback Callback
+	// Continue function to be called by the consumer for further processing the message.
+	Continue Continue
+	// Fail invocation notifies the service that the consumer action event processing failed.
```
troyronda: comment might need some word-smithing (or that the consumer wants to stop processing).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -68,7 +62,7 @@ type provider interface {
 type Service struct {
 	ctx             context
 	store           storage.Store
-	callbackChannel chan didCommChMessage
+	callbackChannel chan string
```
troyronda: it's now hard to understand what is being sent on this `callbackChannel`, since all the code has is `chan string`.

pkg/didcomm/dispatcher/api.go
rolsonquadras: [Done](https://github.com/hyperledger/aries-framework-go/compare/12e97e25c20fc563333b1f7ea8f0de7c09fe1c0e..42c5b3f270b75af9cd92cb0657abbcacffce467f)

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -685,17 +685,19 @@ func startConsumer(t *testing.T, svc *Service, done chan bool) {
 
 	go func() {
 		for e := range actionCh {
+			// assigned to var as lint fails with : Using a reference for the variable on range scope (scopelint)
```
troyronda: Linter fails because each iteration of the loop overwrites e.

When you use a pointer to e, you need to be careful that each iteration actually dereferences the pointer rather than using the pointer as a value.

pkg/didcomm/protocol/didexchange/service.go
troyronda: From discussion, it sounds like this needs to be a struct anyways to internally signal Continue vs Fail.

pkg/didcomm/dispatcher/api.go
```diff
@@ -76,20 +76,20 @@ type OutboundCreator func(prov Provider) (Outbound, error)
...
 	Properties map[string]interface{}
 }
 
-// DIDCommCallback message type to pass service callback in go channels.
-type DIDCommCallback struct {
-	// Set the value in case of any error while processing the DIDComm message event by the consumer.
-	Err error
-}
+// Continue type to process service callbacks.
+type Continue func()
```
llorllale: Suggest we declare `DIDCommAction.Continue` a `func()` directly as opposed to defining this extra type

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -281,8 +275,11 @@ func (s *Service) sendActionEvent(msg dispatcher.DIDCommMsg, aEvent chan<- dispa
 	// TODO pass invitation id #397
 	didCommAction := dispatcher.DIDCommAction{
 		Message: msg,
-		Callback: func(didCommCallback dispatcher.DIDCommCallback) {
-			s.processCallback(id, didCommCallback)
+		Continue: func() {
+			s.processCallback(id)
+		},
+		Fail: func(err error) {
```
llorllale: This `Fail` function should also be marking the state machine instance as 'stale' somehow

pkg/didcomm/dispatcher/api.go
```diff
@@ -76,20 +76,20 @@ type OutboundCreator func(prov Provider) (Outbound, error)
 type DIDCommAction struct {
 	// DIDComm message
 	Message DIDCommMsg
-	// Callback function to be called by the consumer for further processing the message.
-	Callback Callback
+	// Continue function to be called by the consumer for further processing the message.
+	Continue Continue
+	// Fail invocation notifies the service that the consumer action event processing has failed or the consumer wants
+	// to stop the processing.
+	Fail Fail
```
llorllale: Suggest we rename this to an antonym of "continue" - suggest "Abandon" or "Discontinue" or "Stop"

troyronda: I had also mentioned "Stop" as a possibility.

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/42c5b3f270b75af9cd92cb0657abbcacffce467f..d4d6d2ee8a27820203156ad20b4f7a69e8277911#diff-6a4454074ff31ccb82bdbcc343255467R83) to Stop.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Reverted] the changes. Also added a [TODO](https://github.com/hyperledger/aries-framework-go/pull/437/files#diff-f761c91e753eba1c42a68076df06ce12R333)

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Added [TODO](https://github.com/hyperledger/aries-framework-go/pull/437/files#diff-f761c91e753eba1c42a68076df06ce12R333). Also we would need to clean-up temp data.

troyronda: Is this the same TODO @rolsonquadras @llorllale https://github.com/hyperledger/aries-framework-go/pull/437#discussion_r332190493

rolsonquadras: @troyronda yes.

pkg/didcomm/dispatcher/api.go
rolsonquadras: In this case having type doesn't make much sense. Hence [removed](https://github.com/hyperledger/aries-framework-go/compare/cf0815053284baec1dece5480f30d8bd13ec3b36..45f8d6085ca35de12a950470fe29a49a0528996f#diff-6a4454074ff31ccb82bdbcc343255467L88).

PR #447, opened 2019-10-09
refactor: Improve configuration of NewCredential() regarding JWT

### LINKED ISSUE #445: Improve configuration of NewCredential() regarding JWT deserialization
When new `verifiable.Credential` is created, the following algorithm (based on configuration options) of how to interpret input bytes array is used:
- if `issuerPublicKeyFetcher` is defined, try to deserialize JWS, if error, abort
- if `unsecuredJWT` is true, try to deserialize unsecured JWT, if error, abort
- otherwise, unmarshal from JSON.

We need to refactor this and make the algorithm more clear.

kdimak:
closes #445

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -324,6 +324,20 @@ type CredentialTemplate func() *Credential
 // If not defined, JWT encoding is not tested.
 type PublicKeyFetcher func(issuerID, keyID string) (interface{}, error)
 
+// jwtDecoding defines if to decode VC from JWT
+type jwtDecoding int
+
+const (
+	// noJwtDecoding not a JWT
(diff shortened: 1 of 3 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -324,6 +324,20 @@ type CredentialTemplate func() *Credential
 // If not defined, JWT encoding is not tested.
 type PublicKeyFetcher func(issuerID, keyID string) (interface{}, error)
 
+// JWTDecoding defines if to decode VC from JWT
+type JWTDecoding int
```
troyronda: This exported  type doesn’t seem to be currently usable externally as it isn’t used as an argument to an option.

kdimak: True, no need to make it public.

PR #456, opened 2019-10-09
refactor: Move Legacy crypto ops to wallet, use key identifiers to refer to keys needed

### LINKED ISSUE #352: Move Legacy crypter crypto operations under wallet
To ensure the wallet maintains control over secrets, the crypto operations in Crypters need to be inside the wallet. However, the formatting of envelopes doesn't conceptually belong to the wallet, and Crypters/Packers should instead ask the wallet to compute crypto operations needed for pack/unpack.

This issue is for refactoring the Legacy Authcrypter.

Moopli:
Move any Legacy crypter crypto operations which use secrets to the wallet, so secrets are kept inside the wallet.

Fixes #352 


Moopli:
WIP right now because I'd like review from @Baha-sk, specifically on the wallet API

pkg/wallet/wallet.go
```diff
@@ -86,6 +95,186 @@ func (w *BaseWallet) CreateSigningKey() (string, error) {
...
+	encPubB58 := base58.Encode(encPub)
+
+	// Should we check if present in store first, or just convert and persist? Checking store likely slow
+	// checking store also introduces a variable timing dependent on secrets
+
+	kp, err := w.getKey(key)
+	if err != nil {
+		if errors.Is(ErrKeyNotFound, err) {
+			// introduces variable timing dependent on secret being present in store
+			return encPubB58, nil
```
sandrask: return "", nil

pkg/wallet/wallet.go
```diff
@@ -86,6 +95,186 @@ func (w *BaseWallet) CreateSigningKey() (string, error) {
...
+		return "", err
+	}
+
+	return encPubB58, nil
+}
+
+// GetEncryptionPubKey returns the Curve25519 EncryptionPubKey conversion of a given (Ed25519) SigningPubKey
+func (w *BaseWallet) GetEncryptionPubKey(key string) (string, error) {
+	signPub := base58.Decode(key)
+	if len(signPub) != 32 {
```
sandrask: use constant

pkg/wallet/wallet.go
```diff
@@ -86,6 +95,186 @@ func (w *BaseWallet) CreateSigningKey() (string, error) {
...
+	encPubB58 := base58.Encode(encPub)
+	return encPubB58, nil
+}
+
+// BoxEasy seals a message with a provided nonce
+// theirPub is used as a public key, while myPub is used to identify the private key that should be used
+func (w *BaseWallet) BoxEasy(payload, nonce []byte, theirPub, myPub string) ([]byte, error) {
+	recPubSlice := base58.Decode(theirPub)
+
+	var recPubBytes [32]byte
```
sandrask: use constant encKeySize or pubKeySize (comment applies to other places as well)

pkg/wallet/api.go
```diff
@@ -69,6 +69,10 @@ type Crypto interface {
 	//
 	// error: error
 	DecryptMessage(encMessage []byte, toVerKey string) ([]byte, string, error)
+
+	// FindVerKey picks, out of the keys provided, a key present within the wallet which
+	// the wallet wishes to use for decryption
+	FindVerKey(candidateKeys []string) (int, error)
```
troyronda: Find is not the usual prefix.

pkg/wallet/wallet.go
```diff
@@ -87,6 +96,186 @@ func (w *BaseWallet) CreateSigningKey() (string, error) {
...
+	copy(recPubBytes[:], recPubSlice)
+
+	// generate an equivalent nonce to libsodium's (see link above)
+	nonceSlice, err := makeNonce(epk[:], recPubSlice)
+	if err != nil {
+		return nil, err
+	}
+	var nonce [24]byte
```
troyronda: nonce size should also be a constant.

pkg/wallet/api.go
Moopli: I was considering `ChooseVerKey` since this chooses a verfication key, how's that?

pkg/wallet/wallet.go
Moopli: I decided to remove that edge case from this function - there's another function to handle converting public keys that aren't expected to be in the wallet.

pkg/wallet/api.go
Moopli: `SelectVerKey`?

troyronda: Is it returning an index? Is that an index of candidateKeys?

troyronda: Find seems better than Choose or Select.

Note: the standard library has a similar function for searching a byte slice. https://golang.org/pkg/bytes/#IndexByte

Moopli: Yeah, it returns an index into `candidateKeys`, indicating the verkey to use for decryption.

Moopli: Leaving it then, thanks!

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -83,6 +69,7 @@ func (c *Crypter) buildEnvelope(nonce, payload, cek []byte, protected *protected
 
 	protectedB64 := base64.URLEncoding.EncodeToString(protectedBytes)
 
+	// TODO consider: does cek need to be kept secret in wallet?
```
troyronda: Issue number?

pkg/wallet/api.go
```diff
@@ -90,6 +90,16 @@ type DIDCreator interface {
 	CreateDID(method string, opts ...DocOpts) (*did.Doc, error)
 }
 
+// KeyConverter provides methods for converting signing to encryption keys
+type KeyConverter interface {
+	// ConvertToEncryptionKey creates and persists a Curve25519 keypair created from the given SigningPubKey's
+	// Ed25519 keypair, returning the EncryptionPubKey for this new keypair.
+	ConvertToEncryptionKey(key string) (string, error)
+
+	// GetEncryptionPubKey converts the given SigningPubKey to an EncryptionPubKey
```
troyronda: Does this also convert? But not persist?

pkg/wallet/legacy_api.go
```diff
@@ -0,0 +1,28 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package wallet
+
+import "io"
+
+// LegacySecretBox provides methods for wallet to perform secret box sealing and unsealing
+type LegacySecretBox interface {
```
troyronda: Why is there a "Legacy" interface in the **generic** wallet package?

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
Moopli: I'll remove TODO, we did consider, it's not necessary.

pkg/wallet/api.go
Moopli: Yes, I'll move it to cryptoutil

Moopli: Moved, along with a couple other static util functions.

pkg/wallet/legacy_api.go
Moopli: What interface :^)

Replaced with a Box struct that implements these methods, that the wallet instantiates.

pkg/wallet/box.go
```diff
@@ -0,0 +1,146 @@
...
+// Box provides an elliptic-curve-based authenticated encryption scheme
+//
+// Payloads are encrypted using symmetric encryption (XChacha20Poly1305)
+// using a shared key derived from a shared secret created by
+//   Curve25519 Elliptic Curve Diffie-Hellman key exchange.
+//
+// Box is created by a wallet, and reads secret keys from the wallet
+//   for encryption/decryption, so clients do not need to see
+//   the secrets themselves.
+type Box struct {
```
troyronda: FYI: NaCl calls it a crypto box

pkg/wallet/box.go
```diff
@@ -0,0 +1,146 @@
...
+	w *BaseWallet
+}
+
+// BoxEasy seals a message with a provided nonce
+// theirPub is used as a public key, while myPub is used to identify the private key that should be used
+func (b *Box) BoxEasy(payload, nonce []byte, theirPub, myPub string) ([]byte, error) {
```
troyronda: I wonder if these prefixes are still needed.

pkg/wallet/box.go
```diff
@@ -0,0 +1,146 @@
...
+	"errors"
+	"io"
+
+	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
+
+	"github.com/btcsuite/btcutil/base58"
+	"golang.org/x/crypto/nacl/box"
+)
+
+// Box provides an elliptic-curve-based authenticated encryption scheme
```
troyronda: So I assume these are all wrappers of the box methods that need access to private material. One thing that confused me a bit was the method names don’t quite match the underlying wrapped box method.

pkg/wallet/wallet.go
```diff
@@ -81,6 +81,38 @@ func (w *BaseWallet) CreateSigningKey() (string, error) {
...
+	err = w.persistKey(encPubB58, &kpEnc)
+	if err != nil {
+		return "", err
+	}
+
+	return encPubB58, nil
+}
+
+// GetBox returns a Box which provides crypto box encryption which can use this wallet's secret keys
+func (w *BaseWallet) GetBox() *Box {
```
troyronda: I’m not sure that this method “gets” something.

troyronda: I would suggest reversing the injection.

I.e, have a NewBox method at the package level that takes a wallet.

pkg/wallet/wallet.go
```diff
@@ -81,6 +81,38 @@ func (w *BaseWallet) CreateSigningKey() (string, error) {
...
+	kpEnc := cryptoutil.KeyPair{Priv: encPriv, Pub: encPub}
+	err = w.persistKey(encPubB58, &kpEnc)
+	if err != nil {
+		return "", err
+	}
+
+	return encPubB58, nil
+}
+
+// GetBox returns a Box which provides crypto box encryption which can use this wallet's secret keys
```
troyronda: It still seems a bit odd that we have box methods (rather than just normal crypto primitives exposed). I haven’t thought of a good alternative though. We might need to rethink in the future.

troyronda: @Moopli I wonder if it would make sense to have an internal key package that exposes private material only to framework packages.

The wallet could return a handle to the key (pointer) and the box package could take that handle in its constructor... but the accessors would be unaccessible to outside packages. This way box functionality can be extracted from the wallet package.

troyronda:
Created https://github.com/hyperledger/aries-framework-go/issues/493

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
```diff
@@ -48,132 +50,136 @@ func (fw *failReader) Read(out []byte) (int, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal key: %w", err)
+	}
 
-	t.Run("Failure: encrypt with an invalid-size recipient key", func(t *testing.T) {
-		recipientKey := getB58Key(
-			"Bxp2KpXeh6RgXXRVGRQUskT9qT35aSSz1JvdbMUcB2Yc",
-			"2QqgiHtrUtDPpfoZG2C3Qi8a1MbLQuTZaaScu5LzQbUCkw5YnXngKLMJ8VuPgoN3Piqt1PBUACVd6uQRmtayZp2x")
+	err = store.Put(pub, bytes)
+	return err
```
soluchok: simplify it `return store.Put(pub, bytes)`

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
```diff
@@ -7,30 +7,24 @@ SPDX-License-Identifier: Apache-2.0
...
 
 import (
-	"bytes"
 	"crypto/ed25519"
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
```
soluchok: import is not correct `build-in, third party, local
`

pkg/didcomm/crypto/legacy/authcrypt/decrypt.go
```diff
@@ -55,84 +49,80 @@ func (c *Crypter) Decrypt(envelope []byte, recipient cryptoutil.KeyPair) ([]byte
...
 	senderData := base58.Decode(string(senderSlice))
 
 	var senderEdPub [ed25519.PublicKeySize]byte
 	copy(senderEdPub[:], senderData)
 
-	sender, err := publicEd25519toCurve25519((*publicEd25519)(&senderEdPub))
+	sender, err := cryptoutil.GetEncryptionPubKey(base58.Encode(senderData))
 	if err != nil {
-		return nil, err
+		return "", err
```
soluchok: can be simplified or wrap an error `return cryptoutil.GetEncryptionPubKey(base58.Encode(senderData))`

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -7,44 +7,27 @@ SPDX-License-Identifier: Apache-2.0
...
-	"crypto/ed25519"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
 
 	"github.com/btcsuite/btcutil/base58"
+
```
soluchok: again imports are not sorted properly

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
```diff
@@ -60,7 +43,12 @@ func (c *Crypter) Encrypt(payload []byte, sender cryptoutil.KeyPair, recipientPu
 
 	var recipients []recipient
 
-	recipients, err = c.buildRecipients(cek, edSender, recipientEdKeys)
+	var recKeys []string
+	for _, key := range recipientPubKeys {
```
soluchok: use capacity here e.g ``recKeys := make( []string,0,len(recipientPubKeys))

pkg/internal/cryptoutil/utils.go
```diff
@@ -109,6 +116,20 @@ func Derive25519KEK(alg, apu []byte, fromPrivKey, toPubKey *[chacha.KeySize]byte
...
+	signPub := base58.Decode(key)
+	if len(signPub) != ed25519.PublicKeySize {
+		return "", fmt.Errorf("%d-byte key size is invalid", len(signPub))
+	}
+	encPub, err := PublicEd25519toCurve25519(signPub)
+	if err != nil {
+		return "", err
+	}
+	encPubB58 := base58.Encode(encPub)
+	return encPubB58, nil
```
soluchok: no need to create additional variable, just `return base58.Encode(encPub), nil`

pkg/internal/cryptoutil/utils.go
```diff
@@ -117,6 +138,64 @@ func lengthPrefix(array []byte) []byte {
...
+// Curve25519KeySize number of bytes in a Curve25519 public or private key
+const Curve25519KeySize = 32
+
+// NonceSize size of a nonce used by Box encryption (Xchacha20Poly1305)
+const NonceSize = 24
+
+// PublicEd25519toCurve25519 takes an Ed25519 public key and provides the corresponding Curve25519 public key
+//  This function wraps PublicKeyToCurve25519 from Adam Langley's ed25519 repo: https://github.com/agl/ed25519
+func PublicEd25519toCurve25519(pub []byte) ([]byte, error) {
+	if pub == nil {
```
soluchok: better to use len here e.g `len(pub) == 0`

pkg/internal/cryptoutil/utils.go
```diff
@@ -117,6 +138,64 @@ func lengthPrefix(array []byte) []byte {
...
+	if !success {
+		return nil, errors.New("error converting public key")
+	}
+	return pkOut[:], nil
+}
+
+// SecretEd25519toCurve25519 converts a secret key from Ed25519 to curve25519 format
+//  This function wraps PrivateKeyToCurve25519 from Adam Langley's ed25519 repo: https://github.com/agl/ed25519
+func SecretEd25519toCurve25519(priv []byte) ([]byte, error) {
+	if priv == nil {
```
soluchok: the same, better to use len here e.g len(pub) == 0

pkg/wallet/box.go
```diff
@@ -0,0 +1,146 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package wallet
+
+import (
+	"errors"
+	"io"
+
+	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
```
soluchok: imports again

pkg/wallet/box.go
```diff
@@ -0,0 +1,146 @@
...
+	var recPubBytes [cryptoutil.Curve25519KeySize]byte
+	copy(recPubBytes[:], recPubSlice)
+
+	nonce, err := cryptoutil.Nonce(epk[:], recPubSlice)
+	if err != nil {
+		return nil, err
+	}
+
+	var out = make([]byte, len(epk))
+	copy(out, epk[:])
```
soluchok: I think we can remove it 
```
var out = make([]byte, len(epk))
	copy(out, epk[:])
```
epk instead of out
and use `ret := box.Seal(epk, payload, nonce, &recPubBytes, esk)`

pkg/wallet/box.go
```diff
@@ -0,0 +1,146 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package wallet
```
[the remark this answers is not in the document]
Moopli: How does the box get private keys then?

pkg/didcomm/crypto/legacy/authcrypt/encrypt.go
Moopli: I can set its length and set elements instead of appending.
I'll do the same in `buildRecipients` for the slice it constructs.

pkg/wallet/box.go
Moopli: Done

pkg/wallet/box.go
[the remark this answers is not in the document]
troyronda: @Moopli https://github.com/hyperledger/aries-framework-go/pull/456#discussion_r334988449

pkg/internal/cryptoutil/utils.go
```diff
@@ -8,8 +8,13 @@ package cryptoutil
 
 import (
 	"crypto"
+	"crypto/ed25519"
```
Baha-sk: suggest to add these changes to a separate file (legacy_utils.go) for easier decommissioning of code (if it ever happens in the future)

pkg/internal/cryptoutil/utils.go
```diff
@@ -117,6 +122,67 @@ func lengthPrefix(array []byte) []byte {
...
 
+// Curve25519KeySize number of bytes in a Curve25519 public or private key
+const Curve25519KeySize = 32
+
+// NonceSize size of a nonce used by Box encryption (Xchacha20Poly1305)
+const NonceSize = 24
+
+// PublicEd25519toCurve25519 takes an Ed25519 public key and provides the corresponding Curve25519 public key
+//  This function wraps PublicKeyToCurve25519 from Adam Langley's ed25519 repo: https://github.com/agl/ed25519
+func PublicEd25519toCurve25519(pub []byte) ([]byte, error) {
```
Baha-sk: this function can stay in utils.go as it will be used by jwe encryption too (when we move envelope keys encryption to signature keys)

pkg/internal/cryptoutil/utils.go
```diff
@@ -117,6 +122,67 @@ func lengthPrefix(array []byte) []byte {
...
+	success := extra25519.PublicKeyToCurve25519(pkOut, pKIn)
+	if !success {
+		return nil, errors.New("error converting public key")
+	}
+	return pkOut[:], nil
+}
+
+// SecretEd25519toCurve25519 converts a secret key from Ed25519 to curve25519 format
+//  This function wraps PrivateKeyToCurve25519 from Adam Langley's ed25519 repo: https://github.com/agl/ed25519
+func SecretEd25519toCurve25519(priv []byte) ([]byte, error) {
```
Baha-sk: this function can stay in utils.go as it will be used by jwe encryption too (when we move envelope keys encryption to signature keys)

pkg/wallet/crypto_box.go
```diff
@@ -0,0 +1,140 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package wallet
+
+import (
```
troyronda: Needs a TODO to move out of the wallet package.

pkg/internal/cryptoutil/utils.go
Moopli: It's done, only the nonce needed to be moved.

pkg/wallet/crypto_box.go
Moopli: TODO added

pkg/wallet/box.go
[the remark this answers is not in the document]
Moopli: created #511 for this and more

pkg/wallet/wallet.go
Moopli: created #511

llorllale: @troyronda that would increase risks of leaking the secrets - see [Managing Secrets](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0050-wallets/README.md#managing-secrets)

Have other solutions been explored such as the wallet returning a `Signer` or `Encrypter`?

PR #458, opened 2019-10-09
refactor: Properties variable in Action/Msg event struct

### LINKED ISSUE #457: Refactor Properties variable in Action/Msg event struct
While discussing the possible solution for #442 with @troyronda , we came across the [Properties](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/common/service/event.go#L25) variable in Action/Msg event struct.

Currently the type of that is `map[string]interface{}`. The consumer would need to know the keys for each protocol and also they would need to cast to a type for each of those keys.

Updating this to `interface{}` type would require the consumers to know the type for each protocol. The protocols can have an interface that will have functions for different properties. 

```
package service
type StateMsg struct {
	Properties interface{}
}

type DIDExchangeEvent interface {
	GetConnectionID() string
	GetInvitationID() string
}
```

```
package didexconsumer

msg = StateMsg from event

prop, ok := e.Properties.(service.DIDExchangeEvent)
// cast validation
connectionID := prop.GetConnectionID()
invitationID := prop.GetInvitationID()

```

rolsonquadras:
- Currently the type of that is map[string]interface{}. The consumer would need to know the keys for each protocol and also they would need to cast to a type for each of those keys.
- Update this to interface{} type. That would require the consumers to know the type for each protocol. The each protocol can have an interface that will have functions for different properties.

Closes #457 

### DIFF
--- a/pkg/didcomm/common/service/event.go
+++ b/pkg/didcomm/common/service/event.go
@@ -12,30 +12,61 @@ type StateMsgType int
 const (
 	// PreState pre state
 	PreState StateMsgType = iota
+
 	// PostState post state
 	PostState
 )
 
-// StateMsg msg
+// StateMsg is used in MsgEvent to pass the state details to the consumer. Refer service.Event.RegisterMsgEvent
+// for more details.
 type StateMsg struct {
-	Type    StateMsgType
+	// Name of the protocol.
+	//
+	// Supported protocols
+	//   - DID Exchange :  didexchange.DIDExchange
+	ProtocolName string
+
+	// type of the message (pre or post), refer service.StateMsgType
+	Type StateMsgType
+
+	// current state. Refer protocol RFC for possible states.
 	StateID string
-	Msg     DIDCommMsg
-	// Properties contains different values from different protocols
-	Properties map[string]interface{}
+
+	// DIDComm message along with message type
+	Msg DIDCommMsg
+
+	// Properties contains value based on specific protocol. The consumers need to cast this to specific interface
+	// based on the protocol event.
+	//
+	// Cast to following interfaces based on protocol
+	//   - DID Exchange :  service.DIDExchangeEvent
+	Properties interface{}
 }
 
 // DIDCommAction message type to pass events in go channels.
 type DIDCommAction struct {
+	// Name of the protocol.
+	//
+	// Supported protocols
+	//   - DID Exchange :  didexchange.DIDExchange
+	ProtocolName string
+
 	// DIDComm message
 	Message DIDCommMsg
+
 	// Continue function to be called by the consumer for further processing the message.
 	Continue func()
+
 	// Stop invocation notifies the service that the consumer action event processing has failed or the consumer wants
 	// to stop the processing.
 	Stop func(err error)
-	// Properties contains different values from different protocols
-	Properties map[string]interface{}
+
+	// Properties contains value based on specific protocol. The consumers need to cast this to specific interface
+	// based on the protocol event.
+	//
+	// Cast to following interfaces based on protocol
+	//   - DID Exchange :  service.DIDExchangeEvent
+	Properties interface{}
 }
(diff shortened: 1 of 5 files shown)


pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -121,21 +121,9 @@ func TestService_Handle_Inviter(t *testing.T) {
...
-				}
-				if e.StateID == "responded" {
-					respondedFlag <- true
-				}
-			}
-		}
-	}()
+	completedFlag := make(chan struct{})
+	respondedFlag := make(chan struct{})
+	msgEventListener(t, statusCh, respondedFlag, completedFlag)
```
troyronda: personally i like doing go func name rather than embedding the go. it makes it more obvious that there is a goroutine.

rolsonquadras: Linter complained about the complexity (> 10). Hence moved to new function.

pkg/didcomm/common/service/event.go
```diff
@@ -12,30 +12,49 @@ type StateMsgType int
...
+	// current state. Refer protocol RFC for possible states.
 	StateID string
-	Msg     DIDCommMsg
-	// Properties contains different values from different protocols
-	Properties map[string]interface{}
+
+	// DIDComm message along with message type
+	Msg DIDCommMsg
+
+	// Properties contains value based on specific protocol. The consumers need to cast this to specific interface
```
llorllale: How do consumers currently segregate these state messages per protocol? Closest I can see is by inspecting the type on the enclosed `DIDCommMsg`. Consider including an additional identifier whose contents are an exact match against the protocol identifiers (eg. against `didexchange.DIDExchange`, `introduce.Introduce`, and so on)

pkg/didcomm/common/service/event.go
```diff
@@ -56,3 +75,12 @@ type Event interface {
 	// UnregisterMsgEvent on protocol messages. Refer RegisterMsgEvent().
 	UnregisterMsgEvent(ch chan<- StateMsg) error
 }
+
+// DIDExchangeEvent properties related api.
+type DIDExchangeEvent interface {
```
llorllale: Suggest move this to the didexchange package.

rolsonquadras: [Moved](https://github.com/hyperledger/aries-framework-go/compare/ebaa2c60d8b2fce620d906c585bfb49c74ba8111..cc46c9e091f6091fb11e5c1467352ad67d28c0ea#diff-79de382a052bbdb4213b9dd373b0e2f6R10).

pkg/didcomm/protocol/didexchange/service_test.go
rolsonquadras: Misread the point. [Moved](https://github.com/hyperledger/aries-framework-go/compare/ebaa2c60d8b2fce620d906c585bfb49c74ba8111..cc46c9e091f6091fb11e5c1467352ad67d28c0ea#diff-36ac39bebf32030fd7dff6f12aea8d7eR126) go routine from function to caller.

pkg/didcomm/common/service/event.go
rolsonquadras: This is a very valid point. One channel can be registered against multiple protocols and consumers should be able to distinguish protocol at higher level variable than going inside Msg.

I'll add a id in the event message.

rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/cc46c9e091f6091fb11e5c1467352ad67d28c0ea..b836b948f8e18a1341aad88b4edfdb0c0443fb4f#diff-f2de91d20207599740587fa5c3ee13c3R27) a new protocol identifier in event messages.

PR #459, opened 2019-10-09
refactor: Crypter/Wallet crypto operations

### LINKED ISSUE #385: Move JWE crypter crypto operations under wallet
To ensure the wallet maintains control over secrets, the crypto operations in Crypters need to be inside the wallet. However, the formatting of envelopes doesn't conceptually belong to the wallet, and Crypters/Packers should instead ask the wallet to compute crypto operations needed for pack/unpack.

This issue is for refactoring the JWE Authcrypter.

Part of #36

Baha-sk:
The current wallet is calling Pack/Unpack
	This behavior has now changed and delegated to a
	newer Packager interface

	In addition, operations involving private keys have been
	moved to the wallet. Namely, DerviveKEK() is the only function
	in the crypto functionality that deals with private keys

	Aries framework has been updated with new context functions
	to support the new structure. In addtion to WithWallet(),
	2 new function are added to the context provider:
	WithCrypter() and WithPackager().

	JWE Crypto functionality depends on these providers in
	the following order:
	1. WithWallet()
	2. WithCrypter()
	3. WithPackager()

	Finally the recipient key is no longer needed in Crypter.Decrypt()
	as it will be matched with the recipients's list in the envelope
	and the wallet's first key found.

closes #385 


pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
```diff
@@ -18,24 +18,29 @@ import (
...
 // Encrypt will JWE encode the payload argument for the sender and recipients
 // Using (X)Chacha20 encryption algorithm and Poly1305 authenticator
 // It will encrypt using the sender's keypair and the list of recipients arguments
-func (c *Crypter) Encrypt(payload []byte, sender jwecrypto.KeyPair, recipients [][]byte) ([]byte, error) { //nolint:lll,funlen
-	err := verifyKeys(sender, recipients)
-	if err != nil {
-		return nil, fmt.Errorf("failed to encrypt message: %w", err)
+func (c *BaseCrypter) Encrypt(payload, senderKey []byte, recipients [][]byte) ([]byte, error) { //nolint:funlen
+	if len(senderKey) == 0 {
+		return nil, fmt.Errorf("failed to encrypt message")
```
sandrask: error message should include something like empty sender key

pkg/wallet/api.go
```diff
@@ -55,49 +56,29 @@ type Crypto interface {
...
+	//
+	// isFromKeyPriv flag is to direct the function to use fromKey as is if set to true
+	// 		(ie fromKey is an ephemeral private key provided by the caller)
+	// or will fetch the private key corresponding to fromKey from the wallet if set to false
+	// 		(ie fromKey is a public key from the wallet and requires retrieving its private key. This
+	// 		function is part of the wallet for this reason)
+	// returns:
+	// 		kek []byte
+	//		error in case of errors
+	DeriveKEK(alg, apu []byte, fromKey, toPubKey *[chacha.KeySize]byte, isFromKeyPriv bool) ([]byte, error)
```
sandrask: Is there a case where private key would be provided here? I am not clear why we need to expose isFromKeyPriv flag here.

pkg/wallet/api.go
```diff
@@ -143,3 +116,34 @@ func WithServiceType(serviceType string) DocOpts {
...
+var ErrInvalidKey = errors.New("invalid key")
+
+// errEmptyRecipients is used when recipients list is empty
+var errEmptyRecipients = errors.New("empty recipients")
+
+// errInvalidKeypair is used when a keypair is invalid
+var errInvalidKeypair = errors.New("invalid keypair")
+
+// VerifyKeys is a utility function that verifies if sender key pair and recipients keys are valid (not empty)
+func VerifyKeys(sender KeyPair, recipients [][]byte) error {
```
sandrask: Should utility function and non-exported error definitions be in api.go?

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: will update...

pkg/wallet/api.go
Baha-sk: yes, when deriving a KEK from an ephemeral key created in the crypter, in this case `fromKey` will be a private key.. this is why we need `isFromKeyPriv` flag

but deriving a KEK for a CEK doesn't require an ephemeral key, and therefore `fromKey` will be public to tell the function to load its corresponding private key from the wallet

pkg/wallet/api.go
Baha-sk: good point, I can move them to utils.go

pkg/wallet/api.go
Baha-sk: I can probably hide this function and expose 2 functions in the interface:
DeriveKEKUsingPrivKeyFromWallet() which will call `DeriveKEK` with `isFromKeyPriv`= false
DeriveKEKUsingEphemeralPrivKey() which will call `DeriveKEK` with `isFromKeyPriv`= true

this way `isFromKeyPriv` flag is hidden from the interface user. 

Agree?

sandrask: Does it make more sense to create utility function DeriveKEK that will work with private key? In that case we can have have only one method in the wallet that will retrieve private key and call utility function DeriveKEK. And for ephemeral keys since you already have private key you just call utility function DeriveKEK. I am not sure where to put this utility function though.

Baha-sk: I'll move the util function to util.go under wallet package.. will push soon

pkg/wallet/api.go
Baha-sk: done

pkg/wallet/api.go
Baha-sk: member DeriveKEK function takes fromKey as public key and calls a util exported function with the same name that does the real derivation using a private key.

done

pkg/didcomm/crypto/jwe/authcrypt/encrypt.go
Baha-sk: done

pkg/wallet/api.go
```diff
@@ -55,49 +56,24 @@ type Crypto interface {
...
-	//
-	// string: The sender verification key
-	//
-	// error: error
-	DecryptMessage(encMessage []byte, toVerKey string) ([]byte, string, error)
-}
+	// returns:
+	// 		kek []byte the key encryption key used to decrypt a cek (a shared key)
+	//		error in case of errors
+	DeriveKEK(alg, apu []byte, fromPuKey, toPubKey *[chacha.KeySize]byte) ([]byte, error)
```
troyronda: why are we setting the size of the key in this interface. this was previously fixed.

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
```diff
@@ -30,25 +33,14 @@ const XC20P = ContentEncryption("XC20P") // XChacha20 encryption + Poly1305 auth
...
-
-// errRecipientNotFound is used when a recipient is not found
-var errRecipientNotFound = errors.New("recipient not found")
-
 // errUnsupportedAlg is used when a bad encryption algorithm is used
 var errUnsupportedAlg = errors.New("algorithm not supported")
 
-// Crypter represents an Authcrypt Encrypter (Decrypter) that outputs/reads JWE envelopes
-type Crypter struct {
+// BaseCrypter represents an Authcrypt Encrypter (Decrypter) that outputs/reads JWE envelopes
```
troyronda: is there an extended crypter?

i.e., why “base”?

pkg/wallet/api.go
Baha-sk: Key derivation requires the key size used with the curv, which in our case is curve25519 and the key size is fixed..

I can hide the size from the function signature and enforce it in the implementation.

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
Baha-sk: no, there won't be an extended crypter, but this will be considered the `default` crypter implementation of the JWE compliant spec. 

The `base` is technically there to distinguish the implementation name from the interface


shall I remove Base prefix?

pkg/wallet/api.go
troyronda: Well. I don’t really understand why it is hidden in some methods and not others.

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
troyronda: I do not think Base is descriptive. I also do not think you can have no prefix with the Legacy one coming.

pkg/didcomm/envelope/api.go
```diff
@@ -0,0 +1,51 @@
...
+	// envelope: unpack message
+	//
+	// error: error
+	UnpackMessage(encMessage []byte) (*Envelope, error)
+}
+
+// Envelope contain msg, FromVerKey and ToVerKeys
+type Envelope struct {
+	Message    []byte
+	FromVerKey string
```
llorllale: "From" and "To" are relative to the direction the envelope is going (inbound vs outbound) and might get confusing. I suggest we keep it simple: 

* FromVerKey -> MyVerKey
* ToVerKeys -> TheirVerKeys

pkg/didcomm/envelope/packager.go
```diff
@@ -0,0 +1,66 @@
...
+
+// PackMessage Pack a message for one or more recipients.
+func (p *BasePackager) PackMessage(envelope *Envelope) ([]byte, error) {
+	if envelope == nil {
+		return nil, errors.New("envelope argument is nil")
+	}
+
+	var recipients [][]byte
+	for _, verKey := range envelope.ToVerKeys {
+		// TODO It is possible to have different key schemes in an interop situation
```
llorllale: Do we have an issue for this?

llorllale:
Separating packaging concerns (`envelope.Packager`) from encryption (`crypto.Crypter`) seems like a good idea. However, this PR leaves `Crypter` with knowledge that, I think, belongs to the `Packager`. The `Crypter` still knows stuff about envelopes. It's doing the actual packaging of the envelopes.

llorllale:
@Moopli as discussed in today's morning call, it's probably better to separate the concerns of packaging vs. encryption:

* Support at least two envelope formats (legacy and the new one)
* Support different `alg` and `enc`

pkg/didcomm/envelope/packager.go
Baha-sk: this code and comment was there in the wallet originally.. they were moved to envelope as is...

I will remove this comment as different crypto's and key schemes can be implemented later.

Baha-sk:
> @Moopli as discussed in today's morning call, it's probably better to separate the concerns of packaging vs. encryption:
> 
> * Support at least two envelope formats (legacy and the new one)
> * Support different `alg` and `enc`

this will be done as part of a separate PR

Moopli:
@llorllale @Baha-sk opened #475

pkg/didcomm/envelope/api.go
Baha-sk: I don't see how MyVerKey is clearer than FromVerKey? 

This will be more confusing from the agent's perspective since `MyVerKey` for a receiving agent is going to be different from the sending agent's while `FromVerKey` is clear to be the sender's agent key, no matter who is using the Envelope (`From` will always be source agent).

Again, this code was moved from `wallet` into `envelope`, it's not new code.

pkg/wallet/api.go
Moopli: `fromPuKey` -> `fromPubKey`

Baha-sk: will fix

pkg/didcomm/envelope/packager.go
Baha-sk: updated comment

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
Baha-sk: Base has been removed.. implementations will be called Crypter in both jwe and legacy.. packages will have aliases during import ie: jweauthcrypt.Crypter, legacyauthcypt.Crypter

pkg/wallet/api.go
Baha-sk: Wallet.DeriveKEK now uses []byte for keys, no more key size length enforced...

Baha-sk: done

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
Baha-sk: done

pkg/wallet/api.go
Baha-sk: done

pkg/wallet/utils.go
```diff
@@ -0,0 +1,110 @@
...
+func IsKeyPairValid(kp KeyPair) bool {
+	if kp.Priv == nil || kp.Pub == nil {
+		return false
+	}
+
+	return true
+}
+
+// Derive25519KEK is a utility function that will derive an ephemeral symmetric key (kek) using fromPrivKey and toPubKey
+func Derive25519KEK(alg, apu []byte, fromPrivKey, toPubKey *[chacha.KeySize]byte) ([]byte, error) { // nolint:lll
```
troyronda: Is this used outside the package?

Baha-sk: yes

Moopli: It's used in the Crypter - DeriveKEK was split into a util function because the crypter needs to be able to pass an ephemeral private key in at one point.

troyronda: It still seems odd that the **generic** wallet package has a special ed25519 method.

Baha-sk: I will move utils.go to pkg/internal/cryptoutil to separate the utility logic from the wallet.. in the end, they are internal to our go framework and should not be exposed in the wallet.

Baha-sk: done... moved utils.go to `pkg/internal/cryptoutil` package and updated all its references.

pkg/didcomm/envelope/api.go
llorllale: `Envelope` is exported for reasons I don't understand.

If all the complexity of handling `Envelope` structs is contained within the packing/unpacking functionality then it's okay. But if `Envelope` starts percolating into the application layer where code would have to first determine their role (sender vs receiver) before picking the right attribute ("from" vs "to") - that would be an unnecessary burden imposed on the application layer.

PR #461, opened 2019-10-09
Use Pre-Compiled Regex

### LINKED ISSUE #153: regex should be compiled prior to usage
https://github.com/hyperledger/aries-framework-go/pull/152#discussion_r316844064

Ensure pkg/didmethod/peer/did.go uses pre-compiled regular expressions.

jessie-codes:
**Title:**
Use Pre-Compiled Regex

**Description:**
Ensure pkg/didmethod/peer/did.go uses pre-compiled regular expressions.
closes #153 

Summary:
Use `regexp.MustCompile` in `pkg/didmethod/peer/did.go` to ensure the regex is valid at compile time.


pkg/didmethod/peer/did.go
```diff
@@ -64,10 +64,8 @@ func computeDid(doc *did.Doc) (string, error) {
 func validateDID(doc *did.Doc) error {
 	peerDid := doc.ID
 
-	matched, err := regexp.MatchString(`did:peer:11-([a-fA-F0-9]){64}`, peerDid)
-	if err != nil {
-		return fmt.Errorf("regex match string failed %w", err)
-	}
+	didRegex := regexp.MustCompile(`did:peer:11-([a-fA-F0-9]){64}`)
```
soluchok: move it out of function body

troyronda: @jessie-codes Thanks for your contribution. @soluchok suggests that the compiled regular expression is moved out of the function body so that it doesn't need to be recompiled on every call.

jessie-codes: @soluchok, your suggestion definitely makes sense; however, since there is a linting rule for no global variables, there isn't a great place for this to outside of the function. I can't make it a const due to it not being a simple value and this function isn't attached to a struct.

Given those constraints, is there a preference on where you would want the variable for the regex to be defined?

troyronda: @jessie-codes you may add an exception for this case using a nolint comment. there is an example here: https://github.com/hyperledger/aries-framework-go/blob/9582e05f3a7adcb583469da0a0ed074487c9aa22/pkg/doc/verifiable/jwt_unsecured_test.go#L17

jessie-codes: Great, I've updated the branch with the change.

pkg/didmethod/peer/did.go
```diff
@@ -27,6 +27,11 @@ const (
 	peerPrefix = "did:peer:"
 )
 
+// nolint:gochecknoglobals
+var (
+	didRegex = regexp.MustCompile(`did:peer:11-([a-fA-F0-9]){64}`)
```
soluchok: why not in one line? 
`var didRegex = regexp.MustCompile(`did:peer:11-([a-fA-F0-9]){64}`)`

Issue #398, opened 2019-10-02
[Inviter] ExRequest Processing: Store data


rolsonquadras:
- Retrieve recipient publicKey from envelope and get the invitation from Store.
- Create Own Unique ConnectionID (status - Pending) [Use - UUID] :
- Get the thread ID from exRequest and store the mapping to connectionID

key : connectionID, val : {theirDID, theirLabel,invitationID}
key : protocolID=(theirDID+threadID), val : connectionID

protocolID should be the hash.

PR #463, opened 2019-10-10
feat: Verification keys in DIDComm message

### LINKED ISSUE #462: Access to Recipient keys in DID Comm Message
DID Exchange protocol needs to access recipient keys from envelope.
Currently, DID comm message includes only payload from envelope.

Fix: Include 'To Verification keys' from envelope in DID comm message

sudeshrshetty:
- added `ToVerKeys` from envelope to DID comm message, so that it can be
accessed inside service handles.
- DID exchange service needs these keys for persistence related changes
- service handle to accept pointer of DIDCommMsg 
( **reason**: `hugeParam: msg is heavy (80 bytes)`)
- closes #462 

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -54,7 +54,7 @@ const (
 
 // message type to store data for eventing. This is retrieved during callback.
 type message struct {
-	Msg           service.DIDCommMsg
+	Msg           *service.DIDCommMsg
 	ThreadID      string
 	NextStateName string
 }
@@ -107,7 +107,7 @@ func New(didMaker did.Creator, prov provider) (*Service, error) {
 }
 
 // Handle didexchange msg
-func (s *Service) Handle(msg service.DIDCommMsg) error {
+func (s *Service) Handle(msg *service.DIDCommMsg) error {
 	// throw error if there is no action event registered for inbound messages
 	aEvent := s.GetActionEvent()
 
@@ -224,7 +224,7 @@ func (ex *didExchangeEvent) InvitationID() string {
 
 // sendEvent triggers the action event. This function stores the state of current processing and passes a callback
 // function in the event message.
-func (s *Service) sendActionEvent(msg service.DIDCommMsg, aEvent chan<- service.DIDCommAction,
+func (s *Service) sendActionEvent(msg *service.DIDCommMsg, aEvent chan<- service.DIDCommAction,
 	threadID string, nextState state) error {
 	jsonDoc, err := json.Marshal(&message{
 		Msg:           msg,
@@ -327,7 +327,7 @@ func isNoOp(s state) bool {
 	return ok
 }
 
-func threadID(didCommMsg service.DIDCommMsg) (string, error) {
+func threadID(didCommMsg *service.DIDCommMsg) (string, error) {
 	var thid string
 	if !didCommMsg.Outbound && didCommMsg.Type == ConnectionInvite {
 		return uuid.New().String(), nil
--- a/pkg/framework/context/context.go
+++ b/pkg/framework/context/context.go
@@ -83,12 +83,12 @@ func (p *Provider) InboundTransportEndpoint() string {
 
 // InboundMessageHandler return inbound message handler
 func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
-	return func(payload []byte) error {
+	return func(envelope *wallet.Envelope) error {
 		// get the message type from the payload and dispatch based on the services
 		msgType := &struct {
 			Type string `json:"@type,omitempty"`
 		}{}
-		err := json.Unmarshal(payload, msgType)
+		err := json.Unmarshal(envelope.Message, msgType)
 		if err != nil {
 			return fmt.Errorf("invalid payload data format: %w", err)
 		}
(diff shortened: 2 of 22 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -327,7 +327,7 @@ func isNoOp(s state) bool {
 	return ok
 }
 
-func threadID(didCommMsg service.DIDCommMsg) (string, error) {
+func threadID(didCommMsg *service.DIDCommMsg) (string, error) {
```
soluchok: I suggest make threadID function part of service.DIDCommMsg.

soluchok:
>service handle to accept pointer of DIDCommMsg ( reason: hugeParam: msg is heavy (80 bytes))

I have checked it, `DIDCommMsg` has 56 bytes,

pkg/didcomm/protocol/didexchange/service.go
soluchok: I have created PR for that https://github.com/hyperledger/aries-framework-go/pull/465

sudeshrshetty:
> > service handle to accept pointer of DIDCommMsg ( reason: hugeParam: msg is heavy (80 bytes))
> 
> I have checked it, `DIDCommMsg` has 56 bytes

@soluchok After adding `RecKeys` it went beyond 80bytes.

rolsonquadras:
Just a small observation on [RecKeys/VerKeys](https://github.com/hyperledger/aries-framework-go/pull/463/files#diff-9988f5820bf05228e5657f2842fa071eR99) overall (not specific to this PR) : Looks like we are using different terms for these keys at different layers - Recipient Keys, Verification Keys, Sender Keys. May be we need to be consistent with names across the project.

pkg/wallet/wallet.go
```diff
@@ -151,7 +151,7 @@ func (w *BaseWallet) UnpackMessage(encMessage []byte) (*Envelope, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed from decrypt: %w", err)
 		}
-		return &Envelope{Message: bytes}, nil
+		return &Envelope{Message: bytes, ToVerKeys: []string{recipVKeyB58}}, nil
```
Baha-sk: `recipVKeyB58` is not the only recipient key in UnpackeMessage.. we need to get all recipient's keys.. suggest to create an issue for this to be done in a separate pr 

also Decrypt() is currently being refactored to not pass any key in the arguments list. so this code will change a bit (depends who's PR gets merged first)

sudeshrshetty: That was my concern too, as per current logic of `baseWallet.UnpackMessage()` we try to decrypt message from very first recipient key, if we can't decrypt then we throw error, if we can decrypt we return message. Looks odd, created #468

Baha-sk: yes, this logic will be changed in my PR.. we will find the first matching recipient key in the wallet from the list of recipients in the envelope:
https://github.com/hyperledger/aries-framework-go/pull/459/files#diff-e602f67bb01eb068d2869bf2d53256d4R93

so UnpackMessage() will not use the first recipient in the list anymore

Baha-sk: this is why, I believe `ToVerKeys` above should contain the list of recipients keys in the envelope, not just the current decrypting recipient key

pkg/framework/context/context.go
```diff
@@ -83,20 +83,20 @@ func (p *Provider) InboundTransportEndpoint() string {
 
 // InboundMessageHandler return inbound message handler
 func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
-	return func(payload []byte) error {
+	return func(envelope *wallet.Envelope) error {
```
troyronda: Somehow I'm hesitant about the need to send the entire envelope, since these really ought to be two different layers.

Side note: it seems weird that the Envelope type is defined by Wallet @Baha-sk @Moopli

pkg/wallet/wallet.go
troyronda: @sudeshrshetty @Baha-sk is there an action for this PR?

pkg/framework/context/context.go
sudeshrshetty: If service handle needs to access envelope then InboundMessageHandler need entire envelope.

Alternative solution is to pass second argument next to payload to pass 'ToVerKeys',  but arguments list may grow in future as per requirement. --> func(payload, toVerKeys)

or we can a sub-struct inside 'wallet.Envelope' something like 'Keys' and using that struct here along with payload --> func(payload, keys)

rolsonquadras: > Is DID exchange the only protocol that will care about VerKeys? @rolsonquadras @llorllale ?
@troyronda the primary reason for this PR is https://github.com/hyperledger/aries-framework-go/issues/398#issue-501672749 - to fetch the invitation based on recKey. For now it looks to be specific to DIDExchange.

pkg/wallet/wallet.go
sudeshrshetty: no action in this PR, #468 will be addressed once Baha's changes for decrypt() gets merged

pkg/framework/context/context.go
sudeshrshetty: created #474 for envelope type issue

Baha-sk: > Somehow I'm hesitant about the need to send the entire envelope, since these really ought to be two different layers.
> 
> Is DID exchange the only protocol that will care about VerKeys? @rolsonquadras @llorllale ?
> 
> Side note: it seems weird that the Envelope type is defined by Wallet @Baha-sk @Moopli

this envelope a wallet/packager envelope type that has the payload and verification keys. Which is different than the one representing a JWE found here: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go#L55

in my PR, it has moved to a new `envelope` package: https://github.com/hyperledger/aries-framework-go/pull/459/files#diff-2e40038f150e7f436acca5a5949cb54fR46

troyronda: We should followup with a better abstraction (rather than passing the entire envelope to a layer that shouldn't need it).

PR #465, opened 2019-10-10
refactor: made threadID function part of service.DIDCommMsg

soluchok:
This function will be reusable for introduce protocol as well

* threadID function as part of service.DIDCommMsg

### DIFF
--- a/pkg/didcomm/common/service/errors.go
+++ b/pkg/didcomm/common/service/errors.go
@@ -11,6 +11,9 @@ const (
 	ErrChannelRegistered = serviceError("channel is already registered for the action event")
 	ErrNilChannel        = serviceError("cannot pass nil channel")
 	ErrInvalidChannel    = serviceError("invalid channel passed to unregister the action event")
+	ErrThreadIDNotFound  = serviceError("threadID not found")
+	ErrInvalidMessage    = serviceError("invalid message")
+	ErrNoHeader          = serviceError("the header is not provided")
 )
 
 // serviceError defines service error
--- a/pkg/didcomm/common/service/service.go
+++ b/pkg/didcomm/common/service/service.go
@@ -6,6 +6,13 @@ SPDX-License-Identifier: Apache-2.0
 
 package service
 
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+)
+
 // Handler provides protocol service handle api.
 type Handler interface {
 	Handle(msg *DIDCommMsg) error
@@ -20,16 +27,57 @@ type DIDComm interface {
 	Event
 }
 
+// Header helper structure which keeps reusable fields
+type Header struct {
+	ID     string           `json:"@id"`
+	Thread decorator.Thread `json:"~thread"`
+	Type   string           `json:"@type,omitempty"`
+}
+
 // DIDCommMsg did comm msg
 type DIDCommMsg struct {
 	// Outbound indicates the direction of this DIDComm message:
 	//   - outgoing (to another agent)
 	//   - incoming (from another agent)
 	Outbound bool
-	Type     string
 	Payload  []byte
 	// TODO : might need refactor as per the issue-226
 	OutboundDestination *Destination
+	Header              *Header
+}
+
+// NewDIDCommMsg returns DIDCommMsg with Header
+func NewDIDCommMsg(payload []byte) (*DIDCommMsg, error) {
+	msg := &DIDCommMsg{Payload: payload}
+	if err := json.Unmarshal(msg.Payload, &msg.Header); err != nil {
+		return nil, fmt.Errorf("invalid payload data format: %w", err)
+	}
+	return msg, nil
+}
+
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+// NOTE: Header field should be filled before calling ThreadID func
+// it can be done by using NewDIDCommMsg([]byte) func or directly set a Header value
+func (m *DIDCommMsg) ThreadID() (string, error) {
+	if m.Header == nil {
+		return "", ErrNoHeader
+	}
+	// if message has ~thread.thid but @id is absent this is invalid message
+	if len(m.Header.Thread.ID) > 0 && m.Header.ID == "" {
+		return "", ErrInvalidMessage
+	}
+
+	if len(m.Header.Thread.ID) > 0 {
+		return m.Header.Thread.ID, nil
+	}
+
+	// we need to return it only if there is no ~thread.thid
+	if len(m.Header.ID) > 0 {
+		return m.Header.ID, nil
+	}
+
+	return "", ErrThreadIDNotFound
 }
 
 // Destination provides the recipientKeys, routingKeys, and serviceEndpoint populated from Invitation
--- a/pkg/didcomm/common/service/service_test.go
+++ b/pkg/didcomm/common/service/service_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package service
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+)
+
+func TestNewDIDCommMsg(t *testing.T) {
+	tests := []struct {
+		name    string
+		payload []byte
+		err     string
+	}{{
+		name: "No payload",
+		err:  "invalid payload data format: unexpected end of JSON input",
+	}, {
+		name:    "Happy path",
+		payload: []byte(`{"@id":"ID"}`),
+		err:     "",
+	}, {
+		name:    "Empty payload (JSON)",
+		payload: []byte(`{}`),
+		err:     "",
+	}, {
+		name:    "Array payload",
+		payload: []byte(`[]`),
+		err:     `invalid payload data format: json: cannot unmarshal array into Go value of type service.Header`,
+	}}
+	t.Parallel()
+	for _, test := range tests {
+		tc := test
+		t.Run(tc.name, func(t *testing.T) {
+			val, err := NewDIDCommMsg(tc.payload)
+			if err != nil {
+				require.Contains(t, err.Error(), tc.err)
+				require.Nil(t, val)
+			} else {
+				require.NotNil(t, val)
+			}
+		})
+	}
+}
+
+func TestDIDCommMsg_ThreadID(t *testing.T) {
+	tests := []struct {
+		name string
+		msg  DIDCommMsg
+		val  string
+		err  string
+	}{{
+		name: "No header",
+		msg:  DIDCommMsg{},
+		err:  ErrNoHeader.Error(),
+	}, {
+		name: "ID without Thread ID",
+		msg:  DIDCommMsg{Header: &Header{ID: "ID"}},
+		val:  "ID",
+		err:  "",
+	}, {
+		name: "Thread ID with ID",
+		msg:  DIDCommMsg{Header: &Header{ID: "ID", Thread: decorator.Thread{ID: "thID"}}},
+		val:  "thID",
+		err:  "",
+	}, {
+		name: "Thread ID without ID",
+		msg:  DIDCommMsg{Header: &Header{Thread: decorator.Thread{ID: "thID"}}},
+		val:  "",
+		err:  ErrInvalidMessage.Error(),
+	}}
+	t.Parallel()
+	for _, test := range tests {
+		tc := test
+		t.Run(tc.name, func(t *testing.T) {
+			val, err := tc.msg.ThreadID()
+			if err != nil {
+				require.Contains(t, err.Error(), tc.err)
+			}
+			require.Equal(t, tc.val, val)
+		})
+	}
+}
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -119,7 +119,13 @@ func (c *Client) HandleInvitation(invitation *didexchange.Invitation) error {
 	if err != nil {
 		return fmt.Errorf("failed marshal invitation: %w", err)
 	}
-	if err = c.didexchangeSvc.Handle(&service.DIDCommMsg{Type: invitation.Type, Payload: payload}); err != nil {
+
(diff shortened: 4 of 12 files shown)


pkg/didcomm/common/service/service_test.go
```diff
@@ -0,0 +1,57 @@
...
+		val: "thID",
+		err: "",
+	}, {
+		msg: DIDCommMsg{Payload: []byte(`{"~thread":{"thid":"thID"}}`)},
+		val: "thID",
+		err: "",
+	}, {
+		msg: DIDCommMsg{Payload: []byte(`{}`)},
+		val: "",
+		err: "",
```
Baha-sk: how come an empty JSON doesn't return an error while an empty array does (TC flow)

pkg/didcomm/common/service/service.go
```diff
@@ -30,6 +37,39 @@ type DIDCommMsg struct {
...
+	}
+
+	if len(msg.Thread.ID) > 0 {
+		return msg.Thread.ID, nil
+	}
+
+	return msg.ID, nil
+}
+
+// ThreadID returns msg ID
```
Baha-sk: the function also sets the TreadID in DIDCommMsg .. it doesn't just return msg ID

pkg/didcomm/common/service/service_test.go
soluchok: empty JSON is valid json, but values are empty

soluchok: got it ) `{}` vs `[]` 
`{}` - can be structure, map or interface
 `[]` - only array or interface 
we can not unmarshal list into the structure

pkg/didcomm/common/service/service.go
soluchok: `thID *string` is cache,  I did it to avoid unmarshaling twice, the responsibility of this function is to return an ID and no one should care about what it does under the hood

pkg/didcomm/common/service/service.go
```diff
@@ -30,6 +37,39 @@ type DIDCommMsg struct {
...
+func (m *DIDCommMsg) ThreadID() (string, error) {
+	if m.thID != nil {
+		return *m.thID, nil
```
soluchok: If the message does not have an ID it is valid (no error)
I would suggest to check it, and if the ID is empty we can throw an error e.g `message does not have the ID`
@troyronda @rolsonquadras @llorllale what do you think?

rolsonquadras: DIDExchange Invitation and exchangeRequest messages will not have threadID. Not sure throwing error for those cases will make sense. 

Possible solution
- Exclude those 2 message types for error checking in this method.
- Consumer of this API for those message types shouldn't invoke the method.

@llorllale @troyronda any other thoughts ?

llorllale: @soluchok 

> If the message does not have an ID it is valid (no error)
> I would suggest to check it, and if the ID is empty we can throw an error e.g message does not have the ID

I agree that we should throw an error if the threadID cannot be determined. This should probably be a sentinel error as well (`ErrThreadIdNotFound` or something similar).

A message's threadID can be taken from one of two sources:

* Explicit: `~thread.thid`
* Implicit: `@id` (but only if there is no `~thread.thid`)

A message with `~thread.thid` but no `@id` is illegal.

soluchok: Done!

pkg/didcomm/common/service/service.go
soluchok: Done!

pkg/didcomm/common/service/service.go
```diff
@@ -34,6 +41,35 @@ type DIDCommMsg struct {
 	ToVerKeys []string
 }
 
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+func (m *DIDCommMsg) ThreadID() (string, error) {
+	msg := struct {
```
troyronda: we also have a thread definition in https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/decorator/decorator.go#L12

was the plan to have other decorator utilities there? @soluchok @rolsonquadras @llorllale

pkg/didcomm/common/service/service.go
```diff
@@ -34,6 +41,35 @@ type DIDCommMsg struct {
...
+		ID     string           `json:"@id"`
+		Thread decorator.Thread `json:"~thread"`
+	}{}
+
+	if err := json.Unmarshal(m.Payload, &msg); err != nil {
```
troyronda: Will each decorator need to do its own Unmarshal? @rolsonquadras @llorllale @soluchok

pkg/didcomm/common/service/service.go
soluchok: I would not say that it is a decorator utility. It combines id and thread id so, I would say it is message utility.

troyronda:
My thoughts:

- It makes sense to keep the decorator model separate (as is done in this PR).
- A message struct should embed the field for the purpose of unmarshal. Most of the time, this should be the protocol's struct who needs the decorators. 
  - There are some (very few) decorators that are effectively mandatory (including Thread). These could make sense in DIDCommMsg as they will be common anyways.
- There should be a decorator utility function ( in some common package(s) ) that can be used by Protocols to make use of the decorator.

Specific to this PR:
- Unmarshal should not be called every time the ThreadID accessor is used.
  - The thread decorator should be unmarshaled with an existing unmarshal.

rolsonquadras:
> * There should be a decorator utility function ( in some common package(s) ) that can be used by Protocols to make use of the decorator.

@troyronda agree with most of the points. How does this util function look like ?

troyronda:
> > * There should be a decorator utility function ( in some common package(s) ) that can be used by Protocols to make use of the decorator.
> 
> @troyronda agree with most of the points. How would this util function look like ?

@rolsonquadras something like decorator.MyUtilityFunction(msg.DecoratorField)  :)

pkg/didcomm/common/service/service.go
```diff
@@ -32,6 +46,33 @@ type DIDCommMsg struct {
...
 	// ToVerKeys are recipient keys
 	ToVerKeys []string
+	Common    *Common
+}
+
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+func (m *DIDCommMsg) ThreadID() (string, error) {
+	if m.Common == nil {
+		if err := json.Unmarshal(m.Payload, &m.Common); err != nil {
```
troyronda: This should be done when the DIDCommMsg is created and not in the accessor.

See https://github.com/hyperledger/aries-framework-go/blob/3dfb648a8ed5745ecb57dcd946de965e8a5a0d87/pkg/framework/context/context.go#L99

pkg/didcomm/common/service/service.go
```diff
@@ -32,6 +46,33 @@ type DIDCommMsg struct {
 	OutboundDestination *Destination
 	// ToVerKeys are recipient keys
 	ToVerKeys []string
+	Common    *Common
```
troyronda: Do we need a separate struct for these fields?

soluchok: I think yeah, bc it will not overlap with existing fields (probably somehow in the future), and the main reason was to keep a pointer to determine was there unmarshal already or not. 
<nil> - means unmarshal was not executed

pkg/didcomm/common/service/service.go
soluchok: I like a constructor but not in this case :)
* too many dependencies (`Type, Payload, ToVerKeys, etc.`)
* we will execute unmarshal event it is no need

troyronda:
> @troyronda agree with most of the points. 

Note: looking at the structure of DIDCommMsg creation to usage in the protocol handlers, the more efficient solution would be to have each protocol struct individually embed the decorator, such that it is populated when each protocol struct is unmarshaled.

Common utility function would be created that take these protocol structs as an argument (as discussed above).

We could also have a core protocol struct that has the mandatory fields and decorators that each protocol struct embeds.

pkg/didcomm/common/service/service.go
troyronda: The Unmarshal is being performed using this struct: https://github.com/hyperledger/aries-framework-go/blob/3dfb648a8ed5745ecb57dcd946de965e8a5a0d87/pkg/framework/context/context.go#L88

Couldn't we add ID and decorator.Thread to it and, hence, re-use the existing Unmarshal?

soluchok: Done!

pkg/didcomm/common/service/service.go
```diff
@@ -30,6 +45,33 @@ type DIDCommMsg struct {
 	Payload  []byte
 	// TODO : might need refactor as per the issue-226
 	OutboundDestination *Destination
+	Common              *Common
+}
+
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+func (m *DIDCommMsg) ThreadID() (string, error) {
+	if m.Common == nil {
```
troyronda: This logic should be removed.

pkg/didcomm/common/service/errors.go
```diff
@@ -11,6 +11,8 @@ const (
 	ErrChannelRegistered = serviceError("channel is already registered for the action event")
 	ErrNilChannel        = serviceError("cannot pass nil channel")
 	ErrInvalidChannel    = serviceError("invalid channel passed to unregister the action event")
+	ErrThreadIDNotFound  = serviceError("threadID not found")
+	ErrInvalidMessage    = serviceError("invalid message")
```
troyronda: We shouldn’t expose errors types that aren’t expected to be checked and handled by consumer code. Are these errors that will be handled?

pkg/didcomm/common/service/service.go
```diff
@@ -20,6 +27,14 @@ type DIDComm interface {
 	Event
 }
 
+// Common keeps helpful or common fields which can be defined in payload,
+// this structure was created to avoid json.Unmarshal twice
```
troyronda: This comment should be updated.

pkg/didcomm/common/service/service.go
```diff
@@ -30,6 +45,33 @@ type DIDCommMsg struct {
 	Payload  []byte
 	// TODO : might need refactor as per the issue-226
 	OutboundDestination *Destination
+	Common              *Common
```
troyronda: The type field is now duplicated (above).

pkg/didcomm/common/service/service.go
```diff
@@ -20,6 +27,14 @@ type DIDComm interface {
 	Event
 }
 
+// Common keeps helpful or common fields which can be defined in payload,
+// this structure was created to avoid json.Unmarshal twice
+type Common struct {
```
troyronda: This looks a bit like header fields that we are extracting from the message.

I don’t think that common is descriptive.

pkg/didcomm/common/service/errors.go
soluchok: why we shouldn't expose errors? For now, we do not check it, but I do not see any problems with that. These errors are safe. (I made them as constants, nobody can change it)

pkg/didcomm/common/service/service.go
soluchok: Done!

pkg/didcomm/common/service/service.go
soluchok: I agree. Done!

pkg/didcomm/common/service/service.go
soluchok: refactored! but this logic still there, check it out

pkg/didcomm/common/service/errors.go
troyronda: Exposed errors become part of the API surface. Only errors that we expect developers to handle may be exposed. Otherwise do not expose them.

pkg/didcomm/common/service/service.go
```diff
@@ -20,16 +27,61 @@ type DIDComm interface {
...
+	Header              *Header
+}
+
+// GetHeader parses and returns Header from payload
+func (m *DIDCommMsg) GetHeader() (*Header, error) {
+	if m.Header != nil {
+		return m.Header, nil
+	}
+
+	if err := json.Unmarshal(m.Payload, &m.Header); err != nil {
```
troyronda: this should be done when DIDComMsg was constructed not in an accessor.

due this being done in an accessor, thread safety concerns are introduced and complexity is increased.

pkg/didcomm/common/service/service.go
```diff
@@ -20,16 +27,61 @@ type DIDComm interface {
...
 	Outbound bool
-	Type     string
 	Payload  []byte
 	// TODO : might need refactor as per the issue-226
 	OutboundDestination *Destination
+	Header              *Header
+}
+
+// GetHeader parses and returns Header from payload
+func (m *DIDCommMsg) GetHeader() (*Header, error) {
```
troyronda: accessors should not be prefixed with Get. Nothing is being “gotten”.

Just Header in this case.

soluchok: I tried, but we need this field to be public. So, in that case, we have a Header field and Header function which lead us to conflict

troyronda: You don’t need both exposed. If you want to expose Header in the struct then you don’t need an accessor.

pkg/didcomm/common/service/service.go
soluchok: Done!

pkg/didcomm/common/service/service.go
```diff
@@ -20,16 +27,55 @@ type DIDComm interface {
...
+	if err := json.Unmarshal(msg.Payload, &msg.Header); err != nil {
+		return nil, fmt.Errorf("invalid payload data format: %w", err)
+	}
+	return msg, nil
+}
+
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+func (m *DIDCommMsg) ThreadID() (string, error) {
+	if m.Header == nil {
```
troyronda: When would this happen?

soluchok: when someone will create `DIDCommMsg{}` instead of using `NewDIDCommMsg(src)`

rolsonquadras: May be disclaimer in the GoDoc would be helpful imo

PR #477, opened 2019-10-11
refactor: Add signer to did exchange service context

### LINKED ISSUE #476: Add signer to didexchange service context
Did exchange service has to sign connection response. Add signer to did exchange service context.

sandrask:
Did exchange service has to sign connection response. Add signer to did exchange service context.

Closes #476

### DIFF
--- a/pkg/wallet/api.go
+++ b/pkg/wallet/api.go
@@ -13,6 +13,7 @@ import (
 // Wallet interface
 type Wallet interface {
 	Crypto
+	Signer
 	DIDCreator
 }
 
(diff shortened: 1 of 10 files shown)


pkg/wallet/api.go
```diff
@@ -15,6 +15,7 @@ import (
 // Wallet interface
 type Wallet interface {
 	Crypto
```
troyronda: So eventually this should have a Crypto, but rather some key creator / management.

We will also need a Decrypter I guess?

sandrask: Yes, we should end up with key management interface and decrypter interface.

Moopli: It doesn't need to have the Decrypt interface either, baha's getting rid of that in his PR because the way we ended up implementing decryption didn't make sense with a function of that name.

You can keep it or remove it, it should be gone soon either way.

troyronda:
I think this is a "refactor:" rather than a "feat:"

PR #480, opened 2019-10-11
feat: Protocol Failure - Update state to abandoned and trigger message notification

### LINKED ISSUE #442: Protocol Failure Notification
Notify the consumer when there is a failure with protocol processing.

Suggestion:
- Use MsgEvent for this

rolsonquadras:
- Update state to abandoned if the processing fails
- Trigger events if the processing fails

Closes #442 

### DIFF
--- a/pkg/didcomm/common/service/event.go
+++ b/pkg/didcomm/common/service/event.go
@@ -35,12 +35,12 @@ type StateMsg struct {
 	// DIDComm message along with message type
 	Msg *DIDCommMsg
 
-	// Properties contains value based on specific protocol. The consumers need to cast this to specific interface
-	// based on the protocol event.
+	// Properties contains value based on specific protocol. The consumers need to call the protocol client
+	// functions to get the data.
 	//
-	// Cast to following interfaces based on protocol
-	//   - DID Exchange :  service.DIDExchangeEvent
-	Properties interface{}
+	// Clients function to retrieve data based on protocol.
+	//   - DID Exchange :  didexchange.EventProperties
+	Properties EventProperties
 }
 
 // DIDCommAction message type to pass events in go channels.
@@ -61,13 +61,16 @@ type DIDCommAction struct {
 	// to stop the processing.
 	Stop func(err error)
 
-	// Properties contains value based on specific protocol. The consumers need to cast this to specific interface
-	// based on the protocol event.
+	// Properties contains value based on specific protocol. The consumers need to call the protocol client
+	// functions to get the data.
 	//
-	// Cast to following interfaces based on protocol
-	//   - DID Exchange :  service.DIDExchangeEvent
-	Properties interface{}
+	// Clients function to retrieve data based on protocol.
+	//   - DID Exchange :  didexchange.EventProperties
+	Properties EventProperties
 }
+
+// EventProperties type for event related data.
+type EventProperties interface{}
 
 // Event event related apis.
 type Event interface {
--- a/pkg/didcomm/protocol/didexchange/event.go
+++ b/pkg/didcomm/protocol/didexchange/event.go
@@ -14,3 +14,33 @@ type Event interface {
 	// invitation ID
 	InvitationID() string
 }
+
+// didExchangeEvent implements didexchange.Event interface.
+type didExchangeEvent struct {
+	connectionID string
+	invitationID string
+}
+
+// ConnectionID returns DIDExchange connectionID.
+func (ex *didExchangeEvent) ConnectionID() string {
+	return ex.connectionID
+}
+
+// InvitationID returns DIDExchange invitationID.
+func (ex *didExchangeEvent) InvitationID() string {
+	return ex.invitationID
+}
+
+// didExchangeEvent for sending events with processing error.
+type didExchangeEventError struct {
+	*didExchangeEvent
+	err error
+}
+
+// Error implements error interface.
+func (ex *didExchangeEventError) Error() string {
+	if ex.err != nil {
+		return ex.err.Error()
+	}
+	return ""
+}
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -145,7 +145,7 @@ func (s *Service) Handle(msg *service.DIDCommMsg) error {
 	// TODO change from thread id to connection id #397
 	// TODO pass invitation id #397
 	s.sendMsgEvents(&service.StateMsg{
-		Type: service.PreState, Msg: msg, StateID: next.Name(), Properties: s.createEventProperties(thid, "")})
+		Type: service.PreState, Msg: msg, StateID: next.Name(), Properties: createEventProperties(thid, "")})
 	logger.Infof("sent pre event for state %s", next.Name())
 
 	// trigger action event based on message type for inbound messages
@@ -187,7 +187,7 @@ func (s *Service) handle(msg *message) error {
 		// TODO pass invitation id #397
 		s.sendMsgEvents(&service.StateMsg{
 			Type: service.PreState, Msg: msg.Msg, StateID: next.Name(),
-			Properties: s.createEventProperties(msg.ThreadID, "")})
+			Properties: createEventProperties(msg.ThreadID, "")})
 		logger.Infof("sent pre event for state %s", next.Name())
 
 		var action stateAction
@@ -214,7 +214,7 @@ func (s *Service) handle(msg *message) error {
 		// TODO pass invitation id #397
 		s.sendMsgEvents(&service.StateMsg{
 			Type: service.PostState, Msg: msg.Msg, StateID: next.Name(),
-			Properties: s.createEventProperties(msg.ThreadID, "")})
+			Properties: createEventProperties(msg.ThreadID, "")})
 		logger.Infof("sent post event for state %s", next.Name())
 
 		next = followup
@@ -222,24 +222,18 @@ func (s *Service) handle(msg *message) error {
 	return nil
 }
 
-func (s *Service) createEventProperties(connectionID, invitationID string) *didExchangeEvent { //nolint: unparam
-	return &didExchangeEvent{connectionID: connectionID, invitationID: invitationID}
+func createEventProperties(connectionID, invitationID string) *didExchangeEvent { //nolint: unparam
+	return &didExchangeEvent{
+		connectionID: connectionID,
+		invitationID: invitationID,
+	}
 }
 
-// didExchangeEvent implements didexchange.Event interface.
-type didExchangeEvent struct {
-	connectionID string
-	invitationID string
-}
-
-// ConnectionID returns DIDExchange connectionID.
-func (ex *didExchangeEvent) ConnectionID() string {
-	return ex.connectionID
-}
-
-// InvitationID returns DIDExchange invitationID.
-func (ex *didExchangeEvent) InvitationID() string {
-	return ex.invitationID
+func createErrorEventProperties(connectionID, invitationID string, err error) *didExchangeEventError { //nolint: unparam
+	return &didExchangeEventError{
+		err:              err,
+		didExchangeEvent: createEventProperties(connectionID, invitationID),
+	}
 }
 
 // sendEvent triggers the action event. This function stores the state of current processing and passes a callback
@@ -274,7 +268,7 @@ func (s *Service) sendActionEvent(msg *service.DIDCommMsg, aEvent chan<- service
 		Stop: func(err error) {
 			s.processCallback(id, err)
 		},
-		Properties: s.createEventProperties(threadID, ""),
+		Properties: createEventProperties(threadID, ""),
 	}
 
 	// trigger the registered action event
@@ -299,13 +293,60 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
 func (s *Service) startInternalListener() {
 	go func() {
 		for msg := range s.callbackChannel {
-			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
-			if err := s.process(msg); err != nil {
-				// TODO handle error
-				logger.Errorf(err.Error())
+			if msg.Err != nil {
+				// TODO https://github.com/hyperledger/aries-framework-go/issues/438 Cleanup/Update data on Stop Event Action
+				logger.Errorf("client action event processing - msgID:%s error:%s", msg.ID, msg.Err)
+				err := s.processFailure(msg.ID, msg.Err)
+				if err != nil {
+					logger.Errorf("process callback : %s", err)
+				}
+				continue
+			}
+
+			// TODO https://github.com/hyperledger/aries-framework-go/issues/242 - retry logic
+
+			if err := s.process(msg.ID); err != nil {
+				procErr := s.processFailure(msg.ID, err)
+				if procErr != nil {
+					logger.Errorf("process callback : %s", procErr)
+				}
 			}
 		}
 	}()
+}
+
+func (s *Service) processFailure(id string, processErr error) error {
+	// get the transient data
+	data, err := s.getTransientEventData(id)
+	if err != nil {
+		return fmt.Errorf("unable to fetch the event transient data: %w", err)
+	}
+
+	err = s.abandon(data.ThreadID, data.Msg, processErr)
+	if err != nil {
+		return fmt.Errorf("abandon : %w", err)
+	}
+
+	return nil
+}
+
+// abandon updates the state to abandoned and trigger failure event.
+func (s *Service) abandon(thid string, msg *service.DIDCommMsg, processErr error) error {
+	// update the state to abandoned
+	err := s.update(thid, &abandoned{})
+	if err != nil {
+		return fmt.Errorf("unable to update the state to abandoned: %w", err)
+	}
+
+	// send the message event
+	s.sendMsgEvents(&service.StateMsg{
+		Type:       service.PostState,
+		Msg:        msg,
+		StateID:    stateNameAbandoned,
+		Properties: createErrorEventProperties(thid, "", processErr),
+	})
+
+	return nil
 }
 
 func (s *Service) processCallback(id string, err error) {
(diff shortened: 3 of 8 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -211,32 +211,21 @@ func (s *Service) handle(msg *message) error {
...
-
-// ConnectionID returns DIDExchange connectionID.
-func (ex *didExchangeEvent) ConnectionID() string {
-	return ex.connectionID
-}
-
-// InvitationID returns DIDExchange invitationID.
-func (ex *didExchangeEvent) InvitationID() string {
-	return ex.invitationID
+func (s *Service) createEventProperties(connectionID,
```
troyronda: I'm on the fence on if the error version of this method should be split to its own method. I assume you had a rationale for this style?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -296,14 +285,44 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
 	go func() {
 		for msg := range s.callbackChannel {
-			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
-			if err := s.process(msg); err != nil {
-				// TODO handle error
-				logger.Errorf(err.Error())
+			if msg.Err != nil {
+				// TODO https://github.com/hyperledger/aries-framework-go/issues/438 Cleanup/Update data on Stop Event Action
+				logger.Errorf("client action event processing failed - msgID:%s error:%s", msg.ID, msg.Err)
+				s.processCallbackFailure(msg.ID, msg.Err)
```
troyronda: Are these really "callback" failures?

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: I think it makes sense to split this method. ll update the code.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: At the moment, only callback failures are handled. But the method support any protocol failure. Will update this.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/c07ed53802da44eef1d1921b3f421ce1d620ab41..5562ad35e21a3e5cb071507e1e0854c0d9317794#diff-f761c91e753eba1c42a68076df06ce12L222).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/pull/480/files#diff-f761c91e753eba1c42a68076df06ce12R334).

pkg/didcomm/protocol/didexchange/event.go
```diff
@@ -14,3 +14,30 @@ type Event interface {
...
+
+// didExchangeEvent for sending events with processing error.
+type didExchangeEventError struct {
+	*didExchangeEvent
+	err error
+}
+
+// Error implements error interface.
+func (ex *didExchangeEventError) Error() string {
+	return ex.err.Error()
```
sudeshrshetty: probability of panic, is this possible it can be created with nil error?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -215,32 +215,29 @@ func (s *Service) handle(msg *message) error {
...
-// didExchangeEvent implements didexchange.Event interface.
-type didExchangeEvent struct {
-	connectionID string
-	invitationID string
-}
-
-// ConnectionID returns DIDExchange connectionID.
-func (ex *didExchangeEvent) ConnectionID() string {
-	return ex.connectionID
+func createEventProperties(connectionID, invitationID string) interface{} { //nolint: unparam
```
sudeshrshetty: Why we need to return interface here, can't we return - EventProperties type?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -215,32 +215,29 @@ func (s *Service) handle(msg *message) error {
...
+		connectionID: connectionID,
+		invitationID: invitationID,
+	}
 }
 
-// InvitationID returns DIDExchange invitationID.
-func (ex *didExchangeEvent) InvitationID() string {
-	return ex.invitationID
+func createErrorEventProperties(connectionID, invitationID string, err error) interface{} { //nolint: unparam
+	return &didExchangeEventError{
```
sudeshrshetty: call `createEventProperties` and then construct `didExchangeEventError{}` easy to maintain.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -300,47 +297,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
 }
 
+func (s *Service) processCallbackFailure(id string, processErr error) error {
+	// get the transient data
+	data, err := s.getTransientEventData(id)
+	if err != nil {
+		return fmt.Errorf("unable to fetch the event transient data: %w", err)
+	}
+
+	err = s.processFailure(data.ThreadID, data.Msg, processErr)
```
sudeshrshetty: just `return s.processFailure(data.ThreadID, data.Msg, processErr)`?  it will improve coverage.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -300,47 +297,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
+func (s *Service) process(id string) error {
+	data, err := s.getTransientEventData(id)
+	if err != nil {
+		return fmt.Errorf("unable to fetch the event transient data: %w", err)
 	}
 
-	// fetch the record
-	jsonDoc, err := s.store.Get(msg.ID)
+	// continue the processing
+	err = s.handle(data)
```
sudeshrshetty: same here, just `return s.handle(data)`

pkg/client/didexchange/client.go
```diff
@@ -259,3 +259,19 @@ func (c *Client) handleMessageEvent(msg *service.StateMsg) {
 		handler <- *msg
 	}
 }
+
+// EventProperties returns the DIDExchange protocol related data along with any processing errors. Processing error
+// will have value only in case of abandoned state.
+func EventProperties(e service.EventProperties) (didexchange.Event, error, error) {
```
soluchok: Do not do this `(didexchange.Event, error, error)` only one error should be returned

pkg/client/didexchange/client_test.go
```diff
@@ -387,3 +387,49 @@ func TestService_MsgEvents(t *testing.T) {
...
+	ev, svcErr, err := EventProperties(&mockDIDExchangeEvent{})
+	require.NoError(t, err)
+	require.Nil(t, svcErr)
+	require.NotNil(t, ev.ConnectionID())
+	require.NotNil(t, ev.InvitationID())
+
+	// service error
+	ev, svcErr, err = EventProperties(&mockDIDExchangeEventError{})
+	require.NoError(t, err)
+	require.NotNil(t, svcErr)
```
soluchok: I don't like it `require.NotNil(t, svcErr)` 
I would rather use `Contains or EqualError`  functions and check the specific error message. It will be clear to understand what you expect here and it will prevent misunderstanding after some changes in the future.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -299,47 +296,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
+	err := s.update(thid, &abandoned{})
+	if err != nil {
+		return fmt.Errorf("unable to update the state to abandoned: %w", err)
+	}
+
+	// send the message event
+	s.sendMsgEvents(&service.StateMsg{
+		Type:       service.PostState,
+		Msg:        msg,
+		StateID:    (&abandoned{}).Name(),
```
soluchok: Use constant instead of `&abandoned{}).Name()`

pkg/didcomm/protocol/didexchange/event.go
```diff
@@ -14,3 +14,30 @@ type Event interface {
...
+	return ex.connectionID
+}
+
+// InvitationID returns DIDExchange invitationID.
+func (ex *didExchangeEvent) InvitationID() string {
+	return ex.invitationID
+}
+
+// didExchangeEvent for sending events with processing error.
+type didExchangeEventError struct {
```
soluchok: Do not understand why we need `didExchangeEventError` structure, we can keep an error in `didExchangeEvent`

pkg/didcomm/common/service/event.go
```diff
@@ -61,14 +61,17 @@ type DIDCommAction struct {
...
-	// Cast to following interfaces based on protocol
-	//   - DID Exchange :  service.DIDExchangeEvent
-	Properties interface{}
+	// Clients function to retrieve data based on protocol.
+	//   - DID Exchange :  didexchange.EventProperties
+	Properties EventProperties
 }
 
+// EventProperties type for event related data.
+type EventProperties interface{}
```
soluchok: This does nothing, we still have the interface. Just more confusing and dependency in the client pkg e.g `EventProperties(e service.EventProperties)`

pkg/didcomm/protocol/didexchange/service.go
soluchok: @sudeshrshetty  do not write code because of coverage :) sometimes it is hard to detect where errors happen and wrap them with a custom message it's cool. In that case, I do not see the usage of this wrapper so, yeah we can simplify it :)

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -299,47 +296,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
 	go func() {
 		for msg := range s.callbackChannel {
-			// TODO handle error in callback - https://github.com/hyperledger/aries-framework-go/issues/242
-			if err := s.process(msg); err != nil {
-				// TODO handle error
-				logger.Errorf(err.Error())
+			if msg.Err != nil {
+				// TODO https://github.com/hyperledger/aries-framework-go/issues/438 Cleanup/Update data on Stop Event Action
+				logger.Errorf("client action event processing failed - msgID:%s error:%s", msg.ID, msg.Err)
+				err := s.processCallbackFailure(msg.ID, msg.Err)
```
llorllale: Suggest we rename this to "abandon" or something similar

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -299,47 +296,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
+				if err != nil {
+					logger.Errorf("process callback failure: %s", err)
+				}
+				continue
+			}
+
+			// TODO https://github.com/hyperledger/aries-framework-go/issues/242 - retry logic
+
+			if err := s.process(msg.ID); err != nil {
+				procErr := s.processCallbackFailure(msg.ID, err)
```
llorllale: I expect this to be refactored in #242 such that the protocol instance isn't automatically abandoned due to some internal error

pkg/didcomm/protocol/didexchange/service.go
sudeshrshetty: I agree @soluchok

pkg/didcomm/protocol/didexchange/event.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/abea54bccf94c6f84b362bc1673c2639e2ecfa9c..05e7cc5e7e99a90abc09704d05394f0fb8ea61ba#diff-79de382a052bbdb4213b9dd373b0e2f6R42).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/abea54bccf94c6f84b362bc1673c2639e2ecfa9c..05e7cc5e7e99a90abc09704d05394f0fb8ea61ba#diff-f761c91e753eba1c42a68076df06ce12R225) - returned the actual struct.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Done](https://github.com/hyperledger/aries-framework-go/compare/abea54bccf94c6f84b362bc1673c2639e2ecfa9c..05e7cc5e7e99a90abc09704d05394f0fb8ea61ba#diff-f761c91e753eba1c42a68076df06ce12R235).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: @sudeshrshetty I wanted to wrap the error here.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: @sudeshrshetty I wanted to wrap the error here.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: I feel better to use the method as it'll take care of change in state name. If we point to a constant, then we might need to refactor (provided referring to 2 different constants).

pkg/didcomm/protocol/didexchange/event.go
rolsonquadras: didExchangeEventError - implements error interface. This is done as client can differentiate between type ie, error or not. The base struct provides protocol specific info like connectionID and invitationID.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Yes, here is the [TODO](https://github.com/hyperledger/aries-framework-go/pull/480/files#diff-f761c91e753eba1c42a68076df06ce12R306).

pkg/didcomm/common/service/event.go
rolsonquadras: This is consumed by user and they invoke client.EventProperties util to fetch the actual details. This way the framework would have a way to change the actual type without affecting the users.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Any specific suggestions ?

pkg/didcomm/protocol/didexchange/event.go
soluchok: yeah I saw, but my point is that error can also be determined by calling `.Err() error` function, it also can be like protocol-specific info

pkg/client/didexchange/client.go
rolsonquadras: Changed the type of second return type to [service.ErrEventProcess]( https://github.com/hyperledger/aries-framework-go/compare/3c7cd17e91d97a198ad9fbe98046c1373efc7d6c..d215a24d2cb11e4f38c59581ccdcb6a73fc7d364#diff-427522b353a28b1ec350c29c79d387b0R271). 

I didn't add to combine both errors into one (ultimately consumer can switch based on error type) as consumer will ignore the first type (didexchange event props - connectionID and invitationID). Here user can check for error and then process the logic based on correlation in didexchange.Event and service.ErrEventProcess(if any - In case of state type is abandoned).

pkg/client/didexchange/client_test.go
rolsonquadras: I missed to add it. Added in [new patch](https://github.com/hyperledger/aries-framework-go/pull/480/files#diff-415ff83c536dcba8813f0209c1104a1dR399).

pkg/didcomm/protocol/didexchange/event.go
rolsonquadras: Yeah, I think it's design decision based on how we want to do it - validating against type or .Err()!= nil.

pkg/client/didexchange/client.go
soluchok: I still do not like it, I would rather split it into two different functions. The logic here has two different behavior what is a bit confused.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Missed the point. Updated the [name](https://github.com/hyperledger/aries-framework-go/compare/b3d55db4c56ee239eef3f187fef8efba3642ff2b..316bca19317a19ab632e20dc1d3f2dca858ebbb3#diff-f761c91e753eba1c42a68076df06ce12R334).

pkg/client/didexchange/client.go
rolsonquadras: Removed this util function as consumer need to do check.

pkg/didcomm/protocol/didexchange/event.go
troyronda: Validating against type is better.

Create a specific data structure for a particular codepath. (i'm not a fan of optional fields that are only filled in a certain codepath.)

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -299,47 +293,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
 func (s *Service) startInternalListener() {
 	go func() {
```
troyronda: I guess I missed this earlier - I would keep the go statement outside ( no nested go func() {} )

pkg/didcomm/protocol/didexchange/service.go
troyronda: I also think a constant is better here.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -299,47 +293,97 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
+		return fmt.Errorf("unable to fetch the event transient data: %w", err)
 	}
 
-	// fetch the record
-	jsonDoc, err := s.store.Get(msg.ID)
+	// continue the processing
+	err = s.handle(data)
 	if err != nil {
-		return fmt.Errorf("document for the id doesn't exists in the database: %w", err)
+		return fmt.Errorf("processing of the message failed: %w", err)
```
troyronda: do we need to have every wrapping error string say failed? :)

keep it succinct.

test/bdd/agent_steps.go
```diff
@@ -85,10 +85,13 @@ func (a *AgentSteps) registerPostMsgEvent(agentID, statesValue string) error {
 	states := strings.Split(statesValue, ",")
 	a.initializeStates(agentID, states)
 	go func() {
```
troyronda: (from a previous PR) I think it is generally cleaner to create a func and call `go funcname()`, rather than embed like this.

rolsonquadras: yeah, I'll update this existing code.

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Removed](https://github.com/hyperledger/aries-framework-go/compare/807adcf62e0ac91e48d81a5b15a62d91e073bc78..0360003cb9f51624a1493097c2398be2113925f0#diff-f761c91e753eba1c42a68076df06ce12R298) failed word :)

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: [Done](https://github.com/hyperledger/aries-framework-go/compare/807adcf62e0ac91e48d81a5b15a62d91e073bc78..0360003cb9f51624a1493097c2398be2113925f0#diff-f761c91e753eba1c42a68076df06ce12R345).

PR #482, opened 2019-10-11
refactor: Split service.Handle function for Inbound/Outbound type

### LINKED ISSUE #490: Split service.Handle function for Inbound/Outbound type
The [service.Handle](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/service.go#L112) function handles both inbound messages from dispatcher as well as outbound flow as described in #500.  

The service should have below APIs.
- HandleInbound(DIDCommMsg)
- HandleOutbound(DIDCommMsg, Destination)

rolsonquadras:
- Add following APIs to handle the messages in service layer.
	- HandleInbound(DIDCommMsg)
	- HandleOutbound(DIDCommMsg, Destination)
- Remove outbound related fields from DIDCommMsg struct

Closes #490 

### DIFF
--- a/pkg/didcomm/common/service/service.go
+++ b/pkg/didcomm/common/service/service.go
@@ -15,7 +15,8 @@ import (
 
 // Handler provides protocol service handle api.
 type Handler interface {
-	Handle(msg *DIDCommMsg) error
+	// HandleInbound handles inbound didexchange messages.
+	HandleInbound(msg *DIDCommMsg) error
 }
 
 // DIDComm defines service APIs.
@@ -36,14 +37,8 @@ type Header struct {
 
 // DIDCommMsg did comm msg
 type DIDCommMsg struct {
-	// Outbound indicates the direction of this DIDComm message:
-	//   - outgoing (to another agent)
-	//   - incoming (from another agent)
-	Outbound bool
-	Payload  []byte
-	// TODO : might need refactor as per the issue-226
-	OutboundDestination *Destination
-	Header              *Header
+	Header  *Header
+	Payload []byte
 }
 
 // NewDIDCommMsg returns DIDCommMsg with Header
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -69,6 +69,14 @@ type connectionStore interface {
 	GetConnection(connectionID string) (*ConnectionRecord, error)
 }
 
+// stateMachineMsg is an internal struct used to pass data to state machine.
+type stateMachineMsg struct {
+	outbound            bool
+	outboundDestination *service.Destination
+	header              *service.Header
+	payload             []byte
+}
+
 // Service for DID exchange protocol
 type Service struct {
 	service.Action
@@ -108,14 +116,14 @@ func New(didMaker did.Creator, prov provider) (*Service, error) {
 	return svc, nil
 }
 
-// Handle didexchange msg
-func (s *Service) Handle(msg *service.DIDCommMsg) error {
+// HandleInbound handles inbound didexchange messages.
+func (s *Service) HandleInbound(msg *service.DIDCommMsg) error {
 	// throw error if there is no action event registered for inbound messages
 	aEvent := s.GetActionEvent()
 
-	logger.Infof("entered into Handle exchange message : %s", msg.Payload)
+	logger.Infof("entered into HandleInbound exchange message : %s", msg.Payload)
 
-	if !msg.Outbound && aEvent == nil {
+	if aEvent == nil {
 		return errors.New("no clients are registered to handle the message")
 	}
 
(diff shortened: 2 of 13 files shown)


pkg/didcomm/common/service/service.go
```diff
@@ -20,16 +20,24 @@ type DIDComm interface {
...
-// DIDCommMsg did comm msg
+// DIDCommMsg is used by handlers/dispatchers/events to pass DIDComm messages.
 type DIDCommMsg struct {
+	// message type
+	Type string
+	// message
+	Payload []byte
+}
+
+// Msg is used for outbound related features.
```
troyronda: There is a boolean inside this struct that could mean incoming?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -112,7 +112,7 @@ func (s *Service) Handle(msg *service.DIDCommMsg) error {
 
 	logger.Infof("entered into Handle exchange message : %s", msg.Payload)
 
-	if !msg.Outbound && aEvent == nil {
```
troyronda: was the msg.Outbound implied?

pkg/didcomm/common/service/service.go
rolsonquadras: The boolean is not set anywhere outside the service layer. It look to manipulate the state machine operations. e.r. [here](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/states.go#L218).

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/0b5ffafa1afbb9931651ae126eeaf822901dda9f..5826f9a5bccaf6a6bfdbe2bfa90f56f34194147e#diff-75342fbf986feea5a6bb6a33d0caff76R31) the comment (added state machine).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: Outbound flag is used in state machine. The inbound dispatcher never sets this (always false) https://github.com/hyperledger/aries-framework-go/pull/482#discussion_r334157477.

pkg/didcomm/common/service/service.go
```diff
@@ -20,16 +20,24 @@ type DIDComm interface {
...
+// DIDCommMsg is used by handlers/dispatchers/events to pass DIDComm messages.
 type DIDCommMsg struct {
+	// message type
+	Type string
+	// message
+	Payload []byte
+}
+
+// Msg is used for state machine processing.
+type Msg struct {
```
troyronda: Perhaps we should have a TODO for splitting inbound and outbound type.

pkg/didcomm/protocol/didexchange/service.go
troyronda: Hmmm. I think the way this code is being refactored implies that the same type shouldn’t be used... commented above.

pkg/didcomm/common/service/service.go
rolsonquadras: Created an issue #490 and added a [TODO](https://github.com/hyperledger/aries-framework-go/pull/482/files#diff-75342fbf986feea5a6bb6a33d0caff76R32).

rolsonquadras: Updated the changes in this PR.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -69,6 +69,14 @@ type connectionStore interface {
 	GetConnection(connectionID string) (*ConnectionRecord, error)
 }
 
+// stateMachineMsg is an internal struct used to pass data to state machine.
+type stateMachineMsg struct {
```
troyronda: are we planning to keep a combined inbound/outbound structure within the state machine?

rolsonquadras: We can separate [Execute](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/states.go#L54) function into ExecuteInbound and ExecuteOutbound.

rolsonquadras: Created an issue - https://github.com/hyperledger/aries-framework-go/issues/501

pkg/didcomm/common/service/service.go
```diff
@@ -15,7 +15,8 @@ import (
 
 // Handler provides protocol service handle api.
 type Handler interface {
-	Handle(msg *DIDCommMsg) error
+	// HandleInbound handles inbound didexchange messages.
+	HandleInbound(msg *DIDCommMsg) error
```
llorllale: HandleOutbound should've been added to this interface as well.

Issue #464, opened 2019-10-10
Introduce protocol service implementation


soluchok:
Need to implement service methods for introduce protocol

fqutishat:
epic already exist #269

PR #441, opened 2019-10-08
feat: namespace capability for store providers

sudeshrshetty:
- added namespace capabilities for store providers
- GetStoreHandler() changed to OpenStore(name)
- added CloseStore(name) to close specific store handle
- added Close() to close all opened store handles for a store provider
- Each protocol services to have their own store provider namespace
- connection record keys to have prefixes
- closes #433 and #434

Signed-off-by: sudesh.shetty <sudesh.shetty@securekey.com>

PR #498, opened 2019-10-16

soluchok:
* `HandleInbound` and `stateFromName` methods implementation
* state interface was changed (Execute function was separated into ExecuteInbound and ExecuteOutbound)

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -7,8 +7,13 @@ SPDX-License-Identifier: Apache-2.0
 package introduce
 
 import (
+	"errors"
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/common/metadata"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -26,28 +31,173 @@ const (
 	AckMsgType = IntroduceSpec + "ack"
 )
 
+// message type to store data for eventing. This is retrieved during callback.
+type message struct {
```
llorllale: Can we rename this to something more meaningful? I know it's named that way in `didexchange.Service` - that is mistake.

Would this be better named as "EventMetadata"? Or simply "Event"?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -26,28 +31,173 @@ const (
...
 // New returns introduce service
-func New() *Service {
-	return &Service{}
+func New(s storage.Provider) (*Service, error) {
+	store, err := s.OpenStore(Introduce)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: storage should be closed e.g s.CloseStore(Introduce)
```
llorllale: I don't think the store should be closed in these lower layers. A store might be shared across multiple packages/layers.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -26,28 +31,173 @@ const (
...
+// sendEvent triggers the action event. This function stores the state of current processing and passes a callback
+// function in the event message.
+func (s *Service) sendActionEvent(msg *service.DIDCommMsg, aEvent chan<- service.DIDCommAction, nextState state) error {
 	return nil
 }
 
+func stateFromMsgType(msg *service.DIDCommMsg) (state, error) {
+	switch msg.Header.Type {
+	case ProposalMsgType:
+		if msg.Outbound {
```
llorllale: Because of this, suggest rename `stateFromMsgType` -> `nextState`

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -26,28 +31,173 @@ const (
...
+			return &deciding{}, nil
+		}
+		return &arranging{}, nil
+	default:
+		return nil, fmt.Errorf("unrecognized msgType: %s", msg.Header.Type)
```
llorllale: I assume this default behaviour is temporary and that subsequent PRs will add cases for the other message types. Confirm?

pkg/didcomm/protocol/introduce/service.go
soluchok: `Should be closed` I meant we need to provide a server method `Close` or `Stop`  which will close storage as well (and some other things e.g close listener etc.). We opened it and it should be our responsibility to close it (by providing a server method `Stop`). Or we can replace  `storage.Provider to storage.Store` in that case we do not care about closing it.

pkg/didcomm/protocol/introduce/service.go
soluchok: yeah, I did only one to cover the test case scenario. But the logic here is correct 
```
if msg.Outbound {
   return &deciding{}, nil
}
return &arranging{}, nil
```
also, it will depend on `wait_count` as well

pkg/didcomm/protocol/introduce/service.go
soluchok: Done! `eventMetadata` bc it is kind of internal stuff

pkg/didcomm/protocol/introduce/service.go
soluchok: @troyronda @llorllale @rolsonquadras @fqutishat I would rather use `storage.Store`. Is there any purpose why it was done by giving a whole storage provider?
In that case, it is much easier to satisfy the interface and the client can manipulate with the storage.

rolsonquadras: @soluchok store provider was added as part of https://github.com/hyperledger/aries-framework-go/pull/441 to provide namespace support instead of dumping data in one place.

soluchok: @rolsonquadras  Just tell me why I need to receive a `storage.Provider` instead of a `storage.Store`? What is a use-case for that? I see only complexity

soluchok: It looks like that we provide portability to the service to create a few stores instead of one. There is no protection. I do not think that protocols will communicate with each other using that trick.

troyronda: Protocols should use a Store that is namespaced to the protocol. Having a decoupled design is an objective and keeping storage namespaced is part of that goal.

The thought behind a Storage Provider is that the protocol, itself, should supply the namespace string. Hence the protocol calls OpenStore with that namespace string.

It is up to the Storage Provider to decide how to deal with namespaces.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -26,28 +31,180 @@ const (
 	AckMsgType = IntroduceSpec + "ack"
 )
 
+var (
+	_ service.Handler = &Service{}
```
troyronda: This line could use some explanation.

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -70,8 +75,12 @@ func (s *start) CanTransitionTo(next state) bool {
 	return next.Name() == stateNameArranging || next.Name() == stateNameDelivering || next.Name() == stateNameDeciding
 }
 
-func (s *start) Execute(msg *service.DIDCommMsg) (state, error) {
-	return nil, errors.New("start execute: not implemented yet")
+func (s *start) ExecuteInbound(msg *service.DIDCommMsg) (state, error) {
+	return nil, errors.New("start ExecuteInbound: not implemented yet")
+}
+
+func (s *start) ExecuteOutbound(_ *service.DIDCommMsg) (state, error) {
```
troyronda: @rolsonquadras @llorllale @soluchok Are we also planning to split Execute in the DID Exchange protocol implementation? Where does the outbound destination become needed?

pkg/didcomm/protocol/introduce/service.go
soluchok: I added it to be sure that Service satisfies service.Handler. If someone changed the interface service.Handler it will raise an error

troyronda: Why wouldn't that be in a test file?

pkg/didcomm/protocol/introduce/states.go
soluchok: yeah, there is an issue https://github.com/hyperledger/aries-framework-go/issues/501

pkg/didcomm/protocol/introduce/service.go
soluchok: it is not a part of the test, it is a guarantee that service satisfies the common interface.

troyronda: We shouldn't instantiate a variable for this purpose.

pkg/didcomm/protocol/introduce/states.go
rolsonquadras: Execute Outbound should take `service.Destination` arg as well.

pkg/didcomm/protocol/introduce/service.go
soluchok: Done!

PR #510, opened 2019-10-17
refactor: added key type in keypair

### LINKED ISSUE #272: Encrypted interface - add key type argument
current [Crypter](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/crypto/crypter.go#L15) interface needs to be updated to add the key type

this will help implementations to validate the key type prior to proceeding with encryption/decryption

and ensures its users are passing in valid keys.

### LINKED ISSUE #488: store both Encryption and Signing key in the wallet
Currently the wallet is storing encryption key pair, update the store to add signing key pair as well.

When creating a new store entry in the wallet, do the signing to encryption `key conversion` at this time only.

Finally remove any key conversion from LegacyCrypter once the above is done.

Baha-sk:

Now the Packager accepts Signature
	Keys and pass them on to the Crypter
	which will in turn convert them to
	Encrytpion Keys.

Closes #272 
Closes #488

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -96,15 +96,14 @@ func New(ctx provider) (*Client, error) {
 
 // CreateInvitation create invitation
 func (c *Client) CreateInvitation(label string) (*Invitation, error) {
-	verKey, err := c.wallet.CreateEncryptionKey()
+	_, sigPubKey, err := c.wallet.CreateKeySet()
 	if err != nil {
 		return nil, fmt.Errorf("failed CreateSigningKey: %w", err)
 	}
-
 	invitation := &didexchange.Invitation{
 		ID:              uuid.New().String(),
 		Label:           label,
-		RecipientKeys:   []string{verKey},
+		RecipientKeys:   []string{sigPubKey},
 		ServiceEndpoint: c.inboundTransportEndpoint,
 		Type:            didexchange.InvitationMsgType,
 	}
--- a/pkg/didcomm/crypto/jwe/authcrypt/authcrypt_test.go
+++ b/pkg/didcomm/crypto/jwe/authcrypt/authcrypt_test.go
@@ -8,14 +8,14 @@ package authcrypt
 
 import (
 	"bytes"
+	"crypto/ed25519"
+	"crypto/rand"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	chacha "golang.org/x/crypto/chacha20poly1305"
-	"golang.org/x/crypto/nacl/box"
 
 	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
 	mockwallet "github.com/hyperledger/aries-framework-go/pkg/internal/mock/wallet"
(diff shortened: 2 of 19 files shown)


pkg/internal/cryptoutil/utils.go
```diff
@@ -25,6 +25,24 @@ var errEmptyRecipients = errors.New("empty recipients")
 // errInvalidKeypair is used when a keypair is invalid
 var errInvalidKeypair = errors.New("invalid keypair")
 
+// SignatureAlgorithm represents a signature algorithm.
+type SignatureAlgorithm string
+
+// ContentEncryption represents a content encryption algorithm.
+type ContentEncryption string
```
troyronda: The naming compared to L29 seems inconsistent

pkg/internal/cryptoutil/utils.go
```diff
@@ -49,6 +68,8 @@ func IsChachaKeyValid(key []byte) bool {
 
 // KeyPair represents a private/public key pair
 type KeyPair struct {
+	// Type is the type of keys enclosed in this key pair
+	Type string
```
troyronda: Why string?

Baha-sk: because it could be either a signature or encryption key type, we can't enforce struct type in this case

pkg/internal/cryptoutil/utils.go
Baha-sk: updated

pkg/internal/cryptoutil/utils.go
Baha-sk: or we create two KeyPair types  and remove Type altogether

pkg/internal/cryptoutil/utils.go
```diff
@@ -50,20 +69,78 @@ func IsChachaKeyValid(key []byte) bool {
...
+}
+
+// SigKeyPair represents a private/public signature (verification) key pair
+type SigKeyPair struct {
+	KeyPair `json:"keypair,omitempty"`
+	// Alg is the signature algorithm of keys enclosed in this key pair
+	Alg SignatureAlgorithm `json:"alg,omitempty"`
+}
+
+// KeyPairCombo represents a pair of key pairs, one for encryption and one for signature
```
troyronda: Combo doesn't have the right ring to it.

pkg/internal/cryptoutil/utils.go
```diff
@@ -50,20 +69,78 @@ func IsChachaKeyValid(key []byte) bool {
...
 	return true
 }
 
+// IsEncKeyPairValid is a utility function that validates an EncKeyPair
+func IsEncKeyPairValid(kp *EncKeyPair) bool {
+	if !isKeyPairValid(kp.KeyPair) {
+		return false
+	}
+
+	switch kp.Alg {
```
troyronda: Why would the util package know all valid types? If it does know all valid types, why is a string type used?

pkg/wallet/api.go
```diff
@@ -56,6 +47,9 @@ type Crypto interface {
 	//
 	//		in case of error, the index will be -1
 	FindVerKey(candidateKeys []string) (int, error)
+
+	// GetEncryptionKey will return the public encryption key corresponding to the public verKey argument
+	GetEncryptionKey(verKey []byte) ([]byte, error)
```
troyronda: I found this method name a bit confusing.

pkg/client/didexchange/client.go
```diff
@@ -96,15 +99,16 @@ func New(ctx provider) (*Client, error) {
...
 // CreateInvitation create invitation
 func (c *Client) CreateInvitation(label string) (*Invitation, error) {
-	verKey, err := c.wallet.CreateEncryptionKey()
+	logger.Debugf("In CreateInvitation about to call CreateKeyPairsCombo ...")
+	_, sigPubKey, err := c.wallet.CreateKeyPairsCombo()
 	if err != nil {
 		return nil, fmt.Errorf("failed CreateSigningKey: %w", err)
 	}
-
+	logger.Debugf("In CreateInvitation CreateKeyPairsCombo success")
```
troyronda: we shouldn’t treat debug log level like an execution tracer :)

Baha-sk: I will remove logger

pkg/internal/cryptoutil/utils.go
Baha-sk: I will rename to `MessagingKeys`

pkg/internal/cryptoutil/utils.go
Baha-sk: it's `cryptoutil` package which serves as a utility to tell what algorithms are supported by the framework

`Alg` in `EnKeyPair.KeyPair.Alg` is of type `EncryptionAlgorithm`
while in `SigKeyPair.KeyPair.Alg` is of type `SignatureAlgorithm`

so the comment type for both is string, hence `Alg` type is string

does it make sense to pull `Alg` up? this way it will be properly typed in both `EncKeyPair` and `SigKeyPair`.

pkg/wallet/api.go
Baha-sk: it's to get the encryption key corresponding to the signing key

pkg/wallet/api.go
```diff
@@ -20,23 +20,14 @@ type Wallet interface {
 // Crypto interface
 type Crypto interface {
 
-	// CreateEncryptionKey create a new public/private encryption keypair.
+	// CreateKeyPairsCombo create a new public/private encryption and signagure keypairs combo.
```
troyronda: To expand on my "Combo" comment. I would treat this situation more similarly to gpg.

You are simply creating a (messaging) key. Like gpg:
- The encryption subkey happens to be created at the same time. 
- Subkey is associated to a primary key.
- This method should return an identifier for the primary key. This identifier should be derived from the public key.
- You should be able to lookup keys by key identifier.

Baha-sk: sounds good, then I will restructure this to be a messaging key that represents the signing key and having the encryption key-pair as a subkey.

key identifier will be the public signing key as it will be referenced in DID docs and DIDExchange/Invitation messages as well.

Moopli: And it's important that the identifier be derived deterministically from the (signing) public key, since it's the (signing) public keys that will be packed into an envelope header, which the recipient will need to look up in the KMS metadata to find a usable decryption key.

pkg/didcomm/crypto/jwe/authcrypt/authcrypt_test.go
```diff
@@ -8,56 +8,121 @@ package authcrypt
...
+	t.Logf("sender Encryption key priv: %v", base64.RawURLEncoding.EncodeToString(sender.EncKeyPair.Priv))
 
-	ecKeyPub, ecKeyPriv, err = box.GenerateKey(randReader)
+	sigPubKey, sigPrivKey, err = ed25519.GenerateKey(rand.Reader)
 	require.NoError(t, err)
-	recipient1Kp := cryptoutil.KeyPair{Priv: ecKeyPriv[:], Pub: ecKeyPub[:]}
-	t.Logf("recipient1Kp pub: %v", base64.RawURLEncoding.EncodeToString(recipient1Kp.Pub))
-	t.Logf("recipient1Kp priv: %v", base64.RawURLEncoding.EncodeToString(recipient1Kp.Priv))
+	// convert signing keys to encryption keys
+	encPubKey, err = cryptoutil.PublicEd25519toCurve25519(sigPubKey)
```
Moopli: Maybe pull this `KeyPairCombo` construction logic out to a helper function in this file

pkg/internal/cryptoutil/utils.go
```diff
@@ -50,20 +69,78 @@ func IsChachaKeyValid(key []byte) bool {
...
+	case Curve25519:
+		return true
+	default:
+		return false
+	}
+}
+
+// IsSigKeyPairValid is a utility function that validates an EncKeyPair
+func IsSigKeyPairValid(kp *SigKeyPair) bool {
```
Moopli: Validate key sizes too - 64 byte priv, 32 byte pub for Ed25519

pkg/internal/cryptoutil/utils.go
```diff
@@ -50,20 +69,78 @@ func IsChachaKeyValid(key []byte) bool {
...
+func isKeyPairValid(kp KeyPair) bool {
 	if kp.Priv == nil || kp.Pub == nil {
 		return false
 	}
 
 	return true
 }
 
+// IsEncKeyPairValid is a utility function that validates an EncKeyPair
+func IsEncKeyPairValid(kp *EncKeyPair) bool {
```
Moopli: Validate key sizes too - 32 byte priv, 32 byte pub for Curve25519

pkg/wallet/api.go
Moopli: maybe `GetSubKey(keyIdentifier, subKeyType)`?

pkg/internal/cryptoutil/utils.go
Baha-sk: there's already `IsChachaKeyValid()` that checks for this

pkg/wallet/api.go
Baha-sk: created #596 to deal with the new structure and creating a deterministic id for the keys

pkg/internal/cryptoutil/utils.go
Baha-sk: Legacy has this check somewhere.. I believe this can be enhanced when working on KMS and the new keys structure

pkg/wallet/api.go
Baha-sk: this will be refactored as part of #596

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -392,6 +392,8 @@ func (ctx *context) handleInboundInvitation(invitation *Invitation,
 	connRec.MyDID = request.Connection.DID
 	// send the exchange request
 	return func() error {
+		logger.Infof("about to send request in 'handleInboundInvitation' - SendVerKey: %+v - Destination: %+v",
```
troyronda: This shouldn't be Info

pkg/wallet/api.go
```diff
@@ -20,23 +20,14 @@ type Wallet interface {
...
-	//
-	// Returns:
-	//
-	// string: verKey
-	//
+	// string: enc public key of the encryption keypair
+	// string: sig public key of the signature keypair
 	// error: error
-	CreateSigningKey() (string, error)
+	CreateKeyPairSet() (string, string, error)
```
troyronda: TODO: refactor. Don't return two keys but rather a KeySet.

I would consider shortening to CreateKeySet.

pkg/internal/cryptoutil/utils.go
troyronda: So this also needs to be refactored.

- A particular crypto implementation will support some set of algorithms.
- From the available crypto implementation, a particular KMS implementation will support some set of algorithms.
- The framework will get it's supported list from Crypto and KMS. Framework should not care about algorithm IDs - it's simply a string.
- A particular package that uses KMS might care to only allow a subset of what's available. (or fail if nothing it can use is available.)

pkg/didcomm/protocol/didexchange/states.go
Baha-sk: removed tracing log lines

pkg/wallet/api.go
Baha-sk: the issue is I don't want to return a keyset that might include private keys... this will get refactored altogether in #596

pkg/client/didexchange/client.go
```diff
@@ -96,15 +96,14 @@ func New(ctx provider) (*Client, error) {
 
 // CreateInvitation create invitation
 func (c *Client) CreateInvitation(label string) (*Invitation, error) {
-	verKey, err := c.wallet.CreateEncryptionKey()
+	_, sigPubKey, err := c.wallet.CreateKeyPairSet()
```
Moopli: @sandrask it creates the signing key here now.

sandrask: great

pkg/wallet/api.go
sandrask: I also find it confusing to return two keys, especially when you look at usages of this method. It is ok for now since it will be fixed in subsequent PR.

pkg/internal/cryptoutil/utils.go
Baha-sk: sounds good, added these points to #596

pkg/wallet/api.go
Baha-sk: Agreed.

When we switch to the new keys structures (ref #596), this function will return only 1 reference value from which sub keys can be fetched later with only this id.

For now there is only 1 wallet storage and both signing and encryption keys are stored in 1 document, this will change to include a metadata storage to reference keys relationships.

PR #513, opened 2019-10-17
fix: Allow user to accept DID method in DID resolver http binding

### LINKED ISSUE #512: Allow user to accept DID method in DID resolver http binding
accept method in DID resolver http binding  always return true

fqutishat:
- Pass accept method in http binding new

closes #512


pkg/didmethod/httpbinding/resolver.go
```diff
@@ -86,7 +89,7 @@ func containsDIDDocument(resp *http.Response) bool {
 }
 
 // New creates new DID Resolver
-func New(endpointURL string, opts ...ResolverOpt) (*DIDResolver, error) {
+func New(endpointURL string, acceptDID AcceptDID, opts ...ResolverOpt) (*DIDResolver, error) {
```
troyronda: acceptDID isn’t an option?

fqutishat: I thought it's mandatory

troyronda: unless the default is return true :)

pkg/didmethod/httpbinding/resolver.go
```diff
@@ -20,6 +20,9 @@ import (
 
 var logger = log.New("aries-framework/didmethod/httpbinding")
 
+// AcceptDID is method to accept did doc
+type AcceptDID func(method string) bool
```
sandrask: Either Accept or AcceptDIDMethod

pkg/didmethod/httpbinding/resolver.go
fqutishat: ok I will change to option and default value will be true

fqutishat: done

pkg/didmethod/httpbinding/resolver.go
fqutishat: rename to Accept

fqutishat: done

PR #515, opened 2019-10-17
feat: Support public DID in Create/Handle Invitation

### LINKED ISSUE #503: Support public DID in Create/Handle Invitation
Add support for public DID in Create/Handle Invitation

sandrask:
Add support for public DID in Create/Handle Invitation

Closes #503

### DIFF
--- a/pkg/didcomm/protocol/didexchange/persistence.go
+++ b/pkg/didcomm/protocol/didexchange/persistence.go
@@ -48,8 +48,8 @@ type ConnectionRecorder struct {
 // Returns:
 //
 // error: error
-func (c *ConnectionRecorder) SaveInvitation(verKey string, invitation *Invitation) error {
-	k, err := invitationKey(verKey)
+func (c *ConnectionRecorder) SaveInvitation(invitation *Invitation) error {
+	k, err := invitationKey(invitation.ID)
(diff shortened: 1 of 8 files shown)


pkg/didcomm/protocol/didexchange/persistence.go
```diff
@@ -67,14 +67,14 @@ func (c *ConnectionRecorder) SaveInvitation(verKey string, invitation *Invitatio
...
+// id: invitation id
 //
 // Returns:
 //
 // invitation found
 // error: error
-func (c *ConnectionRecorder) GetInvitation(verKey string) (*Invitation, error) {
-	k, err := invitationKey(verKey)
+func (c *ConnectionRecorder) GetInvitation(id string) (*Invitation, error) {
+	k, err := invitationKey(id)
```
sudeshrshetty: You may have to remove computeHash() logic from inside invitationKey() function. We used computeHash since there was a possibility of having very long verKey. @sandrask

sandrask: I was thinking about it yesterday but decided not to do it for two reasons: 
1) UUIDs are long
2) we would be exposing database ids
We can further discuss and create an issue if necessary.

PR #516, opened 2019-10-18

soluchok:
* `HandleOutbound` method implementation
* `handle` internal implementation

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -7,8 +7,12 @@ SPDX-License-Identifier: Apache-2.0
 package introduce
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
+
+	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/common/metadata"
@@ -21,6 +25,9 @@ const (
 	Introduce = "introduce"
 	// IntroduceSpec defines the introduce spec
 	IntroduceSpec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// SkipProposalMsgType defines the skip proposal (the introducer has a public invitation)
+	// skip proposal is not part of protocol specification (internal usage)
+	SkipProposalMsgType = "skip/proposal"
 	// ProposalMsgType defines the introduce proposal message type.
 	ProposalMsgType = IntroduceSpec + "proposal"
 	// RequestMsgType defines the introduce request message type.
@@ -31,12 +38,22 @@ const (
 	AckMsgType = IntroduceSpec + "ack"
 )
 
+const initialWaitCount = 2
+
 var logger = log.New("aries-framework/introduce/service")
 
-// eventMetadata type to store data for eventing. This is retrieved during callback.
-type eventMetadata struct {
-	Msg           *service.DIDCommMsg
-	NextStateName string
+// metaData type to store data for internal usage
+type metaData struct {
+	record
+	Msg      *service.DIDCommMsg
+	ThreadID string
+}
+
+type record struct {
+	StateName string
+	// WaitCount - how many introducees still need to approve the introduction proposal
+	// (initial value = introducee count, e.g. 2)
+	WaitCount int
 }
 
 // Service for introduce protocol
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -21,6 +23,9 @@ const (
 	Introduce = "introduce"
 	// IntroduceSpec defines the introduce spec
 	IntroduceSpec = metadata.AriesCommunityDID + ";spec/introduce/1.0/"
+	// SkipProposalMsgType defines the skip proposal (the introducer has a public invitation)
+	// skip proposal is not part of protocol specification (internal usage)
+	SkipProposalMsgType = "skip/proposal"
```
llorllale: when we spoke about this you mentioned it was going to be some sort of internal implementation detail. Therefore this should definitely not be exported.

soluchok: the problem is that there is no way to know that it is skip proposal

soluchok: I am replacing it to real one https://github.com/hyperledger/aries-framework-go/pull/516/files#diff-43cf3af5d3c6b9acebc5c635ba4d0f7dR174

soluchok: when the client perform this action we need to know somehow that is Skip proposal (and I found the way to create this fake type and then replace it to real one)

soluchok: any suggestion welcome :) but I don't think it is a good idea to extend DIDComm message or proposal model just bc we need to know that this is skip proposal. Or we can do some injection to the DB which is the worst solution ever

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -33,10 +40,16 @@ const (
...
+// metaData type to store data for internal usage
+type metaData struct {
+	record
+	Msg      *service.DIDCommMsg
+	ThreadID string
+}
+
+type record struct {
+	StateName string
+	WaitCount int
```
rolsonquadras: description about this field would be helpful.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -127,31 +179,90 @@ func (s *Service) sendMsgEvents(msg service.StateMsg) {
...
 }
 
-func (s *Service) currentState(thID string) (state, error) {
-	name, err := s.store.Get(thID)
+func (s *Service) currentStateRecord(thID string) (*record, error) {
+	src, err := s.store.Get(thID)
+	if errors.Is(err, storage.ErrDataNotFound) {
+		return &record{
+			StateName: stateNameStart,
+			WaitCount: 2,
```
rolsonquadras: is the value configurable ?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -127,31 +179,90 @@ func (s *Service) sendMsgEvents(msg service.StateMsg) {
...
+	if err := json.Unmarshal(src, &r); err != nil {
+		return nil, err
+	}
+	return r, nil
+}
+
+func (s *Service) save(id string, data interface{}) error {
+	src, err := json.Marshal(data)
+	if err != nil {
+		return fmt.Errorf("failed to mashal: %w", err)
```
rolsonquadras: https://github.com/hyperledger/aries-framework-go/pull/525#discussion_r337080113 applicable here.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -187,8 +298,62 @@ func canTriggerActionEvents(msg *service.DIDCommMsg) bool {
...
+			WaitCount: msg.WaitCount,
+		}); err != nil {
+			return fmt.Errorf("failed to persist state %s %w", next.Name(), err)
+		}
+
+		logger.Infof("persisted the connection using %s and updated the state to %s", msg.ThreadID, next.Name())
+
+		s.sendMsgEvents(service.StateMsg{
+			Type: service.PostState, Msg: msg.Msg, StateID: next.Name(),
+			//Properties: createEventProperties(msg.ThreadID, ""),
```
rolsonquadras: Can we remove the commented code ? Better to have a clean code in master.

pkg/didcomm/protocol/introduce/service.go
soluchok: probably in the future :) I found TODO in spec
`TODO:do we want to support introducing more than 2 at a time?`

PR #522, opened 2019-10-18
feat: Support for PeerDID Generation in Wallet DID Creator

### LINKED ISSUE #485: Wallet.CreateDID() not creating peer DIDs as per the Peer DID spec
https://github.com/hyperledger/aries-framework-go/blob/befeff53c97369842ea84fb67602db155d9e3263/pkg/wallet/wallet.go#L100

`Wallet.CreateDID()` has a `method` argument through which the user could provide "peer" as value, but does not create it as per the Peer DID method spec.

I suspect some refactoring is needed - we already have a package for creating and validating peer dids: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didmethod/peer

rolsonquadras:
- Add support to create Peer DID in Wallet

Closes #485 


pkg/doc/did/doc.go
```diff
@@ -529,3 +529,52 @@ func populateRawProofs(proofs []Proof) []interface{} {
 	}
 	return rawProofs
 }
+
+// DocOption provides options to build DID Doc.
+type DocOption func(opts *Doc)
+
+// WithPublicKey DID doc PublicKey.
+func WithPublicKey(pubKey []PublicKey) DocOption {
```
llorllale: Public keys are not optional for peer DIDs.

pkg/wallet/wallet.go
```diff
@@ -138,51 +140,25 @@ func (w *BaseWallet) CreateDID(method string, opts ...DocOpts) (*did.Doc, error)
...
-
-	didDoc := &did.Doc{
-		Context:   []string{did.Context},
-		ID:        id,
-		PublicKey: []did.PublicKey{pubKey},
-		Service:   service,
-		Created:   &createdTime,
-		Updated:   &createdTime,
+	default:
+		didDoc = w.buildDefaultDIDDoc(method, base58PubKey, docOpts)
```
troyronda: what's a default DID doc?

sandrask:
@llorllale @rolsonquadras Are we supposed to use peer dids in e2e BDD test?

pkg/wallet/wallet.go
rolsonquadras: Thats the existing code and I just added a switch for PeerDID. @llorllale can you confirm if we need to support this ?

troyronda:
> @llorllale @rolsonquadras Are we supposed to use peer dids in e2e BDD test?

we should have e2e BDD tests for both peer DIDs and resolved (ledger/"public") DIDs.

sandrask:
Then we should probably switch current BDD test "e2e" to use peer DIDs instead of default DIDs. We probably don't even need default one.

rolsonquadras:
> > @llorllale @rolsonquadras Are we supposed to use peer dids in e2e BDD test?
> 
> we should have e2e BDD tests for both peer DIDs and resolved (ledger/"public") DIDs.

Peer DID's are created as part of DID Exchange protocol (exchangeRequest and exchangeResponse). In case of Public DID, they are resolved to get the keys/endpoint. Not sure how these are related to BDD tests ?

sandrask:
Your change did not enable peer dids for current e2e test. I was just suggesting that we should do that (instead of using default DID)

rolsonquadras:
> Your change did not enable peer dids for current e2e test. I was just suggesting that we should do that (instead of using default DID)

@sandrask When I read about bdd tests, I was looking at [this](https://github.com/hyperledger/aries-framework-go/blob/master/test/bdd/features/did_resolver_sidetree_node.feature#L15) place. [Updated](https://github.com/hyperledger/aries-framework-go/compare/da9d76ec5334b426019aa180809210d30b48df57..776b761a56ee4466b2c1e361fb38e4b5acca7f38#diff-07171b5046f79c097d56c6687a5a95e3R14) the code to use "PeerDID" from "local" did method.

pkg/wallet/wallet.go
rolsonquadras: @troyronda removed the default case.

PR #524, opened 2019-10-18
refactor: move wallet into pkg/crypto, expose keys to crypto ops

### LINKED ISSUE #511: Allow crypto operations to use private keys
The wallet/keystore holds and manages keys, and the crypters, signer, and various operations related to VCs, DIDs, etc require the use of private keys.

If the keystore is a subpackage of, say, `pkg/crypto`, then `pkg/crypto/internal` can contain functions that access private keys from the keystore, allowing any other package within `pkg/crypto` to access private keys.

This sidesteps the issue where we have to create wrapper functions in the wallet for any crypto operation, so the wallet can call the wrapped function and provide any private keys needed. Crypto operations can be implemented in packages in `pkg/crypto`, and thus have access to private keys.

This way, the rest of the agent library still can't access private keys. We can still be rigorous about minimizing what has access to private keys, by minimizing what will go inside the crypto package, keeping only the primitives/operations that need a private key for their processing, and ensuring they never expose those private keys.

Moopli:
`pkg/crypto` is now the home for any package that does cryptography requiring
 private keys or any other secret stored in the wallet.

- The wallet API in BaseWallet is unchanged, simply moved to `pkg/crypto/wallet`
- The public wallet APIs are moved to `pkg/crypto/walletapi`
- BaseWallet's implementation is now hidden in `pkg/crypto/internal/wallet`, in SecretWallet, and includes PutKey/GetKey methods, only accessible to other packages in `pkg/crypto`
 - All methods meant for the framework or client to access are wrapped by `pkg/crypto/wallet.BaseWallet`

Fixes #511 

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -13,10 +13,10 @@ import (
 
 	"github.com/google/uuid"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/wallet"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
-	"github.com/hyperledger/aries-framework-go/pkg/wallet"
 )
 
 const (
--- a/pkg/common/did/creator_test.go
+++ b/pkg/common/did/creator_test.go
@@ -11,9 +11,9 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
 	d "github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	mockwallet "github.com/hyperledger/aries-framework-go/pkg/internal/mock/wallet"
-	"github.com/hyperledger/aries-framework-go/pkg/wallet"
 )
 
 func getMockDID() *d.Doc {
@@ -54,6 +54,6 @@ func TestPeerDIDCreator(t *testing.T) {
 type mockProvider struct {
 }
 
-func (m *mockProvider) DIDWallet() wallet.DIDCreator {
+func (m *mockProvider) DIDWallet() didcreator.DIDCreator {
 	return &mockwallet.CloseableWallet{MockDID: getMockDID()}
 }
--- a/pkg/crypto/internal/didopts/didopts.go
+++ b/pkg/crypto/internal/didopts/didopts.go
@@ -1,0 +1,12 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didopts
+
+// CreateDIDOpts holds the options for creating DID
+type CreateDIDOpts struct {
+	ServiceType string
+}
--- a/pkg/crypto/internal/wallet/wallet.go
+++ b/pkg/crypto/internal/wallet/wallet.go
@@ -18,6 +18,9 @@ import (
 	chacha "golang.org/x/crypto/chacha20poly1305"
 	"golang.org/x/crypto/nacl/box"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/internal/didopts"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/operator"
 	"github.com/hyperledger/aries-framework-go/pkg/didmethod/peer"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/ed25519signature2018"
@@ -37,15 +40,18 @@ type provider interface {
 	InboundTransportEndpoint() string
 }
 
-// BaseWallet wallet implementation
-type BaseWallet struct {
+// SecretWallet wallet implementation, including access to private keys.
+// Note: this is not exposed to the rest of the library, only to pkg/crypto.
+// Other packages need to use pkg/crypto/wallet, which wraps this,
+//   hiding the access to private keys.
+type SecretWallet struct {
 	keystore                 storage.Store
 	didstore                 storage.Store
 	inboundTransportEndpoint string
 }
 
 // New return new instance of wallet implementation
-func New(ctx provider) (*BaseWallet, error) {
+func New(ctx provider) (*SecretWallet, error) {
 	ks, err := ctx.StorageProvider().OpenStore(keyStoreNamespace)
 	if err != nil {
 		return nil, fmt.Errorf("failed to OpenStore for '%s', cause: %w", keyStoreNamespace, err)
@@ -56,11 +62,11 @@ func New(ctx provider) (*BaseWallet, error) {
 		return nil, fmt.Errorf("failed to OpenStore for '%s', cause: %w", didStoreNamespace, err)
 	}
 
-	return &BaseWallet{keystore: ks, didstore: ds, inboundTransportEndpoint: ctx.InboundTransportEndpoint()}, nil
+	return &SecretWallet{keystore: ks, didstore: ds, inboundTransportEndpoint: ctx.InboundTransportEndpoint()}, nil
 }
 
 // CreateEncryptionKey create a new public/private encryption keypair.
-func (w *BaseWallet) CreateEncryptionKey() (string, error) {
+func (w *SecretWallet) CreateEncryptionKey() (string, error) {
 	pub, priv, err := box.GenerateKey(rand.Reader)
 	if err != nil {
 		return "", fmt.Errorf("failed to GenerateKey: %w", err)
@@ -74,22 +80,29 @@ func (w *BaseWallet) CreateEncryptionKey() (string, error) {
 }
 
 // CreateSigningKey create a new public/private signing keypair.
-func (w *BaseWallet) CreateSigningKey() (string, error) {
+func (w *SecretWallet) CreateSigningKey() (string, error) {
 	pub, priv, err := ed25519.GenerateKey(rand.Reader)
 	if err != nil {
 		return "", fmt.Errorf("failed to GenerateKey: %w", err)
 	}
 	base58Pub := base58.Encode(pub[:])
 	// TODO - need to encrypt the priv before putting them in the store.
-	if err := persist(w.keystore, base58Pub, &cryptoutil.KeyPair{Pub: pub[:], Priv: priv[:]}); err != nil {
+	if e := persist(w.keystore, base58Pub, &cryptoutil.KeyPair{Pub: pub[:], Priv: priv[:]}); e != nil {
+		return "", e
+	}
+
+	// generate the corresponding encryption keypair for any signing keypair
+	_, err = w.convertToEncryptionKey(pub)
+	if err != nil {
 		return "", err
 	}
+
 	return base58Pub, nil
 }
 
-// ConvertToEncryptionKey converts an ed25519 keypair present in the wallet,
+// convertToEncryptionKey converts an ed25519 keypair present in the wallet,
 // persists the resulting keypair, and returns the result public key.
-func (w *BaseWallet) ConvertToEncryptionKey(key []byte) ([]byte, error) {
+func (w *SecretWallet) convertToEncryptionKey(key []byte) ([]byte, error) {
 	encPub, err := cryptoutil.PublicEd25519toCurve25519(key)
 	if err != nil {
 		return nil, err
@@ -97,7 +110,7 @@ func (w *BaseWallet) ConvertToEncryptionKey(key []byte) ([]byte, error) {
 	encPubB58 := base58.Encode(encPub)
 
 	keyB58 := base58.Encode(key)
-	kp, err := w.getKey(keyB58)
+	kp, err := w.GetKey(keyB58)
 	if err != nil {
 		return nil, err
 	}
@@ -115,8 +128,8 @@ func (w *BaseWallet) ConvertToEncryptionKey(key []byte) ([]byte, error) {
 }
 
 // SignMessage sign a message using the private key associated with a given verification key.
-func (w *BaseWallet) SignMessage(message []byte, fromVerKey string) ([]byte, error) {
-	keyPair, err := w.getKey(fromVerKey)
+func (w *SecretWallet) SignMessage(message []byte, fromVerKey string) ([]byte, error) {
+	keyPair, err := w.GetKey(fromVerKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get key: %w", err)
 	}
@@ -124,13 +137,13 @@ func (w *BaseWallet) SignMessage(message []byte, fromVerKey string) ([]byte, err
 }
 
 // Close wallet
-func (w *BaseWallet) Close() error {
+func (w *SecretWallet) Close() error {
 	return nil
 }
 
 // CreateDID returns new DID Document
-func (w *BaseWallet) CreateDID(method string, opts ...DocOpts) (*did.Doc, error) {
-	docOpts := &createDIDOpts{}
+func (w *SecretWallet) CreateDID(method string, opts ...didcreator.DocOpts) (*did.Doc, error) {
+	docOpts := &didopts.CreateDIDOpts{}
 	// Apply options
 	for _, opt := range opts {
 		opt(docOpts)
@@ -164,7 +177,7 @@ func (w *BaseWallet) CreateDID(method string, opts ...DocOpts) (*did.Doc, error)
 }
 
 // GetDID gets already created DID document from underlying store
-func (w *BaseWallet) GetDID(id string) (*did.Doc, error) {
+func (w *SecretWallet) GetDID(id string) (*did.Doc, error) {
 	bytes, err := w.didstore.Get(id)
 	if err != nil {
 		return nil, err
@@ -178,9 +191,9 @@ func (w *BaseWallet) GetDID(id string) (*did.Doc, error) {
 	return &didDoc, nil
 }
 
-// getKey get key
-func (w *BaseWallet) getKey(verkey string) (*cryptoutil.KeyPair, error) {
-	bytes, err := w.keystore.Get(verkey)
+// GetKey gets the keypair associated to the given pubkey
+func (w *SecretWallet) GetKey(pub string) (*cryptoutil.KeyPair, error) {
+	bytes, err := w.keystore.Get(pub)
 	if err != nil {
 		if errors.Is(storage.ErrDataNotFound, err) {
 			return nil, cryptoutil.ErrKeyNotFound
@@ -194,10 +207,23 @@ func (w *BaseWallet) getKey(verkey string) (*cryptoutil.KeyPair, error) {
 	return &key, nil
 }
 
+// PutKey persists a keypair in the keystore
+func (w *SecretWallet) PutKey(pub string, pair *cryptoutil.KeyPair) error {
+	return persist(w.keystore, pub, pair)
+}
+
+// AttachCryptoOperator attaches a crypto operator to this wallet, so the operator can use its private keys.
+func (w *SecretWallet) AttachCryptoOperator(cryptoOp operator.CryptoOperator) error {
+	if cryptoOp == nil {
+		return fmt.Errorf("cannot attach nil crypto operator")
+	}
+	return cryptoOp.InjectKeyHolder(w)
+}
+
 // DeriveKEK will derive an ephemeral symmetric key (kek) using a private key fetched from
 // the wallet corresponding to fromPubKey and derived with toPubKey
 // This implementation is for curve 25519 only
-func (w *BaseWallet) DeriveKEK(alg, apu, fromPubKey, toPubKey []byte) ([]byte, error) { // nolint:lll
+func (w *SecretWallet) DeriveKEK(alg, apu, fromPubKey, toPubKey []byte) ([]byte, error) { // nolint:lll
 	if fromPubKey == nil || toPubKey == nil {
 		return nil, cryptoutil.ErrInvalidKey
 	}
@@ -205,9 +231,9 @@ func (w *BaseWallet) DeriveKEK(alg, apu, fromPubKey, toPubKey []byte) ([]byte, e
 	copy(fromPrivKey[:], fromPubKey)
 
 	// get keypair from wallet store
-	walletKeyPair, err := w.getKey(base58.Encode(fromPubKey))
+	walletKeyPair, err := w.GetKey(base58.Encode(fromPubKey))
 	if err != nil {
-		return nil, fmt.Errorf("failed from getKey: %w", err)
+		return nil, fmt.Errorf("failed from GetKey: %w", err)
 	}
 	copy(fromPrivKey[:], walletKeyPair.Priv)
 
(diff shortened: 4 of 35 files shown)


Moopli:
Welcome feedback in general, but here's something I'm specifically looking for right now:

The idea behind this work is so the wallet can expose secrets to crypto operations, without said crypto operations being inside the same package, and without the rest of the framework able to access the secrets. To do so I had to make a wallet implementation, `pkg/crypto/internal/wallet.SecretWallet`, so anything inside `pkg/crypto` can access said wallet's public methods, including `GetKey`, but nothing outside `pkg/crypto` can do so. Since the wallet needs to be accessed for other purposes outside `pkg/crypto` I made `pkg/crypto/wallet.BaseWallet` a wrapper of the secret wallet.

Here's the problem - using interfaces I can make the wallet connect to a crypto operator (using `AttachCryptoOperator()`), where the crypto operator must be implemented in `pkg/crypto` to implement the interface (which helps enforce that secrets are kept within `pkg/crypto`), but then I can't implement the mockwallet.

I'm not sure if there's a way I can restrict `AttachCryptoOperator()` to only be used on operators defined in `pkg/crypto` while allowing a mock wallet outside `pkg/crypto` to implement the wallet interfaces.

Moopli:
Update: solved the above problem, now I'm fixing up test coverage.

pkg/client/didexchange/client.go
```diff
@@ -12,12 +12,13 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/walletapi"
```
fqutishat: wrong location

pkg/common/did/creator_test.go
```diff
@@ -9,11 +9,12 @@ package did
 import (
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/walletapi"
```
fqutishat: wrong location

pkg/crypto/box/box_test.go
```diff
@@ -4,19 +4,24 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
...
+package box_test
 
 import (
 	"crypto/rand"
+	"fmt"
 	"io"
 	"testing"
 
-	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
+	wallet2 "github.com/hyperledger/aries-framework-go/pkg/internal/mock/wallet"
```
fqutishat: wrong location

pkg/client/didexchange/client.go
Moopli: fixed

pkg/crypto/box/box_test.go
Moopli: fixed all of them I believe

pkg/common/did/creator_test.go
Moopli: fixed

pkg/crypto/internal/walletapi/wallet_api.go
```diff
@@ -0,0 +1,25 @@
...
+
+// CreateDIDOpts holds the options for creating DID
+type CreateDIDOpts struct {
+	ServiceType string
+}
+
+// KeyHolder denotes an object that holds keypairs, indexed by the public key (in base 58)
+type KeyHolder interface {
+	// GetKey gets the keypair associated to the given pubkey
+	GetKey(verkey string) (*cryptoutil.KeyPair, error)
```
sandrask: Should it be GetKey(kid string)? verification key is very specific

pkg/crypto/wallet/wallet_test.go
```diff
@@ -0,0 +1,142 @@
...
+		FindVerKeyErr:            fmt.Errorf("err: FindVerKey"),
+		SignMessageValue:         []byte("signature"),
+		SignMessageErr:           fmt.Errorf("err: SignMessage"),
+		MockDID: &did.Doc{
+			ID: "mock doc id",
+		},
+	}}
+
+	t.Run("CreateEncryptionKey", func(t *testing.T) {
+		ret, err := bw.CreateEncryptionKey()
```
sandrask: It is not correct to to check return values from secret and wrapper wallet for equality. Create encryption key should always return different key.

sandrask: Same applies to tests bellow

pkg/crypto/walletapi/wallet_api.go
```diff
@@ -56,6 +65,33 @@ type Crypto interface {
...
+// interface, which ensures that access to private keys is limited
+// to only packages in pkg/crypto
+type KeyHolderUser interface {
+
+	// InjectKeyHolder injects a KeyHolder which will provide private keys for the KeyHolderUser's crypto computations.
+	// Usage note: InjectKeyHolder is not for the client to call,
+	// the client can't access the secret wallet (KeyHolder) to pass in.
+	// This is used by a wallet implementing the Crypto interface,
+	// in AttachCryptoOperator(KeyHolderUser)
+	InjectKeyHolder(secretapi.KeyHolder) error
```
sandrask: Should we expose this interface in walletapi considering that client should not call this interface? Can you declare it in receiver only?

pkg/didcomm/crypto/legacy/authcrypt/authcrypt_test.go
```diff
@@ -382,9 +396,9 @@ func TestDecrypt(t *testing.T) {
...
 
+	t.Run("Test decrypting python envelope with invalid recipient", func(t *testing.T) {
```
Baha-sk: replace `python` with `reference (python built)`

pkg/crypto/internal/wallet/wallet.go
```diff
@@ -194,20 +199,33 @@ func (w *BaseWallet) getKey(verkey string) (*cryptoutil.KeyPair, error) {
 	return &key, nil
 }
 
+// PutKey persists a keypair in the keystore
+func (w *SecretWallet) PutKey(verkey string, pair *cryptoutil.KeyPair) error {
```
Baha-sk: Probably add a comment for PutKey() and AttachCryptoOperator() are exposed to aid mocking a wallet for testing..

pkg/crypto/walletapi/wallet_api.go
```diff
@@ -56,6 +65,33 @@ type Crypto interface {
...
+	//
+	// Returns:
+	//
+	// []byte: The public key of the encryption keypair created. nil in case of error.
+	//
+	// error:
+	ConvertToEncryptionKey(pubkey []byte) ([]byte, error)
+
+	// AttachCryptoOperator attaches a crypto operator to this wallet, so the operator can use its private keys.
+	AttachCryptoOperator(cryptoOp KeyHolderUser) error
```
Baha-sk: same comment here about AttachCryptoOperator()

pkg/crypto/walletapi/wallet_api.go
```diff
@@ -4,9 +4,12 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package wallet
+package walletapi
 
 import (
+	"io"
+
+	secretapi "github.com/hyperledger/aries-framework-go/pkg/crypto/internal/walletapi"
```
sandrask: Should we use internal api here?

pkg/crypto/walletapi/wallet_api.go
```diff
@@ -56,6 +65,33 @@ type Crypto interface {
...
+}
+
+// KeyHolderUser denotes an object that accesses keys in a KeyHolder
+// A KeyHolderUser must be defined inside pkg/crypto to fulfill the
+// interface, which ensures that access to private keys is limited
+// to only packages in pkg/crypto
+type KeyHolderUser interface {
+
+	// InjectKeyHolder injects a KeyHolder which will provide private keys for the KeyHolderUser's crypto computations.
+	// Usage note: InjectKeyHolder is not for the client to call,
```
Baha-sk: this is a good example of comment.. to be copied for the above functions I mentioned

troyronda:
A general comment:
- We will move away from the wallet naming to KMS. How will that affect the structure in this PR?
- We should allow for different implementations of a KMS (and its sub-packages). Example: A keystore could be software vs hardware, the wrapping key to protect the keystore could be hardware vs software, etc.
   - The idea being that someone could instantiate a framework with different flavours of a KMS.

pkg/crypto/internal/walletapi/wallet_api.go
Baha-sk: this will be replaced with GetKeyPairCombo in my PR later.. it will return a struct that has 2 key pairs, one for encryption and one for signature... naming is not important (for now)

pkg/crypto/internal/wallet/wallet.go
Moopli: PutKey isn't exposed to anything outside pkg/crypto

pkg/crypto/wallet/wallet_test.go
Moopli: these tests use a mock wallet for `secretWallet`, so they can verify that wrapping methods are wrapping the correct methods of `secretWallet`. Functionality tests for the implementation are still present in `pkg/crypto/internal/wallet.SecretWallet`

pkg/crypto/internal/walletapi/wallet_api.go
Moopli: Renamed these, good point

pkg/crypto/walletapi/wallet_api.go
Moopli: This function is safe for external code to call - if a component needs (for example) to encrypt or sign something, and we've created a crypto operator that does that operation, then the component which needs the operation can call `myWallet.AttachCryptoOperator(operator)` to use the operator on their wallet's keys.

Moopli:
Once we've settled on a good way to connect crypto operators to wallets, I'd like to refactor out the Signer, DeriveKEK, etc so they aren't part of the wallet API but are crypto operators instead, which can, with this PR, access private keys for computation.

That should reduce the wallet to being just a keystore, which can be swapped out for another implementation.

> We will move away from the wallet naming to KMS. How will that affect the structure in this PR?

The renaming shouldn't have any effect, the refactoring to turn the wallet into a KMS implementation is partially done by this PR.

Moopli:
I could add the wallet -> KMS renaming to this PR so I'm not putting out another PR touching dozens of files right after this one.

pkg/crypto/didcreator/didcreator.go
```diff
@@ -0,0 +1,53 @@
...
+package didcreator
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/internal/walletapi"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+// DIDCreator provides features to create and query DID document
+type DIDCreator interface {
+	// Creates new DID document.
```
Baha-sk: prefer to start comment with function name... not sure why linter doesn't catch this

pkg/crypto/didcreator/didcreator.go
```diff
@@ -0,0 +1,53 @@
...
+	// opts: options to create DID
+	//
+	// Returns:
+	//
+	// did: DID document
+	//
+	// error: error
+	CreateDID(method string, opts ...DocOpts) (*did.Doc, error)
+
+	// Gets already created DID document by ID.
```
Baha-sk: same as above

pkg/crypto/wallet/wallet_api.go
```diff
@@ -56,6 +66,19 @@ type Crypto interface {
...
+
+	// ConvertToEncryptionKey will create and persist a Curve25519 encryption keypair
+	// created from the ed25519 keypair with the provided public key.
+	//
+	// Returns:
+	//
+	// []byte: The public key of the encryption keypair created. nil in case of error.
+	//
+	// error:
+	ConvertToEncryptionKey(pubkey []byte) ([]byte, error)
```
Baha-sk: this function is specific to legacy crypto, it should not be part of this interface

pkg/crypto/walletapi/wallet_api.go
Moopli: I'm still stuck on this - for the mock wallet to have the correct overrides, it needs an AttachCryptoOperator signature that has a `KeyHolderUser` (which I've renamed to CryptoOperator) as a parameter. If `KeyHolderUser` is internal or consumer-only, then when I mock it inside `mock/wallet` the mockwallet `AttachCryptoOperator()` method has a different signature to the method in the `Crypto` interface, so `mockwallet.CloseableWallet` doesn't implement the wallet interfaces.

pkg/crypto/didcreator/didcreator.go
Moopli: Maybe it didn't complain when I just refactored out code without changing it.

I'll fix.

pkg/internal/mock/wallet/mock_wallet.go
```diff
@@ -18,16 +18,15 @@ type CloseableWallet struct {
 	CreateEncryptionKeyErr   error
 	CreateSigningKeyValue    string
 	CreateSigningKeyErr      error
+	AttachCryptoOperatorErr  error
+	ConvertKeyValue          []byte
+	ConvertKeyErr            error
```
Baha-sk: ConvertKey should not be part of the Crypto interface (it was a mistake to add key conversion in the Python implementation and they admit it)

pkg/internal/mock/wallet/mock_wallet.go
```diff
@@ -51,6 +50,16 @@ func (m *CloseableWallet) FindVerKey(candidateKeys []string) (int, error) {
 	return m.FindVerKeyValue, m.FindVerKeyErr
 }
 
+// ConvertToEncryptionKey converts the given signing key to an encryption key
+func (m *CloseableWallet) ConvertToEncryptionKey(pubkey []byte) ([]byte, error) {
```
Baha-sk: same thing... remove function

pkg/crypto/wallet/wallet_api.go
Moopli: I thought this would be removed by your PR, since keypairs in the wallet won't need to be converted on the fly (by this method) any more, since the Signing and Encryption KeyPairs are going to be generated and persisted together.

pkg/crypto/didcreator/didcreator.go
Baha-sk: linter doesn't catch this error for some reason ... we have the same erroneous comments in other interfaces as well.

pkg/crypto/didcreator/didcreator.go
Moopli: done

pkg/crypto/didcreator/didcreator.go
Moopli: done

pkg/client/didexchange/client.go
```diff
@@ -12,12 +12,13 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/wallet"
```
troyronda: import is grouped wrong.

pkg/common/did/creator_test.go
```diff
@@ -9,11 +9,12 @@ package did
 import (
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
```
troyronda: import is grouped wrong.

pkg/client/didexchange/client.go
Moopli: looks like I need to fix my goland config

pkg/crypto/internal/wallet/wallet_test.go
```diff
@@ -13,10 +13,13 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
```
troyronda: import is grouped wrong.

pkg/crypto/operator/box/box.go
```diff
@@ -4,21 +4,22 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
...
 
-package wallet
+package box
 
 import (
 	"errors"
+	"fmt"
 	"io"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/wallet"
```
troyronda: import is grouped wrong.

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -14,6 +14,8 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/wallet"
```
troyronda: import is grouped wrong.

pkg/internal/mock/common/did/mock_creator.go
```diff
@@ -12,9 +12,10 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
+
```
troyronda: extra line

pkg/framework/context/context.go
```diff
@@ -9,6 +9,10 @@ package context
 import (
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
+
```
troyronda: extra lines.

pkg/framework/aries/api/protocol.go
```diff
@@ -9,12 +9,15 @@ package api
 import (
 	"errors"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
+
```
troyronda: extra line.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -15,13 +15,14 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/didcreator"
```
troyronda: import is grouped wrong.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -11,6 +11,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/wallet"
```
troyronda: import is grouped wrong.

pkg/didcomm/crypto/jwe/authcrypt/authcrypt.go
```diff
@@ -10,10 +10,11 @@ import (
 	"crypto/rand"
 	"errors"
 
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/wallet"
```
troyronda: import is grouped wrong.

pkg/crypto/wallet/wallet_api.go
Moopli: I've made it private, and it's now called inside CreateSigningKey, so encryption keys are available when needed.

pkg/internal/mock/wallet/mock_wallet.go
Moopli: done

pkg/common/did/creator_test.go
Moopli: done

pkg/crypto/wallet/wallet_test.go
Moopli: I've updated these tests to use the real SecretWallet instead of a mock secretWallet - these tests should confirm that the wrapper is correctly calling the wrapped functions.

pkg/crypto/internal/didopts/didopts.go
```diff
@@ -0,0 +1,12 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didopts
+
+// CreateDIDOpts holds the options for creating DID
+type CreateDIDOpts struct {
```
Moopli: Question: this used to be private to the wallet package, and with the package being split up I needed to pull this out into an internal package - question is, does this need to be internal to `pkg/crypto`? If I move this to, for example, `pkg/internal/didutil` then what I've moved into `pkg/crypto/didcreator` can be merged into `pkg/common/did`.

pkg/crypto/wallet/wallet.go
```diff
@@ -0,0 +1,84 @@
...
+
+// GetDID gets already created DID document from underlying store
+func (w *BaseWallet) GetDID(id string) (*did.Doc, error) {
+	return w.secretWallet.GetDID(id)
+}
+
+// DeriveKEK will derive an ephemeral symmetric key (kek) using a private key fetched from
+// the wallet corresponding to fromPubKey and derived with toPubKey
+// This implementation is for curve 25519 only
+func (w *BaseWallet) DeriveKEK(alg, apu, fromPubKey, toPubKey []byte) ([]byte, error) { // nolint:lll
```
Moopli: Having created `CryptoOperator` for doing operations using a wallet's secrets, we can pull methods like this out of the wallet, so the wallet can be reduced to just a keystore, easier to plug in.

pkg/crypto/walletapi/wallet_api.go
Moopli: Talked to @troyronda and he pointed out that we need to allow clients to define their own CryptoOperators to plug in the crypto suites they need to use, so KeyHolder is now part of the public API, and this is now implementable by client code.

pkg/crypto/walletapi/wallet_api.go
Moopli: Removed this dependency in some unrelated refactoring

pkg/crypto/internal/didopts/didopts.go
troyronda: I don't see why this needs to be in the wallet.

pkg/crypto/internal/wallet/wallet.go
```diff
@@ -37,15 +40,18 @@ type provider interface {
 	InboundTransportEndpoint() string
 }
 
-// BaseWallet wallet implementation
-type BaseWallet struct {
+// SecretWallet wallet implementation, including access to private keys.
+// Note: this is not exposed to the rest of the library, only to pkg/crypto.
+// Other packages need to use pkg/crypto/wallet, which wraps this,
+//   hiding the access to private keys.
+type SecretWallet struct {
```
troyronda: Why is did store and inboundTransportEndpoint part of "SecretWallet".

(the secret part is really the keystore and the whole "wallet" name could be dropped).

Moopli: Instead of answering the question I'm going to agree

pkg/crypto/operator/box/box.go
```diff
@@ -29,12 +29,26 @@ import (
...
+	b := CryptoBox{}
+	err := w.AttachCryptoOperator(&b)
+	if err != nil {
+		return nil, err
+	}
+	return &b, nil
+}
+
+// InjectKeyHolder provides a KeyHolder for the CryptoBox to use
+func (b *CryptoBox) InjectKeyHolder(kh operator.KeyHolder) error {
```
troyronda: I'm worried when this type of thing isn't part of the constructor.

pkg/crypto/operator/box/box.go
```diff
@@ -29,12 +29,26 @@ import (
...
+	w operator.KeyHolder
 }
 
-// NewCryptoBox creates a CryptoBox which provides crypto box encryption using the given wallet's keypairs
-func NewCryptoBox(w *BaseWallet) *CryptoBox {
-	return &CryptoBox{w: w}
+// New creates a CryptoBox which provides crypto box encryption using the given KeyHolder's keypairs
+func New(w wallet.Crypto) (*CryptoBox, error) {
+	b := CryptoBox{}
+	err := w.AttachCryptoOperator(&b)
```
troyronda: I'm not really sure of the purpose of this pattern.

If any package can simply request to have the operator "injected", no real implementation is being hidden from outside the framework.

Moopli: This had value when my PR limited CryptoOperators to being defined in the framework itself in `pkg/crypto` - since we need to open that up then this doesn't really have a purpose.

troyronda: As discussed, think of this functionality as the sw implementation of the crypto interface.

something like pkg/crypto/sw could make sense as a parent folder.

Moopli:
Looking at what needs to be done to make our kms/crypto implementation match the target aries architecture, this PR isn't really going to get us there, and since we'd need another refactor after this one there's little point to using this.

PR #536, opened 2019-10-21
refactor: Remove DIDExchange Service Handle API invocation from RestAPI pkg

rolsonquadras:
- Removed didex service.Handle() API invocation from REST API.
- Instead used client.HandleInvitation() API.

### DIFF
--- a/pkg/restapi/operation/didexchange/didexchange.go
+++ b/pkg/restapi/operation/didexchange/didexchange.go
@@ -8,7 +8,6 @@ package didexchange
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -200,6 +173,27 @@ func (c *Operation) AcceptInvitation(rw http.ResponseWriter, req *http.Request)
 	c.writeResponse(rw, response)
 }
 
+// AcceptExchangeRequest swagger:route POST /connections/{id}/accept-request did-exchange acceptRequest
```
rolsonquadras: Rearranged - no change in this function.

PR #564, opened 2019-10-23
refactor: Serialization of Verifiable Credential with extended data model

### LINKED ISSUE #327: Marshalling of Verifiable Credential with extended data model
We are able to decode Verifiable Credential with a custom extended data model. We need to do the opposite - serialize that Verifiable Credential with an extended data model.

Up for now, there are `JSON()` and `JWT()` methods of `Credential` serialization, and we should support them both.

kdimak:
closes #327

This PR brings the following improvements to the Verifiable Credential (VC):
- Keep extra fields of raw VC in a map (motivated by  [issue 22533](https://github.com/golang/go/issues/22533)); this leverages the extensibility of the data model and allows more strict serialization of the VC.
- Use stricter types in `Credential`: issuer, context, type (used generic `interface{}` or `[]interface{}` before).
- Simplify JWT decoding. Previously, JWT decoding resulted in returning `*rawCredential` and serialized JSON (`[]byte`) of VC. Now it returns only serialized JSON which is then unmarshalled to `rawCredential`.

### DIFF
--- a/pkg/doc/verifiable/common.go
+++ b/pkg/doc/verifiable/common.go
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 package verifiable
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/square/go-jose/v3"
@@ -63,7 +64,7 @@ type typedID struct {
 type RefreshService typedID
 
 // TermsOfUse represents terms of use of Verifiable Credential by Issuer or Verifiable Presentation by Holder.
-type TermsOfUse typedID
+type TermsOfUse interface{}
 
 func describeSchemaValidationError(result *gojsonschema.Result, what string) string {
 	errMsg := what + " is not valid:\n"
@@ -71,4 +72,16 @@ func describeSchemaValidationError(result *gojsonschema.Result, what string) str
 		errMsg += fmt.Sprintf("- %s\n", desc)
 	}
 	return errMsg
+}
+
+func stringSlice(values []interface{}) ([]string, error) {
+	strings := make([]string, len(values))
+	for i := range values {
+		t, valid := values[i].(string)
+		if !valid {
+			return nil, errors.New("array element is not a string")
+		}
+		strings[i] = t
+	}
+	return strings, nil
 }
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -208,9 +208,10 @@ type CredentialSchema typedID
 
 // Credential Verifiable Credential definition
 type Credential struct {
-	Context        []interface{}
+	Context        []string
+	CustomContext  []interface{}
 	ID             string
-	Type           interface{}
+	Types          []string
 	Subject        Subject
 	Issuer         Issuer
 	Issued         *time.Time
@@ -221,11 +222,16 @@ type Credential struct {
 	Evidence       *Evidence
 	TermsOfUse     []TermsOfUse
 	RefreshService *RefreshService
+
+	CredentialExtraFields
 }
+
+// CredentialExtraFields is a map of extra fields which extend base data model.
+type CredentialExtraFields map[string]interface{}
 
 // rawCredential is a basic verifiable credential
 type rawCredential struct {
-	Context        []interface{}     `json:"@context,omitempty"`
+	Context        interface{}       `json:"@context,omitempty"`
 	ID             string            `json:"id,omitempty"`
 	Type           interface{}       `json:"type,omitempty"`
 	Subject        Subject           `json:"credentialSubject,omitempty"`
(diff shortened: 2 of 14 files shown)


pkg/doc/verifiable/credential_jwt.go
```diff
@@ -52,107 +53,78 @@ func newJWTCredClaims(vc *Credential, minimizeVc bool) (*JWTCredClaims, error) {
...
+type JWTCredClaimsUnmarshaller func(vcJWTBytes []byte) (*JWTCredClaims, error)
 
-	credJSON, err := credClaimsDecoder.UnmarshalVCClaim(rawJWT)
+// decodeCredJWT parses JWT from the specified bytes array in compact format using unmarshaller.
+// It returns decoded Verifiable Credential refined by JWT Claims in raw byte array form.
+func decodeCredJWT(rawJWT []byte, unmarshaller JWTCredClaimsUnmarshaller) ([]byte, error) {
+	credClaims, err := unmarshaller(rawJWT)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to decode raw Verifiable Credential JWT claims: %w", err)
+		return nil, fmt.Errorf("failed to decode Verifiable Credential JWT claims: %w", err)
```
troyronda: Wrapping messages should be kept succinct. https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -17,7 +18,7 @@ import (
 type JWTCredClaims struct {
 	*jwt.Claims
 
-	Credential *rawCredential `json:"vc,omitempty"`
+	VC map[string]interface{} `json:"vc,omitempty"`
```
troyronda: Is this field intended to be used outside of the package?

troyronda:
This PR/commit appears to be a "refactor: " rather than a "feat: "

pkg/doc/verifiable/credential_jwt.go
kdimak: Agree! I think it deserves the creation of a separate issue, to clean-up the whole `credential` package.
#570

pkg/doc/verifiable/credential_jwt.go
kdimak: We cannot make it unexported as **json** won't take it into account on serialization & deserialization.

kdimak:
@troyronda yes, this is definitely "refactor:".

pkg/doc/verifiable/credential.go
```diff
@@ -340,32 +405,81 @@ func WithUnsecuredJWTDecoding() CredentialOpt {
...
+	case []interface{}:
+		types, err := stringSlice(rType)
+		if err != nil {
+			return nil, fmt.Errorf("failed to map credential types: %w", err)
+		}
+		return types, nil
+	default:
+		return nil, errors.New("credential type of unknown type")
 	}
+}
```
soluchok: I do not like this approach with different types `string and []interface{}`, are we able to change this logic? We can use comma-separated values instead of different types e.g `type1` or  `type1,type3,...` or only slice [type1] or  [type1,type2]
@troyronda @rolsonquadras @fqutishat

pkg/doc/verifiable/credential_jwt.go
troyronda: Is the struct used outside of the package?

kdimak: Yes, it could.

pkg/doc/verifiable/credential.go
kdimak: hey @soluchok 
According to [Verifiable Credentials spec](https://w3c.github.io/vc-data-model/#types), the `type` can be defined as either a single string value or array. We keep it raw in `rawCredential` struct (into which VC JSON is unmarshalled) and then decode it to universal and convenient `[]string` to be used in `Credential` struct.

pkg/doc/verifiable/common.go
```diff
@@ -63,7 +64,7 @@ type typedID struct {
 type RefreshService typedID
 
 // TermsOfUse represents terms of use of Verifiable Credential by Issuer or Verifiable Presentation by Holder.
-type TermsOfUse typedID
+type TermsOfUse interface{}
```
troyronda: I'm not too excited when we need to export types/struct that have interface{}

Can you give more context on this change?

kdimak: It was incorrectly defined before. It can be one or more `TermsOfUse`. Its structure is not precisely defined. [They say](https://w3c.github.io/vc-data-model/#terms-of-use) that **type** field is mandatory and **id** field is optional, and 
> The precise contents of each term of use is determined by the specific termsOfUse type definition.
If we use `typedID` alias for `termsOfUse` and after decoding of VC want to marshal it back, we will loose all the original fields except `type` and `id`.
That was the reason I changed the `TermsOfUse` alias.
We can also stay with `typedID` alias and extend it to keep all extra fields in e.g. `map[string]interface{}` to be used during marshalling.
What do you think @troyronda

troyronda: I thought we had typically used an unexported "raw" struct and filled a properly typed exported struct.

troyronda: I see. It's another situation with extended fields. I would think you would have defined the known parts and keep extra fields in map[string]interface{} like you did elsewhere in the PR.

kdimak: Yes, because of "one or more" and also because of the flexible structure of the `termsOfUse` field.

kdimak: Filed a separate issue #586.

pkg/doc/verifiable/credential.go
troyronda: I would, at least, add some comments in the code. It isn’t entirely clear to a reader when they see these case statements.

pkg/doc/verifiable/credential_jwt.go
```diff
@@ -52,107 +53,78 @@ func newJWTCredClaims(vc *Credential, minimizeVc bool) (*JWTCredClaims, error) {
...
-		refineVCIssuerFromJWTClaims(raw, iss)
+	if iss := claims.Issuer; iss != "" {
+		refineVCIssuerFromJWTClaims(vcMap, iss)
 	}
 
-	if nbf := jcc.NotBefore; nbf != nil {
+	if nbf := claims.NotBefore; nbf != nil {
 		nbfTime := nbf.Time().UTC()
-		raw.Issued = &nbfTime
+		vcMap["issuanceDate"] = nbfTime.Format(time.RFC3339)
```
troyronda: some of these strings could make sense as named constants.

pkg/doc/verifiable/credential.go
kdimak: Sure.

pkg/doc/verifiable/credential_jwt.go
kdimak: Agree! Done.

pkg/doc/verifiable/credential.go
kdimak: Done.

PR #565, opened 2019-10-23
refactor:  Client was extended with service.Action and service.Message

soluchok:
* didexchange client was extended with service.Action and service.Message

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -10,7 +10,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"sync"
 
 	"github.com/google/uuid"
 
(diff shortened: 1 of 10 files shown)


pkg/client/didexchange/client.go
```diff
@@ -191,86 +188,16 @@ func (c *Client) startServiceEventListener() {
 		select {
 		case e := <-c.actionCh:
 			// assigned to var as lint fails with : Using a reference for the variable on range scope (scopelint)
-			msg := e
-			c.handleActionEvent(&msg)
+			c.GetActionEvent() <- e
```
troyronda: This comment should have been in the prior PR:

Simple accessors should not have a Get prefix - they are not Getting anything.

soluchok: I am going to change it! Do not merge!

soluchok: Done!

PR #568, opened 2019-10-23

soluchok:
Added:
* `startInternalListener`
* `processFailure`

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package introduce
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


pkg/didcomm/protocol/introduce/service_test.go
```diff
@@ -69,7 +70,7 @@ func TestService_handle(t *testing.T) {
 	t.Parallel()
 	t.Run("Happy path", func(t *testing.T) {
 		store := mockstore.NewMockStoreProvider()
-		svc, err := New(store)
+		svc, err := New(context.TODO(), store)
```
troyronda: Why TODO rather than Background?

soluchok: I am going to replace it

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -56,21 +57,94 @@ type record struct {
...
+	err error
+}
+
 // Service for introduce protocol
 type Service struct {
 	service.Action
 	service.Message
-	store storage.Store
+	store     storage.Store
+	callbacks chan callback
```
rolsonquadras: callback instead of callbacks ?

PR #575, opened 2019-10-24

soluchok:
* added `dispatcher.Outbound`
* tests improvements
* added *decorator.Thread to `Proposal` model
* `process` implementation

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -7,17 +7,18 @@ SPDX-License-Identifier: Apache-2.0
 package introduce
 
 import (
-	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/common/metadata"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
@@ -66,30 +67,56 @@ type callback struct {
 type Service struct {
 	service.Action
 	service.Message
-	store     storage.Store
-	callbacks chan callback
+	store       storage.Store
+	callbacks   chan callback
+	ctx         internalContext
+	stop        chan struct{}
+	closedMutex sync.Mutex
+	closed      bool
+}
+
+// provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
+type provider interface {
+	OutboundDispatcher() dispatcher.Outbound
+	StorageProvider() storage.Provider
 }
 
 // New returns introduce service
-func New(ctx context.Context, s storage.Provider) (*Service, error) {
-	store, err := s.OpenStore(Introduce)
+func New(p provider) (*Service, error) {
+	store, err := p.StorageProvider().OpenStore(Introduce)
 	if err != nil {
 		return nil, err
 	}
 
 	svc := &Service{
+		ctx: internalContext{
+			Outbound: p.OutboundDispatcher(),
+		},
 		store:     store,
 		callbacks: make(chan callback),
+		stop:      make(chan struct{}),
 	}
 
 	// start the listener
-	go svc.startInternalListener(ctx)
+	go svc.startInternalListener()
 
 	return svc, nil
 }
 
+// Stop stops service (callback listener)
+func (s *Service) Stop() error {
+	s.closedMutex.Lock()
+	defer s.closedMutex.Unlock()
+	if s.closed {
+		return errors.New("server was already stopped")
+	}
+	close(s.stop)
+	s.closed = true
+	return nil
+}
+
 // startInternalListener listens to messages in gochannel for callback messages from clients.
-func (s *Service) startInternalListener(ctx context.Context) {
+func (s *Service) startInternalListener() {
 	for {
 		select {
 		case msg := <-s.callbacks:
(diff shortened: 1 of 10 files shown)


pkg/didcomm/protocol/introduce/states.go
```diff
@@ -53,11 +65,11 @@ func (s *noOp) CanTransitionTo(_ state) bool {
 	return false
 }
 
-func (s *noOp) ExecuteInbound(_ *metaData) (state, error) {
+func (s *noOp) ExecuteInbound(p provider, _ *metaData) (state, error) {
 	return nil, errors.New("cannot execute no-op")
 }
 
-func (s *noOp) ExecuteOutbound(_ *metaData, _ *service.Destination) (state, error) {
+func (s *noOp) ExecuteOutbound(p provider, _ *metaData, _ *service.Destination) (state, error) {
```
talwinder50: Why are we passing provider in execute in bound and outbound ?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -68,16 +69,20 @@ type Service struct {
 	service.Message
 	store     storage.Store
 	callbacks chan callback
+	sender    *providerImpl
 }
 
 // New returns introduce service
-func New(ctx context.Context, s storage.Provider) (*Service, error) {
+func New(ctx context.Context, s storage.Provider, d dispatcher.Outbound) (*Service, error) {
```
rolsonquadras: We can use a provider similar to [DIDExchange Service](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/service.go#L63).

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -29,6 +33,14 @@ const (
 	stateNameWaiting  = "waiting"
 )
 
+type provider interface {
+	dispatcher.Outbound
+}
+
+type providerImpl struct {
```
rolsonquadras: All the dependencies should come from aries.Context(). We shouldn't implement these in protocols.

troyronda: (more specifically, the protocol should define an interface of the subset of Context that it needs).

soluchok: ok, I will do that more clear :)

soluchok: Done!

pkg/didcomm/protocol/introduce/service.go
soluchok: Done!

pkg/didcomm/protocol/introduce/states.go
soluchok: the provider was equal to context (did-exchange), I have renamed it now (provider to internalContext)

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -68,16 +69,26 @@ type Service struct {
...
+// provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
+type provider interface {
+	OutboundDispatcher() dispatcher.Outbound
+	StorageProvider() storage.Provider
 }
 
 // New returns introduce service
-func New(ctx context.Context, s storage.Provider) (*Service, error) {
-	store, err := s.OpenStore(Introduce)
+func New(ctx context.Context, p provider) (*Service, error) {
```
troyronda: Std lib context.Context should be provided to a request not a constructor.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -421,13 +440,13 @@ func (s *Service) handle(msg *metaData, dest *service.Destination) error {
 		)
 
 		if dest != nil {
-			followup, err = next.ExecuteOutbound(msg, dest)
+			followup, err = next.ExecuteOutbound(s.ctx, msg, dest)
```
troyronda: Context is supposed to be defined for a particular request/transaction - not as a member variable used across transactions. See prior comment.

pkg/didcomm/protocol/introduce/service.go
soluchok: I have created it to stop the listener 
https://github.com/hyperledger/aries-framework-go/pull/575/files#diff-43cf3af5d3c6b9acebc5c635ba4d0f7dR97

pkg/didcomm/protocol/introduce/service.go
soluchok: it is internal context

pkg/didcomm/protocol/introduce/service.go
soluchok: usage here https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/introduce/service.go#L110

pkg/didcomm/protocol/introduce/service.go
soluchok: the same logic in did-exchange

pkg/didcomm/protocol/introduce/service.go
troyronda: This seems to be an usual usage of context. Recommend a Stop method instead.

pkg/didcomm/protocol/introduce/service.go
troyronda: internal context is ok.

pkg/didcomm/protocol/introduce/service.go
soluchok: Done!

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -107,16 +134,22 @@ func (s *Service) startInternalListener(ctx context.Context) {
 			if err := s.processFailure(msg.id, msg.err); err != nil {
 				logger.Errorf("process callback : %s", err)
 			}
-		case <-ctx.Done():
+		case <-s.stop:
```
troyronda: the log message is out of date.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -68,28 +69,54 @@ type Service struct {
 	service.Message
 	store     storage.Store
 	callbacks chan callback
+	ctx       internalContext
+	stop      chan struct{}
+	mu        sync.Mutex
```
troyronda: i would name the mutex similar to the var it is protecting.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -130,7 +163,8 @@ func (s *Service) processFailure(id string, pErr error) error {
 }
 
 // abandon updates the state to abandoned and trigger failure event.
-func (s *Service) abandon(thID string, msg *service.DIDCommMsg, _ error) error {
+func (s *Service) abandon(thID string, msg *service.DIDCommMsg, pErr error) error {
+	logger.Infof("go to abandon: %v", pErr)
```
troyronda: i assume you are cleaning up log levels in a future PR?

soluchok: removed!

pkg/didcomm/protocol/introduce/service.go
soluchok: changed to `closedMutex`

pkg/didcomm/protocol/introduce/service.go
troyronda: It was a general comment :). There are several logger lines that should not "INFO" level.

PR #580, opened 2019-10-24
feat: DIDExchange - Integrate Message events with Webhook Notifications

### LINKED ISSUE #551: RestAPI - Integrate Message event with Webhook

rolsonquadras:
- Trigger webhook notifications for DIDExchange Post State change message events

Closes #551 

### DIFF
--- a/pkg/restapi/operation/didexchange/didexchange.go
+++ b/pkg/restapi/operation/didexchange/didexchange.go
@@ -8,6 +8,7 @@ package didexchange
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -31,14 +32,15 @@ import (
 var logger = log.New("aries-framework/did-exchange")
 
 const (
-	operationID           = "/connections"
-	createInvitationPath  = operationID + "/create-invitation"
-	receiveInvtiationPath = operationID + "/receive-invitation"
-	acceptInvitationPath  = operationID + "/{id}/accept-invitation"
-	connections           = operationID
-	connectionsByID       = operationID + "/{id}"
-	acceptExchangeRequest = operationID + "/{id}/accept-request"
-	removeConnection      = operationID + "/{id}/remove"
+	operationID             = "/connections"
+	createInvitationPath    = operationID + "/create-invitation"
+	receiveInvtiationPath   = operationID + "/receive-invitation"
+	acceptInvitationPath    = operationID + "/{id}/accept-invitation"
+	connections             = operationID
+	connectionsByID         = operationID + "/{id}"
+	acceptExchangeRequest   = operationID + "/{id}/accept-request"
+	removeConnection        = operationID + "/{id}/remove"
+	connectionsWebhookTopic = "connections"
 )
 
 // provider contains dependencies for the Exchange protocol and is typically created by using aries.Context()
@@ -350,21 +352,64 @@ func (c *Operation) startClientEventListener() error {
 		return fmt.Errorf("didexchange message event registration failed: %w", err)
 	}
 
-	// auto execute the actions
+	// event listeners
 	go func() {
-		err := service.AutoExecuteActionEvent(c.actionCh)
-		if err != nil {
-			logger.Errorf("auto action event execution failed: %s", err)
+		for {
+			select {
+			case e := <-c.actionCh:
+				c.handleActionEvents(e)
+			case e := <-c.msgCh:
+				err := c.handleMessageEvents(e)
+				if err != nil {
+					logger.Errorf("handle message events failed : %s", err)
+				}
+			}
 		}
 	}()
 
-	go func() {
-		for e := range c.msgCh {
-			// TODO https://github.com/hyperledger/aries-framework-go/issues/551 - Integrate Message event with Webhook
-			// for now, log the messages
-			logger.Infof("message event received : type=%s", e.Type)
+	return nil
+}
+
+func (c *Operation) handleMessageEvents(e service.StateMsg) error {
+	if e.Type == service.PostState {
+		switch v := e.Properties.(type) {
+		case didexchange.Event:
+			props := v
+			err := c.sendConnectionNotification(props.ConnectionID(), e.StateID)
+			if err != nil {
+				return fmt.Errorf("send connection notification failed : %w", err)
+			}
+		case error:
+			return fmt.Errorf("service processing failed : %w", v)
+		default:
+			return errors.New("event is not of DIDExchange event type")
 		}
-	}()
+	}
+	return nil
+}
+
+func (c *Operation) handleActionEvents(e service.DIDCommAction) {
+	// TODO https://github.com/hyperledger/aries-framework-go/issues/579 - Integrate Action events with webhooks
+	e.Continue()
+}
+
+func (c *Operation) sendConnectionNotification(connectionID, stateID string) error {
+	// TODO - fetch connection data from DB (update once https://github.com/hyperledger/aries-framework-go/pull/573
+	// is merged)
+	conn := &webhook.ConnectionMsg{
+		ConnectionID: connectionID,
+		State:        stateID,
+	}
+
+	jsonMessage, err := json.Marshal(conn)
+	if err != nil {
+		return fmt.Errorf("connection notification json marshal : %w", err)
+	}
+
+	err = c.notifier.Notify(connectionsWebhookTopic, jsonMessage)
+	if err != nil {
+		return fmt.Errorf("connection notification webhook : %w", err)
+	}
 
 	return nil
 }
--- a/pkg/restapi/operation/didexchange/didexchange_test.go
+++ b/pkg/restapi/operation/didexchange/didexchange_test.go
@@ -319,13 +319,31 @@ func (m mockWriter) Write([]byte) (int, error) {
 
 func TestServiceEvents(t *testing.T) {
 	store := &mockstore.MockStore{Store: make(map[string][]byte)}
-	recorder := didexsvc.NewConnectionRecorder(store)
+
 	didExSvc, err := didexsvc.New(&did.MockDIDCreator{}, &protocol.MockProvider{CustomStore: store})
 	require.NoError(t, err)
 
+	done := make(chan struct{})
+
 	// create the client
 	op, err := New(&mockprovider.Provider{StorageProviderValue: mockstore.NewMockStoreProvider(),
-		ServiceValue: didExSvc}, webhook.NewHTTPNotifier(nil))
+		ServiceValue: didExSvc},
+		&mockNotifier{
+			notifyFunc: func(topic string, message []byte) error {
+				require.Equal(t, connectionsWebhookTopic, topic)
+
+				conn := didexchange.Connection{}
+				jsonErr := json.Unmarshal(message, &conn)
+				require.NoError(t, jsonErr)
+
+				if conn.State == "responded" {
+					close(done)
+				}
+
+				return nil
+			},
+		},
+	)
 	require.NoError(t, err)
 	require.NotNil(t, op)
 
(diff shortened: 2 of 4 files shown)


pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -350,21 +351,66 @@ func (c *Operation) startClientEventListener() error {
...
+		case error:
+			logger.Warnf("service processing failed : %s", v)
+			return
+		default:
+			logger.Errorf("event is not of DIDExchange event type")
+			return
 		}
-	}()
+
+		err := c.sendConnectionNotification(props.ConnectionID(), e.StateID)
```
DRK3: I see a way to simplify this function. You can just move these four lines inside of the `case didexchange.event` block. Then you don't need that extra `didexchange.Event var` at the top. And then you can also get rid of the `returns` in the `error` and `default` cases.

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -350,21 +351,66 @@ func (c *Operation) startClientEventListener() error {
...
+}
+
+func (c *Operation) handleActionEvents(e service.DIDCommAction) {
+	// TODO https://github.com/hyperledger/aries-framework-go/issues/579 - Integrate Action events with webhooks
+	e.Continue()
+}
+
+func (c *Operation) sendConnectionNotification(connectionID, stateID string) error {
+	// TODO - fetch connection data from DB (update once https://github.com/hyperledger/aries-framework-go/pull/573
+	//  is merged)
```
DRK3: You've got an extra leading space here.

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -350,21 +351,66 @@ func (c *Operation) startClientEventListener() error {
...
+			State:        stateID,
+		},
+	}
+
+	jsonMessage, err := json.Marshal(conn)
+	if err != nil {
+		return fmt.Errorf("connection notification json marshal : %w", err)
+	}
+
+	err = c.notifier.Notify("connections", jsonMessage)
```
DRK3: I think that `"connections"` should be a constant somewhere since it'll be used in quite a few places. Perhaps a constant in the payload.go file?

sudeshrshetty: topic should be constant?

pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -319,9 +318,28 @@ func TestServiceEvents(t *testing.T) {
...
 
+	done := make(chan struct{})
+
 	// create the client
 	op, err := New(&mockprovider.Provider{StorageProviderValue: mockstore.NewMockStoreProvider(),
-		ServiceValue: didExSvc}, webhook.NewHTTPNotifier(nil))
+		ServiceValue: didExSvc},
+		&mockNotifier{
+			notifyFunc: func(topic string, message []byte) error {
+				require.Equal(t, "connections", topic)
```
DRK3: `"connections"` should probably be a constant

pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -319,9 +318,28 @@ func TestServiceEvents(t *testing.T) {
...
+
+				conn := didexchange.Connection{}
+				jsonErr := json.Unmarshal(message, &conn)
+				require.NoError(t, jsonErr)
+
+				if conn.State == "responded" {
```
DRK3: Perhaps these connection states should be constants defined somewhere as well, since I think there's a pre-defined finite number of them right?

pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -319,9 +318,28 @@ func TestServiceEvents(t *testing.T) {
...
+					done <- struct {
+					}{}
```
DRK3: Not sure, but is there something funny going on with the curly braces here?

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: This is a webhook consumer validation (a separate process). So i feel it would make sense to have a validation against a literal rather than against a lib constant.

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: It's passing the empty struct to the channel.

pkg/restapi/operation/didexchange/didexchange.go
rolsonquadras: [Done](https://github.com/hyperledger/aries-framework-go/compare/bbb4d18c35a98651aa199db0c9a11c90d8e3e99f..b7b4d6ceea82fbb36e3bf39e019d95c5cfb30c8c#diff-0ee8eb9e72164ce9fd9b1720100d3b58R328).

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: https://github.com/hyperledger/aries-framework-go/pull/580#discussion_r338765349

pkg/restapi/operation/didexchange/didexchange.go
rolsonquadras: https://github.com/hyperledger/aries-framework-go/pull/580#discussion_r338765349

pkg/restapi/operation/didexchange/didexchange.go
rolsonquadras: [Done](https://github.com/hyperledger/aries-framework-go/compare/bbb4d18c35a98651aa199db0c9a11c90d8e3e99f..b7b4d6ceea82fbb36e3bf39e019d95c5cfb30c8c#diff-5213b774cf71e1cef26c7bb7c48ee94eR374).

pkg/restapi/operation/didexchange/didexchange.go
rolsonquadras: Removed the extra leading space.

pkg/restapi/operation/didexchange/didexchange_test.go
DRK3: True, but it does use the same restAPI + webhook library, so it doesn't look totally separate to me code wise...

pkg/restapi/operation/didexchange/didexchange_test.go
DRK3: It might just be me but I can't figure out what the extra {} is for

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: Webhook server won't use aries library. it'll be a separate server/process. Webhook server and Agent communicate through RestAPIs.

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: This is a good reference - https://stackoverflow.com/questions/45122905/how-do-struct-and-struct-work-in-go.

Note - I have updated the code to close the channel rather than passing the struct.

pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -342,31 +359,17 @@ func TestServiceEvents(t *testing.T) {
...
-			actualState = string(v)
-			if err != nil || expected != string(v) {
-				continue
-			}
-			return
-		}
+	select {
+	case <-done:
+	case <-time.After(1 * time.Second):
+		require.Fail(t, "tests are not validated")
```
DRK3: I think you mean "valid" not "validated". "not validated" basically means "not tested".

pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -389,3 +392,36 @@ func TestOperationEventError(t *testing.T) {
...
 }
+
+func TestHandleMessageEvent(t *testing.T) {
+	op, err := New(&mockprovider.Provider{StorageProviderValue: mockstore.NewMockStoreProvider(),
+		ServiceValue: &protocol.MockDIDExchangeSvc{}}, webhook.NewHTTPNotifier(nil))
+	require.NoError(t, err)
+	require.NotNil(t, op)
+
+	op.handleMessageEvents(service.StateMsg{Type: service.PostState, Properties: "invalid didex prop type"})
+	op.handleMessageEvents(service.StateMsg{Type: service.PostState, Properties: errors.New("err type")})
```
DRK3: This test doesn't check for anything beyond this point, so it looks like it's always going to pass.

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -350,21 +352,62 @@ func (c *Operation) startClientEventListener() error {
...
 
-	go func() {
-		for e := range c.msgCh {
-			// TODO https://github.com/hyperledger/aries-framework-go/issues/551 - Integrate Message event with Webhook
-			// for now, log the messages
-			logger.Infof("message event received : type=%s", e.Type)
+	return nil
+}
+
+func (c *Operation) handleMessageEvents(e service.StateMsg) {
```
DRK3: Consider returning the error so you can test this function?

pkg/restapi/operation/didexchange/didexchange_test.go
DRK3: Although "tests are not valid" also doesn't make sense. I think a more specific message here is needed. Or even just "test failed, no response..." or something like that.

rolsonquadras: test's are valid but those were not validated due to timeout.

pkg/restapi/operation/didexchange/didexchange_test.go
DRK3: Hmm... is there some value to not extracting that string out into the payloads.go file? Or some other common file? If the names are ever changed in the future then we'll have to update in multiple places.

pkg/restapi/operation/didexchange/didexchange_test.go
DRK3: Ah ok, I understand what you mean now. Consider changing the message to "validation could not be completed before timeout" or something similar.

pkg/restapi/operation/didexchange/didexchange.go
rolsonquadras: well, the agent can't meaningfully handle these other than logging as these are just stream events. 

But added return and logged by the caller.

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: https://github.com/hyperledger/aries-framework-go/pull/580#discussion_r338776651

pkg/restapi/operation/didexchange/didexchange.go
DRK3: Yeah, I encountered this issue a bunch when I was unit testing. Returning an error at least allows me to unit test it, and in the non-test files I just logged. Not sure if there's a better way.

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: I understand what you are trying to say. The State names are [internal](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/states.go#L32). And here the literal is defined only once (so it won't make much sense to move it to constant).

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: added 'timeout' to the message.

pkg/restapi/operation/didexchange/didexchange_test.go
DRK3: Gotcha. Makes sense!

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -350,21 +352,64 @@ func (c *Operation) startClientEventListener() error {
...
+
+func (c *Operation) handleActionEvents(e service.DIDCommAction) {
+	// TODO https://github.com/hyperledger/aries-framework-go/issues/579 - Integrate Action events with webhooks
+	e.Continue()
+}
+
+func (c *Operation) sendConnectionNotification(connectionID, stateID string) error {
+	// TODO - fetch connection data from DB (update once https://github.com/hyperledger/aries-framework-go/pull/573
+	// is merged)
+	conn := &webhook.ConnectionMsg{
```
troyronda: I feel like the model location will need a refactoring.

It's weird that ConnectionMsg is at the top level.

rolsonquadras: created an issue https://github.com/hyperledger/aries-framework-go/issues/583.

Issue #587, opened 2019-10-25
Callback logic need to be refactored


soluchok:
`sendActionEvent` function should not store any data related to the callback, according to current logic we can pass object directly to the `processCallback` function

Requirements:
* action event message should keep the copied object of the message (to prevent modifying by user)
* `DIDCommAction.Message` should be without pointer

Issue #438, opened 2019-10-07
Cleanup/Update data on Stop Event Action


rolsonquadras:
Based on TODO - https://github.com/hyperledger/aries-framework-go/compare/42c5b3f270b75af9cd92cb0657abbcacffce467f..d4d6d2ee8a27820203156ad20b4f7a69e8277911#diff-f761c91e753eba1c42a68076df06ce12R333

soluchok:
If I am right this issue will be outdated after https://github.com/hyperledger/aries-framework-go/issues/587

PR #591, opened 2019-10-25
WIP refactor: removed storage usage from Callback logic

soluchok:
Issues: https://github.com/hyperledger/aries-framework-go/issues/587 https://github.com/hyperledger/aries-framework-go/issues/438


pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -320,6 +321,9 @@ func TestServiceEvents(t *testing.T) {
 	require.NoError(t, err)
 
 	done := make(chan struct{})
+	// TODO: the test is failing sometimes, need to review the test logic
+	//       and make sure that `once` solution is fine or fix this test according to logic
+	var once sync.Once
```
soluchok: @rolsonquadras Let me know if this is a solution, looks like we are closing channel more than once

```
 [aries-framework/did-exchange] 2019/10/28 11:36:06 UTC - didexchange.(*Operation).writeResponse -> ERROR Unable to send error response, failed to write
panic: close of closed channel

goroutine 6493 [running]:
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.TestServiceEvents.func1(0xb15a09, 0xb, 0xc0020a3560, 0x120, 0x120, 0x43d4c0, 0xc0026aae00)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange_test.go:336 +0x20a
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*mockNotifier).Notify(0xc0000b9a80, 0xb15a09, 0xb, 0xc0020a3560, 0x120, 0x120, 0x0, 0x431deb)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange_test.go:471 +0x80
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).sendConnectionNotification(0xc002346cd0, 0xc002684d50, 0x24, 0x7f7ca045a008, 0xc002341dc0)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:416 +0x5e4
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).handleMessageEvents(0xc002346cd0, 0xc002703f18, 0x2, 0x1)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:378 +0x182
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).startClientEventListener.func1(0xc002346cd0)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:362 +0x1fe
created by github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).startClientEventListener
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:356 +0x291
FAIL	github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange	1.907s
?   	github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange/models	[no test files]

```

pkg/didcomm/common/service/service.go
```diff
@@ -35,17 +35,44 @@ type Header struct {
...
+}
+
 // NewDIDCommMsg returns DIDCommMsg with Header
-func NewDIDCommMsg(payload []byte) (*DIDCommMsg, error) {
-	msg := &DIDCommMsg{Payload: payload}
+func NewDIDCommMsg(payload []byte) (DIDCommMsg, error) {
+	msg := DIDCommMsg{Payload: payload}
 	if err := json.Unmarshal(msg.Payload, &msg.Header); err != nil {
-		return nil, fmt.Errorf("invalid payload data format: %w", err)
+		return msg, fmt.Errorf("invalid payload data format: %w", err)
```
rolsonquadras: This shouldn't return non-nil err value with a non-nil val ie `return val, err`.

pkg/restapi/operation/didexchange/didexchange_test.go
rolsonquadras: Once will be just a temp solution (looks fine). For some reason events are triggered twice, hence the execution goes inside if clause. There is a issue to investigate/fix - #607

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -209,8 +203,12 @@ func (s *Service) handle(msg *message) error {
 	for !isNoOp(next) {
 		//TODO: Issue-578 is created to consider if we need to create connection ID at pre state level or no
 		s.sendMsgEvents(&service.StateMsg{
-			Type: service.PreState, Msg: msg.Msg, StateID: next.Name(),
-			Properties: createEventProperties("", "")})
+			ProtocolName: DIDExchange,
+			Type:         service.PreState,
+			Msg:          msg.Msg.Clone(),
```
rolsonquadras: Msg is already pass by value. Any specific reason to Clone ?

soluchok: yeah, the payload can be modified by the user (bc Payload is slice and slice keeps a pointer to the array the hood e.g `msg.Payload[0] = 0x1`) and also `Header` is a pointer as well

pkg/didcomm/common/service/service.go
soluchok: yeah, but it is impossible, we can return nil only if `DIDCommMsg` is a pointer

soluchok: ok. I will revert it to pointer

soluchok: Done!

pkg/didcomm/common/service/service.go
```diff
@@ -35,12 +35,39 @@ type Header struct {
 	Type   string           `json:"@type"`
 }
 
+func (h *Header) clone() *Header {
+	if h == nil {
```
llorllale: This can never happen. The caller would get a panic if they attempted to call this method on a nil reference.

pkg/didcomm/common/service/service.go
```diff
@@ -35,12 +35,39 @@ type Header struct {
...
 type DIDCommMsg struct {
 	Header  *Header
 	Payload []byte
 }
 
+// Clone creates new DIDCommMsg with the same data
+// the cloned message is safe for delivering to the client
+// it prevents modifying by the client
+func (m *DIDCommMsg) Clone() *DIDCommMsg {
+	if m == nil {
```
llorllale: Same here - this is an impossible condition inside this method.

pkg/didcomm/common/service/service.go
soluchok: It will not panic I have a test for that
https://github.com/hyperledger/aries-framework-go/pull/591/files#diff-2b388021d16ee766a8d669d2e3d88ffeR98-R101

pkg/didcomm/common/service/service.go
```diff
@@ -35,12 +35,39 @@ type Header struct {
...
 // DIDCommMsg did comm msg
 type DIDCommMsg struct {
 	Header  *Header
 	Payload []byte
 }
 
+// Clone creates new DIDCommMsg with the same data
+// the cloned message is safe for delivering to the client
+// it prevents modifying by the client
+func (m *DIDCommMsg) Clone() *DIDCommMsg {
```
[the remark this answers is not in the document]
soluchok: `the cloned message is safe for delivering to the client` It means Client can not modify our data
without cloning, it can be a tricky hack how to change the protocol logic
e.g Client received an action event than he modified the message data and then he executes the callback (previously we stored msg in DB it was impossible to modify it)

pkg/didcomm/common/service/service.go
troyronda: We should not be nil checking the receiver pointer. This is overkill and unusual.

soluchok: `Clone` is a public method and we do not know how clients will use it. I just provided a reliable solution instead of panic. We should not expect that the message cannot be nil. If the message is nil and it is possible to call the `Clone` method (according to GO) why we should panic instead of handling it and return nil. Copy <nil> DIDCommMsg returns <nil> DIDCommMsg it does not break the logic right? and more reliable. This `Clone` function has nothing in common with Client usage. If the Client should not work with nill msg it is up to him what to do in such case but it should not be related to us (single responsibility principle) 

clients: packages `didexchange, introduce` 
us: `package service`

troyronda: I'm not concerned about clients getting a panic on calling a method on a nil variable. That is quite normal and expected in Go development :).

Is it impossible to have a nil in our current usage?

soluchok: No, it is not, I can remove it and all tests will pass

soluchok: Do you want me to remove it :)?

PR #599, opened 2019-10-25
fix: Execute state handling in a goroutine

rolsonquadras:
- Added goroutine as State execution was executing in calling agents inbound thread. This would resulted in agent invoking next call before calling agent executing previous thread.

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -164,8 +164,15 @@ func (s *Service) HandleInbound(msg *service.DIDCommMsg) error {
 		}
 		return nil
 	}
+
 	// if no action event is triggered, continue the execution
-	return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	go func() {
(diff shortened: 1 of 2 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -164,8 +164,15 @@ func (s *Service) HandleInbound(msg *service.DIDCommMsg) error {
 		}
 		return nil
 	}
+
 	// if no action event is triggered, continue the execution
-	return s.handle(&message{Msg: msg, ThreadID: thid, NextStateName: next.Name()})
+	go func() {
```
troyronda: We shouldn’t be spinning up a new go routine on every message (with no bound and no timeout context on the handling).

 This should be a queue and pool pattern, and needs a cancellable context.

rolsonquadras: Yeah, I'll add it. Since this was a blocker, added a simple go routine. If possible, need to find a better way of handling the inbound request.

rolsonquadras: https://github.com/hyperledger/aries-framework-go/issues/600.

PR #612, opened 2019-10-28
Fix service endpoint in create invitation response

### LINKED ISSUE #572: local address in create invitation response service endpoint
As per current implementation agent reuses `--inbound-host` provided in `aries-agentd` command as service endpoint in create invitation response. It cannot be resolved by invitee agent.

Fix: should be separate mechanism to pass service endpoint to aries agent.

DRK3:
Closes #572 

Added a command line flag for specifying the external inbound host name. The external address is what's returned when you call Endpoint() on an Inbound object.

Fixed some small typos and formatting issues I found.

### DIFF
--- a/cmd/aries-agentd/startcmd/start.go
+++ b/cmd/aries-agentd/startcmd/start.go
@@ -36,7 +36,15 @@ const (
 	AgentInboundHostFlagShorthand = "i"
 
 	// AgentInboundHostFlagUsage is the usage text for the agent inbound host command line argument.
-	AgentInboundHostFlagUsage = "Inbound Host Name:Port"
+	AgentInboundHostFlagUsage = "Inbound Host Name:Port. This is used internally to start the inbound server."
+
+	agentInboundHostExternalFlagName = "inbound-host-external"
+
+	agentInboundHostExternalFlagShorthand = "e"
+
+	agentInboundHostExternalFlagUsage = "Inbound Host External Name:Port." +
+		" This is the URL for the inbound server as seen externally." +
+		" If not provided, then the internal inbound host will be used here."
 
 	// AgentDBPathFlagName is the flag name for the database path command line argument.
 	AgentDBPathFlagName = "db-path"
@@ -50,11 +58,9 @@ const (
 	// AgentWebhookFlagName is the flag name for the webhook command line argument.
 	AgentWebhookFlagName = "webhook-url"
 
-	// AgentWebhookFlagShorthand is the flag shorthand name for the webhook command line argument.
-	AgentWebhookFlagShorthand = "w"
+	agentWebhookFlagShorthand = "w"
 
-	// AgentWebhookFlagUsage is the flag usage text for the webhook command line argument.
-	AgentWebhookFlagUsage = "URL to send notifications to." +
+	agentWebhookFlagUsage = "URL to send notifications to." +
 		" This flag can be repeated, allowing for multiple listeners."
 )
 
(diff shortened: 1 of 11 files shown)


cmd/aries-agentd/startcmd/start.go
```diff
@@ -50,11 +50,9 @@ const (
 	// AgentWebhookFlagName is the flag name for the webhook command line argument.
 	AgentWebhookFlagName = "webhook-url"
 
-	// AgentWebhookFlagShorthand is the flag shorthand name for the webhook command line argument.
-	AgentWebhookFlagShorthand = "w"
+	agentWebhookFlagShorthand = "w"
 
-	// AgentWebhookFlagUsage is the flag usage text for the webhook command line argument.
-	AgentWebhookFlagUsage = "URL to send notifications to." +
+	agentWebhookFlagUsage = "URL to send notifications to." +
```
DRK3: Just a minor refactor - I realized these were exported when they didn't need to be.

pkg/didcomm/transport/http/inbound.go
```diff
@@ -111,7 +113,12 @@ func NewInbound(addr string) (*Inbound, error) {
 		return nil, errors.New("http address is mandatory")
 	}
 
-	return &Inbound{server: &http.Server{Addr: addr}}, nil
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create new Inbound: %s", err)
```
rolsonquadras: I don't see this error being validated in test cases.

pkg/didcomm/transport/http/inbound.go
```diff
@@ -121,6 +128,8 @@ func (i *Inbound) Start(prov transport.InboundProvider) error {
 		return fmt.Errorf("HTTP server start failed: %w", err)
 	}
 
+	i.server = &http.Server{Addr: ":" + i.port}
```
rolsonquadras: Any specific reason to move the server initiation from New() to Start() function ? If Start() is called multiple times, we might have multiple servers created for same instance.

pkg/didcomm/transport/http/inbound_test.go
```diff
@@ -189,6 +189,12 @@ func TestInboundTransport(t *testing.T) {
 	})
 }
 
+func TestPanicStartWithUnsplittableServerAddr(t *testing.T) {
```
sudeshrshetty: Can you add tests to cover different hostname patterns, like ones without port(DNS), ":PORT", etc

pkg/didcomm/transport/http/inbound.go
DRK3: I check for it in TestStartWithUnsplittableServerAddr. (Note that I used to have "panic" in the name, but that was a mistake so I removed that word)

I check for whether the error text contains "too many colons in address", which is the specific error that gets returned by the SplitHostPort function. Should I also be checking for that "unable to create new Inbound" message too?

pkg/didcomm/transport/http/inbound.go
DRK3: Not sure why I did that originally, but you're right. I moved it back to the New function.

pkg/didcomm/transport/http/inbound_test.go
DRK3: I added some tests. Let me know if that was what you were expecting or if you were looking for something else

pkg/didcomm/transport/http/inbound.go
```diff
@@ -111,7 +113,12 @@ func NewInbound(addr string) (*Inbound, error) {
 		return nil, errors.New("http address is mandatory")
 	}
 
-	return &Inbound{server: &http.Server{Addr: addr}}, nil
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create new Inbound: %s", err)
+	}
+
+	return &Inbound{host: host, port: port, server: &http.Server{Addr: ":" + port}}, nil
```
troyronda: I do not understand why the address portion is being removed. This is used to bind to a particular interface.

DRK3: Addressed as discussed in the meeting earlier.

PR #595, opened 2019-10-25
feat: DIDExchange REST MsgEvent - Fetch Connections Details from DB

talwinder50:
closes #590

Signed-off-by: talwinder.kaur <talwinder.kaur@securekey.com>

PR #616, opened 2019-10-28
fix: Get Connection details based on state 

### LINKED ISSUE #607: Unit test failing
[aries-framework/did-exchange] 2019/10/28 11:36:06 UTC - didexchange.(*Operation).writeResponse -> ERROR Unable to send error response, failed to write
panic: close of closed channel

goroutine 6493 [running]:
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.TestServiceEvents.func1(0xb15a09, 0xb, 0xc0020a3560, 0x120, 0x120, 0x43d4c0, 0xc0026aae00)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange_test.go:336 +0x20a
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*mockNotifier).Notify(0xc0000b9a80, 0xb15a09, 0xb, 0xc0020a3560, 0x120, 0x120, 0x0, 0x431deb)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange_test.go:471 +0x80
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).sendConnectionNotification(0xc002346cd0, 0xc002684d50, 0x24, 0x7f7ca045a008, 0xc002341dc0)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:416 +0x5e4
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).handleMessageEvents(0xc002346cd0, 0xc002703f18, 0x2, 0x1)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:378 +0x182
github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).startClientEventListener.func1(0xc002346cd0)
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:362 +0x1fe
created by github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange.(*Operation).startClientEventListener
	/home/soluchok/projects/aries-framework-go/pkg/restapi/operation/didexchange/didexchange.go:356 +0x291
FAIL	github.com/hyperledger/aries-framework-go/pkg/restapi/operation/didexchange	1.907s

rolsonquadras:
Currently, RestAPI message handle the tests fail intermittently. The connectionID retrieved from events used to fetch connection data from database. There is a possibility that the data in db gets updated before consumer calls the
db. This would make previous state data getting lost and consumer would receive same state notification twice.

- Save connection data after state execution.
- Add new API to get connection based on connectionID and stateID.

Closes #607 

### DIFF
--- a/pkg/didcomm/protocol/didexchange/persistence.go
+++ b/pkg/didcomm/protocol/didexchange/persistence.go
@@ -114,7 +114,19 @@ func (c *ConnectionRecorder) GetInvitation(id string) (*Invitation, error) {
 
 // GetConnectionRecord return connection record based on the connection ID
 func (c *ConnectionRecorder) GetConnectionRecord(connectionID string) (*ConnectionRecord, error) {
-	k := connectionKeyPrefix(connectionID)
+	return c.getAndUnmarshal(connectionKeyPrefix(connectionID))
+}
+
+// GetConnectionRecordAtState return connection record based on the connection ID and state.
+func (c *ConnectionRecorder) GetConnectionRecordAtState(connectionID, stateID string) (*ConnectionRecord, error) {
+	if stateID == "" {
+		return nil, errors.New("stateID can't be empty")
+	}
+
+	return c.getAndUnmarshal(connectionKeyPrefix(connectionID + stateID))
+}
+
+func (c *ConnectionRecorder) getAndUnmarshal(k string) (*ConnectionRecord, error) {
 	connRecordBytes, err := c.store.Get(k)
 	if err != nil {
 		return nil, fmt.Errorf("get connection record: %w", err)
@@ -139,12 +151,26 @@ func (c *ConnectionRecorder) GetConnectionRecordByNSThreadID(nsThreadID string) 
 }
 
 // saveConnectionRecord saves the connection record against the connection id  in the store
+// TODO - https://github.com/hyperledger/aries-framework-go/issues/622 Transient v Permanent Data store
 func (c *ConnectionRecorder) saveConnectionRecord(record *ConnectionRecord) error {
-	k := connectionKeyPrefix(record.ConnectionID)
-	bytes, err := json.Marshal(record)
+	if err := c.marshalAndSave(connectionKeyPrefix(record.ConnectionID), record); err != nil {
+		return fmt.Errorf("save connection record: %w", err)
+	}
+
+	if record.State != "" {
+		return c.marshalAndSave(connectionKeyPrefix(record.ConnectionID+record.State), record)
+	}
+
+	return nil
+}
+
+func (c *ConnectionRecorder) marshalAndSave(k string, v *ConnectionRecord) error {
+	bytes, err := json.Marshal(v)
+
 	if err != nil {
 		return fmt.Errorf("save connection record: %w", err)
 	}
+
 	return c.store.Put(k, bytes)
 }
 
@@ -179,6 +205,7 @@ func (c *ConnectionRecorder) saveNSThreadID(thid, namespace, connectionID string
 
 	return c.store.Put(k, []byte(connectionID))
 }
+
 func prepareConnectionRecord(connRecBytes []byte) (*ConnectionRecord, error) {
 	connRecord := &ConnectionRecord{}
 	err := json.Unmarshal(connRecBytes, connRecord)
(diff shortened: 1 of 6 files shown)


pkg/didcomm/common/model/didexchange/models.go
```diff
@@ -0,0 +1,20 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
```
troyronda: Why isn’t this part of client package

rolsonquadras: it creates cyclic dependency as the implementation is present in [service pkg](https://github.com/hyperledger/aries-framework-go/pull/616/files#diff-79de382a052bbdb4213b9dd373b0e2f6R15). Also, @fqutishat and I were thinking that the [common models](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/client/didexchange/models.go#L47) (currently client models which wrap service models) between client and service could be moved to this common package.

troyronda:
> There is a possibility that the data in db gets updated before consumer calls the
db.

Should we be concerned about a bug in this logic? Somehow so get the feeling that we might be hiding a bug in this PR.

rolsonquadras:
> > There is a possibility that the data in db gets updated before consumer calls the
> > db.
> 
> Should we be concerned about a bug in this logic? Somehow I get the feeling that we might be hiding a bug in this PR.

We are trying to fix the bug. Here is the flow when Agent receives ExchangeRequest.
1. Service triggers message event when status updates to requested (`state - requested`)
2. Consumer receives event and calls client to get the Connection data based on connectionID from event message.
3. Meanwhile, the service sends exchangeResponse message to other agent and updates the state (`state - responded`).
4. After executing pt3, the service triggers an event (`state - responded`)

Now there is a race condition. If pt3 completes before pt2, then the stateID received in pt2 would be `responded`. This will result in RestAPI client receiving 2 `responded` events. 

In this fix, we are sending whole connection record as part of the event rather than consumer looking up for the connection record by invoking the client (at that point data might have changed).

troyronda:
@rolsonquadras Instead of sending an incomplete connection record and creating a new model package, why not simply add a state field to the event?

rolsonquadras:
> @rolsonquadras Instead of sending an incomplete connection record and creating a new model package, why not simply add a state field to the event?

@troyronda We already have that [data (connectionID and state)](https://github.com/hyperledger/aries-framework-go/pull/595/files#diff-5213b774cf71e1cef26c7bb7c48ee94eR399) from event. The [webhook](https://github.com/hyperledger/aries-cloudagent-python/blob/master/AdminAPI.md#pairwise-connection-record-updated-connections) expects the [connection record](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/restapi/operation/didexchange/models.go#L9) with each update.

soluchok:
As I understand when we execute `GetConnection Record(connection string)` we can catch this bug again in different places  (`Meanwhile, the service sends exchangeResponse message to other agent and updates the state`). This bug is pretty similar to  `Execute(...) (... stateAction ..)` when we return stateAction to update DB before stateAction execution. I agree with @troyronda  that probably we are hiding some bug (probably related to architecture). I would recommend reviewing application logic and think about a reliable solution that can fix these issues. 
I think that the best way is to provide one thread (goroutine) per protocol thread ID where this goroutine will consistently handle messages related to this thread

pkg/didcomm/protocol/didexchange/persistence.go
```diff
@@ -140,11 +152,24 @@ func (c *ConnectionRecorder) GetConnectionRecordByNSThreadID(nsThreadID string)
...
 // saveConnectionRecord saves the connection record against the connection id  in the store
 func (c *ConnectionRecorder) saveConnectionRecord(record *ConnectionRecord) error {
-	k := connectionKeyPrefix(record.ConnectionID)
-	bytes, err := json.Marshal(record)
+	if err := c.marshalAndSave(connectionKeyPrefix(record.ConnectionID), record); err != nil {
+		return fmt.Errorf("save connection record: %w", err)
+	}
+
+	if record.State != "" {
+		return c.marshalAndSave(connectionKeyPrefix(record.ConnectionID+record.State), record)
```
troyronda: TODO: separate the storage for in-progress states.
TODO: activity log

PR #618, opened 2019-10-28
refactor: WIP Move did creator outside of wallet

### LINKED ISSUE #577: Wallet should not know about service endpoint
Currently wallet requires knowledge of service endpoint in order to create peer dids. This should not be the case. Keys should be created inside of the wallet but did document should be created outside of wallet. Q: Should we still store it in the wallet or have separate storage?

sandrask:
Closes #577


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -88,14 +88,14 @@ type Service struct {
 
 type context struct {
 	outboundDispatcher dispatcher.Outbound
-	didCreator         did.Creator
+	didCreator         didcreator.Creator
```
sudeshrshetty: stutter??

pkg/framework/aries/api/didcreator/didcreator.go
```diff
@@ -0,0 +1,73 @@
...
+*/
+
+package didcreator
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+// Creator allows for creation of DID document
+type Creator interface {
```
sudeshrshetty: Question: What about to storage and get DID part inside wallet?

sandrask: Another PR #620 that is currently in progress will expose peer storage functionality so the sequence of calls will be Create followed by Store. This way we can store 'my' and 'their' dids in the same storage. Later on we can probably consolidate creator, storage and resolver  in the same peer method (VDR). This will have to be discussed and handled separately.

PR #621, opened 2019-10-29
refactor: Keep extra fields of termsOfUse of VC in a map

### LINKED ISSUE #586: Keep extra fields of termsOfUse of VC in a map.
Currently we use `interface{}` type for `termsOfUse` what is too generic. According to [VC spec](https://w3c.github.io/vc-data-model/#terms-of-use), `type` field is mandatory and `id` field is optional, and that "precise contents of each term of use is determined by the specific termsOfUse type definition".
We should use 'typedID' type and keep extra fields in `map[string]interface{}`.

kdimak:
Closes #586

### DIFF
--- a/pkg/doc/verifiable/common.go
+++ b/pkg/doc/verifiable/common.go
@@ -55,16 +55,47 @@ func (ja JWSAlgorithm) jose() (jose.SignatureAlgorithm, error) {
 // Proof defines embedded proof of Verifiable Credential
 type Proof interface{}
 
-type typedID struct {
+// ExtraFields is a map of extra fields of struct build when unmarshalling JSON which are not
+// mapped to the struct fields.
+type ExtraFields map[string]interface{}
+
+// TypedID defines a flexible structure with id and name fields and arbitrary extra fields
+// kept in ExtraFields.
+type TypedID struct {
 	ID   string `json:"id,omitempty"`
 	Type string `json:"type,omitempty"`
+
+	ExtraFields `json:"-"`
 }
 
-// RefreshService provides a way to automatic refresh of expired Verifiable Credential
-type RefreshService typedID
+// MarshalJSON defines custom marshalling of TypedID to JSON.
+// TODO hide this exported method
+func (tid *TypedID) MarshalJSON() ([]byte, error) {
+	type Alias TypedID
+	alias := (*Alias)(tid)
 
-// TermsOfUse represents terms of use of Verifiable Credential by Issuer or Verifiable Presentation by Holder.
-type TermsOfUse interface{}
+	data, err := marshalWithExtraFields(alias, tid.ExtraFields)
+	if err != nil {
+		return nil, fmt.Errorf("marshal TypedID: %w", err)
+	}
+
+	return data, nil
+}
+
+// UnmarshalJSON defines custom unmarshalling of TypedID from JSON.
+// TODO hide this exported method
+func (tid *TypedID) UnmarshalJSON(data []byte) error {
+	type Alias TypedID
+	alias := (*Alias)(tid)
+
+	tid.ExtraFields = make(ExtraFields)
+	err := unmarshalWithExtraFields(data, alias, tid.ExtraFields)
+	if err != nil {
+		return fmt.Errorf("unmarshal TypedID: %w", err)
+	}
+
+	return nil
+}
 
 func describeSchemaValidationError(result *gojsonschema.Result, what string) string {
 	errMsg := what + " is not valid:\n"
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -200,12 +200,6 @@ type Issuer struct {
 // Subject of the Verifiable Credential
 type Subject interface{}
 
-// CredentialStatus defines status of Verifiable Credential
-type CredentialStatus typedID
-
-// CredentialSchema defines a link to data schema which enforces a specific structure of Verifiable Credential.
-type CredentialSchema typedID
-
 // Credential Verifiable Credential definition
 type Credential struct {
 	Context        []string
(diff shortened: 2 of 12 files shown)


pkg/doc/verifiable/json.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Complement all fields with known fields to become a set of extra fields.
+	for k := range kf {
+		delete(ef, k)
+	}
+
+	return ef, nil
```
Baha-sk: it looks like this function returns unmarshalled extra fields in `data` that are not in `v`

retrieving these extra fields should be conveyed by the function name, `unmarshalJSON` is too broad

pkg/doc/verifiable/json.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	// Supplement value map with extra fields.
+	for k, v := range ef {
+		if _, exists := kf[k]; !exists {
+			kf[k] = v
+		}
+	}
+
+	return kf, nil
```
Baha-sk: same thing for the function name here.. it appends `ef` to `v`.. need a better name.. `valeToMap` doesn't convey this

pkg/doc/verifiable/common.go
```diff
@@ -55,16 +55,31 @@ func (ja JWSAlgorithm) jose() (jose.SignatureAlgorithm, error) {
...
+
+	return nil
 }
 
 // RefreshService provides a way to automatic refresh of expired Verifiable Credential
 type RefreshService typedID
 
 // TermsOfUse represents terms of use of Verifiable Credential by Issuer or Verifiable Presentation by Holder.
-type TermsOfUse interface{}
+type TermsOfUse typedID
```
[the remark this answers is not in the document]
kdimak: Agree. I made that exportable. Also, I started to use `TypedID` type directly for **TermsOfUse** and **RefreshService** fields.

pkg/doc/verifiable/json.go
kdimak: Yeah, I will try to find a better name.

kdimak:
@troyronda @Baha-sk guys I reworked the PR a lot. I think I had to mark this as "WIP" initially. Sorry about that!
The main change touches the extendable types like `rawCredential` or `TypedID` for which a number of standard fields are defined but we should keep extra fields in the map as well.
I finally found out a way how to implement `json.Marshaller` and `json.Unmarshaler` interfaces by those type and do not get into infinite recursion.
Please take your time and review more this PR if possible. Meanwhile, I will try to find better names for the functions from `json.go`.

pkg/doc/verifiable/json.go
kdimak: The name is changed to `mergeValueWithMap()`. What do you fill about it @Baha-sk ?

pkg/doc/verifiable/json.go
kdimak: Now it's `unmarshalValueWithMap()`.

pkg/doc/verifiable/json.go
Baha-sk: as discussed @kdimak, you will change this name again to make it clearer ..

kdimak: done!

pkg/doc/verifiable/common.go
```diff
@@ -55,16 +55,47 @@ func (ja JWSAlgorithm) jose() (jose.SignatureAlgorithm, error) {
...
+	type Alias TypedID
+	alias := (*Alias)(tid)
+
+	ef := make(TypedIDExtraFields)
+	err := unmarshalWithExtraFields(data, alias, ef)
+	if err != nil {
+		return fmt.Errorf("unmarshal TypedID: %w", err)
+	}
+
+	*tid = *(*TypedID)(alias)
```
troyronda: what's this?

pkg/doc/verifiable/credential_test.go
```diff
@@ -601,6 +601,7 @@ func TestCustomCredentialJsonSchemaValidator2018(t *testing.T) {
 
 	missingReqFieldSchema, mErr := json.Marshal(raw)
 	require.NoError(t, mErr)
+	fmt.Println(string(missingReqFieldSchema))
```
troyronda: why Println instead of t.Logf?

https://golang.org/pkg/testing/#B.Logf

pkg/doc/verifiable/common.go
Baha-sk: it seems UnmarshalJSON changes the value of the wrapped TypedID struct.. this is an update function, it does not do `only` unmarshal

pkg/doc/verifiable/credential.go
```diff
@@ -242,83 +242,36 @@ type rawCredential struct {
...
-
-	// Complement all fields with known fields to become a set of unknown fields.
-	for k := range kf {
-		delete(af, k)
+		return err
 	}
 
-	// Save unknown fields.
-	rc.CredentialExtraFields = af
+	*rc = *(*rawCredential)(alias)
```
Baha-sk: same here, Unmarshal changes the (pointer) value of the main struct *rawCredential ..

pkg/doc/verifiable/common.go
kdimak: Yes, that's right. We would like to handle unmarshalling of TypedID in a different to standard way as want to keep extra fields map. `json` package allows doing so by the implementation of `json.Unmarshaler`. We had to add aliases trick at the very beginning to avoid infinite recursion (as `unmarshalWithExtraFields()` calls `json.Unmarshall()` on typedID which in turn calls `TypedID::UnmarshalJSON` etc.

Will it be clearer if we put the following assignments:
```
tid.ID = alias.ID
tid.Type = alias.Type
```
instead of
```
*tid = *(*TypedID)(alias)
```

pkg/doc/verifiable/credential_test.go
kdimak: This is my negligence, I will fix it.

pkg/doc/verifiable/common.go
Baha-sk: in either case, this is an update operation on `tid`, the meaning of Unmarshal is to take a marshalled content and unmarshal it `without` modifying the struct.

kdimak: I think the opposite - we are doing unmarshalling and should update the struct. Otherwise, there will be no effect of the operation - we return just `error` and take no arguments (which we could update).

kdimak: hey, just understood how stupid was that line of code
removed it

pkg/doc/verifiable/credential.go
kdimak: fixed

troyronda:
TODO: UnmarshalJSON should not be a member method, rather it should return a fresh structure.

kdimak:
> TODO: UnmarshalJSON should not be a member method, rather it should return a fresh structure.

That's done because of the intention to implement `json.Unmarshaller` so that `encoding/json` lib delegates unmarshalling.

pkg/doc/verifiable/common.go
```diff
@@ -55,16 +55,44 @@ func (ja JWSAlgorithm) jose() (jose.SignatureAlgorithm, error) {
...
+// TypedIDExtraFields is a map of extra fields which extend typed ID field.
+type TypedIDExtraFields map[string]interface{}
+
+// TypedID defines a flexible structure with id and name fields and arbitrary extra fields
+// kept in TypedIDExtraFields.
+type TypedID struct {
 	ID   string `json:"id,omitempty"`
 	Type string `json:"type,omitempty"`
+
+	TypedIDExtraFields `json:"-"`
```
DRK3: Repetition of names here: TypedID.TypedIDExtraFields

Can this just be TypedID.ExtraFields?

pkg/doc/verifiable/credential.go
```diff
@@ -201,10 +201,10 @@ type Issuer struct {
 type Subject interface{}
 
 // CredentialStatus defines status of Verifiable Credential
-type CredentialStatus typedID
+type CredentialStatus TypedID
 
 // CredentialSchema defines a link to data schema which enforces a specific structure of Verifiable Credential.
-type CredentialSchema typedID
+type CredentialSchema TypedID
```
DRK3: Is this going to cause a shadowing, since the name of this variable matched the name of the type "TypedID". May want to change this name

pkg/doc/verifiable/json.go
```diff
@@ -0,0 +1,82 @@
...
+	kf, err := mergeExtraFields(v, ef)
+	if err != nil {
+		return nil, err
+	}
+
+	// Marshal extended known fields map.
+	return json.Marshal(kf)
+}
+
+// unmarshalWithExtraFields unmarshals JSON into value and puts all JSON fields which do not belong to value
```
DRK3: "value" refers to "v" here, and "extra fields" is "ef"? Perhaps it could be clearer to just use the argument names in the comment instead.

pkg/doc/verifiable/common.go
kdimak: Sure, this will sound better, no need for type embedding. I will change
```
TypedIDExtraFields `json:"-"`
```
to 
```
ExtraFields TypedIDExtraFields `json:"-"`
```

pkg/doc/verifiable/credential.go
kdimak: Good catch @DRK3 ! I cleaned this in other places but miss those assignments.

pkg/doc/verifiable/json.go
kdimak: for clarity, I'll put **value v** and **extra fields map ef**.

pkg/doc/verifiable/common.go
DRK3: Looks good! (I assume the extra "TypedIDExtraFields" in the second line is a typo)

pkg/doc/verifiable/json.go
DRK3: Even better. Thanks :)

pkg/doc/verifiable/credential.go
DRK3: Gotcha! Thanks

pkg/doc/verifiable/common.go
DRK3: Oh I see, you wanted to extra TypedIDExtraFields into a type

pkg/doc/verifiable/common.go
```diff
@@ -55,16 +55,44 @@ func (ja JWSAlgorithm) jose() (jose.SignatureAlgorithm, error) {
 // Proof defines embedded proof of Verifiable Credential
 type Proof interface{}
 
-type typedID struct {
+// TypedIDExtraFields is a map of extra fields which extend typed ID field.
+type TypedIDExtraFields map[string]interface{}
```
DRK3: Is there a reason that this is an extra type? instead of just making ExtraFields in TypedID a map[string]interface{} directly?

pkg/doc/verifiable/common.go
kdimak: Ah, right. Extra fields are defined for two types - `rawCredential` and `TypedID`. Why not have single `ExtraField` type declaration used by these two types? I will declare it as 
```
type ExtraFields map[string]interface{}
```

pkg/doc/verifiable/common.go
kdimak: It's just done to improve readability. No methods are defined for `ExtraFields`.

troyronda:
> > TODO: UnmarshalJSON should not be a member method, rather it should return a fresh structure.
> 
> That's done because of the intention to implement `json.Unmarshaller` so that `encoding/json` lib delegates unmarshalling.

We should have a TODO to hide this exposed method (e.g., by using a transiently created unexported structure during this operation).

pkg/doc/verifiable/common.go
DRK3: Gotcha, makes sense. You can disregard my other question I made asking about this since you answered it here

kdimak:
> > > TODO: UnmarshalJSON should not be a member method, rather it should return a fresh structure.
> > 
> > 
> > That's done because of the intention to implement `json.Unmarshaller` so that `encoding/json` lib delegates unmarshalling.
> 
> We should have a TODO to hide this exposed method (e.g., by using a transiently created unexported structure during this operation).

I did not get this completely but put the TODO :)

kdimak:
@DRK3 just updated the PR

PR #627, opened 2019-10-29
refactor: rename wallet to KMS

Baha-sk:
part of #494

This PR deals with pure rename only.
Follow up PRs will update the KMS with the new structure


pkg/client/didexchange/client.go
```diff
@@ -42,7 +42,7 @@ type Client struct {
 	service.Action
 	service.Message
 	didexchangeSvc           service.DIDComm
-	wallet                   wallet.Crypto
+	kms                      kms.Crypto
```
troyronda: Note: Crypto should be a sibling to kms.

Baha-sk: I will refactor this further...

Baha-sk: renamed to `KeyManager`

PR #628, opened 2019-10-30
feat: OpenAPI user interface for demo

sudeshrshetty:
Added below targets
- _`make generate-demo-specs`_ : generates demo specs for each demo agents
- _`make run-openapi-demo`_ : starts swagger ui docker containers using generated
spec as base.
- Part of #94

Once started, below links can be used to play with agents
**Alice Demo Agent :** http://localhost:8089/openapi/
**Bob Demo Agent :** http://localhost:9089/openapi/

Existing images/containers built for bdd tests are reused for demo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ A Go framework for Aries
 ##### Prerequisites (General)
 - Go 1.13
 
-##### Prerequisites (for running tests)
+##### Prerequisites (for running tests and demos)
 - Go 1.13
 - Docker
 - Docker-Compose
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -32,6 +32,15 @@ make checks
 # run unit tests
 make unit-test
 ```
+##### Running demo locally
```
troyronda: should update L18 as prerequisites are also need for demos.

scripts/generate-demo-specs.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
troyronda: general comment. we should prefix "demo" filenames and targets with the name of the demo (e.g., openapi).

sudeshrshetty: done

llorllale:
Nice one

PR #638, opened 2019-10-30
RestAPI - Get label value in CreateInvitation

### LINKED ISSUE #552: RestAPI - Get label value in CreateInvitation
Based on [TODO](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/restapi/operation/didexchange/didexchange.go#L94).

DRK3:
Closes #552 

Invitation label is now passed via a command line argument.

### DIFF
--- a/cmd/aries-agentd/startcmd/start.go
+++ b/cmd/aries-agentd/startcmd/start.go
@@ -62,6 +62,12 @@ const (
 
 	agentWebhookFlagUsage = "URL to send notifications to." +
 		" This flag can be repeated, allowing for multiple listeners."
+
+	agentDefaultLabelFlagName = "agent-default-label"
+
+	agentDefaultLabelFlagShorthand = "l"
+
+	agentDefaultLabelFlagUsage = "Default Label for this agent. Defaults to blank if not set."
 )
 
 // ErrMissingHost is the error when the user provides a blank host argument.
@@ -73,9 +79,9 @@ var ErrMissingInboundHost = errors.New("unable to start aries agentd, HTTP Inbou
 var logger = log.New("aries-framework/agentd")
 
 type agentParameters struct {
-	server                                                 server
-	host, inboundHostInternal, inboundHostExternal, dbPath string
-	webhookURLs                                            []string
+	server                                                               server
+	host, inboundHostInternal, inboundHostExternal, dbPath, defaultLabel string
+	webhookURLs                                                          []string
 }
 
 type server interface {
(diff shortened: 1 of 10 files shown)


pkg/client/didexchange/client.go
```diff
@@ -74,6 +75,7 @@ func New(ctx provider) (*Client, error) {
 		actionCh:        make(chan service.DIDCommAction, 10),
 		msgCh:           make(chan service.StateMsg, 10),
 		connectionStore: didexchange.NewConnectionRecorder(store),
+		invitationLabel: invitationLabel,
```
rolsonquadras: I feel we don't need to save it as client variable. We can pass as an argument in [CreateInvitation](https://github.com/hyperledger/aries-framework-go/pull/638/files#diff-427522b353a28b1ec350c29c79d387b0R108), as request can have different label for invitations.

sudeshrshetty: @rolsonquadras Since we are following arcpy pattern, they use label from agent (cmd argument --label here) and client function accepts only alias name.

rolsonquadras: @sudeshrshetty I think the question is whether we want to support dynamic labels per request or static one ?

cmd/aries-agentd/startcmd/start_test.go
```diff
@@ -104,7 +104,7 @@ func TestStartAriesDRequests(t *testing.T) {
 
 	go func() {
 		parameters := &agentParameters{&HTTPServer{}, testHostURL, testInboundHostURL,
-			"", path, []string{}}
+			"", path, "", []string{}}
```
rolsonquadras: Not directly related to this PR - But it's better to initialize using field names. If we don't need to initialize, we can just ignore the field. Without name, we need to pass values sequentially.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -119,8 +125,13 @@ func Cmd(server server) (*cobra.Command, error) {
...
 			}
 
+			invitationLabel, err := cmd.Flags().GetString(agentInvitationLabelFlagName)
+			if err != nil {
+				return fmt.Errorf("agent DB path flag not found: %s", err)
+			}
+
 			parameters := &agentParameters{server, host, inboundHost,
-				inboundHostExternal, dbPath, webhookURLs}
+				inboundHostExternal, dbPath, invitationLabel, webhookURLs}
```
rolsonquadras: Same as https://github.com/hyperledger/aries-framework-go/pull/638#discussion_r341137683.

pkg/restapi/restapi.go
```diff
@@ -27,6 +28,13 @@ func WithWebhookURLs(webhookURLs ...string) Opt {
 	}
 }
 
+// WithLabel is an option allowing for the label to be set.
+func WithLabel(label string) Opt {
```
troyronda: To me this is a default label.
I assume we will eventually extend the REST API to allow for overriding the default.

pkg/client/didexchange/client.go
troyronda: @rolsonquadras @sudeshrshetty  we should allow for default label and dynamic labels.

pkg/restapi/restapi.go
DRK3: Done!

sudeshrshetty: Btw, arcpy REST API doesn’t allow to override labels. They use ‘alias’ field for that purpose and alias is separate  field in request/response models & persisted records  .

PR #640, opened 2019-10-30
feat: Add signature to exchange response

### LINKED ISSUE #24: did-exchange: exchange response needs to be signed
We should be signing the [connection response](https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange#2-exchange-response). Specifically, the `connection` object needs to be replaced by `connection-sig`.

The reason for this is that the keys shared in the initial [invitation to exchange](https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange#0-invitation-to-exchange) will likely NOT be the keys present in the DID shared in the connection response:

* The keys in the initial invitation to exchange will likely be ephemeral since they are sent in plaintext (by design) and because DIDComm uses keys as identifiers. Identifiers should not be revealed to outside parties.
* The party has to prove that they own the DID that they are sharing and this is a mechanism to do so in a single message (as opposed to a later challenge-response protocol).

sandrask:
Add signature to outbound exchange response and verify signature in inbound exchange response

Closes #24

### DIFF
--- a/pkg/didcomm/protocol/didexchange/states.go
+++ b/pkg/didcomm/protocol/didexchange/states.go
@@ -15,12 +15,14 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/btcsuite/btcutil/base58"
 	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/ed25519signature2018"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -12,15 +12,18 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
```
fqutishat: extra space

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -710,6 +712,48 @@ func (ctx *context) handleInboundResponse(response *Response) (stateAction,
 	}, connRecord, nil
 }
 
+// verifySignature verifies connection signature and returns connection
+func verifySignature(connSignature *ConnectionSignature) (*Connection, error) {
```
Baha-sk: what's the reason behind adding this function (hidden) here when there's a `Signer` interface, [found here](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/kms/api.go#L53), in the KMS (formerly wallet)? Anything sensitive in this function that shouldn't be exposed?

sandrask: Verify signature doesn't require private key. We can discuss weather it belongs to wallet.

PR #644, opened 2019-10-31

soluchok:
* added possibility to decrement wait count when introducer receives a request message

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -268,6 +268,24 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
 	}
 }
 
+// Opt option for Continue function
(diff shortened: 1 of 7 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -268,6 +268,24 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
 	}
 }
 
+// Opt option for Continue function
+type Opt func(msg *metaData)
+
+// WithSkipProposal is used when the client received request message and he has a public invitation
+func WithSkipProposal() Opt {
```
rolsonquadras: The consumer of framework shouldn't have any references to service pkg ie `pkg/didcomm/protocol/...`. They need to refer only client pkg ie, `pkg/client/...`.

Issue #657, opened 2019-11-01
Introduce protocol client implementation


soluchok:
Need to implement a client for Introduce protocol.

PR #658, opened 2019-11-01
feat: Introduce protocol client implementation

soluchok:
* `didexchange/client.go` was refactored
* added Introduce protocol client, following methods were defined:
`SendProposal(dest1, dest2 *service.Destination)`
`SendProposalWithInvitation(inv *didexchange.Invitation, dest *service.Destination)`
`SendRequest(dest *service.Destination)`
`HandleRequest(dest1, dest2 *service.Destination)`
`HandleRequestWithInvitation(inv *didexchange.Invitation, dest *service.Destination)`

Issues https://github.com/hyperledger/aries-framework-go/issues/657


pkg/client/introduce/client.go
```diff
@@ -0,0 +1,138 @@
...
+	StorageProvider() storage.Provider
+}
+
+// Client enable access to introduce API
+type Client struct {
+	service.Action
+	service.Message
+	service                  service.DIDComm
+	kms                      kms.KeyManager
+	inboundTransportEndpoint string
```
rolsonquadras: Is this required in introduce protocol ?

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	// start listening for action/message events
+	go c.startServiceEventListener()
+
+	return c, nil
+}
+
+// startServiceEventListener listens to action and message events from introduce service.
+func (c *Client) startServiceEventListener() {
```
llorllale: this code is almost exactly the same as the one in the client/didexchange package. Can it be extracted into a common function somehow?

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,138 @@
...
+	didMsg, err := service.NewDIDCommMsg(payload)
+	if err != nil {
+		return err
+	}
+
+	return c.service.HandleOutbound(didMsg, &service.Destination{})
+}
+
+// CreateProposal creates proposal
+func (c *Client) CreateProposal() error {
```
llorllale: Should be renamed to "SendProposal" or something similar ("Propose"? :D).

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,138 @@
...
+	return c.handleOutbound(&introduce.Proposal{
+		Type: introduce.ProposalMsgType,
+		ID:   uuid.New().String(),
+	})
+}
+
+// CreateSkipProposal creates proposal
+// Skip Proposal should be created when client has public invitation
+func (c *Client) CreateSkipProposal() error {
```
llorllale: This should be removed. There is no "SkipProposal" message type in the RFC. As per our discussion this morning I understood the "skipping" (ie. decrementing wait_count) was going to be handled via other means.

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,138 @@
...
+	return c.handleOutbound(&introduce.Proposal{
+		Type: introduce.SkipProposalMsgType,
+		ID:   uuid.New().String(),
+	})
+}
+
+// CreateRequest creates request
+func (c *Client) CreateRequest() error {
```
llorllale: Same here - rename to "SendRequest" or something similar

pkg/client/introduce/client.go
llorllale: Can `introduce.SkipProposalMsgType` be deleted?

pkg/client/introduce/client.go
soluchok: Done! Code is redundant I have removed it

pkg/client/introduce/client.go
soluchok: Not sure, anyway I have deleted it!

pkg/client/introduce/client.go
soluchok: Done! Also, I added some comments to this function, it is about how we can get rid of `WithInvitation and WithCandidate`

pkg/client/introduce/client.go

pkg/client/introduce/client.go
soluchok: Done! I am going to delete ` introduce.SkipProposalMsgType` (but not in that PR)

pkg/restapi/operation/didexchange/didexchange_test.go
```diff
@@ -455,26 +454,6 @@ func TestServiceEvents(t *testing.T) {
...
-
-	ops := &Operation{client: client, actionCh: make(chan service.DIDCommAction)}
-
-	aCh := make(chan service.DIDCommAction)
-	err = client.RegisterActionEvent(aCh)
-	require.NoError(t, err)
-
-	err = ops.startClientEventListener()
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "didexchange action event registration failed: channel is already "+
```
rolsonquadras: Why is this test case removed ? I don't see any API changes in client pkg.

pkg/client/didexchange/client_test.go
```diff
@@ -500,93 +500,21 @@ func TestAcceptExchangeRequest(t *testing.T) {
...
-		ProtocolName:           didexchange.DIDExchange,
-		RegisterActionEventErr: errors.New("action event registration failed"),
-		RegisterMsgEventErr:    errors.New("msg event registration failed"),
-	}
-
-	// register action event on service throws error
-	_, err := New(&mockprovider.Provider{StorageProviderValue: mockstore.NewMockStoreProvider(),
-		ServiceValue: &didExSvc})
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "didexchange action event "+
```
rolsonquadras: Same as https://github.com/hyperledger/aries-framework-go/pull/658#discussion_r342085399. I understand we are reusing the events related code from service, but API surface for consumers hasn't changed. Just curious about removal of these test cases.

pkg/client/introduce/client.go
rolsonquadras: ok, better to avoid adding redundant code. That string is used by one of the DIDExchange protocol message. Hence it was added in DIDExchange client.

pkg/client/didexchange/client_test.go
soluchok: previously the function `New` did registration for  Action and Msg event. I have removed it,  no code no test

https://github.com/hyperledger/aries-framework-go/pull/658/files/3698253ec6234300a5133a5fec475986dac2484e#diff-427522b353a28b1ec350c29c79d387b0L87-L97

pkg/restapi/operation/didexchange/didexchange_test.go
soluchok: We are using `ServiceValue: &protocol.MockDIDExchangeSvc{}` it mocks our `RegisterActionEvent` and it is always without error.  This test does not make sense.

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,137 @@
...
+//		ThreadID: 8618e938-b68f-4712-b59c-f86f9a80d3c5,
+// 	})
+// 	Forwards an invitation from one introducee to another (Request was received)
+// 	SendProposal(&Options{
+// 		Destinations: []*service.Destination{
+// 			destination1,
+// 			destination1,
+// 		},
+//		ThreadID: 8618e938-b68f-4712-b59c-f86f9a80d3c5,
+// 	})
```
soluchok: @llorllale @troyronda @sandrask  This approach assumes that we have all data before the start state machine, but there is only one exception when we need an invitation from one of the introducees (it should be done through Continue(WithInvintation())), also after we received a request we should not execute Continue() (I am gonna make it panicked) instead of it we should SendProposal with threadID (threadID is equal to request.ID we received) we need this threadID to avoid doing Continue(WithInvintation()) I mean when we are sending request invitation can be saved to the DB

Summary:
* Continue() will be executed almost every time without args
* Continue(WithInvintation()) - will be executed only once by one of the introducees (if the introduction was started without request message)
* Continue() can not be executed after receiving the request

pkg/restapi/operation/didexchange/didexchange_test.go
soluchok: reverted it

troyronda:
This PR will need a rebase.

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,172 @@
...
+	Invitation *didexchange.Invitation
+	// Destinations contain one or two elements
+	// one element - for a public Invitation, otherwise two elements
+	Destinations []*service.Destination
+	// ThreadID must be an ID from incoming Request
+	ThreadID string
+}
+
+// sendProposal sends proposal
+// the function may receive an invitation or <nil> (Options.Invitation)
```
troyronda: nil?

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,172 @@
...
+// 		Invitation: invitation,
+// 		Destinations: []*service.Destination{
+// 			destination,
+// 		},
+// 	})
+// 	Forwards an invitation from one introducee to another
+// 	sendProposal(&Options{
+// 		Destinations: []*service.Destination{
+// 			destination1,
+// 			destination1,
```
troyronda: what's the idea behind having destination1 twice?

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,172 @@
...
+	didMsg, err := service.NewDIDCommMsg(payload)
+	if err != nil {
+		return err
+	}
+
+	return c.service.HandleOutbound(didMsg, &service.Destination{})
+}
+
+// Options keeps the information needed for sending a proposal
+type Options struct {
```
troyronda: Is this planned to be used outside this package? (i.e., why exported?)

Suggest a more specific name in case you have need for more Option types.

pkg/client/introduce/client.go
troyronda: I had to read this line a few times to grok that you were talking about the Invitation field :).

pkg/client/introduce/client.go
soluchok: typo `destination2`

pkg/client/introduce/client.go
soluchok: it is an internal thing, I am gonna make in unexported

pkg/client/introduce/client.go
```diff
@@ -0,0 +1,147 @@
...
+// 		},
+// 	})
+// 	Forwards an invitation from one introducee to another
+// 	sendProposal(&options{
+// 		Destinations: []*service.Destination{
+// 			destination1,
+// 			destination2,
+// 		},
+// 	})
+func (c *Client) sendProposal(_ options) error {
```
troyronda: why the ununsed argument? what happened to thread id?

soluchok: I added thread to Proposal structure bc I did not know how to handle it without thread, now I came up with some idea (it will be done in the next PR). The idea is that the client will keep an agent invitation, any time we want to accept the proposal this invitation will be passing to the Continue function, it allows us to proceed working with protocol without handling additional logic(I mean handle specific msg type)

PR #665, opened 2019-11-01
feat: bdd tests for get all connections

### LINKED ISSUE #429: Add ability to query all connections from did exchange service

sudeshrshetty:
- added one more verification step in controller tests to verify
completed connection against connections list.
- moved state based filter logic from REST API to client
- removed openapi annotations from client model
- reverted exported service model `ConnectionRecord` from client 
- closes #429

### DIFF
--- a/test/bdd/agent_controller_steps.go
+++ b/test/bdd/agent_controller_steps.go
@@ -202,7 +202,6 @@ func (a *AgentWithControllerSteps) validateConnection(agentID, stateValue string
 	logger.Debugf(" Getting connection by ID %s from %s", connectionID, destination)
 	// call controller
 	var response models.QueryConnectionResponse
-	// strings.Replace(connectionsByID, "{id}", connectionID)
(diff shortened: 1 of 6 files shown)


test/bdd/agent_controller_steps.go
```diff
@@ -215,6 +214,43 @@ func (a *AgentWithControllerSteps) validateConnection(agentID, stateValue string
 		return fmt.Errorf("Expected state[%s] for agent[%s], but got[%s]", stateValue, agentID, response.Result.State)
 	}
 
+	// Also make sure new connection is available in list of connections for given agent
```
troyronda: which feature step is this function included in?

sudeshrshetty: While validating connection state after did-exchange, we already have one verification using getConnectionByID. This is just one more layer of verification, so that if something goes wrong with query connections, this bddtest step should fail.

PR #675, opened 2019-11-04
feat: Allow injectable transient storage

### LINKED ISSUE #672: Allow injectable transient storage
part of #622

fqutishat:
- Add mem store as default transient storage
- Allow user to inject custom transient storage

closes #672

### DIFF
--- a/pkg/framework/aries/framework.go
+++ b/pkg/framework/aries/framework.go
@@ -24,6 +24,7 @@ import (
 type Aries struct {
 	transport                 api.TransportProviderFactory
 	storeProvider             storage.Provider
+	transientStoreProvider    storage.Provider
 	protocolSvcCreators       []api.ProtocolSvcCreator
 	services                  []dispatcher.Service
 	inboundTransport          transport.InboundTransport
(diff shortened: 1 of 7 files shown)


pkg/storage/mem/mem_store.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return nil
+}
+
+type memStore struct {
+	db map[string][]byte
+}
+
+// Put stores the key and the record
+func (s *memStore) Put(k string, v []byte) error {
```
troyronda: TODO - we will need to consider automatic eviction based on TTL.

pkg/storage/mem/mem_store.go
```diff
@@ -0,0 +1,108 @@
...
+	if k == "" || v == nil {
+		return errors.New("key and value are mandatory")
+	}
+	s.db[k] = v
```
sandrask: Do we need lock here for synchronization issues with map?

fqutishat: yes

fqutishat: i will fix it

fqutishat: done

pkg/storage/mem/mem_store.go
fqutishat: done

pkg/framework/aries/framework.go
```diff
@@ -24,6 +24,7 @@ import (
 type Aries struct {
 	transport                 api.TransportProviderFactory
 	storeProvider             storage.Provider
```
troyronda: TODO - will we need to fix the naming (right now we have "storeProvider" and "transientStoreProvider" - a bit inconsistent).

PR #689, opened 2019-11-04
test: Log event messages with connectionID in webhook server

rolsonquadras:

### DIFF
--- a/test/bdd/webhook/main.go
+++ b/test/bdd/webhook/main.go
@@ -7,15 +7,19 @@ SPDX-License-Identifier: Apache-2.0
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net/http"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/bdd/webhook/main.go
```diff
@@ -30,6 +34,17 @@ func connections(w http.ResponseWriter, r *http.Request) {
...
 		w.WriteHeader(http.StatusBadRequest)
 	}
+
+	connMsg := didexchange.ConnectionMsg{}
+	err = json.Unmarshal(msg, &connMsg)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+	}
+	// check for action event and log the connectionID
+	if connMsg.State == "null" {
```
troyronda: we send null on the web hook? That seems like it should be reviewed.

rolsonquadras: Created an issue - https://github.com/hyperledger/aries-framework-go/issues/690.

PR #701, opened 2019-11-05
refactor: Automatic detection for decoding of VC and VP

### LINKED ISSUE #514: Automatic detection for decoding of Verifiable Credential & Presentation
Current variants:
- JWS
- Unsecured JWT
- JSON (JSON-LD)

kdimak:
Automatic detection of decoding of Verifiable Credential and Verifiable Presentation.

A user is not obligated to explicitly define the nature of input bytes when creating new VC or VP.

closes #514

### DIFF
--- a/pkg/doc/verifiable/jwt_unsecured.go
+++ b/pkg/doc/verifiable/jwt_unsecured.go
@@ -67,3 +67,22 @@ func unmarshalUnsecuredJWT(rawJWT []byte) (joseHeaders map[string]string, bytesC
 
 	return headers, bytesPayload, nil
 }
+
+func isJWTUnsecured(data []byte) bool {
+	parts := strings.Split(string(data), ".")
+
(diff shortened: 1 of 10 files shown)


pkg/doc/verifiable/jwt_unsecured.go
```diff
@@ -67,3 +67,22 @@ func unmarshalUnsecuredJWT(rawJWT []byte) (joseHeaders map[string]string, bytesC
...
+		if err != nil {
+			return false
+		}
+		var j map[string]interface{}
+		err = json.Unmarshal(b, &j)
+		return err == nil
+	}
+
+	return len(parts) == 3 &&
+		isValidJSON(parts[0]) &&
```
troyronda: It seems a bit expensive to check every section for valid JSON (by Unmarshaling).
Why not just check the parts and the header section?

Issue #692, opened 2019-11-05
REST accept-invitation doesn't seem to be needed


troyronda:
By only calling receive-invitation on Agent Bob and accept-connection on Agent Alice, the connection still seems to enter completed state.

What is accept-invitation supposed to do?

rolsonquadras:
@troyronda The receive invitation API saves and approves the request in aries-framework-go. The Accept Invitation API is not used. We can either remove or send an error message when this API is invoked.

sudeshrshetty:
@troyronda  Receive invitation usually is 2 step process, (Receive + Accept) unless 'auto accept' is true. 
Receive invitation receives invitation object and later asks business logic's approval to accept it.

Currently we don't support that in REST API, like SDK we handle it like single step process.

troyronda:
FYI @andrewwhitehead @swcurran

andrewwhitehead:
In ACA-py we have the two step process to allow for invitations to be received by different means and create the connection record, while not necessarily invoking any outbound communication. The controller would still receive a webhook and have the chance to accept or remove the new connection (unless auto-accept is enabled, or accept=auto is passed into the receive-invitation method). This matches the pattern used in other protocols where the initial creation of the state record is separate from the triggering of a response (again, unless auto-* is turned on), although in this case the initial trigger isn't received via agent messaging.

I think there might be scenarios where the controller would want to create a new connection record so that related records can be tied to it, before actually accepting the invitation (such as if a proof request needs to be manually authorized first), but that might be a niche use case.

PR #708, opened 2019-11-05
feat: Add support for public dids in did-exchange protocol

### LINKED ISSUE #561: Add support for public dids in exchange request/response
Currently we support "peer" DIDs in exchange request and response. We should also support public dids.

sandrask:
Currently we support "peer" DIDs in exchange request and response. We should also support public dids.

Closes #561

### DIFF
--- a/pkg/didcomm/common/service/event.go
+++ b/pkg/didcomm/common/service/event.go
@@ -55,7 +55,7 @@ type DIDCommAction struct {
 	Message *DIDCommMsg
 
 	// Continue function to be called by the consumer for further processing the message.
-	Continue func()
+	Continue func(args interface{})
 
 	// Stop invocation notifies the service that the consumer action event processing has failed or the consumer wants
 	// to stop the processing.
@@ -102,8 +102,12 @@ type Event interface {
 //	go service.AutoExecuteActionEvent(actionCh)
 func AutoExecuteActionEvent(ch chan DIDCommAction) error {
 	for msg := range ch {
-		msg.Continue()
+		msg.Continue(&Empty{})
 	}
 
 	return nil
+}
+
+// Empty is used if there are no arguments to Continue
+type Empty struct {
 }
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -49,6 +49,7 @@ const (
 type message struct {
 	Msg           *service.DIDCommMsg
 	ThreadID      string
+	PublicDID     string
 	NextStateName string
 	ConnRecord    *ConnectionRecord
 	// err is used to determine whether callback was stopped
@@ -72,6 +73,7 @@ type stateMachineMsg struct {
 	header     *service.Header
 	payload    []byte
 	connRecord *ConnectionRecord
+	publicDID  string
 }
 
 // Service for DID exchange protocol
@@ -90,6 +92,12 @@ type context struct {
 	didResolver        didresolver.Resolver
 	connectionStore    *ConnectionRecorder
 	didStore           didstore.Storage
+}
+
+// opts are used to provide client properties to DID Exchange service
+type opts interface {
+	// PublicDID allows for setting public DID
+	PublicDID() string
 }
 
 // New return didexchange service
@@ -253,7 +261,10 @@ func (s *Service) handle(msg *message) error {
 		)
 
 		connectionRecord, followup, action, err = next.ExecuteInbound(&stateMachineMsg{
-			header: msg.Msg.Header, payload: msg.Msg.Payload, connRecord: msg.ConnRecord}, msg.ThreadID, s.ctx)
+			header: msg.Msg.Header, payload: msg.Msg.Payload,
+			connRecord: msg.ConnRecord, publicDID: msg.PublicDID},
+			msg.ThreadID, s.ctx)
+
 		if err != nil {
 			return fmt.Errorf("failed to execute state %s %w", next.Name(), err)
 		}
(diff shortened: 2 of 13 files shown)


troyronda:
This PR will need a rebase.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -92,6 +94,12 @@ type context struct {
 	didStore           didstore.Storage
 }
 
+// opts are used to provide client properties to DID Exchange service
+type opts interface {
+	// PublicDID allows for setting public DID
+	PublicDID() string
+}
```
soluchok: It does not look like options, I would rather call it `actionDependency`

pkg/didcomm/common/service/event.go
```diff
@@ -102,7 +102,7 @@ type Event interface {
 //	go service.AutoExecuteActionEvent(actionCh)
 func AutoExecuteActionEvent(ch chan DIDCommAction) error {
 	for msg := range ch {
-		msg.Continue()
+		msg.Continue(nil)
```
troyronda: Not a fan of nil arguments, particularly when passing nil into an interface.

troyronda: I think we should either:
- Create a specific Empty type (and use instead of nil).
- Use two methods.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -450,6 +439,48 @@ func (ctx *context) getDestination(invitation *Invitation) (*service.Destination
 	}, nil
 }
 
+func (ctx *context) getDidDocAndConnection(pubDID string) (*did.Doc, *Connection, error) {
```
troyronda: s/Did/DID

pkg/didcomm/protocol/didexchange/states_test.go
```diff
@@ -793,6 +803,43 @@ func TestGetPublicKey(t *testing.T) {
 	})
 }
 
+func TestGetDidDocAndConnection(t *testing.T) {
```
troyronda: s/Did/DID

test/bdd/features/didexchange_e2e_sdk.feature
```diff
@@ -18,7 +18,8 @@ Feature: Decentralized Identifier(DID) exchange between the agents using SDK
     And   "Bob" registers to receive notification for post state event "completed"
     And   "Alice" creates invitation
     And   "Bob" receives invitation from "Alice"
-    And   "Alice" approves did exchange request
+    And   "Alice" approves invitation request
```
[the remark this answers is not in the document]
sandrask: Will fix the order

pkg/restapi/operation/didexchange/didexchange.go
```diff
@@ -392,6 +392,11 @@ func (c *Operation) handleActionEvents(e service.DIDCommAction) error {
 	case didexchange.Event:
 		props := v
 
+		if e.Message.Header.Type == didexchange.InvitationMsgType {
```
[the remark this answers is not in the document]
sandrask: What do you want me to put in TODO: I am not sure what action item is here for rest API?

test/bdd/features/didexchange_e2e_sdk.feature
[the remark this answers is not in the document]
sandrask: Yes, we are currently just proceeding with requests hence they both call same function acceptRequest. There was no distinguishing between types of  requests before either.

pkg/didcomm/protocol/didexchange/service.go
sandrask: I would prefer to leave it as opts since it is optional.

pkg/didcomm/common/service/event.go
sandrask: Done

pkg/restapi/operation/didexchange/didexchange.go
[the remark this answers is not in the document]
rolsonquadras: With this change, there would be mismatch between SDK client and RestAPI client. In case of SDK client, the consumer would need to approve invitation and exchange request. But at the moment (due to this internally calling continue for invitation action item), RestAPI would need approval only for exchange request. We can add this feature as part of https://github.com/hyperledger/aries-framework-go/issues/692 to match the functionality added in this PR to RestAPI.

pkg/didcomm/protocol/didexchange/service.go
soluchok: ok

pkg/didcomm/common/service/event.go
```diff
@@ -102,8 +102,12 @@ type Event interface {
 //	go service.AutoExecuteActionEvent(actionCh)
 func AutoExecuteActionEvent(ch chan DIDCommAction) error {
 	for msg := range ch {
-		msg.Continue()
+		msg.Continue(&Empty{})
```
soluchok: We should not pass the `Empty` structure here, instead of this I suggest adding an additional parameter to the function signature, we can pass the `Empty` structure on the caller side
e.g `func AutoExecuteActionEvent(ch chan DIDCommAction, opts interface{}) error`

troyronda: This suggestion might need an example for this case.

soluchok: e.g Introduce client keeps agent’s invitation (it is a constant) and we will always pass it to the Continue function, on the service side we can decide whether to use it or no

sandrask: In case of using peer dids only we never need this functionality. Should we have different function for other protocols e.g. AutoExecuteActionEventWithOpts.

soluchok: ok, you can ignore this comment, I will come up with something when it will be needed

troyronda: Let's have a TODO to improve this Empty argument concept.

llorllale:
Note: needs followup - see #718

Issue #269, opened 2019-09-17
RFC 0028: Introduce Protocol


llorllale:
Add support for the [`introduce`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0028-introduce/README.md) protocol.

Should follow the same overall design as the existing did-exchange state machine: https://github.com/hyperledger/aries-framework-go/tree/master/pkg/didcomm/protocol/didexchange

troyronda:
Basic support for Introduce is now enabled. #882 lists additional TODO items.

PR #711, opened 2019-11-06
fix: fixed data race in introduce protocol

soluchok:
* data race
* goroutines leak (in tests)

### DIFF
--- a/pkg/didcomm/protocol/introduce/service_test.go
+++ b/pkg/didcomm/protocol/introduce/service_test.go
@@ -72,6 +72,9 @@ func TestService_handle(t *testing.T) {
 	t.Run("Happy path", func(t *testing.T) {
 		svc, err := New(&protocol.MockProvider{})
 		require.NoError(t, err)
+
+		defer stop(t, svc)
+
 		require.EqualError(t, svc.handle(&metaData{
 			Msg: &service.DIDCommMsg{},
 		}, nil), "state from name: invalid state name ")
@@ -80,6 +83,9 @@ func TestService_handle(t *testing.T) {
 	t.Run("Happy path", func(t *testing.T) {
 		svc, err := New(&protocol.MockProvider{})
 		require.NoError(t, err)
+
+		defer stop(t, svc)
+
 		ch := make(chan service.StateMsg, 2)
 		require.NoError(t, svc.RegisterMsgEvent(ch))
 		done := make(chan struct{})
@@ -132,8 +138,10 @@ func TestService_abandon(t *testing.T) {
 	store.Store.ErrPut = errors.New(errMsg)
 
 	svc, err := New(&protocol.MockProvider{StoreProvider: store})
-	require.NotNil(t, svc)
 	require.NoError(t, err)
+	require.NotNil(t, svc)
+
+	defer stop(t, svc)
 
 	const errStr = "save abandoning sate: " + errMsg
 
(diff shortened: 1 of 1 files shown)


rolsonquadras:
just a suggestion - better to have a specific commit title for rather than generic one (feat: Introduce protocol server implementation). While going through commit history, the title doesn't give specific info and looks [same](https://github.com/hyperledger/aries-framework-go/commits?author=soluchok).

.github/workflows/build.yml
```diff
@@ -59,6 +59,11 @@ jobs:
     timeout-minutes: 10
     runs-on: ubuntu-18.04
     steps:
+      - name: Setup Go 1.13
```
rolsonquadras: Looks like this commit contains same code as in PR - https://github.com/hyperledger/aries-framework-go/pull/658/files.

pkg/didcomm/protocol/introduce/persistence.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, err
+	}
+
+	return rec, nil
+}
+
+// SaveRecord saves data
+// this function usually is executed by the introduce client
+// NOTE: if the record already exists in the store it will rewrite it
+func (p *Persistence) SaveRecord(thID string, rec *Record) error {
```
rolsonquadras: Ideally client shouldn't save protocol related records and it should be done through service pkg (knowing of lower layer primitives like threadID). What's the use case here ?

pkg/didcomm/protocol/introduce/persistence.go
```diff
@@ -0,0 +1,77 @@
...
+	return &Persistence{store: store}
+}
+
+// GetRecord returns data by given threadID
+// this function should be used in introduce service
+// state machine behavior depends on this record's data
+func (p *Persistence) GetRecord(thID string) (*Record, error) {
+	var rec *Record
+	if err := p.getData(recordKey(thID), &rec); err != nil {
+		return nil, err
```
rolsonquadras: May be wrapping error message here would help with error message stack.

.github/workflows/build.yml
soluchok: I mentioned in the description `Depends on PR #658`

pkg/didcomm/protocol/introduce/persistence.go
soluchok: the client has a bunch of data needed for service, we have similar logic in `didexchange` client e.g `connectionStore: didexchange.NewConnectionRecorder(transientStore, store),`

rolsonquadras: Yes, but not save record - https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/persistence.go#L195

soluchok: `Record` it is just a name :) which keeps dependency needed for the state machine https://github.com/hyperledger/aries-framework-go/pull/711/files#diff-2be7cff09231a8b321a1f1aed45a4913R24-R28

troyronda: @fqutishat @llorllale @rolsonquadras @sandrask

troyronda: Is the issue that the method should be named to something more specific?
@soluchok @rolsonquadras

rolsonquadras: @troyronda My comment was about the 2nd item that you pointed out. ie - service should be persisting thread related data. ie. un-export the function.

I see the db operations from different layers as follows.
1. Service layer - handle db operations for did-comm message related data. ie, based on threadID or messageID etc.
2. Client layer - handle db operation for agent/application specific data. ie, based on connectionID(which is specific to the agent)

Note - service layer can handle pt#2 as well.

PR #714, opened 2019-11-06
chore: aries-agentd go mod

### LINKED ISSUE #648: separate go mod for aries-agentd pkg

sudeshrshetty:
- aries-agentd as new module so that it can be replaced to with
customized start
- closes #648


cmd/aries-agentd/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/hyperledger/aries-framework-go/cmd/aries-agentd
```
fqutishat: You need to remove the dependency from go.mod root

sudeshrshetty: done

PR #715, opened 2019-11-06
test: JWS from VC/VP using EdDSA alg

### LINKED ISSUE #712: JWS from Verifiable Credential using EdDSA alg
Add tests to verify marshaling of Verifiable Credential into JWS using EdDSA (Ed25519 signature algorithm). Also to verify the decoding of VC from EdDSA JWS.

kdimak:
Tests are added to verify the marshaling of VC/VP into JWS using EdDSA (Ed25519 signature algorithm).

closes #712

### DIFF
--- a/pkg/doc/verifiable/presentation.go
+++ b/pkg/doc/verifiable/presentation.go
@@ -173,8 +173,8 @@ func (vp *Presentation) MarshalJSON() ([]byte, error) {
 
 // JWTClaims converts Verifiable Presentation into JWT Presentation claims, which can be than serialized
 // e.g. into JWS.
-func (vp *Presentation) JWTClaims(audience []string, minimizeVc bool) *JWTPresClaims {
-	return newJWTPresClaims(vp, audience, minimizeVc)
+func (vp *Presentation) JWTClaims(audience []string, minimizeVP bool) *JWTPresClaims {
+	return newJWTPresClaims(vp, audience, minimizeVP)
 }
 
 // Credentials provides Verifiable Credentials enclosed into Presentation in raw byte array format.
--- a/pkg/doc/verifiable/presentation_jwt_proof_test.go
+++ b/pkg/doc/verifiable/presentation_jwt_proof_test.go
@@ -6,6 +6,8 @@ SPDX-License-Identifier: Apache-2.0
 package verifiable
 
 import (
+	"crypto/ed25519"
+	"crypto/rand"
 	"errors"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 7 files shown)


pkg/doc/verifiable/presentation.go
```diff
@@ -173,8 +173,8 @@ func (vp *Presentation) MarshalJSON() ([]byte, error) {
 
 // JWTClaims converts Verifiable Presentation into JWT Presentation claims, which can be than serialized
 // e.g. into JWS.
-func (vp *Presentation) JWTClaims(audience []string, minimizeVc bool) *JWTPresClaims {
-	return newJWTPresClaims(vp, audience, minimizeVc)
+func (vp *Presentation) JWTClaims(audience []string, minimizeVp bool) *JWTPresClaims {
```
DRK3: (small detail)
VP is an acronym right? So I think we usually would capitalize the "p" in "minimizeVp", kind of like how we capitalize "HTTP" in inbound.validateHTTPMethod.

pkg/doc/verifiable/presentation_jwt_proof_test.go
```diff
@@ -77,29 +79,53 @@ func TestNewPresentationFromJWS(t *testing.T) {
 	})
 }
 
+func TestNewPresentationFromJWS_EdDSA(t *testing.T) {
+	vpBytes := []byte(validPresentation)
+
+	pubKey, privKey, err := ed25519.GenerateKey(rand.Reader)
+	require.NoError(t, err)
+
+	// create presentation
```
DRK3: I think this comment pretty much just restates the line after it (NewPresentation). So I think you can delete this comment

pkg/doc/verifiable/presentation.go
DRK3: More specifically, I think acronyms generally need to have the same capitalization (so "vp *Presentation" is OK since "vp" is kind of like one word. Hope that makes sense)

kdimak: Good one! It should be `minimizeVP` of course. I'll fix that.

pkg/doc/verifiable/presentation_jwt_proof_test.go
kdimak: Yes, it's almost useless. I'll drop this comment.

pkg/doc/verifiable/presentation.go
DRK3: Thanks! Looks good now

Issue #475, opened 2019-10-11
Separate all crypto implementation details from Pack/Unpack


Moopli:
Extending #352 and #385, make pack/unpack invoke an injectable crypto-operator (could call it a crypter and do some renaming) for all encryption, decryption, AEAD, etc, using format-agnostic parameters, so the envelope format can be algorithm-independent.

Moopli:
Likely to be needed for the JWE DIDcomm envelope format to be accepted by the IETF, probably not necessary for the legacy envelope format. Uncertain how soon we'll need it.

Moopli:
What our Packers need as part of this:

For JWE:
- Remove specific uses of chacha or Ed/Curve25519 key size
- Crypter needs to return values of parameters like nonce size, key sizes (pub, priv, symmetric), encryption name string (eg, `chacha20poly1305_ietf` or `ECDH-SS+XC20PKW`) (or values which can be used to determine the name string)
- Create generic AEAD encryption functions, including one that can return the nonce used, and the tag and ciphertext separately; and one that can be given the nonce. (nonce is encoded in envelope in some places)
- Provide a function that can provide correct encryption keys given signing keys (conversion of ed25519 -> curve25519, or a lookup for RSA, for example)
- Use generic Crypter interface for crypto functions in creating/decoding SPK, JWK, etc

For Legacy:
- Remove dependency on specific key sizes
- same generic AEAD functions in Crypter
- same signing->encryption public key conversion in Crypter
- DH+AEAD encryption - crypto box and ability to implement and use logically-equivalent functions (like RSA + AES-GCM equivalents to crypto box). "Box Easy" interface needs a decrypt that can accept a nonce, and encrypt that returns the nonce generated (some nonces are part of the envelope).

A note on public key conversion - key types won't necessarily support this (or be able to do it safely), but the pack/unpack algorithms could be modified to operate using only encryption keys, with some updates to the draft specs.

troyronda:
We should also support all algorithms that are being specified in the W3C DID WG & DIF Interop Group. This is specifically:
- RSA
- ed25519
- sec256k1

Moopli:
[...]

PR #717, opened 2019-11-06
feat: Packager now holds multiple Packers for runtime format selection

### LINKED ISSUE #160: pkg/crypto is misleading
The packages under pkg/crypto do not implement crypto primitives.

Rather they are implementing JOSE support for encryption envelopes (e.g., [RFC0019](https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope))

Suggest that we should use either the name envelope or jose instead of "crypto".

### LINKED ISSUE #273: Create a higher-level Crypter to switch encryption based on key type/algorithm
The current [Crypter](https://github.com/hyperledger/aries-framework-go/blob/5bf052af3c6d6d6d6ff8ccd132c93717ab5d9c97/pkg/didcomm/crypto/crypter.go#L15) interface does not take into account the algorithm or the key type.

We need to create a higher-level Crypter that takes this info into account and use the correspopnding Crypter implementation.

Moopli:
Renames Crypters to Packers, and gives the Packager a list of Packers, for all formats the agent supports (elements are configurable and injectable by client), and a reference to the default Packer (also configurable and injectable) to use for encryption.

By default, the JWE and Legacy Packers are both enabled, available for decrypting messages. 

Fixes #160 
Fixes #273


Moopli:
Note: need to reconcile urlencoding vs rawurlencoding, or allow `Packager.getEncodingType()` to test both encodings, since JWE uses raw and legacy doesn't.

pkg/didcomm/envelope/api.go
```diff
@@ -6,12 +6,38 @@ SPDX-License-Identifier: Apache-2.0
...
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/packer"
+)
+
 // Package envelope manages the handling of DIDComm raw messages in JWE compliant envelopes.
 // The aim of this package is build these envelopes and parse them. They are mainly used as
 // wire-level wrappers of 'payloads' used in DID Exchange flows.
 
+/*
+// JWE returns the encoding type for the JWE format
```
troyronda: ?

pkg/didcomm/envelope/api.go
```diff
@@ -26,7 +52,7 @@ type Packager interface {
 	// []byte: The packed message
 	//
 	// error: error
-	PackMessage(envelope *Envelope) ([]byte, error)
+	PackMessage(envelope *packer.Envelope) ([]byte, error)
```
troyronda: There is a package called envelope but this signature has an Envelope defined in packer?

What is the intended purpose of the Envelope package?

pkg/framework/context/context.go
```diff
@@ -83,11 +84,23 @@ func (p *Provider) Packager() envelope.Packager {
...
+	return p.packers
 }
 
+// DefaultPacker returns the default packer service
+func (p *Provider) DefaultPacker() packer.Packer {
+	return p.defaultPacker
+}
+
+/*
+// Packer returns the defaultPacker service to be used by the packager
```
troyronda: commented out?

pkg/framework/aries/api/protocol.go
```diff
@@ -28,7 +27,7 @@ type Provider interface {
 	Service(id string) (interface{}, error)
 	StorageProvider() storage.Provider
 	KMS() kms.KeyManager
-	Crypter() crypto.Crypter
+	// Packer() packer.Packer
```
troyronda: commented out?

pkg/didcomm/packer/packer.go
```diff
@@ -4,35 +4,45 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
...
 */
 
-package crypto
+package packer
 
 import "github.com/hyperledger/aries-framework-go/pkg/kms"
 
-// Provider interface for Crypter ctx
-type Provider interface {
+// KMSProvider interface for Packer ctx
```
troyronda: are you sure this shouldn’t be a generic name?

pkg/didcomm/packer/jwe/authcrypt/authcrypt.go
```diff
@@ -28,6 +28,9 @@ const C20P = ContentEncryption("C20P") // Chacha20 encryption + Poly1305 authent
 // XC20P XChacha20Poly1305 algorithm
 const XC20P = ContentEncryption("XC20P") // XChacha20 encryption + Poly1305 authenticator cipher (192 bits nonce)
 
+// encodingType is the `typ` string identifier in a message that identifies the format as being JWE
+const encodingType string = "prs.hyperledger.aries-auth-message"
```
troyronda: suggest a const block.

pkg/didcomm/envelope/api.go
Moopli: I could swap the package names - I needed to pull Envelope out of envelope to avoid an import cycle.

pkg/framework/context/context.go
Moopli: Needs deletion, thanks for catching it.

pkg/framework/aries/api/protocol.go
Moopli: Needs deletion, thanks for catching it

pkg/didcomm/packer/packer.go
Moopli: Fair point

pkg/didcomm/envelope/api.go
```diff
@@ -6,12 +6,16 @@ SPDX-License-Identifier: Apache-2.0
...
 // Package envelope manages the handling of DIDComm raw messages in JWE compliant envelopes.
 // The aim of this package is build these envelopes and parse them. They are mainly used as
 // wire-level wrappers of 'payloads' used in DID Exchange flows.
 
 // PackagerCreator method to create new outbound dispatcher service
-type PackagerCreator func(prov Provider) (Packager, error)
+type PackagerCreator func(prov PackerProvider) (Packager, error)
 
 // Packager provide methods to pack and unpack msg
 type Packager interface {
```
Baha-sk: now that Crypter is renamed to Packer, this Packager interface becomes an extra layer that we can skim.

pkg/framework/context/context.go
```diff
@@ -83,9 +84,14 @@ func (p *Provider) Packager() envelope.Packager {
 	return p.packager
```
Baha-sk: now we have Packager(), Packers() and DefaultPackager().. back to my first comment above.. Packager interface becomes redundant

pkg/framework/context/context.go
```diff
@@ -225,10 +231,18 @@ func WithPackager(p envelope.Packager) ProviderOption {
...
+// WithDefaultPacker injects a packer into the context
+func WithDefaultPacker(p packer.Packer) ProviderOption {
+	return func(opts *Provider) error {
+		opts.defaultPacker = p
+		return nil
+	}
+}
+
+// WithPackers injects a variable number of packer services into the Aries framework
+func WithPackers(packers ...packer.Packer) ProviderOption {
```
Baha-sk: same thing with options, the user now has options: WithPackager, WithDefaultPacker and WithPackers.. it will confuse to the consumer of the framework as to what is required... another reason why Packager interface becomes redundant with this PR.

removing it can be done in a future PR.. add TODOs in the right locations if doing it separately.

pkg/didcomm/envelope/api.go
```diff
@@ -6,12 +6,16 @@ SPDX-License-Identifier: Apache-2.0
 
 package envelope
```
troyronda: Is it correct to observe that this package is only needed by the framework (supplying the injectable API)?

pkg/didcomm/envelope/package_test.go
```diff
@@ -79,7 +111,7 @@ func TestBaseKMSInPackager_UnpackMessage(t *testing.T) {
 		// It should fail since Recipient keys are not found in the KMS
 		_, err = packager.UnpackMessage(packMsg)
 		require.Error(t, err)
-		require.EqualError(t, err, "failed from decrypt: failed to decrypt message: key not found")
```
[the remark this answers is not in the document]
Moopli: Now that the message is partially decoded before the Packer unpacks it, the failure case is different, and occurs in `getEncodingType()`

pkg/didcomm/envelope/package_test.go
```diff
@@ -27,18 +30,49 @@ func TestBaseKMSInPackager_UnpackMessage(t *testing.T) {
 		}}))
 		require.NoError(t, err)
 
-		mockedProviders := &mprovider.Provider{
-			KMSValue: w,
-		}
-		crypter, err := authcrypt.New(mockedProviders, authcrypt.XC20P)
+		mockedProviders := &mockProvider{nil, w, nil, nil}
```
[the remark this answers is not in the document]
Moopli: Will fix

pkg/didcomm/envelope/package_test.go
```diff
@@ -27,18 +30,49 @@ func TestBaseKMSInPackager_UnpackMessage(t *testing.T) {
...
 		require.NoError(t, err)
 
-		mockedProviders.CrypterValue = crypter
-		packager, err := New(mockedProviders)
+		mockedProviders.defaultPacker = testPacker
+		packager, err := envelope.New(mockedProviders)
 		require.NoError(t, err)
 		_, err = packager.UnpackMessage(nil)
 		require.Error(t, err)
-		require.EqualError(t, err, "failed from decrypt: failed to decrypt message: unexpected end of JSON input")
```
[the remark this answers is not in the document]
Moopli: This is another case where the error now happens inside `getEncodingType()`

pkg/didcomm/envelope/api.go
Moopli: It now has the purpose of selecting which pack/unpack algorithm to use, allowing the agent to support communicating using multiple message formats for compatibility - so it's not a layer we can pull out.

pkg/framework/context/context.go
Moopli: I agree that the interface is redundant - with packing algorithm configuration available through `WithDefaultPacker()` and `WithPackers()`, we don't need to allow the `Packager` layer to be injectable.

pkg/framework/context/context.go
Moopli: I've removed framework `WithPackager()`

pkg/didcomm/envelope/api.go
Moopli: In talking with @Baha-sk we've decided to go the other way - Packager does not need to be injectable because pack/unpack format configuration is done through `WithDefaultPacker()` and `WithPackers()`. Where do you think Packager should move? Without Packager injection we only need the interface for mocking.

pkg/didcomm/envelope/api.go
Moopli: I renamed `envelope` (containing just the packager) to `packager` and `packer` (previously `crypto`) to `envelope`

pkg/didcomm/envelope/package_test.go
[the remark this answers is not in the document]
Moopli: done

pkg/didcomm/packer/jwe/authcrypt/decrypt_jwk.go
```diff
@@ -60,31 +60,31 @@ func (c *Crypter) decryptSPK(recipientPubKey *[chacha.KeySize]byte, spk string)
 		return nil, err
 	}
 
-	sharedKey, err := c.decryptJWKSharedKey(cipherKEK, headersJSON, recipientPubKey[:])
+	sharedKey, err := p.decryptJWKSharedKey(cipherKEK, headersJSON, recipientPubKey[:])
```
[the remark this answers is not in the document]
Moopli: I'll add comments to the relevant issue #475 instead.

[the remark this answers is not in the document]
Moopli: Or maybe just have a few TODOs instead of having a whole bunch.

pkg/didcomm/envelope/api.go
Baha-sk: Packager will hold the list of Packers including a default one, it will also decide which one to use based on the envelope type. Users don't need to inject it as it's a utility interface that's required by the framework.

Users can inject Packers and a DefaultPacker to support their own implementation.

pkg/framework/context/context.go
Moopli: I've removed the injection.

pkg/didcomm/envelope/jwe/authcrypt/encrypt_jwk_test.go
```diff
@@ -66,7 +66,7 @@ func TestNilEncryptSenderJwk(t *testing.T) {
 	require.Empty(t, l)
 	require.Empty(t, m)
 
-	pld, err := crypter.Encrypt([]byte(""), someKey[:], [][]byte{someKey[:]})
+	pld, err := crypter.Pack([]byte(""), someKey[:], [][]byte{someKey[:]})
```
troyronda: why is the variable name still `crypter` ?

pkg/didcomm/packager/api.go
```diff
@@ -0,0 +1,47 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package packager
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/envelope"
+)
+
+// Package packager manages the handling of DIDComm raw messages in JWE compliant envelopes.
```
troyronda: I think the top level Packager interface is agnostic of JWE?

pkg/didcomm/packager/api.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package packager
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/envelope"
```
troyronda: Is there a reason for a separate envelope package? Does it now only have a struct?

pkg/didcomm/packager/packager.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Provider contains dependencies for the base packager and is typically created by using aries.Context()
+type Provider interface {
+	envelope.Provider
+	Packers() []envelope.Packer
+	DefaultPacker() envelope.Packer
+}
+
+// BasePackager is the basic implementation of Packager
+type BasePackager struct {
```
troyronda: Is there a need for the prefix `Base` ?

pkg/framework/aries/framework.go
```diff
@@ -208,7 +211,8 @@ func (a *Aries) Context() (*context.Provider, error) {
 		context.WithInboundTransportEndpoint(a.inboundTransport.Endpoint()),
 		context.WithStorageProvider(a.storeProvider),
 		context.WithTransientStorageProvider(a.transientStoreProvider),
-		context.WithCrypter(a.crypter),
+		context.WithDefaultPacker(a.packer),
```
troyronda: Should I assume that `a.packer` must also be in the `a.packers` list in L215?

pkg/framework/aries/framework_test.go
```diff
@@ -92,11 +90,20 @@ func TestFramework(t *testing.T) {
 			WithKMS(func(ctx api.Provider) (api.CloseableKMS, error) {
 				return &mockkms.CloseableKMS{SignMessageValue: []byte("mockValue")}, nil
 			}),
-			WithCrypter(func(ctx crypto.Provider) (crypto.Crypter, error) {
-				return &didcomm.MockAuthCrypt{EncryptValue: nil}, nil
+			WithDefaultPacker(func(ctx envelope.Provider) (envelope.Packer, error) {
```
troyronda: Is this really "Default" or is it the outbound packer?

I mean the framework puts defaults in for the Packers list too :).

pkg/didcomm/packager/api.go
Moopli: Envelope struct, Packer interface, and Packer Creator.

pkg/didcomm/packager/packager.go
Moopli: No, since it's not injectable we don't expect others.

pkg/framework/aries/framework.go
Moopli: Packager handles that - if the default packer is not in the list, it's added.

pkg/framework/aries/framework_test.go
Moopli: Good point, I'll rename.

pkg/didcomm/envelope/jwe/authcrypt/encrypt_jwk_test.go
Moopli: I'll go over all other uses of 'crypter' too.

pkg/didcomm/packager/api.go
Moopli: what do you think I should say instead for "DID Exchange flows"? Since this format is used for all protocol communications.

pkg/didcomm/packager/packager.go
Moopli: Then again there is the distinction between this and the interface (the interface is `Packager`), since we also have a mock packager. The interface is what everything consumes.

I could make the struct definition private.

pkg/didcomm/packager/api.go
Moopli: It's separate mainly to avoid an import cycle - which iirc was a cycle only because of tests, so I'll see if I can fix that.

pkg/framework/aries/framework_test.go
troyronda: If the outbound packer isn't specified, which packer from `s.packers` is used?

pkg/didcomm/packager/packager.go
troyronda: The interface and struct shouldn't be in the same package.

pkg/framework/aries/framework.go
Moopli: Specifically - each Packer defines an encoding type string (returned by `Packer.EncodingType()`) which is used to key into the packer map. So if `a.packer`'s type is the type of one of the elements of `a.packers`, then it's assumed to be in the list already.

pkg/framework/aries/framework_test.go
Moopli: Though it is also a default in the sense that it's included in the list of inbound packers, since it makes little sense to send messages using a format you're unwilling to read.

pkg/didcomm/packager/api.go
troyronda: I don't see "DID Exchange" - I saw "DIDComm".

pkg/didcomm/packager/api.go
Moopli: I htink I should keep the packages separate, but I can rename `envelope` to `packer` - I'd named it `packer` earlier. I'll work out how to put the Envelope struct inside `pkg/didcomm/packager` without an import cycle, which I believe only came about due to tests.

pkg/didcomm/packager/packager.go
Moopli: I can return the interface instead of a pointer to the struct

pkg/framework/aries/framework_test.go
Moopli: If the outbound packer isn't specified, then the framework falls back to initializing the default, which isn't inside `s.packers`. `packager.New` requires a default packager to be provided. I'll add an error check in the constructor if the value is nil.

pkg/didcomm/packager/packager.go
Moopli: Where should I move the interface? It's consumed all over the codebase, particularly in tests which usually use the mock packager.

pkg/didcomm/packager/api.go
Moopli: A few lines down.

pkg/didcomm/packager/api.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// Package packager manages the handling of DIDComm raw messages in JSON envelopes.
+// The aim of this package is build these envelopes and parse them. They are mainly used as
+// wire-level wrappers of 'payloads' used in DID Exchange flows.
```
Moopli: @troyronda here it says 'DID Exchange flows'

pkg/didcomm/packer/jwe/authcrypt/authcrypt.go
```diff
@@ -112,9 +115,14 @@ func New(ctx crypto.Provider, alg ContentEncryption) (*Crypter, error) {
 		return nil, errUnsupportedAlg
 	}
 
-	return &Crypter{
+	return &Packer{
 		alg:       alg,
 		nonceSize: nonceSize,
```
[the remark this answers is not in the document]
Moopli: I've added some TODOs

pkg/didcomm/envelope/jwe/authcrypt/encrypt_jwk_test.go
Moopli: this should be done now.

pkg/didcomm/packager/api.go
Moopli: I've resolved this by pulling `Envelope` out into `pkg/didcomm/common/transport`

pkg/didcomm/packager/packager.go
Moopli: I've resolved this by pulling the `Packager` interface out into `pkg/didcomm/common/transport`

pkg/framework/aries/framework_test.go
Moopli: instead of "Default" or "Outbound" it is now just "Packer", since it's used for both in/outbound.
The packer list is now InboundPackers, since they're used inbound only right now. #725 is for making those packers usable outbound as well, by remembering the format to use for a particular connection.

pkg/didcomm/packager/api.go
Moopli: I reworded this.

pkg/didcomm/packager/packager.go
```diff
@@ -0,0 +1,154 @@
...
+	Packer() packer.Packer
+}
+
+// Creator method to create new packager service
+type Creator func(prov Provider) (transport.Packager, error)
+
+// Packager is the basic implementation of Packager
+type Packager struct {
+	packer         packer.Packer
+	inboundPackers map[string]packer.Packer
```
Baha-sk: why `inbound`Packers? it's just a list of Packers

a Packer does pack/unpack, regardless of the direction of the message. 

any reason for the direction word in the variable `inoundPackers`?

pkg/didcomm/packager/package_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+func TestBaseKMSInPackager_UnpackMessage(t *testing.T) {
+	t.Run("test failed to unmarshal encMessage", func(t *testing.T) {
+		w, err := kms.New(newMockKMSProvider(&mockstorage.MockStoreProvider{Store: &mockstorage.MockStore{
+			Store: make(map[string][]byte),
+		}}))
+		require.NoError(t, err)
+
+		mockedProviders := &mockProvider{
+			storage:        nil,
```
Baha-sk: no need to set nil variables

pkg/didcomm/packager/package_test.go
```diff
@@ -0,0 +1,294 @@
...
+func newMockKMSProvider(storagePvdr *mockstorage.MockStoreProvider) *mockProvider {
+	return &mockProvider{storagePvdr, nil, nil, nil}
+}
+
+// mockProvider mocks provider for KMS
+type mockProvider struct {
+	storage        *mockstorage.MockStoreProvider
+	kms            kms.KeyManager
+	packers        []packer.Packer
+	outboundPacker packer.Packer
```
Baha-sk: not sure `outbound`Packer has a meaning, it's just a `Packer` that should be able to do Pack/Unpack operations. 

why add the direction in the variable name?

pkg/didcomm/packager/packager.go
troyronda: The packer seems to be the only argument doing both pack and unpack.

The inbound packers seems to be unpack only. Perhaps a better name would be unpackers :).

pkg/didcomm/packager/package_test.go
troyronda: See reply below.

pkg/didcomm/packager/package_test.go
troyronda: Agree.

pkg/didcomm/packager/packager.go
Baha-sk: the map of packers should not be meant for `inbound` only, during a connection, if Agent A talks with B in JWE compliant and Agent C in legacy packer, then Agent A should use the right `outbound` Packer for each connection (not just for replying to incoming messages). The the map of Packers is not meant to be for `inbound` only (Unpack()) calls.

also, `packer` field above was agreed to be the DefaultPacker which is mean to be used for new connections. For pre-existing/saved connections, the agent should store the corresponding Packer to be used for all subsequent calls (outbound and inbound).

so the word `inbound` in `inboundPacker` does not correctly represent the map of Packers as they should not be used for inbound calls only.

troyronda: Default doesn’t seem like a good prefix  -if we use that then there is a default default packer set by the framework package if the option is not provided.

pkg/didcomm/packager/package_test.go
troyronda: I think we should combine these options into one to reduce confusion (and API surface).
 https://github.com/hyperledger/aries-framework-go/issues/764.

Issue #727, opened 2019-11-07
Investigate refactor DID exchange client issue


troyronda:
This refactor will need to be reapplied: https://github.com/hyperledger/aries-framework-go/pull/726 
on top of: https://github.com/hyperledger/aries-framework-go/pull/708 

(and fixed to be compatible).

PR #730, opened 2019-11-07
refactor: DID exchange client

soluchok:
* added buffered channel (test)

Closes https://github.com/hyperledger/aries-framework-go/issues/727


test/bdd/agent_sdk_steps.go
```diff
@@ -116,7 +116,7 @@ func (a *AgentSDKSteps) createDIDExchangeClient(agentID string) error {
 		return fmt.Errorf("failed to create new didexchange client: %w", err)
 	}
 
-	actionCh := make(chan service.DIDCommAction)
+	actionCh := make(chan service.DIDCommAction, 1)
```
sandrask: Why 1? Are we going to ran into same issue if we add more action events?

soluchok: the listener starts after write, according to that we got a deadlock. We are not going to run into the same issue. In our code we already have this logic for `statusCh := make(chan service.StateMsg, 1)`  but we missed it for `actionCh := make(chan service.DIDCommAction)`

sandrask: How did bdd tests for you previous commit work? You didn't have 1 there.

soluchok: bc you added this code which did that deadlock https://github.com/hyperledger/aries-framework-go/commit/aa0bf5895cbb4a5d137da55d8a8cf49d676b663f#diff-36702ab66cf0f951b3103d5c6e033346R200

soluchok: Sorry, probably goroutine with listener was executed first

rolsonquadras: On analyzing these changes with https://github.com/hyperledger/aries-framework-go/pull/708, I discovered the following.

Before the PR https://github.com/hyperledger/aries-framework-go/pull/708:
1. BDD test creates client for Alice and Bob
2. Alice creates request and Bob receives it. 
3. During step2, Alice starts listening to exchange approval event in-parallel.
3. Alice gets the event and flow completes successfully.

After PR https://github.com/hyperledger/aries-framework-go/pull/708:  
Change : Bob needs to approve the invitation
1. BDD test creates client for Alice and Bob
2. Alice creates request and Bob receives it and it triggers an event. 
3. During step2, Alice and Bob starts listening to exchange approval event in-parallel.
4. But the agent triggers event before Bob starts listening.
5. Since it's an Unbuffered channel, the service becomes blocked/deadlocked since there are no listeners(go routines).

After adding size= 1 (buffered) to the channel, the service pushes the event to the buffered channel. It holds it till 
listener/go routine becomes available and processes the message.

This is due to unbuffered channel behaves as full as there is no goroutine to receive it. Here is a reference to one of the [stackoverflow discussion](https://stackoverflow.com/questions/18660533/why-does-the-use-of-an-unbuffered-channel-in-the-same-goroutine-result-in-a-dead).

troyronda:
@soluchok needs rebase.

PR #734, opened 2019-11-07
Support environment variables in addition to command line arguments

### LINKED ISSUE #473: Support environment variables in addition to command line arguments
It would be useful for Ops people if they can set the command line arguments via environment variables as an alternative to command line arguments. This task is to reintroduce support for environment variables. The command line arguments introduced in #250 and #200 should be supported.

DRK3:
Closes #473

### DIFF
--- a/cmd/aries-agentd/startcmd/start.go
+++ b/cmd/aries-agentd/startcmd/start.go
@@ -9,6 +9,8 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
+	"strings"
 
 	"github.com/gorilla/mux"
 	"github.com/rs/cors"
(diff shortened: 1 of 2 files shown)


cmd/aries-agentd/startcmd/start.go
```diff
@@ -9,6 +9,8 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
```
fqutishat: @troyronda  @sudeshrshetty  @rolsonquadras I think we should use config file to pass the configuration instead of passing directly into CLI

DRK3: (already discussed in daily standup, won't be doing)

cmd/aries-agentd/startcmd/start.go
```diff
@@ -21,61 +23,69 @@ import (
...
 
 	agentDefaultLabelFlagShorthand = "l"
 
-	agentDefaultLabelFlagUsage = "Default Label for this agent. Defaults to blank if not set."
+	agentDefaultLabelFlagUsage = "Default Label for this agent. Defaults to blank if not set." +
+		" Alternatively, this can be set with the following environment variable: " + agentDefaultLabelEnvKey
+
+	agentDefaultLabelEnvKey = "ARIESD_DEFAULT_LABEL"
+
+	unableToStartAgentErrMsg = "unable to start agent"
```
DRK3: I realised a bunch of these were exported when they didn't need to be, so I fixed that. Removed those comments since they were really only there because the linter forced me to - they didn't really add anything.

cmd/aries-agentd/startcmd/start.go
```diff
@@ -99,104 +109,182 @@ func (s *HTTPServer) ListenAndServe(host string, router http.Handler) error {
...
+	startCmd.Flags().StringP(agentDBPathFlagName, agentDBPathFlagShorthand, "", agentDBPathFlagUsage)
+	startCmd.Flags().StringSliceVarP(webhookURLs, agentWebhookFlagName, agentWebhookFlagShorthand, []string{},
+		agentWebhookFlagUsage)
+	startCmd.Flags().StringP(agentInboundHostExternalFlagName, agentInboundHostExternalFlagShorthand,
+		"", agentInboundHostExternalFlagUsage)
+	startCmd.Flags().StringP(agentDefaultLabelFlagName, agentDefaultLabelFlagShorthand, "",
+		agentDefaultLabelFlagUsage)
+}
+
+func getHost(cmd *cobra.Command) (string, error) {
```
DRK3: A lot of repeated code here - can refactor a lot into a common method.

PR #735, opened 2019-11-08
feat: Add thread decorator in the exchange request and sign the exchange response with Invitation RecipientKeys

### LINKED ISSUE #676: Add thread decorator in the exchange request
"~thread": { "pthid": "<id of invitation>" },
https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange

Implement the following spec:

https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange#correlating-requests-to-invitations

### LINKED ISSUE #728: Sign response using recepientKey
As per spec we should sign exchange response using recipientKey from invitation if the flow was started from invitation. This can be done upon completing #676 that allows for correlation between exchange-request and invitation using pthid.

We also need to handle the case of implicit invitation #718

talwinder50:
closes #676 
closes #728 
 
### DIFF
--- a/pkg/client/didexchange/client_test.go
+++ b/pkg/client/didexchange/client_test.go
@@ -6,15 +6,19 @@ SPDX-License-Identifier: Apache-2.0
 package didexchange
 
 import (
+	"crypto/ed25519"
+	"crypto/rand"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcutil/base58"
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	mockprotocol "github.com/hyperledger/aries-framework-go/pkg/internal/mock/didcomm/protocol"
 	mockkms "github.com/hyperledger/aries-framework-go/pkg/internal/mock/kms"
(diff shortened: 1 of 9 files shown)


pkg/didcomm/common/service/service_test.go
```diff
@@ -9,6 +9,8 @@ package service
 import (
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+
```
rolsonquadras: extra space.

pkg/didcomm/common/service/service.go
```diff
@@ -83,6 +86,23 @@ func NewDIDCommMsg(payload []byte) (*DIDCommMsg, error) {
 	return msg, nil
 }
 
+// GetPthid fetches the parent threadID from the public did document for implicit invitations
+func GetPthid(didServices []did.Service) ([]did.Service, error) {
```
rolsonquadras: function name is GetPthid (get parent thread ID), but returns did.Service. Looks like name is doesn't match the functionality.

pkg/client/didexchange/client_test.go
```diff
@@ -474,19 +475,24 @@ func TestServiceEvents(t *testing.T) {
 	newDidDoc, err := (&mockcreator.MockDIDCreator{}).Create("test")
 	require.NoError(t, err)
 
+	didService, err := service.GetPthid(newDidDoc.Service)
```
sandrask: Is there a need to expose GetPthid at service level? This looks more like private utility function.

pkg/client/didexchange/client_test.go
```diff
@@ -474,19 +475,24 @@ func TestServiceEvents(t *testing.T) {
...
+	require.NoError(t, err)
+
 	request, err := json.Marshal(
 		&didexchange.Request{
 			Type:  didexchange.RequestMsgType,
 			ID:    id,
 			Label: "test",
+			Thread: &decorator.Thread{
```
sandrask: Currently all parent thread test cases refer to public did. Test cases should be mixture of invitations and public dids. You could use invitation here since this is what our first release is about.

Issue #724, opened 2019-11-07
Provide necessary data for service state machine


soluchok:
The client should provide necessary information to the service by passing the proper interface to the Continue function. 

Requirements:

* add `ProtocolDependency(thID) Dependency` function to the client
* Dependency interface looks like following
```
type Dependency interface {
	Invitation() didexchange.Invitation
	Destinations() []service.Destination
}
```

Invitation() function returns an invitation related to the introduction, the invitation can be `public` or `private`
public - means that the agent has another agent's invitation
private - means own agent invitation (unique)
`public` and `private` means nothing to the service (it is just an explanation)
The private invitation should be registered once when the `Client` is created and it cannot be changed.
The public invitation can be received from `Client API` e.g `SendProposalWithInvitation(inv *didexchange.Invitation, dest *service.Destination)` previously saved it to the store and retrieved it by  thID. 
The public invitation has more priority.If we have a public invitation for the given thID it must be returned, otherwise, we need to return the private invitation.
What if there is no public or private invitation?
It is ok, in that case, we should return a nil value.

Destinations() - returns the destination value (to whom we need to send it)
e.g sending a public proposal we will have only one destination
e.g sending proposal (no public invitation) we will have two destinations

Part of https://github.com/hyperledger/aries-framework-go/issues/269

llorllale:
@soluchok 

I want to clarify something:

> public - means that the agent has another agent's invitation
> private - means own agent invitation (unique)

The invitations that the *Introducer* handles are *never their own*. They are always of other agents (introducees). What the introducer does is forward these invitations around.

soluchok:
@llorllale correct! What I want to do is to prepare data before the state machine will be started! 
The idea is to provide one Dependency interface which should provide all necessary data to the state machine (better to collect it before the protocol starts).  For instance, state machine calls `Invitation(thID)` function  (see dependency interface) and Client must provide the right invitation
According to a state machine, `Invitation(thID)` function will return a different result. 
e.g for introducer we will get an invitation from DB by given thID, for introducee DB will be empty, so we will return a `private` invitation (this logic on the client-side)
It allows us to avoid handling the action events by their type (on the client-side or even layer up) and do something like that b.Continue(WithSomeParam(),WithSomethingElse())

llorllale:
@soluchok 

Nevermind the naming which doesn't seem right to me: how does that usage allow the Introducer to specify the invitation when they receive a `request`?

soluchok:
[...]

PR #737, opened 2019-11-08
feat: Added dependency interface to the Client

soluchok:
* added dependency interface to the Client 

Closes https://github.com/hyperledger/aries-framework-go/issues/724

### DIFF
--- a/pkg/client/introduce/client.go
+++ b/pkg/client/introduce/client.go
@@ -9,14 +9,20 @@ package introduce
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 
 	"github.com/google/uuid"
 
-	"github.com/hyperledger/aries-framework-go/pkg/client/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
+
+const invitationEnvelopePrefix = "invitation_envelope_"
+
+var logger = log.New("aries-framework/introduce/client")
 
 // Provider contains dependencies for the introduce protocol and is typically created by using aries.Context()
 type Provider interface {
@@ -27,12 +33,14 @@ type Provider interface {
 // Client enable access to introduce API
 type Client struct {
 	service.Event
-	service service.DIDComm
-	store   storage.Store
+	service    service.DIDComm
+	store      storage.Store
+	defaultInv *didexchange.Invitation
+	newUUID    func() string
 }
 
 // New return new instance of introduce client
-func New(ctx Provider) (*Client, error) {
+func New(ctx Provider, inv *didexchange.Invitation) (*Client, error) {
 	svc, err := ctx.Service(introduce.Introduce)
 	if err != nil {
 		return nil, err
@@ -49,80 +57,141 @@ func New(ctx Provider) (*Client, error) {
 	}
 
 	return &Client{
-		Event:   introduceSvc,
-		service: introduceSvc,
-		store:   store,
+		Event:      introduceSvc,
+		service:    introduceSvc,
+		store:      store,
+		defaultInv: inv,
+		newUUID:    func() string { return uuid.New().String() },
 	}, nil
 }
 
-// SendProposal sends proposal to introducees
+// InvitationEnvelope is a helper function that returns the dependency needed for the
+// service to proceed with the protocol. Dependency should be passed to the service through `Continue` function.
+// The function should never return an error, instead of error we provide the callable interface
+// and the state machine will act according to the provided data.
+// Dependency is populated after executing the following functions:
+//  - SendProposal
+//  - SendProposalWithInvitation
+//  - HandleRequest
+//  - HandleRequestWithInvitation
+// usage: e.Continue(c.InvitationEnvelope(threadID))
+func (c *Client) InvitationEnvelope(thID string) *InvitationEnvelope {
+	opts, err := c.getInvitationEnvelope(thID)
+	if err != nil {
+		opts = &InvitationEnvelope{}
+
+		logger.Errorf("invitation envelope: %v", err)
+	}
+
+	if opts.Invitation() == nil {
+		opts.Inv = c.defaultInv
+	}
+
+	return opts
+}
+
+// SendProposal sends proposal to introducees (the client does not have a public Invitation)
 func (c *Client) SendProposal(dest1, dest2 *service.Destination) error {
-	return c.sendProposal(options{
-		Destinations: []*service.Destination{dest1, dest2},
+	return c.sendProposal(InvitationEnvelope{
+		Dests: []*service.Destination{dest1, dest2},
 	})
 }
 
-// SendProposalWithInvitation sends proposal to introducee
-// executes when the client has a public Invitation
+// SendProposalWithInvitation sends proposal to introducee (the client has a public Invitation)
 func (c *Client) SendProposalWithInvitation(inv *didexchange.Invitation, dest *service.Destination) error {
-	return c.sendProposal(options{
-		Invitation:   inv,
-		Destinations: []*service.Destination{dest},
+	return c.sendProposal(InvitationEnvelope{
+		Inv:   inv,
+		Dests: []*service.Destination{dest},
 	})
 }
 
-// SendRequest sends request
+// SendRequest sends a request
 // sending a request means that introducee is willing to share its invitation
 func (c *Client) SendRequest(dest *service.Destination) error {
 	return c.handleOutbound(&introduce.Request{
 		Type: introduce.RequestMsgType,
-		ID:   uuid.New().String(),
+		ID:   c.newUUID(),
+	}, InvitationEnvelope{
+		Dests: []*service.Destination{dest},
 	})
 }
 
-// HandleRequest executes after receiving a request (the client does not have a public Invitation)
-func (c *Client) HandleRequest(dest1, dest2 *service.Destination) error {
-	return c.sendProposal(options{
-		Destinations: []*service.Destination{dest1, dest2},
+// HandleRequest is a helper function to prepare the right protocol dependency interface
+// It can be executed after receiving a Request action message (the client does not have a public Invitation)
+func (c *Client) HandleRequest(msg service.DIDCommMsg, dest1, dest2 *service.Destination) error {
+	thID, err := msg.ThreadID()
+	if err != nil {
+		return fmt.Errorf("handle request threadID: %w", err)
+	}
+
+	return c.saveInvitationEnvelope(thID, InvitationEnvelope{
+		Dests: []*service.Destination{dest1, dest2},
 	})
 }
 
-// HandleRequestWithInvitation executes after receiving a request when the client has a public Invitation
-func (c *Client) HandleRequestWithInvitation(inv *didexchange.Invitation, dest *service.Destination) error {
-	return c.sendProposal(options{
-		Invitation:   inv,
-		Destinations: []*service.Destination{dest},
+// HandleRequestWithInvitation is a helper function to prepare the right protocol dependency interface
+// It can be executed after receiving a Request action message (the client has a public Invitation)
+// nolint: lll
+func (c *Client) HandleRequestWithInvitation(msg service.DIDCommMsg, inv *didexchange.Invitation, dest *service.Destination) error {
+	thID, err := msg.ThreadID()
+	if err != nil {
+		return fmt.Errorf("handle request with invitation threadID: %w", err)
+	}
+
+	return c.saveInvitationEnvelope(thID, InvitationEnvelope{
+		Inv:   inv,
+		Dests: []*service.Destination{dest},
 	})
 }
 
-func (c *Client) handleOutbound(msg interface{}) error {
+func (c *Client) handleOutbound(msg interface{}, o InvitationEnvelope) error {
 	payload, err := json.Marshal(msg)
 	if err != nil {
-		return err
+		return fmt.Errorf("marshal outbound msg: %w", err)
 	}
 
 	didMsg, err := service.NewDIDCommMsg(payload)
 	if err != nil {
+		return fmt.Errorf("new outbound DIDCommMsg msg: %w", err)
+	}
+
+	thID, err := didMsg.ThreadID()
+	if err != nil {
+		return fmt.Errorf("outbound threadID: %w", err)
+	}
+
+	if err := c.saveInvitationEnvelope(thID, o); err != nil {
 		return err
 	}
 
-	return c.service.HandleOutbound(didMsg, &service.Destination{})
+	return c.service.HandleOutbound(didMsg, o.Dests[0])
 }
 
-// options keeps the information needed for sending a proposal
-type options struct {
+// InvitationEnvelope keeps the information needed for sending a proposal
+type InvitationEnvelope struct {
 	// Invitation must be set when we have a public Invitation
-	Invitation *didexchange.Invitation
+	Inv *didexchange.Invitation `json:"inv,omitempty"`
 	// Destinations contain one or two elements
 	// one element - for a public Invitation, otherwise two elements
-	Destinations []*service.Destination
+	Dests []*service.Destination `json:"dests,omitempty"`
+}
+
+// Invitation returns an Invitation needed for the service (state machine depends on it)
+func (o *InvitationEnvelope) Invitation() *didexchange.Invitation {
+	return o.Inv
+}
+
+// Destinations returns destinations needed for the service (state machine depends on it)
+func (o *InvitationEnvelope) Destinations() []*service.Destination {
+	return o.Dests
 }
 
 // sendProposal sends proposal
-// the function may receive an `Invitation` or <nil> (e.g options.Invitation)
+// the function may receive an `Invitation` or <nil> (e.g InvitationEnvelope.Invitation)
 // if the invitation is not <nil> it means we have a public invitation.
-// Destinations (e.g options.Destinations) should have one or two values, if destinations have only one value it means
-// that it is public invitation (an invitation `options.Invitation` is required), otherwise if destinations have two
+// Destinations (e.g InvitationEnvelope.Destinations) should have one or two values, if destinations
+// have only one value it means that it is public invitation
+// (an invitation `InvitationEnvelope.Invitation` is required), otherwise destinations have two
 // values, it means we do not have a public invitation (we will wait for it from one of introducees)
 // Usage:
 //  Sends public invitation to the destination
@@ -139,9 +208,36 @@ type options struct {
 // 			destination2,
 // 		},
 // 	})
-func (c *Client) sendProposal(_ options) error {
+func (c *Client) sendProposal(o InvitationEnvelope) error {
 	return c.handleOutbound(&introduce.Proposal{
 		Type: introduce.ProposalMsgType,
-		ID:   uuid.New().String(),
-	})
+		ID:   c.newUUID(),
+	}, o)
+}
+
+func (c *Client) saveInvitationEnvelope(thID string, o InvitationEnvelope) error {
+	data, err := json.Marshal(o)
+	if err != nil {
+		return fmt.Errorf("marshal invitation envelope: %w", err)
+	}
+
+	return c.store.Put(invitationEnvelopeKey(thID), data)
+}
+
+func (c *Client) getInvitationEnvelope(thID string) (*InvitationEnvelope, error) {
+	data, err := c.store.Get(invitationEnvelopeKey(thID))
+	if err != nil {
+		return nil, fmt.Errorf("get invitation envelope: %w", err)
+	}
+
+	var o *InvitationEnvelope
+	if err := json.Unmarshal(data, &o); err != nil {
+		return nil, fmt.Errorf("unmarshal invitation envelope: %w", err)
+	}
+
+	return o, nil
+}
+
+func invitationEnvelopeKey(thID string) string {
+	return invitationEnvelopePrefix + thID
 }
--- a/pkg/client/introduce/client_test.go
+++ b/pkg/client/introduce/client_test.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package introduce
 
 import (
+	"encoding/json"
 	"errors"
 	"testing"
 
@@ -14,7 +15,9 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/aries-framework-go/pkg/client/introduce/mocks"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	serviceMocks "github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service/mocks"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
 	"github.com/hyperledger/aries-framework-go/pkg/internal/mock/didcomm/protocol"
 	storageMocks "github.com/hyperledger/aries-framework-go/pkg/storage/mocks"
@@ -29,7 +32,7 @@ func TestNew(t *testing.T) {
 	t.Run("get service error", func(t *testing.T) {
 		provider := mocks.NewMockProvider(ctrl)
 		provider.EXPECT().Service(introduce.Introduce).Return(nil, errors.New(errMsg))
-		_, err := New(provider)
+		_, err := New(provider, nil)
 		require.EqualError(t, err, errMsg)
 	})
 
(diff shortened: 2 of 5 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -201,6 +210,30 @@ func (s *Service) doHandle(msg *service.DIDCommMsg, outbound bool) (*metaData, e
...
+		},
+		Msg:      msg,
+		ThreadID: thID,
+	})
+	// there is no way to execute Continue or Stop functions
+	// RequestMsgType is not a part of the state machine
+	// and Client should handle this message type and act according to that (send a proposal)
+	aMsg.Continue = func(args interface{}) { panic("action is forbidden") }
+	aMsg.Stop = func(err error) { panic("action is forbidden") }
+	aEvent <- aMsg
```
soluchok: @llorllale @troyronda @rolsonquadras  What do you think about it?
When we received a Request message we cannot proceed with the protocol. First of all Request message is not a part of the state machine (from the doc). The request message is a kind of ping message or trigger message, but we should not respond to that. There is no even response message for Request. So, my suggestion is to forbid this execution (Continue()), and instead of calling `Continue()` client should send a proposal that does not have anything in common with the request message. 
e.g When Client receives a Request message he may:
* ignore that message
* send a proposal to any agent
* call Continue() and get panic (next time he will handle it properly :))

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -276,13 +295,22 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
-func (s *Service) sendActionEvent(msg *metaData, aEvent chan<- service.DIDCommAction) {
+func (s *Service) newDIDCommActionMsg(msg *metaData) service.DIDCommAction {
 	// create the message for the channel
 	// trigger the registered action event
-	aEvent <- service.DIDCommAction{
+	return service.DIDCommAction{
 		ProtocolName: Introduce,
 		Message:      msg.Msg.Clone(),
 		Continue: func(args interface{}) {
+			// there is no way to receive another interface
```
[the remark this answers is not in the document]
soluchok: yeah, it is correct. I have added `ProtocolDependency(thID string) introduce.ActionDependency` function to the client. The usage must be `e.Continue(c.ProtocolDependency())`

pkg/client/introduce/client.go
```diff
@@ -52,22 +60,42 @@ func New(ctx Provider) (*Client, error) {
...
+// ProtocolDependency returns dependency for the service
+// the function should never return an error, instead of error we provide the callable interface
+// and the state machine will act according to the provided data
+// usage: e.Continue(c.ProtocolDependency())
+func (c *Client) ProtocolDependency(thID string) introduce.ActionDependency {
+	opts, err := c.getOptions(thID)
+	if err != nil {
+		opts = &options{}
+
+		logger.Errorf("protocol dependency: %v", err)
```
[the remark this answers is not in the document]
soluchok: Why? It can be `%s` or `%v` but there is no reason for `%w`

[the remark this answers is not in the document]
sandrask: Never mind, %w is for wrapping errors

pkg/didcomm/protocol/introduce/service.go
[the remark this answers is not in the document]
sandrask: ok

pkg/client/introduce/client.go
```diff
@@ -139,9 +181,36 @@ type options struct {
...
-		ID:   uuid.New().String(),
-	})
+		ID:   c.newUUID(),
+	}, o)
+}
+
+func (c *Client) saveOptions(thID string, o options) error {
+	data, err := json.Marshal(o)
+	if err != nil {
+		return fmt.Errorf("save options: %w", err)
```
Baha-sk: this is a little picky but probably put a more accurate message here, or no message at all by returning err as is. This error is due to marshalling, not saving.

store.Put will will return a `save`.  If you want to send `save options`, then catch the error of `c.store.Put()` below and wrap it with `save options: `

pkg/client/introduce/client.go
```diff
@@ -139,9 +181,36 @@ type options struct {
...
+
+func (c *Client) getOptions(thID string) (*options, error) {
+	data, err := c.store.Get(optionsKey(thID))
+	if err != nil {
+		return nil, fmt.Errorf("get options: %w", err)
+	}
+
+	var o *options
+	if err := json.Unmarshal(data, &o); err != nil {
+		return nil, fmt.Errorf("unmarshal options: %w", err)
```
Baha-sk: the error message is right here :)

pkg/client/introduce/client_test.go
```diff
@@ -53,106 +56,250 @@ func TestNew(t *testing.T) {
...
+
+	opts := options{
+		Dests: []*service.Destination{
+			{ServiceEndpoint: "service/endpoint1"},
+			{ServiceEndpoint: "service/endpoint2"},
+		},
+	}
+
+	store := storageMocks.NewMockStore(ctrl)
+	store.EXPECT().Put("client_options_"+UUID, toBytes(t, opts)).Return(nil)
```
Baha-sk: any reason for not using `optionsPrefix` instead of the string literal `"client_options_"` here?

pkg/client/introduce/client_test.go
```diff
@@ -53,106 +56,250 @@ func TestNew(t *testing.T) {
...
+		Inv: &didexchange.Invitation{
+			ID: UUID,
+		},
+		Dests: []*service.Destination{{
+			ServiceEndpoint: "service/endpoint",
+		}},
+	}
+
+	store := storageMocks.NewMockStore(ctrl)
+	store.EXPECT().Put("client_options_"+UUID, toBytes(t, opts)).Return(nil)
```
Baha-sk: same here

pkg/client/introduce/client_test.go
```diff
@@ -53,106 +56,250 @@ func TestNew(t *testing.T) {
...
+
+	opts := options{
+		Dests: []*service.Destination{
+			{ServiceEndpoint: "service/endpoint1"},
+			{ServiceEndpoint: "service/endpoint2"},
+		},
+	}
+
+	store := storageMocks.NewMockStore(ctrl)
+	store.EXPECT().Put("client_options_"+UUID, toBytes(t, opts)).Return(nil)
```
Baha-sk: same here

pkg/client/introduce/client_test.go
```diff
@@ -53,106 +56,250 @@ func TestNew(t *testing.T) {
...
-	provider.EXPECT().Service(introduce.Introduce).Return(DIDComm, nil)
+
+	opts := options{
+		Dests: []*service.Destination{{
+			ServiceEndpoint: "service/endpoint",
+		}},
+	}
+
+	store := storageMocks.NewMockStore(ctrl)
+	store.EXPECT().Put("client_options_"+UUID, toBytes(t, opts)).Return(nil).Times(1)
```
Baha-sk: I see the same string literal used here.. using the constant is beneficial for easier referencing.

pkg/didcomm/protocol/introduce/service.go
Baha-sk: doing panic means the whole agent will go down... I get that this is not supposed to happen.. but is there a way to tell the agent (or support team) what's the root cause of this state? (maybe simply logging the error with a meaningful message is enough)

But providing hints to the ops, it will be easier to investigate what went wrong. I'm not familiar with the state machine, but adding hints will definitely help. And personally, I would avoid panics unless the agent is genuinely in a non recoverable state. 

I guess @troyronda @llorllale @rolsonquadras might have a better opinion about it.

pkg/didcomm/protocol/introduce/service_test.go
```diff
@@ -568,8 +547,146 @@ func TestService_save(t *testing.T) {
...
+		checkStateMsg(t, sCh, service.PostState, ProposalMsgType, stateNameDeciding)
+		checkStateMsg(t, sCh, service.PreState, ProposalMsgType, stateNameWaiting)
+		checkStateMsg(t, sCh, service.PostState, ProposalMsgType, stateNameWaiting)
+
+		// creates Ack msg
+		ackMsg, err := service.NewDIDCommMsg(toBytes(t, model.Ack{
+			Type: AckMsgType,
+			ID:   thID,
+		}))
+		require.NoError(t, err)
```
Baha-sk: this is a general question, not just specific to this line:

I don't see error (failure test cases in the tests), there's only require.NoError but no require.Error... 

Adding them will increase the test coverage as well. As a general rule we try to increase the code coverage in the project with each patch.

soluchok: in this package, only one line is not covered
```
data, err := json.Marshal(o)
if err != nil {
	return fmt.Errorf("marshal options: %w", err)
}
```

pkg/client/introduce/client.go

pkg/didcomm/protocol/introduce/service_test.go
Baha-sk: Unless `o` can be injected.. then it's fine as we have no control over JSON marshalling a struct.

pkg/didcomm/protocol/introduce/service.go
soluchok: First of all this comment related to the design (panic or logging is not a problem), the question is should we proceed with protocol after receiving this message (`Continue()` or `Stop()`) or forbid it (not possible to use `Continue()` or `Stop()`) and start sending a proposal independently (no matter what we received in the Request)

llorllale: @soluchok 

> When we received a Request message we cannot proceed with the protocol. [...] Request message is not a part of the state machine (from the doc)

I opened hyperledger/aries-rfcs#309 regarding this - I have my doubts.

troyronda: We should not have panics.

soluchok: @troyronda @Baha-sk Done!

pkg/client/introduce/client.go
```diff
@@ -52,22 +60,42 @@ func New(ctx Provider) (*Client, error) {
 		Event:   introduceSvc,
 		service: introduceSvc,
 		store:   store,
+		newUUID: func() string { return uuid.New().String() },
```
troyronda: are you expecting a new client to be created for each invocation of introduce?
how does the recipient listen for introduction requests? certainly they cannot creating new clients on every potentially parallel request.

we are building a controller model and we also need to consider other bindings like rest.

pkg/client/introduce/client.go
```diff
@@ -27,8 +33,10 @@ type Provider interface {
 // Client enable access to introduce API
 type Client struct {
 	service.Event
-	service service.DIDComm
-	store   storage.Store
+	service    service.DIDComm
+	store      storage.Store
+	privateInv *didexchange.Invitation
```
troyronda: please use default rather than private.

pkg/client/introduce/client.go
```diff
@@ -52,22 +60,42 @@ func New(ctx Provider) (*Client, error) {
 		Event:   introduceSvc,
 		service: introduceSvc,
 		store:   store,
+		newUUID: func() string { return uuid.New().String() },
 	}, nil
 }
 
+// ProtocolDependency returns dependency for the service
```
troyronda: i think this is confusing in a developer facing API. i don't grok it from the description.

pkg/client/introduce/client.go
```diff
@@ -9,15 +9,21 @@ package introduce
...
 
-	"github.com/hyperledger/aries-framework-go/pkg/client/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
+const optionsPrefix = "client_options_"
```
troyronda: what are these options that are being stored? is there a more descriptive name?

pkg/client/introduce/client.go
troyronda: note: we also need to allow for deferred continues in controller APIs (particularly REST). See DID exchange allowing for continue operations by connection ID.

pkg/client/introduce/client.go

pkg/client/introduce/client.go
soluchok: renamed it to `const proposalOptionsPrefix = "proposal_options_"`, when we create a proposal there are a bunch of options (arguments), we cannot pass it directly to the service. We save options to DB and then we are passing them through `Continue(introduce.ActionDependency)` function.

pkg/client/introduce/client.go
soluchok: I have changed it a bit. ` the event I got from the client anyways. it is a bit surprising that this is needed.` there are two options: you can handle action event and pass your dependency (managing by yourself) or you can always use ProtocolDependency(msg.TreadID()) function.

Or we can add ProtocolDependency function to the service (like a service dependency) so, in this way, we can get rid of e.Continue(c.ProtocolDependency(thID)) bc service can call it by yourself

soluchok: I would like to pass it as a dependency function to the service. What do you think?

pkg/didcomm/protocol/introduce/service.go
llorllale: @soluchok I pushed PR hyperledger/aries-rfcs#313 to address hyperledger/aries-rfcs#309. Basically, there is agreement that the request message *is* part of the state machine.

soluchok: @llorllale just to clarify `requesting` state is still optional right?

llorllale: @soluchok yes, the introducee can begin from either the requesting state or the deciding state

pkg/client/introduce/client.go
```diff
@@ -52,77 +60,136 @@ func New(ctx Provider) (*Client, error) {
...
 	})
 }
 
-// HandleRequest executes after receiving a request (the client does not have a public Invitation)
-func (c *Client) HandleRequest(dest1, dest2 *service.Destination) error {
-	return c.sendProposal(options{
-		Destinations: []*service.Destination{dest1, dest2},
+// PrepareDependency is a helper function to prepare the right protocol dependency interface
+// Usually, it executes after receiving a Request action message (the client does not have a public Invitation)
+func (c *Client) PrepareDependency(msg service.DIDCommMsg, dest1, dest2 *service.Destination) error {
```
troyronda: the previous Handle method names made more sense to me.

prepare dependency doesn’t make sense to an end developer. they care about the action they are taking - eg., HandleFoo.

pkg/client/introduce/client.go
```diff
@@ -52,77 +60,136 @@ func New(ctx Provider) (*Client, error) {
...
+// service to proceed with the protocol. Dependency should be passed to the service through `Continue` function.
+// The function should never return an error, instead of error we provide the callable interface
+// and the state machine will act according to the provided data.
+// Dependency is populated after executing the following functions:
+//  - SendProposal
+//  - SendProposalWithInvitation
+//  - PrepareDependency
+//  - PrepareDependencyWithInvitation
+// usage: e.Continue(c.ProtocolDependency(threadID))
+func (c *Client) ProtocolDependency(thID string) introduce.ActionDependency {
```
troyronda: we haven’t found a way to hide this complexity or rename? i do not think end developers will easily understand this method.

pkg/client/introduce/client.go
```diff
@@ -52,77 +60,136 @@ func New(ctx Provider) (*Client, error) {
...
-// HandleRequestWithInvitation executes after receiving a request when the client has a public Invitation
-func (c *Client) HandleRequestWithInvitation(inv *didexchange.Invitation, dest *service.Destination) error {
-	return c.sendProposal(options{
-		Invitation:   inv,
-		Destinations: []*service.Destination{dest},
+// PrepareDependencyWithInvitation is a helper function to prepare the right protocol dependency interface
+// Usually, it executes after receiving a Request action message (the client has a public Invitation)
+// nolint: lll
+func (c *Client) PrepareDependencyWithInvitation(msg service.DIDCommMsg, inv *didexchange.Invitation, dest *service.Destination) error {
+	thID, err := msg.ThreadID()
```
troyronda: again here. previous name was better.

pkg/client/introduce/client.go
soluchok: The previous method was sending a message using `HandleOutbound` method, now, we do not have this logic. What I can do just rename it to `PrepareRequestDependency`. Handle is not what this method does.

pkg/client/introduce/client.go
soluchok: The right name for me can be `ActionDependency`. Is it ok?

soluchok: I was avoiding this name because of the function result which clearly says that this is `introduced.ActionDependency`

troyronda: i would rather not expose it at all.

pkg/client/introduce/client.go
troyronda: name according to what the developer understands. prepare dependency is not meaningful.

pkg/client/introduce/client.go
troyronda: i do not understand these “options” and why they are being stored in this manner. is this the introduce request?

pkg/client/introduce/client.go

pkg/client/introduce/client.go

pkg/client/introduce/client.go
soluchok: When we send a proposal request I want to prepare all the data before the state machine starts. The protocol starts from proposal message, according to that logic, a proposal has a bunch of options (combinations). We do not need this data when we send a proposal but we need it on the next `Continue` iteration. To avoid handling action messages by end-user(client) we saved necessary information to the DB and we provided the method `ProtocolDependency(thID string)` which will supply the service with the data (options are converted to dependency).

```
//  Sends public invitation to the destination
// 	sendProposal(&options{
// 		Invitation: invitation,
// 		Destinations: []*service.Destination{
// 			destination,
// 		},
// 	})
// 	Forwards an invitation from one introducee to another
// 	sendProposal(&options{
// 		Destinations: []*service.Destination{
// 			destination1,
// 			destination2,
// 		},
// 	})
```

pkg/client/introduce/client.go
soluchok: At least I need this for the tests (for gomock generation), also it is clear for the user what interface he should implement. There is nothing wrong with exposing interfaces, it is even better.

rolsonquadras: We shouldn't expose service pkg dependencies to the business layer. ie. `introduce.ActionDependency`.

pkg/client/introduce/client.go
```diff
@@ -52,77 +60,136 @@ func New(ctx Provider) (*Client, error) {
...
 	})
 }
 
-// HandleRequest executes after receiving a request (the client does not have a public Invitation)
-func (c *Client) HandleRequest(dest1, dest2 *service.Destination) error {
-	return c.sendProposal(options{
-		Destinations: []*service.Destination{dest1, dest2},
+// HandleRequest is a helper function to prepare the right protocol dependency interface
+// It can be executed after receiving a Request action message (the client does not have a public Invitation)
+func (c *Client) HandleRequest(msg service.DIDCommMsg, dest1, dest2 *service.Destination) error {
```
rolsonquadras: Will there be only 2 destinations (dest1 and dest2) ?

pkg/client/introduce/client.go
troyronda: @sandrask ^^^

pkg/client/introduce/client.go
troyronda: @sandrask @rolsonquadras ^^^

pkg/client/introduce/client.go
```diff
@@ -27,8 +33,10 @@ type Provider interface {
 // Client enable access to introduce API
 type Client struct {
 	service.Event
-	service service.DIDComm
-	store   storage.Store
+	service    service.DIDComm
+	store      storage.Store
+	defaultInv *didexchange.Invitation
```
rolsonquadras: Where is this set ? I see only one [reference](https://github.com/hyperledger/aries-framework-go/pull/737/files#diff-53f48d9400b55325b3f6a88c6129625eR86) where we are assigning this to opts.

pkg/client/introduce/client.go
```diff
@@ -52,77 +60,136 @@ func New(ctx Provider) (*Client, error) {
 		Event:   introduceSvc,
 		service: introduceSvc,
 		store:   store,
+		newUUID: func() string { return uuid.New().String() },
```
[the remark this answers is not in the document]
troyronda: I previously asked this question. This was for testing (similar to what @Moopli has done elsewhere for rand).

pkg/client/introduce/client.go
soluchok: Correct, there is no possibility to set it. It will be added on the next PR.

pkg/client/introduce/client.go
soluchok: For now only two destinations. There is some TODO about it https://github.com/hyperledger/aries-rfcs/tree/master/features/0028-introduce#roles

pkg/client/introduce/client.go
soluchok: @rolsonquadras It is `ActionDependency` and the logic is based on that. This is how the end-user communicates with the protocol. e.g passing the invitation and destinations

pkg/client/introduce/client.go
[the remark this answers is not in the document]
soluchok: If you really don't like it I can remove it. It was done to have more control over the tests.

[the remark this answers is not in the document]
rolsonquadras: @soluchok  just wanted to understand the need for it.

pkg/client/introduce/client.go
rolsonquadras: @soluchok my comments were not on the use of this function. I was merely pointing out that the return type `introduce.ActionDependency` is coming from service package. Till now we have designed in such a way that the consumers will not have any imports from service pkg.

pkg/client/introduce/client.go
rolsonquadras: ok. I saw the suffix '1' and '2'. hence was just curious.

pkg/client/introduce/client.go
soluchok: Currently, the consumer may impact on the protocol logic. According to that, I must provide the right interface for that with the description of how it works. This is a way how they communicate (end-user with the service) and I cannot just hide it.
What if I need one more function in my dependency? In that case, consumer, client whatever must implement the interface. In my case service requires this interface, the protocol will not work without it. But how can we implement a hidden interface and make sure we satisfy it?
And this is exactly why it is public.

soluchok: To be honest I am not excited about it as well. I would rather do something like that:
* `ProtocolDependency(thID string) Options`
* `Continue(introduce.ActionDependency)`
But the problem is I cannot specify the specific interface for the `Continue` function.

pkg/client/introduce/client.go
soluchok: @rolsonquadras I've added a function `SetDefaultInvitation`

troyronda: I don't think we should mutate default invitation after creation of the client.

pkg/client/introduce/client.go
troyronda: @rolsonquadras comment is that you are returning a type from the service package. I would also say that you should prefer to return a struct rather than an interface.

This function could return an exported *Options struct.

pkg/client/introduce/client.go
soluchok: @troyronda totally agree I was thinking about it. My decision was based on flexibility.

pkg/client/introduce/client.go

pkg/client/introduce/client.go
```diff
@@ -9,15 +9,21 @@ package introduce
...
 
-	"github.com/hyperledger/aries-framework-go/pkg/client/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
+const invitationEnvelopePrefix = "proposal_options_"
```
troyronda: is this correct?

soluchok: no :) Done!

pkg/client/introduce/client.go
```diff
@@ -49,80 +57,141 @@ func New(ctx Provider) (*Client, error) {
...
+	return c.service.HandleOutbound(didMsg, o.Dests[0])
 }
 
-// options keeps the information needed for sending a proposal
-type options struct {
+// InvitationEnvelope keeps the information needed for sending a proposal
+type InvitationEnvelope struct {
 	// Invitation must be set when we have a public Invitation
-	Invitation *didexchange.Invitation
+	Inv *didexchange.Invitation `json:"inv,omitempty"`
```
rolsonquadras: Do we need to expose these fields ?

soluchok: Yeah, for marshall, this structure is being stored in DB

PR #742, opened 2019-11-08
feat: DIDExchange RestAPI - Update state name for action callback and Support for Accept Invitation

### LINKED ISSUE #690: DIDExchange - State for Action Events
Currently, when action events are triggered the state is ["null"](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/states.go#L312). The actual state transition happens after action events are triggered. Need to review this logic.

### LINKED ISSUE #692: REST accept-invitation doesn't seem to be needed

rolsonquadras:
- Currently action events are triggered before updating the state for the incoming message(invited/requested). Moved action event triggering point after state is updated, ie, invited and requested based on agent role (invitee/inviter).
- Added support to Accept Invitation through RestAPI client. Now, RestAPI triggers accept-invitation when it receives "invited" and accept-request on "requested" state from webhook notification.

Closes #690 
Closes #692 

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -60,6 +60,9 @@ type protocolService interface {
 
 	// Accepts/Approves exchange request
 	AcceptExchangeRequest(connectionID string) error
+
+	// Accepts/Approves exchange invitation
+	AcceptInvitation(connectionID string) error
 }
 
 // New return new instance of didexchange client
@@ -152,6 +155,17 @@ func (c *Client) HandleInvitation(invitation *Invitation) (string, error) {
 	}
 
 	return connectionID, nil
+}
+
+// TODO https://github.com/hyperledger/aries-framework-go/issues/754 - e.Continue v Explicit API call for action events
+
+// AcceptInvitation accepts/approves exchange invitation.
+func (c *Client) AcceptInvitation(connectionID string) error {
+	if err := c.didexchangeSvc.AcceptInvitation(connectionID); err != nil {
+		return fmt.Errorf("did exchange client - accept exchange invitation: %w", err)
+	}
+
+	return nil
 }
 
 // AcceptExchangeRequest accepts/approves exchange request.
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -135,6 +135,7 @@ func New(didMaker didcreator.Creator, prov provider) (*Service, error) {
 
 // HandleInbound handles inbound didexchange messages.
 func (s *Service) HandleInbound(msg *service.DIDCommMsg) (string, error) {
+	// https://github.com/hyperledger/aries-framework-go/issues/753 -- DIDExchange Svc - Action Event refactor
 	// throw error if there is no action event registered for inbound messages
 	aEvent := s.ActionEvent()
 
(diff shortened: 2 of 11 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -238,7 +228,7 @@ func (s *Service) nextState(msgType, thID string) (state, error) {
 	return next, nil
 }
 
-func (s *Service) handle(msg *message) error {
+func (s *Service) handle(msg *message, aEvent chan<- service.DIDCommAction) error { //nolint funlen
```
troyronda: maybe the function should be split ;)

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -164,18 +164,8 @@ func (s *Service) HandleInbound(msg *service.DIDCommMsg) (string, error) {
...
-			return "", fmt.Errorf("handle inbound : %w", err)
-		}
-
-		return connRecord.ConnectionID, nil
-	}
-
-	// if no action event is triggered, continue the execution
 	go func(msg *message) {
-		if err = s.handle(msg); err != nil {
+		if err = s.handle(msg, aEvent); err != nil {
```
troyronda: any reason aEvent isn't an argument like msg?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -164,18 +164,8 @@ func (s *Service) HandleInbound(msg *service.DIDCommMsg) (string, error) {
 
 	internalMsg := &message{Msg: msg, ThreadID: thID, NextStateName: next.Name(), ConnRecord: connRecord}
 
-	// trigger action event based on message type for inbound messages
-	if canTriggerActionEvents(msg.Header.Type) {
```
troyronda: this code isn't needed anymore?

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: missed this. I'll update it.

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/0acce2085bbdff34cadccb9d86d74957ba12b028..c3fbb046885a794db280ed57a43d363cda1b2742#diff-f761c91e753eba1c42a68076df06ce12R171).

pkg/didcomm/protocol/didexchange/service.go
rolsonquadras: When I look at this [function](https://github.com/hyperledger/aries-framework-go/blob/c3fbb046885a794db280ed57a43d363cda1b2742/pkg/didcomm/protocol/didexchange/service.go#L231), there are 20 lines for struct field value setting, 7 lines for return error, 5 lines for logging. There are no logic code(it's invoking other functions). So not sure it would make much sense to split this function.

pkg/client/didexchange/client.go
```diff
@@ -154,6 +157,15 @@ func (c *Client) HandleInvitation(invitation *Invitation) (string, error) {
 	return connectionID, nil
 }
 
+// AcceptInvitation accepts/approves exchange invitation.
+func (c *Client) AcceptInvitation(connectionID string) error {
```
troyronda: I'm not understanding the usage of this function in the Go binding. It's a bit confusing?

troyronda: TODO: determine what to do about currently having two ways to handle action events.

pkg/didcomm/protocol/didexchange/service.go
troyronda: Is the suggestion to have a longer funlen maximum ... or this one just happens to be an exception?

rolsonquadras: more of an exception. But yeah, we might need to increase the funclen config as we have wsl linter now.

pkg/client/didexchange/client.go
rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/c3fbb046885a794db280ed57a43d363cda1b2742..a400b15ddf9c1cfd918fcc3adfe90ded902ee5c6#diff-427522b353a28b1ec350c29c79d387b0R160).

PR #767, opened 2019-11-11
refactor: Introduce - replace custom mock with gomock

soluchok:
* gomock instead of  custom mock (introduce protocol)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,8 @@ license:
 	@scripts/check_license.sh
 
 .PHONY: unit-test
-unit-test: mocks
+unit-test:
+	@make -s mocks
 	@scripts/check_unit.sh
 
 .PHONY: bdd-test
@@ -120,8 +121,8 @@ comma:= ,
 semicolon:= ;
 
 define create_mock
-  mkdir -p $(1)/mocks && rm -rf $(1)/mocks/*
-  $(MOCKGEN) -destination $(1)/mocks/mocks.go -self_package mocks -package mocks $(PROJECT_ROOT)/$(1) $(subst $(semicolon),$(comma),$(2))
+  mkdir -p $(1)/gomocks && rm -rf $(1)/gomocks/*
+  $(MOCKGEN) -destination $(1)/gomocks/mocks.go -self_package mocks -package mocks $(PROJECT_ROOT)/$(1) $(subst $(semicolon),$(comma),$(2))
 endef
 
 build-mockgen:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -130,7 +130,8 @@ build-mockgen:
 .PHONY: mocks
 mocks: build-mockgen
 	$(call create_mock,pkg/client/introduce,Provider)
-	$(call create_mock,pkg/didcomm/protocol/introduce,InvitationEnvelope)
+	$(call create_mock,pkg/didcomm/protocol/introduce,Provider;InvitationEnvelope)
```
fqutishat: we need to find way to clean mock file

fqutishat: currently we have mock folder in each sub package

rolsonquadras:
@soluchok a suggestion - Can we update the commit title to better align with the changes ie prefix with feature/protocol that we are changing ? eg. "refactor: Introduce - replace custom mock with gomock'. We use custom mocks in many places and the title without prefix might infer that we are replacing all of custom mocks with gomock.

Makefile
soluchok: sorry not sure I understand, I did not commit mock files, we only generate it on CI and local development (to run the tests). My point is to not commit mocks in the repository.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -85,19 +85,20 @@ type Service struct {
...
+	wg          sync.WaitGroup
 	stop        chan struct{}
 	closedMutex sync.Mutex
 	closed      bool
 }
 
-// provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
-type provider interface {
+// Provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
+type Provider interface {
```
rolsonquadras: Does this need to be exposed for gomocks ? just trying to understand the reason for this change.

soluchok: Yep

Makefile
troyronda: @soluchok make clean should ideally remove all generated and build artifacts.

soluchok: How about to replace it with `gomock` ? Or another name that is not so generic

soluchok: I thought all mocks are located here `aries-framework-go/pkg/internal/mock`

soluchok: @fqutishat @troyronda @rolsonquadras ^

Makefile
```diff
@@ -130,10 +130,15 @@ build-mockgen:
...
+	$(call create_mock,pkg/didcomm/dispatcher,Outbound)
 	$(call create_mock,pkg/storage,Provider;Store)
 	$(call create_mock,pkg/didcomm/common/service,DIDComm)
 
+.PHONY: clean-mocks
+clean-mocks:
```
fqutishat: Add clean-mocks to clean target

Makefile
fqutishat: looks good just add clean-mocks to clean target

Makefile
soluchok: there is an issue, looking for cause

PR #771, opened 2019-11-11
refactor: Combine WithPacker and WithInboundPackers

### LINKED ISSUE #764: Combine WithPacker options
- Instead of WithPacker and WithInboundPacker, just have WithPacker
- The first argument of WithPacker is required. It specifies the packer that will be used when initiating DID exchanges with a new party.
- The second argument is variadic. It specifies additional packers that are supported when other parties are initiating DID exchanges.

Moopli:
WithPacker now takes at least 1 Packer(Creator for framework).
The first Packer is used for outbound, all Packers are used for
inbound.

Fixes #764

### DIFF
--- a/pkg/didcomm/packager/package_test.go
+++ b/pkg/didcomm/packager/package_test.go
@@ -32,15 +32,15 @@ func TestBaseKMSInPackager_UnpackMessage(t *testing.T) {
 		require.NoError(t, err)
 
 		mockedProviders := &mockProvider{
-			storage:        nil,
-			kms:            w,
-			outboundPacker: nil,
-			packers:        nil,
+			storage:       nil,
+			kms:           w,
+			primaryPacker: nil,
+			packers:       nil,
 		}
 		testPacker, err := jwe.New(mockedProviders, jwe.XC20P)
 		require.NoError(t, err)
 
-		mockedProviders.outboundPacker = testPacker
+		mockedProviders.primaryPacker = testPacker
 		packager, err := New(mockedProviders)
 		require.NoError(t, err)
 		_, err = packager.UnpackMessage(nil)
@@ -55,15 +55,15 @@ func TestBaseKMSInPackager_UnpackMessage(t *testing.T) {
 		require.NoError(t, err)
 
 		mockedProviders := &mockProvider{
-			storage:        nil,
-			kms:            w,
-			outboundPacker: nil,
-			packers:        nil,
+			storage:       nil,
+			kms:           w,
+			primaryPacker: nil,
+			packers:       nil,
 		}
 		testPacker, err := jwe.New(mockedProviders, jwe.XC20P)
 		require.NoError(t, err)
 
-		mockedProviders.outboundPacker = testPacker
+		mockedProviders.primaryPacker = testPacker
 		packager, err := New(mockedProviders)
 		require.NoError(t, err)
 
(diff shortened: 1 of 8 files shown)


pkg/framework/context/context.go
```diff
@@ -232,9 +232,10 @@ func WithPackager(p commontransport.Packager) ProviderOption {
 }
 
 // WithPacker injects a Packer into the context as the outbound Packer
```
troyronda: Comment needs to be updated.

pkg/framework/aries/framework.go
```diff
@@ -170,20 +170,13 @@ func WithKMS(k api.KMSCreator) Option {
 	}
 }
 
-// WithPacker injects a Packer service to the Aries framework
-// to pack outbound messages and be available for unpacking inbound messages.
-func WithPacker(c packer.Creator) Option {
+// WithPacker injects Packer services to the Aries framework.
+// The first parameter is used for outbound and inbound messages,
+// and the optional additional packers are for inbound messages.
+func WithPacker(c packer.Creator, inboundPackers ...packer.Creator) Option {
```
Baha-sk: we should remove inbound/outbound words from packer

c is the default/primary packer for packing/unpacking

inboundPackers is a list of packers to accommodate different types of envelopes.. yes, initially for inbound envelopes, but if in future communications 2 agents talk to each other with a different type than the primary one defined in `c`, then a packer from this list should be referenced.

this is why the word `inbound` can be misleading

pkg/framework/aries/framework.go
```diff
@@ -170,20 +170,13 @@ func WithKMS(k api.KMSCreator) Option {
...
-		return nil
-	}
-}
-
-// WithInboundPackers injects a variable number of Packer services into the Aries framework
-// for unpacking inbound messages.
-func WithInboundPackers(packers ...packer.Creator) Option {
-	return func(opts *Aries) error {
-		opts.inboundPackerCreators = append(opts.inboundPackerCreators, packers...)
+		opts.inboundPackerCreators = append(opts.inboundPackerCreators, inboundPackers...)
```
Baha-sk: same here... I would remove `inbound` everywhere

pkg/framework/context/context.go
```diff
@@ -232,9 +232,10 @@ func WithPackager(p commontransport.Packager) ProviderOption {
...
-func WithPacker(p packer.Packer) ProviderOption {
+func WithPacker(p packer.Packer, inboundPackers ...packer.Packer) ProviderOption {
```
Baha-sk: remove `inbound`

pkg/framework/aries/framework.go
Moopli: Good point, I'll call them primary and additional packers.

pkg/framework/context/context.go
Moopli: done

pkg/didcomm/packager/package_test.go
```diff
@@ -289,6 +289,6 @@ func (m *mockProvider) InboundTransportEndpoint() string {
 	return "sample-endpoint.com"
 }
 
-func (m *mockProvider) Packer() packer.Packer {
-	return m.outboundPacker
+func (m *mockProvider) PrimaryPacker() packer.Packer {
+	return m.packer
```
Baha-sk: might as well call the member field `primaryPacker`

pkg/framework/context/context.go
```diff
@@ -231,18 +231,13 @@ func WithPackager(p commontransport.Packager) ProviderOption {
...
-		opts.packer = p
-		return nil
-	}
-}
-
-// WithInboundPackers injects a variable number of Packer services into the Aries framework
-func WithInboundPackers(packers ...packer.Packer) ProviderOption {
-	return func(opts *Provider) error {
-		opts.inboundPackers = append(opts.inboundPackers, packers...)
+		opts.packer = primary
```
Baha-sk: thinking about this.. primaryPacker is not optional.. it is mandatory
or we can say it's mandatory if WithPackers() is not used

another thought:
If WithPacker is not set and WithPackers is set, then use the first packer in the WithPackers list as primary packer. There must always be one packer created regardless if both With options were used or not (hence fallback to default packer).

Moopli: There should be no more `WithPackers()` - `WithPacker()` takes extra packers as variadic args after the primary one.

pkg/didcomm/packager/package_test.go
Moopli: Done

pkg/framework/context/context.go
Moopli: done

pkg/didcomm/packager/package_test.go
Moopli: done

Issue #739, opened 2019-11-08
0.1.0 Documentation


troyronda:
README Structure

- [x] Project badges
- [x] Project scope and overview (#90) (@troyronda, ...)
- [x] Project structure and usage pattern (markdown files in docs + link) @Moopli 
- [x] Supported specs list  (markdown files in docs + link) @sandrask @llorllale 
- [x] Controller API bindings (markdown files in docs + link)
  - [x] Go (including link to GoDoc, usage, examples) @sandrask 
  - [x] REST (including link to OpenAPI/Swagger, usage, deamon/docker image usage) @rolsonquadras @sudeshrshetty @DRK3 
  * Future: WASM, C 
- [x]  BDD tests (markdown files in docs + link) @fqutishat 
- [x]  Contributing (markdown files in docs + link)
  - [x] Guidelines
  - [x] Build and test @Baha-sk 
  - [x] Test suites @Baha-sk 
- [x] License

GoDoc targeted to end developers
- [x] Framework folder @rolsonquadras 
- [x] Client folder @sandrask 
- [x] Root project folder @fqutishat

PR #772, opened 2019-11-11
docs: Documentation, examples for didexchange client

sandrask:
Documentation and examples for didexchange client.

part #739

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -96,7 +96,9 @@ func New(ctx provider) (*Client, error) {
 	}, nil
 }
 
-// CreateInvitation create invitation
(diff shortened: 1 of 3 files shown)


pkg/client/didexchange/client.go
```diff
@@ -4,6 +4,21 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
+// Package didexchange enables relationship between two agents via DID Exchange Protocol.
```
rolsonquadras: I feel it would be better to have package comments in pkg level doc.go file. Here is a [reference](https://github.com/golang/go/blob/master/src/fmt/doc.go).

sandrask: Done

PR #778, opened 2019-11-11
docs: Aries framework pkg godocs

rolsonquadras:
Part of #739 


pkg/framework/aries/doc.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+/*
+Package aries provides access to the library consumers to create a framework with options. The framework
```
troyronda: Would be an idea to update this text a bit to align with the README.

> A pluggable dependency framework, where implementors can customize primitives via Service Provider Interfaces (SPIs). We have a "batteries included" model where default primitives are included -- such as a key management system (KMS), crypto, data storage, digital hub integration, etc.

pkg/framework/aries/framework.go
```diff
@@ -46,7 +46,8 @@ type Aries struct {
 // Option configures the framework.
 type Option func(opts *Aries) error
 
-// New initializes the Aries framework based on the set of options provided.
+// New initializes the Aries framework based on the set of options provided. This function returns a framework
+// which can be used to manage Aries clients by getting the framework context.
```
troyronda: Should also create examples. See DID exchange godoc.

pkg/framework/aries/doc.go
rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/1a0fabad4135a7495c46315752ad06c326249a89..6d8821c860b3f4c2f12b4e5048eacf470b90bcd2#diff-d8e0c85fc8df82bb6530d2620c04e3daR8).

pkg/framework/aries/framework.go
rolsonquadras: [Added example](https://github.com/hyperledger/aries-framework-go/compare/1a0fabad4135a7495c46315752ad06c326249a89..6d8821c860b3f4c2f12b4e5048eacf470b90bcd2#diff-6b0dce919a202a9ca9791c186a715b5cR17).

rolsonquadras:
@sandrask I agree with both the points. Will update the example and push the changes.

PR #781, opened 2019-11-11
Documentation for webhook

DRK3:


pkg/restapi/webhook/readme_webhook.txt
```diff
@@ -0,0 +1,5 @@
+Aries-agentd uses a webhook mechanism to communicate events back to the controller.
```
troyronda: Please use markdown format not txt files.

fqutishat: Add it under docs and in the same file with aries-agentd executable docs

DRK3: Will do. I will also close this and add it to my other PR as suggested by @fqutishat.

PR #794, opened 2019-11-12
fix: externalAddr scheme can be set by user

### LINKED ISSUE #747: externalAddr scheme can be set by user

DRK3:
Closes #747

Removed the hardcoded "http://" scheme from Inbound Endpoint.


troyronda:
missing semantic prefix in title.

pkg/didcomm/transport/http/inbound.go
```diff
@@ -157,5 +157,5 @@ func (i *Inbound) Stop() error {
 func (i *Inbound) Endpoint() string {
 	// return http prefix as framework only supports http
 	// TODO https://github.com/hyperledger/aries-framework-go/issues/747 configure scheme for external address
```
rolsonquadras: We can remove this TODO.

DRK3:
> missing semantic prefix in title.

Done!

pkg/didcomm/transport/http/inbound.go
DRK3: done

PR #805, opened 2019-11-14
feat: Implement service behavior according to Client dependency

### LINKED ISSUE #802: Implement service behavior according to Client dependency (introduce protocol)
After implementation https://github.com/hyperledger/aries-framework-go/issues/724 service should behave (act) according to the dependency which it receives through the `e.Continue(c.ProtocolDependency(thID))` function.

soluchok:
* added tests for state machine behavior (`SkipProposal`, `Proposal`, `ProposalUnusual`)
* service logic according to dependency (`ExecuteInbound`,`ExecuteOutbound`)

Closes #802 
Part of #269

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -34,7 +34,10 @@ const (
 	AckMsgType = IntroduceSpec + "ack"
 )
 
-const initialWaitCount = 2
+const (
+	initialWaitCount    = 2
+	introduceeIndexNone = -1
+)
 
 var logger = log.New("aries-framework/introduce/service")
 
@@ -72,10 +75,12 @@ type metaData struct {
 }
 
 type record struct {
-	StateName string
+	StateName string `json:"state_name,omitempty"`
 	// WaitCount - how many introducees still need to approve the introduction proposal
 	// (initial value = introducee count, e.g. 2)
-	WaitCount int
+	WaitCount       int                     `json:"wait_count,omitempty"`
+	IntroduceeIndex int                     `json:"introducee_index,omitempty"`
+	Invitation      *didexchange.Invitation `json:"invitation,omitempty"`
 }
 
 // Service for introduce protocol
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -26,7 +26,7 @@ GOBIN_PATH=$(abspath .)/build/bin
 MOCKGEN = $(GOBIN_PATH)/mockgen
 
 .PHONY: all
-all: checks generate-openapi-spec unit-test bdd-test
+all: generate-openapi-spec checks unit-test bdd-test
```
fqutishat: why checks is moved

fqutishat: just add as first target

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -95,6 +100,9 @@ type Service struct {
 type Provider interface {
 	OutboundDispatcher() dispatcher.Outbound
 	StorageProvider() storage.Provider
+	// this method should be provided by the introduce client
+	// method should be implemented in didexchange Client
+	SendInvitation(inv *didexchange.Invitation, dest *service.Destination) error
```
rolsonquadras: The services are initialized during framework creation by passing the context (eg didExchange service [initialization](https://github.com/hyperledger/aries-framework-go/blob/539fbb4c2c29535d2b531997425aaa2cc95420c1/pkg/framework/aries/default.go#L77)). At that point, I'm not sure we would have this dependency.

rolsonquadras:
@soluchok https://github.com/hyperledger/aries-framework-go/pull/767#issuecomment-556091695 similar comment applies here.

pkg/didcomm/protocol/introduce/service.go
soluchok: @llorllale @troyronda @rolsonquadras  I am going to get it directly from the context, Are you ok with that?
```
svc, err := ctx.Service(didexchange.DIDExchange)
```

PR #823, opened 2019-11-15
feat: error handling in controller REST API

### LINKED ISSUE #756: Controller REST API : error codes
Implement error handling and return error codes for each REST API failures.

-  current implementation returns HTTP response 200 for business errors, may need to return different HTTP response codes for errors.
- current implementation following only generic error json  `{"code": "1", "message": "sample erro r message"}` . This may need to be updated as per requirements.

sudeshrshetty:
- different HTTP Status code for error scenarios
- generic error message body
- closes #756

### DIFF
--- a/pkg/restapi/errors/errors.go
+++ b/pkg/restapi/errors/errors.go
@@ -1,0 +1,75 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package errors
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+)
+
+var logger = log.New("aries-framework/rest")
+
+// genericError is aries rest api error response
+// swagger:response genericError
+type genericError struct {
+	// in: body
+	Code Code `json:"code"`
+	// in: body
+	Message string `json:"message"`
+}
+
+// Group is the error groups.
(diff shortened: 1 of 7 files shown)


sudeshrshetty:
Ignore codecov `pkg/restapi/operation/didexchange/didexchange.go` I replaced error handling lines which aren't covered since before. No easy way to reproduce error on json umarshal of request body.

pkg/restapi/errors/errors.go
```diff
@@ -0,0 +1,83 @@
...
+	UnknownStatus Code = iota
+
+	// InvalidRequest is typically a validation error for invalid requests
+	InvalidRequest
+
+	// CreateInvitationError is for failures in create invitation endpoint
+	CreateInvitationError
+
+	// ReceiveInvitationError is for failures in receive invitation endpoint
+	ReceiveInvitationError
```
rolsonquadras: These error codes are Protocol specific. Should we define these protocol specific error messages in common package (/pkg/restapi/errors) ?

sudeshrshetty: These are business errors, not at all protocol specific. There will be collision if we define them in separate packages.
I didn't export `genericError{}` indirectly forcing developers to use one of the `SendXXX()` functions to send errors.

sudeshrshetty: As discussed, I moved protocol errors to protocol specific implementations.
Introduced error Group and error code will be (Group + error code).
Example: DIDExchange will use `2000` series, Introduce will use `3000` series.

PR #824, opened 2019-11-18
refactor: creation of VC which extends base data model

### LINKED ISSUE #263: Selection of Verifiable Credential extensions based on the basic Credential
As a follow-up of #227, let's put to production code an algorithm of selection of Verifiable Credential extensions based on the basic Credential. 

The initial implementation idea can be found in **credential_extension_switch_test.go**, `CustomCredentialProducer` struct.

kdimak:
refactor: Selection of Verifiable Credential extensions based on the basic Credential

Simplified `NewCredential()` which decodes a base VC. Extensions can make extra decoding based on the VC and/or decoded JSON bytes returned.
Added `NewExtCredential()` which can be used when there are several VC with extended data model and we can build concrete extended VC depending on the e.g. `type` and `context` of the VC. 

closes #263

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -288,8 +288,6 @@ type CredentialTemplate func() *Credential
 type credentialOpts struct {
 	schemaDownloadClient   *http.Client
 	disabledCustomSchema   bool
-	decoders               []CredentialDecoder
-	template               CredentialTemplate
 	issuerPublicKeyFetcher PublicKeyFetcher
 }
 
@@ -309,20 +307,6 @@ func WithSchemaDownloadClient(client *http.Client) CredentialOpt {
 func WithNoCustomSchemaCheck() CredentialOpt {
 	return func(opts *credentialOpts) {
 		opts.disabledCustomSchema = true
-	}
-}
-
-// WithDecoders option is for adding extra JSON decoders into Verifiable Credential data model.
-func WithDecoders(decoders []CredentialDecoder) CredentialOpt {
-	return func(opts *credentialOpts) {
-		opts.decoders = append(opts.decoders, decoders...)
-	}
-}
-
-// WithTemplate option is for setting a custom factory method to create new Credential instance.
-func WithTemplate(template CredentialTemplate) CredentialOpt {
-	return func(opts *credentialOpts) {
-		opts.template = template
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -454,88 +438,115 @@ func decodeCredentialSchema(data []byte) ([]TypedID, error) {
...
 // NewCredential creates an instance of Verifiable Credential by reading a JSON document from bytes.
 // It also applies miscellaneous options like custom decoders or settings of schema validation.
-func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, error) {
-	// Apply options
-	crOpts := defaultCredentialOpts()
-
-	for _, opt := range opts {
-		opt(crOpts)
-	}
+// It returns decoded Credential and its marshalled JSON. They can be used for extensions
```
troyronda: It isn't completely obvious what the return values represent. I get that there is a Credential and marshalled JSON. It's not so clear in the comment how that differs from the input? We should make this text very clear as it is the documentation for the package.

pkg/doc/verifiable/credential.go
```diff
@@ -452,90 +436,121 @@ func decodeCredentialSchema(data []byte) ([]TypedID, error) {
...
-		err = decoder(vcDataDecoded, cred)
-		if err != nil {
-			return nil, err
+// ExtCredentialProducer is a factory for Credentials with extended data model.
+type ExtCredentialProducer interface {
+	Accept(vc *Credential) bool
+	Apply(vc *Credential, dataJSON []byte) (interface{}, error)
+}
+
+// NewExtCredential creates custom extended credentials from bytes which could be marshalled JSON or serialized JWT.
```
troyronda: This sounds like the same Input Bytes as NewCredential. This is confusing to the end developer. e.g.,

why does NewCredential need to return these bytes? are these byes different somehow?
why have a second new function?

pkg/doc/verifiable/credential.go
kdimak: Done! Could you check the new description, please?

pkg/doc/verifiable/credential.go
kdimak: I've changed the description of `NewCredential()`, there is the rationale why we return the bytes as second return value. Are you thinking about merging `NewCredential()` and `NewExtCredential` together? The reason I added these two functions is that in the second case we return `interface{}` just because we return the value which `ExtCredentialProducer::Apply()` returns (what is extended VC).

troyronda: - Yes I suspected that this new function was created so that an untyped interface could be returned.
- I wasn't entirely sure about the second return value. (e.g., is it different than the inputs, should it be a field vs a second return value, ...)
  - I was also wondering if `NewExtCredential` should take bytes or a *Credential.
- I also wonder if the name `NewExtCredential` is correct.

pkg/doc/verifiable/credential.go
```diff
@@ -452,90 +436,121 @@ func decodeCredentialSchema(data []byte) ([]TypedID, error) {
...
-
-	for _, opt := range opts {
-		opt(crOpts)
-	}
+// NewCredential decodes Verifiable Credential from bytes which could be marshalled JSON or serialized JWT.
+// It also applies miscellaneous options like settings of schema validation.
+// It returns decoded Credential and its marshalled JSON.
+// For JSON bytes input, the output marshalled JSON is the same value.
+// For serialized JWT input, the output is the result of decoding `vc` claim from JWT.
+// The output Credential and marshalled JSON can be used for extensions of the base data model
```
troyronda: At first, I thought that this second return value might be used as an input to NewExtCredential. But that doesn't make sense since NewExtCredential calls NewCredential itself.

So I think we could potentially create confusion with the current pattern.

talwinder50: I agree it is bit confusing

pkg/doc/verifiable/credential.go
haardikk21: - If the second return value is only used for extension and/or unmarshalling to custom structures, I think `NewCredential` should never have been called and rather `NewExtCredential` should've been used. I don't see why the business logic should have the option of reinventing the wheel in this scenario.

- @troyronda `NewExtCredential` doesn't return an `ExtCredential` type because it can't have a typed struct encompassing all possibilities. It does, however, return an "extended credential" through the interface. I would however change this to `NewExtendedCredential` instead as more verbose naming, in this case, would be better for readability.

kdimak: - to extend NewCredential(), `decoders` and `templates` were passed previously as `credentialOpts`. The `decoder` defined extra decoding actions. The `template` allowed to define a factory method of `Credential`.
- I decided to simplify that and make the extension of `Credential` an explicit two-step. The first step is to decode base `Credential`. The second one is to make extra decodings if needed. I found useful to return 2 results of decoding at `NewCredential()` - the `Credential` struct and `[]byte`. For extension, it's really easy to unmarshal the rest of the fields (which are missing at base `Credential`) from that `[]byte` rather than check `ExtraFields` map of `Credential`. Note that return `[]byte` is the same as input `[]byte` (of `NewCredential()`) for JSON case, but is a result of parsing and refinement in the case of JWT.
- `NewExtCredential()`  was added as actually the goal of the task. This pattern was added at `credential_extension_switch_test.go` initially, to show the case when we can (e.g. as **Holder** side) received credentials of several different kinds and want to apply appropriate extension based on e.g. `context` and `type`. I agree that we shouldn't use `New...()` name, instead use smth like `Create...()`. I don't know how to name this better, maybe `CreateCustomCredential()` will be more clear for the user.

Windforcer: Same thoughts as Haardik, the very minor suggestions. Overall it looks good, and these features will allow us to progress with our OTBS demo using VC's

kdimak: I think we should have 2 public functions `NewCredential()` and `CreateCustomCredential()`.
- `NewCredential()` can be used when we don't want to introduce extra structures for extensions and instead want to have an output of explicit type (`Credential`) and get access to all custom fields at `ExtraFields` map of `Credential`.
- `CreateCustomCredential()` can be used for explicit extensions. A special use case is when  we expect to receive two or more different extension types.

pkg/doc/verifiable/credential.go
```diff
@@ -452,90 +436,127 @@ func decodeCredentialSchema(data []byte) ([]TypedID, error) {
...
+
+// CreateCustomCredential creates custom extended credentials from bytes which could be marshalled JSON
+// or serialized JWT. It decodes input bytes to the base Verifiable Credential using NewCredential().
+// It then checks all producers to find the appropriate which is capable of building extended Credential data model.
+// If none of producers accept the credential, the base credential is returned.
+func CreateCustomCredential(
+	vcData []byte,
+	producers []CustomCredentialProducer,
+	opts ...CredentialOpt) (interface{}, error) {
+	vcBase, vcBytes, credErr := NewCredential(vcData, opts...)
```
troyronda: I guess you are thinking the second return value is useful beyond its usage in CreateCustomCredential?

kdimak: I think - yes, because in case of VC data model extension we can either use `ExtraFields` of `verifiable.Credential` or make unmarshaling of bytes to a custom structure that defines extra fields. An example can be found [here](https://github.com/kdimak/aries-framework-go/blob/5fa6c7f7402094770ac7597021b1da58f26e86b1/pkg/doc/verifiable/credential_extension_test.go#L40).
But if I'm alone in this opinion, I can revert and return just `Credential`  (without bytes).

troyronda: I think we should have another TODO to rename ExtraFields.

PR #825, opened 2019-11-18
refactor: Integrate did-communication service conventions

### LINKED ISSUE #822: refactor: Integrate did-communication service conventions
We are currently picking first key of type Ed25519VerificationKey2018 from did document public keys for:
1) sender verification key
2) recipient keys in destination 

Refactor current logic to read above mentioned keys from recipient keys in did-communication service for both peer and public did documents.

See following spec:
https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions

sandrask:
We are currently picking first key of type Ed25519VerificationKey2018 from did document public keys for:
-- sender verification key
-- recipient keys in destination
Refactor current logic to read above mentioned keys from recipient keys in did-communication service for both peer and public did documents.

Closes #822

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service_test.go
+++ b/pkg/didcomm/protocol/didexchange/service_test.go
@@ -549,15 +549,26 @@ func getMockDID() *did.Doc {
 		Service: []did.Service{
 			{
 				ServiceEndpoint: "https://localhost:8090",
+				Type:            "did-communication",
+				Priority:        0,
+				RecipientKeys:   []string{"did:example:123456789abcdefghi#keys-2"},
+			},
+			{
+				ServiceEndpoint: "https://localhost:8090",
+				Type:            "did-communication",
+				Priority:        1,
+				RecipientKeys:   []string{"did:example:123456789abcdefghi#keys-1"},
 			},
 		},
-		PublicKey: []did.PublicKey{{
-			ID:         "did:example:123456789abcdefghi#keys-1",
-			Controller: "did:example:123456789abcdefghi",
-			Type:       "Secp256k1VerificationKey2018",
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -428,52 +431,87 @@ func (ctx *context) getDestinationFromDID(id string) (*service.Destination, erro
...
+	for i, s := range didDoc.Service {
+		if s.Type == didCommServiceType {
+			if index == notFound || didDoc.Service[index].Priority > s.Priority {
+				index = i
+			}
+		}
+	}
+
+	if index == notFound {
+		return nil, fmt.Errorf("service not found in DID document: %s", didCommServiceType)
```
troyronda: Is error the correct behaviour if the service endpoint isn't defined in the DID document?

sandrask: How do we pick recipient keys and sender verification key then?

sandrask: I double-checked with @llorllale : As per did-comm convention did document has to have "did-communication" service.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -549,16 +549,27 @@ func getMockDID() *did.Doc {
 		Service: []did.Service{
 			{
```
troyronda: TODO: What shall we do with the ;did-communication suffix?
Example from the Indy version: https://github.com/hyperledger/aries-rfcs/issues/317#issuecomment-556257803

sandrask: Ok, created issue #842 to address this.

PR #829, opened 2019-11-18
refactor: key structure

### LINKED ISSUE #596: Refactor Key structure in KMS (from KeyPair to KeySet)
part of #494 

currently keys are stored in the current (old) `wallet` as a keypair with public and private keys stored as is in the wallet storage.

This key storage struct must change to a more flexile format that allows a pluggable (new) KMS system to work with any pluggable storage system.

We need to move away from direct key referencing format in the storage by using the following structure:
```
type KeySet struct{
	ID string // reference ID of KeySet
	Keys []SomeKeyStruct // list of sub keys in this set
	PrimaryKey SomeKeyStruct // the primary key in this key set (usually a signing key)
}
```
and
```
SomeKeyStruct{
	ID struct // reference ID of the key struct
	Pub string // 
	Priv string
	Alg KeyAlgorithm
	Capability KeyCapability
}
```

where KeyAlgorithm is a predefined type (pre-typed as a string) representing the algorithm used for the key (Ed25519, Curv25519, etc.)
and KeyCapability is a predefined type (iota) representing the key capability (ENCRYPTION, SIGNING)

Implementation specific logic of storing and retrieving a KeySet:
KMS should be able to store minimal information in the metadata storage about `KeySet` like the Keys IDs and the primary Key ID only (no keys to be stored in the metadata store) and each key will be stored separately in the secret storage.

and when retrieving a KeySet the Unmarshal function should rebuild the full structure as defined above and fetching the keys from the secret store per ID.

> And it's important that the identifier be derived deterministically from the public key, since it's the public keys that will be packed into an envelope header, which the recipient will need to look up in the KMS metadata to find a usable decryption key.

part of key validation refactoring:
> *A particular crypto implementation will support some set of algorithms.
> *From the available crypto implementation, a particular KMS implementation will support some set of algorithms.
> *The framework will get it's supported list from Crypto and KMS. Framework should not care about algorithm IDs - it's simply a string.
> *A particular package that uses KMS might care to only allow a subset of what's available. (or fail if nothing it can use is available.)

Baha-sk:
This change refactors the KeyPair structure into a SimpleKey
	structure that has information about only 1 key.
	Related Signing and Encryption keys are stored in a KeySet.

	This change also introduces a metadatastore in the KMS.
	The keystore now has SimpleyKey instances that contain raw
	key values while the metadatastore has key relationship
	information stored in a KeySet.

	Prior to storing a KeySet into metadatastore, it is stripped
	from all the raw key information. Only SimpleKey.ID info is
	persisted in a KeySet when storing in a metadatastore.

	The KMS will fetch the right keys by ID.

	The Packer/Crypter interface remain unchanged to allow
	retrieving keys from the envelope. The KMS will fetch the
	corresponding keys related to the ones set in the envelope.

close #596


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/google/uuid v1.1.1
 	github.com/gorilla/mux v1.7.3
 	github.com/kr/pretty v0.1.0 // indirect
+	github.com/mr-tron/base58 v1.1.2
```
troyronda: what is this new import?

pkg/internal/cryptoutil/utils.go
```diff
@@ -66,79 +74,66 @@ func IsChachaKeyValid(key []byte) bool {
...
-
-// isKeyPairValid is a utility function that validates a KeyPair
-func isKeyPairValid(kp KeyPair) bool {
-	if kp.Priv == nil || kp.Pub == nil {
-		return false
-	}
-
-	return true
+// SimpleKey represents a key with an ID and a value
+type SimpleKey struct {
```
troyronda: what does the prefix Simple imply?

pkg/kms/crypto_box.go
```diff
@@ -133,7 +151,9 @@ func (b *CryptoBox) SealOpen(cipherText, myPub []byte) ([]byte, error) {
 		return nil, errors.New("message too short")
 	}
 
-	kp, err := b.km.getKeyPairSet(base58.Encode(myPub))
+	keySetID := hashKeySetID(base58.Encode(myPub))
```
troyronda: the detail of encoding format should be hidden from the rest of the code. 

we should not see so many calls to base58.encode.
rather we should see some variety of a KeyID function that hides this detail.

go.mod
Baha-sk: this must've been added by the IDE without me noticing it.. will fix

pkg/internal/cryptoutil/utils.go
Baha-sk: I will rename to Key... for some tests.. initially I didn't want to confuse it with key variable..I'll push another commit soon..

pkg/kms/crypto_box.go
Baha-sk: that function question is hashKeyID() and hashKeySetID() ... it's not exposed as it should be only called within the kms package..

the encoding is an issue as sometimes we use key.Value which is already base58 encoded... if I move base58.Encode inside the ID function then everywhere we get the key from key.Value we will need to do base58.Decode()...  it may make more sense since Value is encoded but will end up doing Decode(Value) and then inside hashKeyID/hashKeySetID() we will do Encode(key) .. so we're duplicating instruction unnecessarily... this is why the encoding is kept for the caller to make.. again, this code is internal to kms and should not be exposed to the rest of the code.

troyronda: Even when this is internal functionality, we should still hide this detail.

Baha-sk: ok... will assume the base key argument is in raw bytes and do encoding inside the functions

Baha-sk: after refactoring, I noticed that the Packer interface expects base58 encoded keys, this clashes with hashKeyID refactoring of base58 keys to raw bytes keys, we can do an update of the interface and hashKey() in a later pr to expect raw bytes instead of base58 encoded key across the board.

For now, the expectation of verification keys stored in DID documents is to be base58 encoded, so public keys are expected to be encoded between calls in the framework. So it's safe to assume keys passed in arguments to be pre encoded.

Baha-sk: updated hashKeyID to take raw bytes keys.. I'll update hashKeySetID as well..

Moopli: > I noticed that the Packer interface expects base58 encoded keys, this clashes with hashKeyID refactoring of base58 keys to raw bytes keys, we can do an update of the interface and hashKey() in a later pr to expect raw bytes instead of base58 encoded key across the board.

The Packer interface expects raw byte keys, the Packager interface currently converts its base58 input to raw bytes for the Packer it invokes. It would simplify things to change the Packager's methods to take raw byte keys, and skip the conversion step.

Baha-sk: yes I'm aware of that... the rest of the code interacts with Packager while Packer is shielded by Packager (now the names are more confusing as both mean the same thing)

we can change this in a separate PR as this one is already too big

pkg/internal/cryptoutil/utils.go
Baha-sk: done

pkg/kms/crypto_box.go
Baha-sk: done.. hashKeySetID now accepts raw key bytes instead of the base58 encoding of the key.

There are still a few places where the KMS API either returns base58 encoded keys like CreateKeySet() or accept base58 encoded key argument like SignMessage(). 

I have not changed these API functions as they have a wider impact on code outside of KMS. This can be done in a separate PR (if we choose to do it).

troyronda:
Will reopen later.

PR #836, opened 2019-11-19
feat: DIDComm Outbound Transport - Support for WebSocket

### LINKED ISSUE #827: DIDComm Outbound Transport - Support for WebSocket
Aries framework Go should support WebSocket as one of the DIDComm transport. - https://github.com/hyperledger/aries-rfcs/tree/master/features/0025-didcomm-transports#websocket.

rolsonquadras:
Closes #827 

### DIFF
--- a/pkg/didcomm/transport/ws/outbound.go
+++ b/pkg/didcomm/transport/ws/outbound.go
@@ -1,0 +1,66 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ws
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/gorilla/websocket"
+)
+
+const webSocketScheme = "ws"
+
+// OutboundClient websocket outbound.
+type OutboundClient struct {
+}
+
+// NewOutbound creates a client for Outbound WS transport.
+func NewOutbound() *OutboundClient {
+	return &OutboundClient{}
+}
+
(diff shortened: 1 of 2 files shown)


pkg/didcomm/transport/ws/outbound.go
```diff
@@ -0,0 +1,59 @@
...
+	return &OutboundClient{}
+}
+
+// Send sends a2a data via WS.
+func (cs *OutboundClient) Send(data []byte, url string) (string, error) {
+	if url == "" {
+		return "", errors.New("url is mandatory")
+	}
+
+	client, _, err := websocket.DefaultDialer.Dial(url, nil)
```
troyronda: dial on every Send?

rolsonquadras: https://github.com/hyperledger/aries-rfcs/tree/master/features/0025-didcomm-transports#websocket "Websockets are considered transmit only by default. Messages will only flow from the agent that opened the socket." -- this is what the RFC says.

For WASM requirement, we have to figure out a way to keep connections in a pool or something when they connect to the server (inbound) and can be used to send/push the messages without creating a socket (also, wouldn't be possible to dial as wasm will not have an endpoint).

troyronda: Yes. Probably a TTL scheme. I guess a TODO :).

rolsonquadras: I didn't add TODO as it's a separate feature and might need to revisit this based on the solution.

PR #846, opened 2019-11-21
feat: Create Public DID (VDRI & REST support)

### LINKED ISSUE #845: VDRI support & REST endpoint for creating public DIDs
Add create DID feature in VDRI http binding and also add REST endpoint to create public DIDs

sudeshrshetty:
- add rest endpoint `/vdri/create-public-did` to created public DIDs
using VDRI of agent
- VDRI Build() for httpbinding to create public DIDs using custom request
builders
- `/vdri/create-public-did` to accept optional `request header` parameter to add custom headers to 
request being sent to http binding endpoint.
- closes #845

### DIFF
--- a/pkg/framework/aries/api/vdri/vdri.go
+++ b/pkg/framework/aries/api/vdri/vdri.go
@@ -8,6 +8,7 @@ package vdri
 
 import (
 	"errors"
+	"io"
 	"time"
 
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
@@ -15,6 +16,9 @@ import (
 
 // ErrNotFound is returned when a DID resolver does not find the DID.
 var ErrNotFound = errors.New("DID not found")
+
+// DIDCommServiceType default DID Communication service endpoint type
+const DIDCommServiceType = "did-communication"
 
 // Registry vdri registry
 type Registry interface {
@@ -87,6 +91,7 @@ type CreateDIDOpts struct {
 	ServiceType     string
 	KeyType         string
 	ServiceEndpoint string
+	RequestBuilder  func([]byte) (io.Reader, error)
 }
 
 // DocOpts is a create DID option
@@ -110,6 +115,14 @@ func WithKeyType(keyType string) DocOpts {
 func WithServiceEndpoint(serviceEndpoint string) DocOpts {
 	return func(opts *CreateDIDOpts) {
 		opts.ServiceEndpoint = serviceEndpoint
+	}
+}
+
+// WithRequestBuilder allows to supply request builder
+// which can be used to add headers to request stream to be sent to HTTP binding URL
+func WithRequestBuilder(builder func(payload []byte) (io.Reader, error)) DocOpts {
+	return func(opts *CreateDIDOpts) {
+		opts.RequestBuilder = builder
 	}
 }
 
--- a/pkg/restapi/operation/common/models.go
+++ b/pkg/restapi/operation/common/models.go
@@ -1,0 +1,44 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package common
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+// CreatePublicDIDRequest model
+//
+// This is used for operation to create public DID
+//
+// swagger:parameters createPublicDID
+type CreatePublicDIDRequest struct {
+	// Params for creating public DID
+	//
+	// in: path
+	*CreatePublicDIDParams
+}
+
+// CreatePublicDIDParams contains parameters for creating new public DID
+type CreatePublicDIDParams struct {
+	// Params for creating public DID
+	Method string `json:"method"`
+
+	// RequestHeader to be included while submitting request to http binding URL
+	RequestHeader string `json:"header"`
+}
+
+// CreatePublicDIDResponse model
+//
+// This is used for returning public DID created
+//
+// swagger:response createPublicDIDResponse
+type CreatePublicDIDResponse struct {
+
+	// in: body
+	// TODO return base64-encoded raw bytes of the DID doc [Issue: #855]
+	DID *did.Doc `json:"did"`
+}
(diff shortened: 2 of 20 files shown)


troyronda:
This doesn’t seem like general path but rather vdri?

pkg/restapi/operation/common/operation.go
```diff
@@ -0,0 +1,173 @@
...
+	b, err := json.Marshal(args)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(b, v)
+}
+
+// buildSideTreeRequest request builder for sidetree public DID creation
+func buildSideTreeRequest(didDoc *did.Doc) (io.Reader, error) {
```
fqutishat: aries code should not know about sidetree

sudeshrshetty:
> This doesn’t seem like general path but rather vdri?
> 
> Is this an endpoint in aca-py?

Updated, aca-py doesn't have it.

pkg/restapi/operation/common/operation.go
sudeshrshetty: Removed, I just added feature to inject optional header to http binding request. Using that header to inject side tree specific minimal headers, and it is working. 
The current create request in REST API follows `{header:xxy, payload:xxz}` pattern. We can modify it anytime in future to add support to other public DIDs.

pkg/framework/aries/api/vdri/vdri.go
```diff
@@ -113,6 +118,14 @@ func WithServiceEndpoint(serviceEndpoint string) DocOpts {
 	}
 }
 
+// WithRequestBuilder allows to supply request builder
+// which can be used to add headers to request stream to be sent to HTTP binding URL
+func WithRequestBuilder(builder func(payload []byte) (io.Reader, error)) DocOpts {
+	return func(opts *CreateDIDOpts) {
+		opts.RequestBuilder = builder
+	}
+}
```
llorllale: I'm wondering if this option is even needed at all. I would have assumed the specific VDRI implementation would take care of building the request and serializing whatever needs to be serialized. The VDRI implementation for the DID method can be picked with the `VDRI.Accept(method string)` method.

pkg/vdri/httpbinding/vdri.go
```diff
@@ -57,12 +65,103 @@ func (v *VDRI) Accept(method string) bool {
...
+	resDoc, err := v.sendCreateRequest(reqBody)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send create DID request: %s", err)
+	}
+
+	return resDoc, nil
+}
+
+func (v *VDRI) sendCreateRequest(req io.Reader) (*did.Doc, error) {
+	httpReq, err := http.NewRequest(http.MethodPost, v.endpointURL, req)
```
llorllale: please leave a TODO for configuring timeouts properly on these external calls

pkg/restapi/operation/common/models.go
```diff
@@ -0,0 +1,43 @@
...
+
+// CreatePublicDIDResponse model
+//
+// This is used for returning public DID created
+//
+// swagger:response createPublicDIDResponse
+type CreatePublicDIDResponse struct {
+
+	// in: body
+	DID *did.Doc `json:"did"`
```
llorllale: I believe it's better to return the base64-encoded raw bytes of the DID doc for the reasons laid out in hyperledger/aries-rfcs#243

pkg/vdri/httpbinding/vdri.go
```diff
@@ -57,12 +65,103 @@ func (v *VDRI) Accept(method string) bool {
...
+	}
+
+	var reqBody io.Reader
+	if docOpts.RequestBuilder != nil {
+		reqBody, err = docOpts.RequestBuilder(docBytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build request : %s", err)
+		}
+	} else {
+		bytes.NewReader(docBytes)
```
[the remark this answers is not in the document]
sudeshrshetty: done

pkg/restapi/operation/common/operation.go
```diff
@@ -0,0 +1,160 @@
...
+		return
+	}
+
+	if request.CreatePublicDIDParams == nil || request.Method == "" {
+		errors.SendHTTPBadRequest(rw, InvalidRequestErrorCode, fmt.Errorf("invalid method name"))
+		return
+	}
+
+	logger.Debugf("creating public DID for method[%s]", request.Method)
+	fmt.Println("creating public DID for method[]", request.Method)
```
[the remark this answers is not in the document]
sudeshrshetty: done

pkg/framework/aries/api/vdri/vdri.go
sudeshrshetty: we thought about the same when we started doing this change. SDK has feature of injecting custom VDRI implementation through opts/context. It cannot be done with controller APIs.

But this option can be used to modify request being sent to http binding URL.  REST API use this option to add http binding endpoint specific headers.

pkg/vdri/httpbinding/vdri.go
sudeshrshetty: Agreed, will be addressed in #855 , added a TODO

pkg/restapi/operation/common/models.go
sudeshrshetty: Agreed, will be addressed in #855 , added a TODO

pkg/framework/aries/api/vdri/vdri.go
llorllale: @sudeshrshetty 

> It cannot be done with controller APIs.

Sounds like some kind of refactor is needed to support this use case?

sudeshrshetty: This is what this PR is about. HTTP Binding implementation creates DID with minimal features like public keys and service endpoint (from current agent) and submits it to http binding endpoint URL. This option will help us to add custom header, in long run we have to go as you suggested.

This PR is to unblock other public DID related features in controller API

sudeshrshetty: Will be handled as part of next PR for #860

Issue #847, opened 2019-11-21
Improve Verifiable Credential extension via type embedding 


kdimak:
Extension of VC via type embedding is quite natural and easy to write in Go. When writing [examples of VC extension](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/doc/verifiable/credential.go), the following problems were detected:

- we should override `MarshalJSON()` of extended structure, otherwise `MarshalJSON()` of `verifiable.Credential` is invoked which serializes only base VC and does not consider the extra fields of the extension.

- when extended VC is marshaled to JWS and decoded back, extra fields are missed.

These limitations make the extension approach using `ExtraFields` map the only properly working option, while it seems to be less comfortable to use compared to the type embedding approach.

PR #848, opened 2019-11-21
test: Construct Verifiable Credential object (encoding side)

### LINKED ISSUE #831: Construct Verifiable Credential object (encoding side)
We need to add examples of VC encoding flow when we construct a verifiable credential object and then serialize it to JWS. To consider the case of VC with an extended data model.

Simplified flow:
```
	vc := &Credential{
		// initialize 
	}
	jwtCredClaims, err := vc.JWTClaims(true)
	jws, err := jwtCredClaims.MarshalJWS(RS256, privateKey, "key id")
```

kdimak:
Here we add examples of VC encoding. VC base model extension is shown by 2 ways:
- struct embedding
- using `ExtraFields` map

Struct embedding does not work properly for now (#847 is created), so it's recommended to use `ExtraFields` map for extensions.

closes #831

### DIFF
--- a/pkg/doc/verifiable/common.go
+++ b/pkg/doc/verifiable/common.go
@@ -53,17 +53,17 @@ func SingleKey(pubKey interface{}) PublicKeyFetcher {
 // Proof defines embedded proof of Verifiable Credential
 type Proof interface{}
 
-// ExtraFields is a map of extra fields of struct build when unmarshalling JSON which are not
+// CustomFields is a map of extra fields of struct build when unmarshalling JSON which are not
 // mapped to the struct fields.
-type ExtraFields map[string]interface{}
+type CustomFields map[string]interface{}
 
 // TypedID defines a flexible structure with id and name fields and arbitrary extra fields
-// kept in ExtraFields.
+// kept in CustomFields.
 type TypedID struct {
 	ID   string `json:"id,omitempty"`
 	Type string `json:"type,omitempty"`
 
-	ExtraFields `json:"-"`
+	CustomFields `json:"-"`
 }
 
 // MarshalJSON defines custom marshalling of TypedID to JSON.
@@ -73,7 +73,7 @@ func (tid *TypedID) MarshalJSON() ([]byte, error) {
 
 	alias := (*Alias)(tid)
 
-	data, err := marshalWithExtraFields(alias, tid.ExtraFields)
+	data, err := marshalWithCustomFields(alias, tid.CustomFields)
 	if err != nil {
 		return nil, fmt.Errorf("marshal TypedID: %w", err)
 	}
@@ -88,9 +88,9 @@ func (tid *TypedID) UnmarshalJSON(data []byte) error {
 
 	alias := (*Alias)(tid)
 
-	tid.ExtraFields = make(ExtraFields)
+	tid.CustomFields = make(CustomFields)
 
-	err := unmarshalWithExtraFields(data, alias, tid.ExtraFields)
+	err := unmarshalWithCustomFields(data, alias, tid.CustomFields)
 	if err != nil {
 		return fmt.Errorf("unmarshal TypedID: %w", err)
 	}
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -217,7 +217,7 @@ type Credential struct {
 	TermsOfUse     []TypedID
 	RefreshService *TypedID
 
-	ExtraFields ExtraFields
+	CustomFields CustomFields
 }
 
 // rawCredential is a basic verifiable credential
@@ -237,7 +237,7 @@ type rawCredential struct {
 	RefreshService *TypedID    `json:"refreshService,omitempty"`
 
 	// All unmapped fields are put here.
-	ExtraFields `json:"-"`
+	CustomFields `json:"-"`
 }
 
 // MarshalJSON defines custom marshalling of rawCredential to JSON.
@@ -246,7 +246,7 @@ func (rc *rawCredential) MarshalJSON() ([]byte, error) {
 
 	alias := (*Alias)(rc)
 
-	return marshalWithExtraFields(alias, rc.ExtraFields)
+	return marshalWithCustomFields(alias, rc.CustomFields)
 }
 
 // UnmarshalJSON defines custom unmarshalling of rawCredential from JSON.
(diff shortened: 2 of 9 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -674,9 +674,11 @@ func (vc *Credential) JWTClaims(minimizeVC bool) (*JWTCredClaims, error) {
 	return newJWTCredClaims(vc, minimizeVC)
 }
 
-// SubjectID gets ID of single subject if present or
+// subjectID gets ID of single subject if present or
 // returns error if there are several subjects or one without ID defined
-func (vc *Credential) SubjectID() (string, error) {
+// subjectID gets ID of single subject if present or
+// returns error if there are several subjects or one without ID defined
+func subjectID(subject interface{}) (string, error) {
```
kdimak: I found the issue with `subjectID()` as it did not support custom structures case. So I had to fix this in this PR to enable the VC encoding example with custom `UniversityCredentialSubject` subject.

haardikk21:
LGTM :+1:

troyronda:
Do we have a better name for "Extra Fields" @haardikk21 @kdimak @Windforcer

pkg/doc/verifiable/example_credential_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	cp.ExtraFields = map[string]interface{}{
+		"referenceNumber": udc.ReferenceNumber,
+	}
+
+	return json.Marshal(&cp)
+}
+
+//nolint:gochecknoglobals
+var privKey = ed25519.PrivateKey{56, 237, 176, 143, 247, 162, 167, 111, 85, 161, 158, 14, 243, 173, 144, 51, 157, 109, 155, 228, 77, 170, 238, 85, 220, 144, 158, 51, 14, 40, 153, 141, 193, 179, 12, 234, 125, 193, 60, 56, 198, 150, 80, 93, 30, 58, 14, 152, 205, 6, 50, 98, 125, 212, 65, 17, 15, 11, 230, 3, 226, 187, 7, 89} //nolint:lll
```
talwinder50: Why we are creating the priv key this way ?

pkg/doc/verifiable/credential.go
```diff
@@ -674,9 +674,11 @@ func (vc *Credential) JWTClaims(minimizeVC bool) (*JWTCredClaims, error) {
 	return newJWTCredClaims(vc, minimizeVC)
 }
 
-// SubjectID gets ID of single subject if present or
+// subjectID gets ID of single subject if present or
 // returns error if there are several subjects or one without ID defined
-func (vc *Credential) SubjectID() (string, error) {
+// subjectID gets ID of single subject if present or
```
talwinder50: duplication of line , subjectID gets ID of single subject if present

pkg/doc/verifiable/example_credential_test.go
kdimak: If we generate the key, we will get different JWS each time we are running the example. And we are checking the JWS output (2-nd **Outcome**). That's why I made the keys static.

pkg/doc/verifiable/credential.go
kdimak: Ah, I will fix that! thanks!

kdimak:
> Do we have a better name for `ExtraFields` @haardikk21 @kdimak @Windforcer

Does `CustomFields`sound better?

kdimak:
I don't know how to reach 85% of patch diff (now it's about 82%)... 
I've changed `subjectID()` in prod code to fix the issue which blocked the successful run of the example, added a test case, but cannot cover all the lines,

haardikk21:
> > Do we have a better name for `ExtraFields` @haardikk21 @kdimak @Windforcer
> 
> Does `CustomFields`sound better?

Yeah `CustomFields` seems more in line with the new structure.

troyronda:
when I read the spec, I would understand this `ExtraFields` to be extension claims (and claims also seems to called properties once in the extensions section). Is this correct?

kdimak:
> when I read the spec, I would understand this `ExtraFields` to be extension claims (and claims also seems to called properties once in the extensions section). Is this correct?

At [Extensibility section](https://www.w3.org/TR/vc-data-model/#extensibility), they use `new terms` for 2 new fields.  I didn't find any proper term in the spec, that's why we are trying to find the clearest term ourselves. 
For me, "claim" is a whole thing about VC (e.g. "Issuer is a role an entity performs by asserting claims about one or more subjects") or specific JWT term.

kdimak:
I changed `ExtraFields` to `CustomFields`. Also found a way how to make the patch diff happy :)

troyronda:
Is there a followup test demonstrating signing of the created credential?

pkg/doc/verifiable/common.go
```diff
@@ -73,7 +73,7 @@ func (tid *TypedID) MarshalJSON() ([]byte, error) {
 
 	alias := (*Alias)(tid)
```
haardikk21: We should change the function name to match `CustomFields`

pkg/doc/verifiable/common.go
```diff
@@ -88,9 +88,9 @@ func (tid *TypedID) UnmarshalJSON(data []byte) error {
...
 
-	tid.ExtraFields = make(ExtraFields)
+	tid.CustomFields = make(CustomFields)
 
-	err := unmarshalWithExtraFields(data, alias, tid.ExtraFields)
+	err := unmarshalWithExtraFields(data, alias, tid.CustomFields)
```
haardikk21: Same thing for this function.

pkg/doc/verifiable/common.go
```diff
@@ -73,7 +73,7 @@ func (tid *TypedID) MarshalJSON() ([]byte, error) {
 
 	alias := (*Alias)(tid)
 
-	data, err := marshalWithExtraFields(alias, tid.ExtraFields)
+	data, err := marshalWithExtraFields(alias, tid.CustomFields)
```
haardikk21: We should change this to match `CustomFields`

kdimak:
> Is there a followup test demonstrating signing of the created credential?

yes, just created #856  covering this demand.

pkg/doc/verifiable/common.go
kdimak: I was thinking about this renaming and came to the conclusion that `marshalWithExtraFields()` reflects better the purpose of the function - marshall a struct with adding extra fields to the JSON.

haardikk21: If we're replacing the term "extra fields" with "custom fields" everywhere, I think this should be renamed for consistency reasons. It's also equally valid t say the function Marshalls a struct and adds custom fields to the JSON.

kdimak: Completed the renaming! Now it's more consistent, thanks!

PR #872, opened 2019-11-22
feat: public DIDs in Accept Invitation & Accept Request

### LINKED ISSUE #871: DID Exchange Accept Invitation/Request: Controller & Client to accept public DIDs
Add public DID field in accept invitation and accept exchange request.

sudeshrshetty:
For  `Accept Exchange Request` and ` Accept Invitation`,
- added public DID in controller API for below endpoints
- added public DID and label for client & service
- closes #871

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service_test.go
+++ b/pkg/didcomm/protocol/didexchange/service_test.go
@@ -958,7 +958,67 @@ func TestAcceptExchangeRequest(t *testing.T) {
 		for e := range actionCh {
 			prop, ok := e.Properties.(event)
 			require.True(t, ok, "Failed to cast the event properties to service.Event")
-			require.NoError(t, svc.AcceptExchangeRequest(prop.ConnectionID()))
+			require.NoError(t, svc.AcceptExchangeRequest(prop.ConnectionID(), "", ""))
+		}
+	}()
+
+	statusCh := make(chan service.StateMsg, 10)
+	err = svc.RegisterMsgEvent(statusCh)
+	require.NoError(t, err)
+
+	done := make(chan struct{})
+
+	go func() {
+		for e := range statusCh {
+			if e.Type == service.PostState && e.StateID == stateNameResponded {
+				done <- struct{}{}
+			}
+		}
+	}()
+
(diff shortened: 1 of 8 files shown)


sudeshrshetty:
Please ignore the codecov, I have covered all scenarios except unmarshal failures of requests which is hard to cover.

pkg/internal/mock/didcomm/protocol/mock_didexchange.go
```diff
@@ -30,6 +30,7 @@ type MockDIDExchangeSvc struct {
 	UnregisterActionEventErr error
 	RegisterMsgEventErr      error
 	UnregisterMsgEventErr    error
+	AcceptError              error
```
rolsonquadras: Separating AcceptError based on exchange type might help to have more control on the mock.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -1118,7 +1118,7 @@ func TestEventTransientData(t *testing.T) {
 		svc, err := New(&protocol.MockProvider{})
 		require.NoError(t, err)
 
-		err = svc.AcceptExchangeRequest(generateRandomID())
+		err = svc.AcceptExchangeRequest(generateRandomID(), "", "")
```
rolsonquadras: All test cases here pass empty values. Can we verify non-empty cases as well ?

pkg/internal/mock/didcomm/protocol/mock_didexchange.go
sudeshrshetty: But we don't need it now for current test cases, I added general mock error for all accept cases together since we test exchange/invitation separately (didn't want to have one extra field in private function of creating test handler which already has too many args).

rolsonquadras: Ok. I was thinking about the cases where we want one accept to succeed and other fail. We might need to refactor this and other usages/consumers of this mock.

pkg/didcomm/protocol/didexchange/service_test.go
sudeshrshetty: done

PR #875, opened 2019-11-25
WIP feat: issue-credential service skeleton

talwinder50:

### DIFF
--- a/pkg/didcomm/common/model/attachement.go
+++ b/pkg/didcomm/common/model/attachement.go
@@ -1,0 +1,18 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package model
+
+// Attachment defines the struct for request and issue in the protocol
+type Attachment struct {
+	ID       string `json:"@id,omitempty"`
+	MimeType string `json:"mime-type,omitempty"`
+	Data     Data   `json:"data,omitempty"`
+}
+
+// Data defines the type of the data to be part of attachment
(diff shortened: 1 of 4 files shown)


DRK3:
Typo in "skeleton" in the title

pkg/didcomm/common/model/attachement.go
```diff
@@ -0,0 +1,18 @@
+/*
```
DRK3: Typo in file name - should be "attachment"

pkg/didcomm/protocol/issuecredential/service_test.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/google/uuid"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
```
soluchok: extra lines

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -0,0 +1,140 @@
...
+		return true
+	}
+
+	return false
+}
+
+// stateFromMsgType returns the state by given name.
+func stateFromMsgType(msgType string) (string, error) {
+	var s string
+	switch msgType {
```
soluchok: function says `returns the state by the given name` but actually it returns state name, probably it should be renamed to `stateNameFromMsgType` or return actual state instead of the state name

talwinder50:
Not required right now

PR #880, opened 2019-11-25
refactor: Pass SenderVerKey to inbound transport from Unpacker along with msg

### LINKED ISSUE #876: Inbound Transport handler - Access to Sender Verification Key
The inbound transport needs access to the sender verification key for Transport Route Return protocol. 

Required for implementing #858.

rolsonquadras:
Closes #876 

### DIFF
--- a/pkg/didcomm/packer/jwe/authcrypt/unpack.go
+++ b/pkg/didcomm/packer/jwe/authcrypt/unpack.go
@@ -22,22 +22,22 @@ import (
 // encrypted CEK.
 // The current recipient is the one with the sender's encrypted key that successfully
 // decrypts with recipientKeyPair.Priv Key.
-func (p *Packer) Unpack(envelope []byte) ([]byte, error) {
+func (p *Packer) Unpack(envelope []byte) ([]byte, []byte, error) {
 	jwe := &Envelope{}
 
 	err := json.Unmarshal(envelope, jwe)
 	if err != nil {
-		return nil, fmt.Errorf("unpack: %w", err)
+		return nil, nil, fmt.Errorf("unpack: %w", err)
 	}
 
 	recipientPubKey, recipient, err := p.findRecipient(jwe.Recipients)
 	if err != nil {
-		return nil, fmt.Errorf("unpack: %w", err)
+		return nil, nil, fmt.Errorf("unpack: %w", err)
 	}
 
 	senderKey, err := p.decryptSPK(recipientPubKey, recipient.Header.SPK)
 	if err != nil {
-		return nil, fmt.Errorf("unpack: sender key: %w", err)
+		return nil, nil, fmt.Errorf("unpack: sender key: %w", err)
 	}
 
 	// senderKey must not be empty to proceed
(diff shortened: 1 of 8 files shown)


pkg/didcomm/packer/jwe/authcrypt/unpack.go
```diff
@@ -47,18 +47,18 @@ func (p *Packer) Unpack(envelope []byte) ([]byte, error) {
...
 		}
 
 		symOutput, er := p.decryptPayload(sharedKey, jwe)
 		if er != nil {
-			return nil, fmt.Errorf("unpack: %w", er)
+			return nil, nil, fmt.Errorf("unpack: %w", er)
 		}
 
-		return symOutput, nil
+		return symOutput, senderPubKey[:], nil
```
Moopli: You could return `senderKey` instead but it makes no difference here.

Moopli: (It's simpler, but it's fine this way)

rolsonquadras: Valid point (missed this during refactoring). [Updated](https://github.com/hyperledger/aries-framework-go/compare/91457d1236b61875c4b0ce2b529d75fe2851dbc3..b34c5cd85fb496c143d5c08f96451194c349a7c8#diff-798c9657f10f55743907e471a65aee97R58) the code to use `senderKey`.

pkg/didcomm/packer/legacy/authcrypt/unpack.go
```diff
@@ -66,68 +68,70 @@ func getCEK(recipients []recipient, km kms.KeyManager) (*[chacha.KeySize]byte, e
...
+		return nil, nil, err
 	}
 
-	senderData := base58.Decode(string(senderSlice))
+	senderData := base58.Decode(string(senderPub))
+
+	senderPubCurve, err := cryptoutil.PublicEd25519toCurve25519(senderData)
 
-	return cryptoutil.PublicEd25519toCurve25519(senderData)
+	return senderPub, senderPubCurve, err
```
Baha-sk: `senderPubCurve` is raw bytes of the public encryption key
while `senderPub` seems to be the byte representation of the `base58` encoding of the public signing (verification) key

I guess the tests are passing as senderPub is sent to the caller of Unpack which expects a base58 encoded key... either make this clear in the comment, or return senderData (raw bytes of public signing key) and do base58.Encode after  getting the output of Unpack().

but returning keys should be of the same type (either return both base58 encoded or both raw bytes, but not a mix of 1 base58 encoded and the other not.. we prefer returning both raw bytes).

in short, return `senderData` instead of `senderPub` and call base58.Encode() at https://github.com/hyperledger/aries-framework-go/pull/880/files#diff-68bdfb240e9da84a7ae7cdd37c85a8e4R154 

and make sure the `jwe` Packer implementation returns the same raw byte (not base58 encoded) to keep the same behaviour as in `legacy` implementation

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/b34c5cd85fb496c143d5c08f96451194c349a7c8..c11501105b1b19784aa4a95fe77f516372579ac5#diff-b6253114caf4247940f4442d6515375aR134).

PR #884, opened 2019-11-27
feat: Introduce - finish state machine

### LINKED ISSUE #883: Introduce - possibility to finish state machine
Need to add a method to the introduce service `InvitationReceived` which should be executed by the did-exchange service or client to finish the state machine

soluchok:
* added `InvitationReceived` function to  finish state machine
* the behavior of the state machine is completed

Closes #883 
Part of #269

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -12,9 +12,13 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/google/uuid"
+
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -108,7 +112,7 @@ type Provider interface {
 // Forwarder provides the possibility to forward an invitation
 type Forwarder interface {
 	// method should be implemented in didexchange service
-	SendInvitation(inv *didexchange.Invitation, dest *service.Destination) error
+	SendInvitation(pthID string, inv *didexchange.Invitation, dest *service.Destination) error
```
troyronda: Are you planning to create a new Invitation object based on `inv` and `pthID`?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -253,6 +257,28 @@ func (s *Service) doHandle(msg *service.DIDCommMsg, outbound bool) (*metaData, e
 	}, nil
 }
 
+// InvitationReceived is used to finish the state machine
+// the function should be called by didexchange after receiving an invitation
```
troyronda: How does DID Exchange know to do this?

pkg/didcomm/protocol/introduce/service.go
soluchok: To close the state machine I need the same thread id. According to that, I assume that did-exchange will start protocol (did-exchange) with parent thread id.

pkg/didcomm/protocol/introduce/service.go
soluchok: I don't know yet. Probably after receiving an invitation did-exchange should check parent thread id. Somehow it should determine that pthid was provided by the introduce protocol.

troyronda: We will need to avoid a direct coupling.

Issue #890, opened 2019-11-28
Introduce - Add validation to the service for incoming messages


soluchok:
Add validation to the service for incoming messages (e.g response message approval false)

PR #894, opened 2019-11-28
feat: Introduce - Add validation to the service for incoming messages

soluchok:
* added ProblemReport structure (a few fields)
* now, we can handle one more use case (no one responded with invitation)
* tests were refactored

Part one of issue #890 

### DIFF
--- a/pkg/didcomm/common/model/problem_report.go
+++ b/pkg/didcomm/common/model/problem_report.go
@@ -1,0 +1,17 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package model
+
+// ProblemReport problem report definition
+// TODO: need to provide full ProblemReport structure https://github.com/hyperledger/aries-framework-go/issues/912
+type ProblemReport struct {
+	Type        string `json:"@type"`
+	ID          string `json:"@id"`
+	Description struct {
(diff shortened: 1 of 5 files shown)


pkg/didcomm/common/model/problem_report.go
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package model
+
+// ProblemReport problem report definition
+// TODO: need to provide structure like it is described here
```
troyronda: issue number?

soluchok: Done! https://github.com/hyperledger/aries-framework-go/issues/912

pkg/didcomm/protocol/introduce/service_test.go
```diff
@@ -1423,42 +1020,365 @@ func TestService_ProposalUnusualWithRequest(t *testing.T) {
 
 	wg.Add(3)
 
-	go handleIntroducer(&flow{
+	go checkAndHandle(&flow{
```
troyronda: it would be good to comment each usage of checkAndHandle, since it is now a generic func.

soluchok: Done!

Issue #858, opened 2019-11-22
Transport Return Route Protocol (RFC 0092)


rolsonquadras:
Implement Transport Return Route Protocol in Aries Framework Go - https://github.com/hyperledger/aries-rfcs/tree/master/features/0092-transport-return-route.

rolsonquadras:
Implementation of this protocol can be broken down into following issues:
1. Get access to sender verification key in Inbound transport (To store the transport based on verKey) - https://github.com/hyperledger/aries-framework-go/issues/876
2. Support to configure framework with Transport return route option - To add values to the transport decorator.
3. Make the transport duplex - Use the stored transport based on the verKey of the recipient
(sender in case of inbound) during outbound calls - https://github.com/hyperledger/aries-framework-go/issues/877

rolsonquadras:
Core framework support is implemented as part of PR #896. 

Remaining issues for completing basic e2e implementation - #877 #893

PR #896, opened 2019-11-28
feat: Transport Return Route - Framework Configuration

rolsonquadras:
- Aries framework level option to set the transport return route value [none, all, thread]; Added the option at framework level(edge agent doesn't have inbound capability) rather than each DIDComm message type.
- Pass the framework option to Outbound Dispatcher
- Outbound dispatcher injects the transport decorator to the outbound message before packing

Part of #858 

### DIFF
--- a/pkg/didcomm/transport/http/outbound_test.go
+++ b/pkg/didcomm/transport/http/outbound_test.go
@@ -13,6 +13,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 )
 
 func TestWithOutboundOpts(t *testing.T) {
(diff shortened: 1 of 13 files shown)


pkg/didcomm/transport/http/outbound_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
```
fqutishat: wrong location

pkg/framework/aries/framework_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
```
fqutishat: wrong location

pkg/didcomm/transport/http/outbound_test.go
rolsonquadras: [Fixed](https://github.com/hyperledger/aries-framework-go/compare/cd67ce4b092da2155a6a31cb89a25d78f07e5543..106115c2eb1724054c5a79866f440e5ea66d22d9#diff-9eab1127a5a2c8101c28c65b686ce256R17).

pkg/framework/aries/framework_test.go
rolsonquadras: [Fixed](https://github.com/hyperledger/aries-framework-go/compare/cd67ce4b092da2155a6a31cb89a25d78f07e5543..106115c2eb1724054c5a79866f440e5ea66d22d9#diff-315e22328726a0c9da38ac055e2663f4R26).

pkg/framework/aries/framework.go
```diff
@@ -128,6 +130,21 @@ func WithInboundTransport(inboundTransport transport.InboundTransport) Option {
 	}
 }
 
+// WithTransportReturnRoute injects transport return route option to the Aries framework. Acceptable values - "none",
```
troyronda: Do we need a TODO for more dynamically setting the TransportReturnRoute (in the protocol clients) or do we expect to always use the same setting (for a particular framework?

pkg/framework/aries/framework.go
```diff
@@ -128,6 +130,21 @@ func WithInboundTransport(inboundTransport transport.InboundTransport) Option {
...
+// "all" or "thread". RFC - https://github.com/hyperledger/aries-rfcs/tree/master/features/0092-transport-return-route
```
troyronda: Does thread actually work at a global level? Also I don't see return_route_thread attribute :).

> thread: Send all messages matching the cryptographic key and thread specified in the return_route_thread attribute.

pkg/framework/aries/framework.go
rolsonquadras: We can enhance this to support route option at message/protocol layer by passing a `{messageType, returnValue}` config. eg, {`didexchange, all},{introduce, none}`. Created an issue - https://github.com/hyperledger/aries-framework-go/issues/899

pkg/framework/aries/framework.go
rolsonquadras: We can intercept `Thread` at inbound layer as it's a decorator. Currently planning to use senderKey for all to map it with incoming conection. For thread, we can map senderKey-threadID combo.

PR #898, opened 2019-11-28
chore: No-op Packer, useful for testing.

Moopli:

### DIFF
--- a/pkg/internal/mock/packer/noop.go
+++ b/pkg/internal/mock/packer/noop.go
@@ -1,0 +1,91 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package packer
+
+import (
+	"encoding/base64"
+	"encoding/json"
+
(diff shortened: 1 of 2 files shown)


troyronda:
@Moopli Where do you plan to use the no-op packer?

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/golang/mock v1.3.1
 	github.com/google/uuid v1.1.1
 	github.com/gorilla/mux v1.7.3
+	github.com/mr-tron/base58 v1.1.2
```
troyronda: We are using a different base58 package elsewhere.

troyronda: "github.com/btcsuite/btcutil/base58"

Moopli: Good catch, thought I'd gotten them all.

Moopli: Ah, I did get them all, but I didn't tidy the mod file.

Moopli:
@troyronda it's for testing, to make tests results easier to parse - I've moved it to mock since it's not intended for prod.

Issue #879, opened 2019-11-25
Linter Warning : WebSocket package


rolsonquadras:
https://github.com/hyperledger/aries-framework-go/pull/870/checks?check_run_id=316707181#step:4:66

```
level=warning msg="[runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [/opt/workspace/pkg/didcomm/transport/ws/inbound.go:137:22: Accept not declared by package websocket]"
level=warning msg="[runner] Can't run linter goanalysis_metalinter: assign: failed prerequisites: inspect@github.com/hyperledger/aries-framework-go/pkg/didcomm/transport/ws [github.com/hyperledger/aries-framework-go/pkg/didcomm/transport/ws.test]"
```

The WebSocket library was updated as part of #867 - https://github.com/hyperledger/aries-framework-go/pull/867/files#diff-a920dee998c6d38ee04585f60a90626bR137

PR #910, opened 2019-12-02
feat: WebSocket - Support Duplex communication based on Transport Return Route Feature

### LINKED ISSUE #877: DIDComm Transport(WS) - Support Duplex communication
The inbound transport need to save the connections if the clients wants to use the same connection for the return route. The outbound response should make use of same connection to send the response message

Required for implementing #858.

rolsonquadras:
- Add new function Start() to Outbound Transport : Called by framework by passin the context (similar to Inbound Transport)
- Add new function AriesFrameworkID() to Context : Framework uses this ID to map between inbound and outbound transport for duplex communication
- Update WebSocket transport implementation to support Duplex communication based on Transport Return Route config

closes #877  

### DIFF
--- a/pkg/didcomm/transport/ws/inbound.go
+++ b/pkg/didcomm/transport/ws/inbound.go
@@ -20,12 +20,11 @@ import (
 
 var logger = log.New("aries-framework/ws")
 
-const processFailureErrMsg = "failed to process the message"
-
 // Inbound http(ws) type.
 type Inbound struct {
 	externalAddr string
 	server       *http.Server
+	pool         *connPool
 }
 
 // NewInbound creates a new WebSocket inbound transport instance.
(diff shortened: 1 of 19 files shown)


pkg/didcomm/transport/ws/inbound.go
```diff
@@ -73,77 +75,22 @@ func (i *Inbound) Endpoint() string {
...
-		err = c.Write(context.Background(), websocket.MessageText, []byte(resp))
-		if err != nil {
-			logger.Errorf("error writing the message: %v", err)
-		}
-	}
+	i.pool.listener(c)
 }
 
-func upgradeConnection(w http.ResponseWriter, r *http.Request) (*websocket.Conn, func(), error) {
+func upgradeConnection(w http.ResponseWriter, r *http.Request) (*websocket.Conn, error) {
```
fqutishat: Can you add // +build !js

pkg/didcomm/transport/ws/outbound.go
```diff
@@ -15,57 +15,95 @@ import (
...
+	return "", nil
 }
 
 // Accept checks for the url scheme.
 func (cs *OutboundClient) Accept(url string) bool {
 	return strings.HasPrefix(url, webSocketScheme)
 }
+
+func (cs *OutboundClient) getConnection(destination *service.Destination) (*websocket.Conn, func(), error) {
+	if cs.pool != nil {
```
sandrask: Is this condition valid? Pool should have been set in start. Otherwise, you will have to put locks around it.

pkg/didcomm/transport/ws/inbound.go
rolsonquadras: To resolve https://github.com/hyperledger/aries-framework-go/issues/879, adding this tag will not suffice. Need to update the dependencies. I'll add it as a separate PR (this PR is already > 500 lines).

pkg/didcomm/transport/ws/outbound.go
rolsonquadras: Very valid point. Pool is already set in start(). This condition is not necessary and [removed](https://github.com/hyperledger/aries-framework-go/compare/5ba26394828508c6b430b55d5c3960891ca1dc81..89537285cd7d33cf2940c005e4a583162a90d55b#diff-84ebb4df3b345595955c03cc5f620fd9L70) in new patch.

PR #917, opened 2019-12-03
feat: Create and verify JWS when creating new presentation

### LINKED ISSUE #862: Create and verify JWS when creating new presentation
**What I'm trying to do**

When creating a new Presentation from JSON data, we should actually verify the VC proof and the VP proof before returning. 

When creating a presentation from a VC, we should create the Proof for it before returning.

kdimak:
closes #862

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -402,7 +402,7 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, []byte, e
 	crOpts := parseCredentialOpts(opts)
 
 	// Decode credential (e.g. from JWT).
-	vcDataDecoded, err := decodeRaw(vcData, crOpts)
+	vcDataDecoded, err := decodeRaw(vcData, crOpts.issuerPublicKeyFetcher)
 	if err != nil {
 		return nil, nil, fmt.Errorf("decode new credential: %w", err)
 	}
@@ -513,13 +513,13 @@ func newCredential(raw *rawCredential, schemas []TypedID) (*Credential, error) {
 	}, nil
 }
 
-func decodeRaw(vcData []byte, crOpts *credentialOpts) ([]byte, error) {
+func decodeRaw(vcData []byte, pubKeyFetcher PublicKeyFetcher) ([]byte, error) {
 	if isJWS(vcData) {
-		if crOpts.issuerPublicKeyFetcher == nil {
+		if pubKeyFetcher == nil {
 			return nil, errors.New("public key fetcher is not defined")
 		}
 
-		vcDecodedBytes, err := decodeCredJWS(vcData, crOpts.issuerPublicKeyFetcher)
+		vcDecodedBytes, err := decodeCredJWS(vcData, pubKeyFetcher)
 		if err != nil {
 			return nil, fmt.Errorf("JWS decoding: %w", err)
 		}
@@ -756,12 +756,16 @@ func (vc *Credential) MarshalJSON() ([]byte, error) {
 }
 
 // Presentation encloses credential into presentation.
-func (vc *Credential) Presentation() *Presentation {
+func (vc *Credential) Presentation() (*Presentation, error) {
 	vp := Presentation{
-		Context:    vc.Context,
-		Type:       []string{vpType},
-		Credential: []Credential{*vc},
+		Context: vc.Context,
+		Type:    []string{vpType},
 	}
 
-	return &vp
+	err := vp.SetCredentials(vc)
+	if err != nil {
+		return nil, fmt.Errorf("build presentation from credential: %w", err)
+	}
+
+	return &vp, nil
 }
--- a/pkg/doc/verifiable/credential_test.go
+++ b/pkg/doc/verifiable/credential_test.go
@@ -101,15 +101,19 @@ func TestNewCredential(t *testing.T) {
 
 	t.Run("test a try to create a new Verifiable Credential from JSON with invalid structure", func(t *testing.T) {
 		emptyJSONDoc := "{}"
-		_, _, err := NewCredential([]byte(emptyJSONDoc))
+		vc, vcBytes, err := NewCredential([]byte(emptyJSONDoc))
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "verifiable credential is not valid")
+		require.Nil(t, vc)
+		require.Nil(t, vcBytes)
 	})
 
 	t.Run("test a try to create a new Verifiable Credential from non-JSON doc", func(t *testing.T) {
-		_, _, err := NewCredential([]byte("non json"))
+		vc, vcBytes, err := NewCredential([]byte("non json"))
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "new credential")
+		require.Nil(t, vc)
+		require.Nil(t, vcBytes)
 	})
 }
 
@@ -592,9 +596,11 @@ func TestCustomCredentialJsonSchemaValidator2018(t *testing.T) {
 	require.NoError(t, mErr)
 
 	t.Run("Applies custom JSON Schema and detects data inconsistency due to missing new required field", func(t *testing.T) { //nolint:lll
-		_, _, err := NewCredential(missingReqFieldSchema, WithSchemaDownloadClient(&http.Client{}))
+		vc, vcBytes, err := NewCredential(missingReqFieldSchema, WithSchemaDownloadClient(&http.Client{}))
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "referenceNumber is required")
+		require.Nil(t, vc)
+		require.Nil(t, vcBytes)
 	})
 
 	t.Run("Applies custom credentialSchema and passes new data inconsistency check", func(t *testing.T) {
@@ -627,9 +633,11 @@ func TestCustomCredentialJsonSchemaValidator2018(t *testing.T) {
 		schemaWithInvalidURLToCredentialSchema, err := json.Marshal(raw)
 		require.NoError(t, err)
 
-		_, _, err = NewCredential(schemaWithInvalidURLToCredentialSchema, WithSchemaDownloadClient(&http.Client{}))
+		vc, vcBytes, err := NewCredential(schemaWithInvalidURLToCredentialSchema, WithSchemaDownloadClient(&http.Client{}))
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "load of custom credential schema")
+		require.Nil(t, vc)
+		require.Nil(t, vcBytes)
 	})
 
 	t.Run("Uses default schema if custom credentialSchema is not of 'JsonSchemaValidator2018' type", func(t *testing.T) { //nolint:lll
@@ -681,9 +689,10 @@ func TestDownloadCustomSchema(t *testing.T) {
 		}))
 		defer func() { testServer.Close() }()
 
-		_, err := loadCredentialSchema(testServer.URL, &http.Client{})
+		customSchema, err := loadCredentialSchema(testServer.URL, &http.Client{})
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "load credential schema")
+		require.Nil(t, customSchema)
 	})
 
 	t.Run("HTTP GET request to download custom credentialSchema returns not OK", func(t *testing.T) {
@@ -693,9 +702,10 @@ func TestDownloadCustomSchema(t *testing.T) {
 		}))
 		defer func() { testServer.Close() }()
 
-		_, err := loadCredentialSchema(testServer.URL, &http.Client{})
+		customSchema, err := loadCredentialSchema(testServer.URL, &http.Client{})
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "credential schema endpoint HTTP failure")
+		require.Nil(t, customSchema)
 	})
 }
 
@@ -747,27 +757,30 @@ func TestCredentialSubjectId(t *testing.T) {
 					"spouse": "did:example:ebfeb1f712ebc6f1c276e12ec21",
 				},
 			}}
-		_, err := subjectID(vcWithMultipleSubjects.Subject)
+		subjectID, err := subjectID(vcWithMultipleSubjects.Subject)
 		require.Error(t, err)
 		require.EqualError(t, err, "more than one subject is defined")
+		require.Empty(t, subjectID)
 	})
 
 	t.Run("With no Subject", func(t *testing.T) {
 		vcWithNoSubject := &Credential{
 			Subject: nil,
 		}
-		_, err := subjectID(vcWithNoSubject.Subject)
+		subjectID, err := subjectID(vcWithNoSubject.Subject)
 		require.Error(t, err)
 		require.EqualError(t, err, "subject id is not defined")
+		require.Empty(t, subjectID)
 	})
 
 	t.Run("With empty Subject", func(t *testing.T) {
 		vcWithNoSubject := &Credential{
 			Subject: []map[string]interface{}{},
 		}
-		_, err := subjectID(vcWithNoSubject.Subject)
+		subjectID, err := subjectID(vcWithNoSubject.Subject)
 		require.Error(t, err)
 		require.EqualError(t, err, "no subject is defined")
+		require.Empty(t, subjectID)
 	})
 
 	t.Run("With non-string Subject ID", func(t *testing.T) {
@@ -778,9 +791,10 @@ func TestCredentialSubjectId(t *testing.T) {
 				"name": "Bachelor of Science and Arts",
 			},
 		}}
-		_, err := subjectID(vcWithNotStringID.Subject)
+		subjectID, err := subjectID(vcWithNotStringID.Subject)
 		require.Error(t, err)
 		require.EqualError(t, err, "subject id is not string")
+		require.Empty(t, subjectID)
 	})
 
 	t.Run("Subject without ID defined", func(t *testing.T) {
@@ -795,9 +809,10 @@ func TestCredentialSubjectId(t *testing.T) {
 				},
 			},
 		}
-		_, err := subjectID(vcWithSubjectWithoutID.Subject)
+		subjectID, err := subjectID(vcWithSubjectWithoutID.Subject)
 		require.Error(t, err)
 		require.EqualError(t, err, "subject id is not defined")
+		require.Empty(t, subjectID)
 	})
 
 	t.Run("Get subject id from custom structure", func(t *testing.T) {
@@ -830,9 +845,10 @@ func TestCredentialSubjectId(t *testing.T) {
 	})
 
 	t.Run("Get subject id from unmarshalable structure", func(t *testing.T) {
-		_, err := subjectID(make(chan int))
+		subjectID, err := subjectID(make(chan int))
 		require.Error(t, err)
 		require.EqualError(t, err, "subject of unknown structure")
+		require.Empty(t, subjectID)
 	})
 }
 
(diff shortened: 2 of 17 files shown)


pkg/doc/verifiable/presentation.go
```diff
@@ -313,6 +327,49 @@ func NewPresentation(vpData []byte, opts ...PresentationOpt) (*Presentation, err
 	return vp, nil
 }
 
+func decodeCredentials(rawCred interface{}, opts *presentationOpts) ([]interface{}, error) {
```
kdimak: It's the place we check for JWT signature.
In the verifiable presentation, `verifiableCredential` field could be either single string (JWT) or list of strings (several JWTs) or struct (verifiable credential) or list structs (several verifiable credentials).

Baha-sk: it would be good to have a doc comment defining the possible/accepted rawCred values..

kdimak: Good idea Baha! I added some comments.

pkg/doc/verifiable/credential.go
```diff
@@ -402,7 +402,7 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, []byte, e
 	crOpts := parseCredentialOpts(opts)
 
 	// Decode credential (e.g. from JWT).
-	vcDataDecoded, err := decodeRaw(vcData, crOpts)
+	vcDataDecoded, err := decodeRaw(vcData, crOpts.issuerPublicKeyFetcher)
```
talwinder50: isnt issuerPublicKey is sufficient, do we need to add fetcher word ??

kdimak: It's not a public key, it's a mechanism of its retrieval based on issuer/holder id (should be DID) and key id. We get the DID and the key id from received JWS in order to load the public key and then verify the signature of JWS. The real mechanism of `PublicKeyFetcher` goes beyond the verifiable credential library. I believe you will need to implement retrieval of the key from the verifiable credential registry (see https://www.w3.org/TR/vc-data-model/#dfn-verifiable-data-registries).

talwinder50: when you say 
"you will need to implement retrieval of the key from the verifiable credential registry" can you elaborate if its related to the consumption of this vc library ?

kdimak: Indirectly - yes :)
I think it will be implemented by simply DID resolution.

pkg/doc/verifiable/presentation.go
```diff
@@ -178,49 +179,57 @@ func (vp *Presentation) JWTClaims(audience []string, minimizeVP bool) *JWTPresCl
...
-		if err != nil {
-			return nil, fmt.Errorf("marshal credentials from presentation: %w", err)
-		}
+// Credentials returns current credentials of presentation.
+func (vp *Presentation) Credentials() []interface{} {
+	return vp.credentials
+}
 
-		return credBytes, nil
+// SetCredentials defines credentials of presentation. It could be Credential struct or encoded form (e.g. JWT).
```
talwinder50: You can edit the comment saying cred can be either string/byte (probably serialized JWT) or Credential structure. It will just made it more readable. Just the minor comment

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -163,3 +167,290 @@ func ExampleCredential_Presentation() {
...
 }
+
+func ExamplePresentation_SetCredentials() {
```
kdimak: Added example of Verifiable.SetCredentials() - pass Credential and JWS string.

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -163,3 +167,290 @@ func ExampleCredential_Presentation() {
...
+	//				"referenceNumber": 83294847
+	//			}
+	//		}
+	//	],
+	//	"holder": "did:example:ebfeb1f712ebc6f1c276e12ec21"
+	//}
+}
+
+//nolint:gocyclo
+func ExamplePresentation_MarshalledCredentials() {
```
kdimak: Added example of Verifiable.MarshalledCredentials(). In this example, it's shown:
- Holder prepares VP in JWS form where VP is defined in JWS form as well.
- Verifier decodes VP and then decodes its VC.

pkg/doc/verifiable/presentation.go
kdimak: Done!

pkg/doc/verifiable/presentation_jwt_proof_test.go
```diff
@@ -129,6 +130,167 @@ func TestNewPresentationFromUnsecuredJWT(t *testing.T) {
...
+
+		holderPubKey, holderPrivKey, err := ed25519.GenerateKey(rand.Reader)
+		r.NoError(err)
+
+		vpJWS, err := vp.JWTClaims([]string{}, true).MarshalJWS(
+			EdDSA, holderPrivKey, "holder-key")
+		r.NoError(err)
+
+		// Decode VP
+		_, err = NewPresentation([]byte(vpJWS), WithPresPublicKeyFetcher(
```
sandrask: you should check that presentation is nil

kdimak:
> This is correct, you need fetcher/resolver. We implemented similar in did-exchange (it is called keyresolver

Yeah, but we need first to resolve DID of the entity which signed the JWT (`issuerID` parameter) and then find appropriate public key (`keyID` parameter) as it's done in `keyResolver.resolve()`.

pkg/doc/verifiable/presentation_jwt_proof_test.go
kdimak: Well, in case of expected error I usually just check error presence and error message. I will change this for all error expecting checks then.

sandrask: That would be perfect. In case of an error we should also check return value to make sure that it is nil (or empty)

kdimak: Done! @sandrask

pkg/doc/verifiable/presentation.go
```diff
@@ -178,49 +179,58 @@ func (vp *Presentation) JWTClaims(audience []string, minimizeVP bool) *JWTPresCl
...
 
-		return credBytes, nil
+// SetCredentials defines credentials of presentation.
+// The credential could be string/byte (probably serialized JWT) or Credential structure.
+func (vp *Presentation) SetCredentials(creds ...interface{}) error {
+	for i := range creds {
+		switch creds[i].(type) {
+		default:
+			return errors.New("unsupported credential format")
+		case []byte, string, *Credential, Credential:
```
DRK3: Empty case statement?

pkg/doc/verifiable/presentation.go
```diff
@@ -178,49 +179,58 @@ func (vp *Presentation) JWTClaims(audience []string, minimizeVP bool) *JWTPresCl
...
-		}
+// Credentials returns current credentials of presentation.
+func (vp *Presentation) Credentials() []interface{} {
+	return vp.credentials
+}
 
-		return credBytes, nil
+// SetCredentials defines credentials of presentation.
+// The credential could be string/byte (probably serialized JWT) or Credential structure.
+func (vp *Presentation) SetCredentials(creds ...interface{}) error {
```
DRK3: I'm a bit confused by this method... why not just do `vp.credentials = cred` directly wherever you call `SetCredentials`? For other types of credentials, do you think they won't assign creds to vp.creds? Perhaps you can clarify for me.

pkg/doc/verifiable/credential.go
```diff
@@ -756,12 +756,16 @@ func (vc *Credential) MarshalJSON() ([]byte, error) {
 }
 
 // Presentation encloses credential into presentation.
-func (vc *Credential) Presentation() *Presentation {
+func (vc *Credential) Presentation() (*Presentation, error) {
```
DRK3: Can the name be a bit more descriptive of what the function is actually doing?

pkg/doc/verifiable/credential_test.go
```diff
@@ -681,9 +689,10 @@ func TestDownloadCustomSchema(t *testing.T) {
 		}))
 		defer func() { testServer.Close() }()
 
-		_, err := loadCredentialSchema(testServer.URL, &http.Client{})
+		sBytes, err := loadCredentialSchema(testServer.URL, &http.Client{})
```
DRK3: I think "s" as a short form for "schema" here is a bit unintuitive. I would just write out the whole word "schemaBytes" instead.

pkg/doc/verifiable/credential_test.go
```diff
@@ -693,9 +702,10 @@ func TestDownloadCustomSchema(t *testing.T) {
...
```
DRK3: I think "s" as a short form for "schema" here is a bit unintuitive. I would just write out the whole word "schemaBytes" instead.

pkg/doc/verifiable/credential_test.go
```diff
@@ -747,27 +757,30 @@ func TestCredentialSubjectId(t *testing.T) {
 					"spouse": "did:example:ebfeb1f712ebc6f1c276e12ec21",
 				},
 			}}
-		_, err := subjectID(vcWithMultipleSubjects.Subject)
+		sID, err := subjectID(vcWithMultipleSubjects.Subject)
```
DRK3: Same thing here. This is me nitpicking though, so consider these "s" short form changes optional - I'm okay with whatever you decide here

pkg/doc/verifiable/credential_test.go
```diff
@@ -884,34 +900,38 @@ func TestDecodeIssuer(t *testing.T) {
...
 			"name": 55,
 		})
 		require.Error(t, err)
 		require.EqualError(t, err, "value of key 'name' is not a string")
+		require.Empty(t, iID)
 	})
 
 	t.Run("Decode Issuer of invalid type", func(t *testing.T) {
-		_, err := decodeIssuer(77)
+		iID, err := decodeIssuer(77)
```
DRK3: Interesting name - iID. Again, still optional!

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -163,3 +167,290 @@ func ExampleCredential_Presentation() {
...
+		CustomFields: map[string]interface{}{
+			"referenceNumber": 83294847,
+		},
+	}
+
+	// The second VC is provided in JWS form (e.g. kept in the wallet in that form).
+	vcJWS := "eyJhbGciOiJFZERTQSIsImtpZCI6IiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Nzc5MDY2MDQsImlhdCI6MTI2MjM3MzgwNCwiaXNzIjoiZGlkOmV4YW1wbGU6NzZlMTJlYzcxMmViYzZmMWMyMjFlYmZlYjFmIiwianRpIjoiaHR0cDovL2V4YW1wbGUuZWR1L2NyZWRlbnRpYWxzLzE4NzIiLCJuYmYiOjEyNjIzNzM4MDQsInN1YiI6ImRpZDpleGFtcGxlOmViZmViMWY3MTJlYmM2ZjFjMjc2ZTEyZWMyMSIsInZjIjp7IkBjb250ZXh0IjpbImh0dHBzOi8vd3d3LnczLm9yZy8yMDE4L2NyZWRlbnRpYWxzL3YxIiwiaHR0cHM6Ly93d3cudzMub3JnLzIwMTgvY3JlZGVudGlhbHMvZXhhbXBsZXMvdjEiXSwiY3JlZGVudGlhbFNjaGVtYSI6W10sImNyZWRlbnRpYWxTdWJqZWN0Ijp7ImRlZ3JlZSI6eyJ0eXBlIjoiQmFjaGVsb3JEZWdyZWUiLCJ1bml2ZXJzaXR5IjoiTUlUIn0sImlkIjoiZGlkOmV4YW1wbGU6ZWJmZWIxZjcxMmViYzZmMWMyNzZlMTJlYzIxIiwibmFtZSI6IkpheWRlbiBEb2UiLCJzcG91c2UiOiJkaWQ6ZXhhbXBsZTpjMjc2ZTEyZWMyMWViZmViMWY3MTJlYmM2ZjEifSwiaXNzdWVyIjp7Im5hbWUiOiJFeGFtcGxlIFVuaXZlcnNpdHkifSwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCIsIlVuaXZlcnNpdHlEZWdyZWVDcmVkZW50aWFsIl19fQ.AHn2A2q5DL1heX3_izq_2yrsBDhoZ6BGGKhoRvhfMnMUuuOnBOdekdTg-dfUMJgipXRql_6WzBUIj4wTFehXCw" // nolint:lll
+
+	// Set the credentials.
```
DRK3: Since the method name below is precisely "SetCredentials", this comment is redundant

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -163,3 +167,290 @@ func ExampleCredential_Presentation() {
...
+
+	// json.MarshalIndent() calls Presentation.MarshalJSON()
+	vpJSON, err := json.MarshalIndent(vp, "", "\t")
+	if err != nil {
+		fmt.Println(fmt.Errorf("failed to marshal VP to JSON: %w", err))
+	}
+
+	fmt.Println(string(vpJSON))
+
+	// Output: {
```
DRK3: Just want to confirm - is this comment the expected output? If so, it might be a little clearer to label it as "expected output" instead of just "output".

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -163,3 +167,290 @@ func ExampleCredential_Presentation() {
...
+	vcJWS := "eyJhbGciOiJFZERTQSIsImtpZCI6IiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Nzc5MDY2MDQsImlhdCI6MTI2MjM3MzgwNCwiaXNzIjoiZGlkOmV4YW1wbGU6NzZlMTJlYzcxMmViYzZmMWMyMjFlYmZlYjFmIiwianRpIjoiaHR0cDovL2V4YW1wbGUuZWR1L2NyZWRlbnRpYWxzLzE4NzIiLCJuYmYiOjEyNjIzNzM4MDQsInN1YiI6ImRpZDpleGFtcGxlOmViZmViMWY3MTJlYmM2ZjFjMjc2ZTEyZWMyMSIsInZjIjp7IkBjb250ZXh0IjpbImh0dHBzOi8vd3d3LnczLm9yZy8yMDE4L2NyZWRlbnRpYWxzL3YxIiwiaHR0cHM6Ly93d3cudzMub3JnLzIwMTgvY3JlZGVudGlhbHMvZXhhbXBsZXMvdjEiXSwiY3JlZGVudGlhbFNjaGVtYSI6W10sImNyZWRlbnRpYWxTdWJqZWN0Ijp7ImRlZ3JlZSI6eyJ0eXBlIjoiQmFjaGVsb3JEZWdyZWUiLCJ1bml2ZXJzaXR5IjoiTUlUIn0sImlkIjoiZGlkOmV4YW1wbGU6ZWJmZWIxZjcxMmViYzZmMWMyNzZlMTJlYzIxIiwibmFtZSI6IkpheWRlbiBEb2UiLCJzcG91c2UiOiJkaWQ6ZXhhbXBsZTpjMjc2ZTEyZWMyMWViZmViMWY3MTJlYmM2ZjEifSwiaXNzdWVyIjp7Im5hbWUiOiJFeGFtcGxlIFVuaXZlcnNpdHkifSwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCIsIlVuaXZlcnNpdHlEZWdyZWVDcmVkZW50aWFsIl19fQ.AHn2A2q5DL1heX3_izq_2yrsBDhoZ6BGGKhoRvhfMnMUuuOnBOdekdTg-dfUMJgipXRql_6WzBUIj4wTFehXCw" // nolint:lll
+
+	// Set the credentials.
+	err := vp.SetCredentials(vc, vcJWS)
+	if err != nil {
+		fmt.Println(fmt.Errorf("failed to set credentials of VP: %w", err))
+	}
+
+	//Output:
```
DRK3: Empty comment? Looks like you meant to put something here.

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -18,7 +18,7 @@ import (
 //nolint:gochecknoglobals
 var privHolderKey = ed25519.PrivateKey{137, 207, 34, 115, 168, 80, 44, 145, 236, 206, 165, 152, 211, 195, 240, 205, 232, 37, 216, 101, 58, 133, 198, 107, 232, 119, 30, 80, 176, 137, 10, 251, 109, 144, 158, 93, 189, 195, 0, 24, 15, 29, 166, 185, 169, 69, 246, 25, 182, 179, 115, 54, 107, 4, 123, 30, 5, 88, 175, 94, 109, 15, 34, 113} //nolint:lll
 
-func ExamplePresentation() {
+func ExampleNewPresentation() {
```
DRK3: Since this isn't new code, this is optional - but, I noticed that this file ends in "_test" but doesn't seem to actually have tests in it? I wonder if it's possible to either turn these examples in unit tests... or perhaps just remove the word "test" from the file name.

pkg/doc/verifiable/json_test.go
```diff
@@ -49,8 +49,9 @@ func Test_marshalJSON(t *testing.T) {
 
 	t.Run("Failed JSON marshall", func(t *testing.T) {
 		// artificial example - pass smth which cannot be marshalled
-		_, err := marshalWithCustomFields(make(chan int), map[string]interface{}{})
+		jb, err := marshalWithCustomFields(make(chan int), map[string]interface{}{})
```
DRK3: What does "jb" stand for?

pkg/doc/verifiable/jwt_unsecured_test.go
```diff
@@ -62,28 +63,36 @@ func TestUnmarshalUnsecuredJWT(t *testing.T) {
 
 	t.Run("rejects serialized JWT headers not in base64 format", func(t *testing.T) {
 		invalidJWT := fmt.Sprintf("%s.eyJwcm9kdWN0SWRzIjpbMSwyXSwic3ViIjoidXNlcjEyMyJ9.", notBase64)
-		_, _, err := unmarshalUnsecuredJWT([]byte(invalidJWT))
+		h, b, err := unmarshalUnsecuredJWT([]byte(invalidJWT))
```
DRK3: What do "h" and "b" stand for?

pkg/doc/verifiable/presentation.go
```diff
@@ -313,6 +328,57 @@ func NewPresentation(vpData []byte, opts ...PresentationOpt) (*Presentation, err
 	return vp, nil
 }
 
+// decodeCredentials decodes credential(s) embedded into presentation.
+// It could be defined as:
+// 1) string - it could be credential decoded into e.g. JWS.
+// 2) the same as 1) but as array - e.g. zero ore more JWS
+// 3) struct (should be map[string]interface{}) representing credential data model
+// 4) the same as 3) but as array - i.e. zero or more credentials structs.
```
DRK3: These are listing different possible formats for rawCred right? If so, perhaps you can make it explictly say "rawCred must be one of the following..."

pkg/doc/verifiable/example_presentation_test.go
troyronda: @DRK3 Example prefixes in Go are also counted as tests.

They create testable documentation.

DRK3: Gotcha - good to know!

pkg/doc/verifiable/presentation.go
kdimak: Yes, it's done on purpose in order to separate OK case (do nothing) with default exceptional one (do not accept VC of such kind).
I will make this more precise by putting `default: ` at the bottom of `switch` (what is everyone used for) and add a comment on empty case.

pkg/doc/verifiable/presentation.go
kdimak: The reason is to get control of what the users are passing.

pkg/doc/verifiable/credential.go
kdimak: It's doing an enclosure of VC into VP. Don't know how to name this better...

pkg/doc/verifiable/credential_test.go
kdimak: Yeah, just shorten the names in the tests... I will improve it.

pkg/doc/verifiable/credential_test.go
kdimak: OK

pkg/doc/verifiable/credential_test.go
kdimak: OK

pkg/doc/verifiable/credential_test.go
kdimak: This is really not well readable :)

pkg/doc/verifiable/example_presentation_test.go
kdimak: It's a requirement of Golang examples - see [here](https://blog.golang.org/examples).

pkg/doc/verifiable/example_presentation_test.go
kdimak: It's done to make sure the test outputs nothing - otherwise, some error messages were printed.

pkg/doc/verifiable/json_test.go
kdimak: It's `jsonBytes`, that's easy, isn't it? :) just kidding, I will rename it!

pkg/doc/verifiable/jwt_unsecured_test.go
kdimak: :) `headers` and `body`. I'll rename it.

PR #938, opened 2019-12-06
feat: Protocols can use DIDs to identify the agents they communicate with

### LINKED ISSUE #725: Allow DIDComm protocols to use DIDs to identify agents sending/receiving messages
For one-to-one agent communication, protocol code shouldn't need to include code for retrieving destination parameters when sending messages, and shouldn't need to decipher the identity of a sender when receiving messages. Instead, the outbound dispatcher or helper functions should retrieve destination parameters given a DID, and inbound transport should provide the sender's DID, so protocols can operate solely using DIDs if they don't need more customized functionality.

Moopli:
Outbound: create a Destination struct from a DID using helper functions, so an agent can provide the DID of the target.
- One function resolves a DID using your VDRI
- The other uses a DID doc if it's already available

Inbound: receive the DID of the sender as a parameter to the message handler, if the sender has a DID doc in your VDRI.

Fixes #725


Moopli:
BDD tests are currently broken, but review from @soluchok is needed, for the interface this provides to protocols.

Moopli:
Current API this PR creates:

This saves the correlation/connection between another agent's peer DID/keys, and the peer DID your agent uses to communicate with them. DIDExchange uses this to save the connection information when the protocol succeeds.
`pkg/didcomm/common/didconnection.DIDConnectionStore.SaveDIDConnection()`:
`context.DIDConnectionStore().SaveDIDConnection(myDID, theirDID string, theirKeys []string) error`

To get the `Destination` struct for sending outbound messages, provide the target's DID to `pkg/didcomm/common/service.GetDestination()`:
`service.GetDestination(did, serviceType, keyType string, vdr vdri.Registry) (*Destination, error)`

But why would you, when you can call SendToDID on the outbound dispatcher, and also skip determining your sender key:
```
// SendToDID sends a message from myDID to the agent who owns theirDID
func (o *OutboundDispatcher) SendToDID(msg interface{}, myDID, theirDID string) error
```

And for inbound messages, `HandleInbound()` has been extended to provide DID parameters for the sender's DID and the recipient DID (since you need to know which of your DIDs the message was sent to).
`HandleInbound(msg *DIDCommMsg, myDID string, theirDID string) (string, error)`

I still have to write a lot of tests and then bugfix, but any feedback is welcome.

pkg/didcomm/common/service/service.go
```diff
@@ -16,7 +16,7 @@ import (
 // Handler provides protocol service handle api.
 type Handler interface {
 	// HandleInbound handles inbound messages.
-	HandleInbound(msg *DIDCommMsg) (string, error)
+	HandleInbound(msg *DIDCommMsg, myDID string, theirDID string) (string, error)
```
soluchok: It looks like my myDID is redundant. We can get my myDID using a  `GetDID` function. e.g 
`myDID := GetDID(theirDID)`. According to that line https://github.com/hyperledger/aries-framework-go/pull/938/files#diff-b429c177e8c9b423d01fcc0d7d2383e9R81

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -15,29 +15,55 @@ import (
...
 	return &OutboundDispatcher{
 		outboundTransports:   prov.OutboundTransports(),
 		packager:             prov.Packager(),
 		transportReturnRoute: prov.TransportReturnRoute(),
+		vdRegistry:           prov.VDRIRegistry(),
+	}
+}
+
+// SendToDID sends a message from myDID to the agent who owns theirDID
+func (o *OutboundDispatcher) SendToDID(msg interface{}, myDID, theirDID string) error {
```
soluchok: This function also can be optimized. We can get myDID from `DIDConnectionStore` by theirDID. It would be easy to use.

pkg/didcomm/common/service/service.go
troyronda: This shouldn’t be guaranteed. An agent could have multiple connections to the same “theirDIDs” with different “myDIDs”.

soluchok: Yeah, but it is a map. I mean only two DIDs can be related to each other. According to code we have. See link ^

soluchok: Otherwise GetDID(theirDID) should return a list of myDids. Which means you can use any of them. Right?

soluchok: I am a bit confused. Why do we need multiple myDIDs by on theirDid ? If connection is established doesn’t make sense to create one more.

Moopli: It's useful for communicating with a public DID - you might want multiple DIDs you can use to communicate with them, for privacy.

pkg/didcomm/dispatcher/outbound.go
Moopli: Public DIDs, which we can use multiple peer DIDs to communicate with, means we can't actually do this mapping in all cases. I've added support for providing "myDID" to the inbound handler in a service.

Moopli:
Tests are passing, meaning it's time to write more tests.
I will take this out of draft when code is better tested.

pkg/didcomm/common/didconnection/didconnection.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	diddoc "github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// BaseDIDConnectionStore stores DIDs indexed by key
```
troyronda: why the prefix Base?

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -15,29 +15,55 @@ import (
...
+	if err != nil {
+		return err
 	}
+
+	src, err := service.GetDestination(myDID, didCommServiceType, ed25519KeyType, o.vdRegistry)
+	if err != nil {
+		return err
+	}
+
+	key := src.RecipientKeys[0]
```
troyronda: why index 0 could use a comment.

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -15,29 +15,55 @@ import (
...
 // OutboundDispatcher dispatch msgs to destination
 type OutboundDispatcher struct {
 	outboundTransports   []transport.OutboundTransport
 	packager             commontransport.Packager
 	transportReturnRoute string
+	vdRegistry           vdri.Registry
 }
 
+const (
+	ed25519KeyType     = "Ed25519VerificationKey2018"
```
troyronda: I assume hardcoding the key type is temporary?

pkg/didcomm/packager/packager.go
```diff
@@ -52,11 +55,13 @@ func New(ctx Provider) (*Packager, error) {
 
 	basePackager.primaryPacker = ctx.PrimaryPacker()
 	if basePackager.primaryPacker == nil {
-		return nil, fmt.Errorf("need primary primaryPacker to initialize packager")
+		return nil, fmt.Errorf("need primary packer to initialize packager")
 	}
 
 	basePackager.addPacker(basePackager.primaryPacker)
 
+	basePackager.connectionStore = ctx.DIDConnectionStore()
```
troyronda: it seems odd to see this direct mutation of the field.

why isn't packager constructed with the store?

pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
```diff
@@ -331,10 +331,14 @@ func TestEncrypt(t *testing.T) {
 			[][]byte{base58.Decode(recSign)})
 		require.NoError(t, err)
 
-		msgOut, sendKey, err := recPacker.Unpack(enc)
+		msgOut, sendKey, recKey, err := recPacker.Unpack(enc)
 		require.NoError(t, err)
 		require.Equal(t, msgIn, msgOut)
 		require.Equal(t, base58.Encode(sendKey), encSign)
+		// this won't work, since input keys and output keys to this Packer
```
troyronda: what’s the resolution to this comment?

pkg/didcomm/packer/jwe/authcrypt/unpack.go
```diff
@@ -22,22 +22,22 @@ import (
 // encrypted CEK.
 // The current recipient is the one with the sender's encrypted key that successfully
 // decrypts with recipientKeyPair.Priv Key.
-func (p *Packer) Unpack(envelope []byte) ([]byte, []byte, error) {
+func (p *Packer) Unpack(envelope []byte) ([]byte, []byte, []byte, error) {
```
troyronda: there were already too many return values. perhaps a struct would be in order?

pkg/didcomm/packer/legacy/authcrypt/authcrypt_test.go
```diff
@@ -698,7 +703,8 @@ func Test_getCEK(t *testing.T) {
 		},
 	}
 
-	_, _, err := getCEK(recs, &k)
+	// nolint: dogsled
```
troyronda: perhaps a struct return would help?

pkg/didcomm/protocol/didexchange/persistence.go
```diff
@@ -54,14 +55,15 @@ func (r *ConnectionRecord) isValid() error {
...
-	return &ConnectionRecorder{transientStore: transientStore, store: store}
+func NewConnectionRecorder(transientStore, store storage.Store, didMap didconnection.Store) *ConnectionRecorder {
+	return &ConnectionRecorder{transientStore: transientStore, store: store, didMap: didMap}
 }
 
 // ConnectionRecorder takes care of connection related persistence features
 type ConnectionRecorder struct {
 	transientStore storage.Store
 	store          storage.Store
+	didMap         didconnection.Store
```
troyronda: why didMap rather than didStore?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -636,7 +640,11 @@ func (s *Service) CreateImplicitInvitation(inviterLabel, inviterDID, inviteeLabe
 		return "", fmt.Errorf("resolve public did[%s]: %w", inviterDID, err)
 	}
 
-	dest, err := prepareDestination(didDoc)
+	dest, err := service.MakeDestination(didDoc, didCommServiceType, ed25519KeyType)
```
troyronda: todo: hardcoded key type.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -404,6 +405,11 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
 			return nil, nil, fmt.Errorf("resolve public did[%s]: %w", pubDID, err)
 		}
 
+		err = ctx.connectionStore.didMap.SaveDIDFromDoc(didDoc, didCommServiceType, ed25519KeyType)
```
troyronda: 3 references above the func call might call for a refactor.

pkg/didcomm/common/didconnection/didconnection.go
Moopli: Another name would be welcome - `Store` is the interface name, and I didn't want a stuttering name. I can pull the interface out into an api package in the framework (which would allow this struct type to be named `Store`), though the interface is only necessary to allow mocking, and isn't intended to be implemented and injected by the client.

pkg/didcomm/dispatcher/outbound.go
Moopli: I can make a TODO and issue - this question extends beyond the scope of this PR, and I aimed to just carry over the current design with these constants. The key type needs to be flexible, and our resolution of keys from DID docs seems to already support multiple key types, but these constants need to be replaced by another system (the communication layer recognizing the key type being used by the connection, and fetching the appropriate keys from DID docs).

pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
Moopli: The JWE packer would need to be changed to use the same key format as `Pack()` parameters and `Unpack()` return values, so the keys returned by `Unpack()` are keys passed into `Pack()` rather than the Curve25519 conversions of given Ed25519 keys. With an agent that supports multiple formats for encryption keys, we could do this - enforce that the JWE packer uses Curve25519 keys as parameters/return values, and that it only be chosen for DIDComm channels that have Curve25519 pubkeys available on both ends.

pkg/didcomm/packer/jwe/authcrypt/unpack.go
Moopli: Also solves dogsledding in tests.

pkg/didcomm/protocol/didexchange/persistence.go
Moopli: This is an alternate name I had for it. Consistency suggests that they should match, but I'm still unsure what name to use. DIDMap? didconnection.Store? didlookup.Store? All alternatives I've tried using already.

pkg/didcomm/common/didconnection/didconnection.go
```diff
@@ -0,0 +1,126 @@
...
+	err = json.Unmarshal(bytes, &record)
+	if err != nil {
+		return "", err
+	}
+
+	return record.DID, nil
+}
+
+// SaveDIDConnection saves a connection between this agent's did and another agent
+func (c *BaseDIDConnectionStore) SaveDIDConnection(myDID, theirDID string, theirKeys []string) error {
```
troyronda: Where are the Lookup functions?

pkg/didcomm/common/service/service.go
troyronda: A Lookup function is needed in the didconnection package to make this more useful. See prior comment.

pkg/didcomm/common/didconnection/didconnection.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	err = c.SaveDID(doc.ID, keys...)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
soluchok: could be simplified or wrapped `return c.SaveDID(doc.ID, keys...)`

pkg/didcomm/common/didconnection/didconnection_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestBaseConnectionStore(t *testing.T) {
+	prov := ctx{
+		store: mockstorage.NewMockStoreProvider(),
+		vdr: &mockvdri.MockVDRIRegistry{
+			CreateErr:    nil,
+			CreateValue:  mockdiddoc.GetMockDIDDoc(),
+			MemStore:     nil,
+			PutErr:       nil,
+			ResolveErr:   nil,
```
soluchok: better to remove these fields with nil e.g `PutErr:       nil,`

pkg/didcomm/common/service/destination.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return MakeDestination(didDoc, serviceType, keyType)
+}
+
+// MakeDestination makes a Destination object from a DID Doc, following the given parameters.
+func MakeDestination(didDoc *diddoc.Doc, serviceType, keyType string) (*Destination, error) {
+	didCommService, err := diddoc.GetDIDCommService(didDoc, serviceType)
```
soluchok: FYI: I don't need these functions anymore :) I am going to use `SendToDID(msg interface{}, myDID, theirDID string)`

pkg/didcomm/packager/packager.go
```diff
@@ -146,10 +150,27 @@ func (bp *Packager) UnpackMessage(encMessage []byte) (*transport.Envelope, error
...
 	if err != nil {
 		return nil, fmt.Errorf("unpack: %w", err)
 	}
 
-	return &transport.Envelope{Message: data, FromVerKey: base58.Encode(senderVerKey)}, nil
+	theirDID, err := bp.connectionStore.GetDID(base58.Encode(senderVerKey))
+	if err != nil {
+		//	ignore error - agents can communicate without using DIDs - for example, in DIDExchange
+		theirDID = ""
+	}
```
soluchok: theirDID should be empty by default
```
// ignore error - agents can communicate without using DIDs - for example, in DIDExchange
theirDID, _ := bp.connectionStore.GetDID(base58.Encode(senderVerKey))
```

pkg/didcomm/packager/packager.go
```diff
@@ -146,10 +150,27 @@ func (bp *Packager) UnpackMessage(encMessage []byte) (*transport.Envelope, error
...
+
+	myDID, err := bp.connectionStore.GetDID(base58.Encode(recVerKey))
+	if err != nil {
+		// ignore error - at beginning of DIDExchange, you might be about to generate a DID
+		myDID = ""
+	}
```
soluchok: same or log the error (if it is needed)

pkg/didcomm/packer/jwe/authcrypt/unpack.go
```diff
@@ -22,22 +22,22 @@ import (
...
 	err := json.Unmarshal(envelope, jwe)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unpack: %w", err)
+		return nil, nil, nil, fmt.Errorf("unpack: %w", err)
 	}
 
 	recipientPubKey, recipient, err := p.findRecipient(jwe.Recipients)
 	if err != nil {
-		return nil, nil, fmt.Errorf("unpack: %w", err)
+		return nil, nil, nil, fmt.Errorf("unpack: %w", err)
```
soluchok: please, change the error message. The error above has the same message.

pkg/didcomm/common/didconnection/didconnection.go
Moopli: Every function saves DIDs under some type of key - `didconnection.Store.GetDID()` is the lookup for all of these.

pkg/didcomm/common/service/destination.go
Moopli: I'm using GetDestination inside SendToDID now :)

And MakeDestination is still useful for DIDExchange.

soluchok: just to clarify, `GetDID` function always returns myDID, right?.
e.g `GetDID(myDID) returns myDID` or `GetDID(theirDID) returns myDID`

pkg/didcomm/common/didconnection/didconnection.go
soluchok: just to clarify, `GetDID` function always returns myDID, right?.
e.g `GetDID(myDID)` returns `myDID` or `GetDID(theirDID)` returns `myDID`

Moopli: `GetDID(myDID)`, as the code is right now, would not find anything, since `myDID` is not used as a key for the store.

Moopli:
I've split the commits, and will make a separate PR for the first portion.

pkg/didcomm/common/didconnection/didconnection.go
troyronda: If a package's only purpose is to implement a struct, it is okay to have that stutter.
I don't think `Base` adds any value.

(I'm also not understanding the conflict with the Store interface defined in another package?)

Moopli:
Coverage almost there. Next, will merge in updates (generic connection store), and test that DIDs are saved correctly for resolution.

pkg/didcomm/common/didconnection/didconnection.go
Moopli: The conflict would be with the interface in the same package - `didconnection.Store` in `didconnection/api.go`. But I've renamed to ConnectionStore so there's no conflict.

pkg/didcomm/packer/jwe/authcrypt/unpack.go
Moopli: Struct now used - the same Envelope struct that the packager returns is now created by the packer.

pkg/didcomm/packer/legacy/authcrypt/authcrypt_test.go
Moopli: Made a hidden struct for these return values.

pkg/didcomm/protocol/didexchange/service.go
Moopli: Hardcoding is no longer here.

pkg/didcomm/protocol/didexchange/states.go
Moopli: I will make an issue for this one - (permanent) connectionStore and the DIDConnection Store should be merged in another PR.

pkg/didcomm/protocol/didexchange/persistence.go
Moopli: Renamed to didStore.

pkg/didcomm/packer/jwe/authcrypt/unpack.go
Moopli: changed the error above.

pkg/didcomm/common/didconnection/didconnection.go
```diff
@@ -0,0 +1,146 @@
...
+
+type provider interface {
+	StorageProvider() storage.Provider
+	VDRIRegistry() vdri.Registry
+}
+
+// New returns a new did lookup Store
+func New(ctx provider) (*ConnectionStore, error) {
+	store, err := ctx.StorageProvider().OpenStore("con-store")
```
troyronda: Would "didconnection" make sense as the string for the store?

pkg/didcomm/common/service/destination.go
```diff
@@ -24,7 +24,8 @@ type Destination struct {
 
 const (
 	didCommServiceType = "did-communication"
-	ed25519KeyType     = "Ed25519VerificationKey2018"
+	// TODO: hardcoded key type
```
troyronda: Is there more context on this TODO? Why hardcoded?

pkg/didcomm/common/service/service.go
Moopli: Since myDID and theirDID are parameters to HandleInbound, we don't need the theirDID -> myDID map. As parameters to HandleInbound, the DIDs can now be any values for any connection, even if theirDID is used in multiple connections.

pkg/didcomm/common/transport/envelope.go
```diff
@@ -6,9 +6,14 @@ SPDX-License-Identifier: Apache-2.0
...
 package transport
 
-// Envelope contain msg, FromVerKey and ToVerKeys
+// Envelope holds message data and metadata for inbound and outbound messaging
 type Envelope struct {
 	Message    []byte
-	FromVerKey string
-	ToVerKeys  []string
+	FromVerKey []byte
+	// ToVerKeys stores string (base58) verification keys for an outbound message
```
troyronda: It's probably unrelated to this issue, but should better understand the selection of base58.

pkg/didcomm/dispatcher/outbound.go
Moopli: Commented, it's part of the same hardcoded key type issue.

pkg/didcomm/common/transport/envelope.go
Moopli: Bitcoin-format base58 is our default way to encode keys as strings in the codebase. Moving to a key handle format would allow us to have a `ToString()` method and FromString constructor, for example, so all conversions back and forth are consistent.

Except when they aren't because sometimes we need to convert to Base64, but those can be the exception.

pkg/didcomm/common/service/destination.go
Moopli: Figured out that this is not necessary at all, we want all keys in a DID doc to point to the DID, so we don't want to filter at all, here.

Moopli: When we're using another key type for the communication, we need to remove the hardcoding of the key type and instead set it based on some connection parameters.

pkg/didcomm/protocol/didexchange/states.go
Moopli: #1004

pkg/didcomm/common/transport/envelope.go
Moopli: Created #1007 for this.

pkg/didcomm/common/service/destination.go
Moopli: created issue #1008

pkg/didcomm/packager/packager.go
```diff
@@ -146,10 +150,21 @@ func (bp *Packager) UnpackMessage(encMessage []byte) (*transport.Envelope, error
...
-	data, senderVerKey, err := p.Unpack(encMessage)
+	envelope, err := p.Unpack(encMessage)
 	if err != nil {
 		return nil, fmt.Errorf("unpack: %w", err)
 	}
 
-	return &transport.Envelope{Message: data, FromVerKey: base58.Encode(senderVerKey)}, nil
+	//	ignore error - agents can communicate without using DIDs - for example, in DIDExchange
+	// nolint: errcheck
+	theirDID, _ := bp.connectionStore.GetDID(base58.Encode(envelope.FromVerKey))
```
troyronda: Suspicious about not even examining the error for the allowed errors in this situation.
Similarly with L164.

pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
troyronda: Is this a TODO?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -404,6 +405,12 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
 			return nil, nil, fmt.Errorf("resolve public did[%s]: %w", pubDID, err)
 		}
 
+		// TODO: x.y.z.foo
```
troyronda: ?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -415,6 +422,16 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
 		return nil, nil, fmt.Errorf("create %s did: %w", didMethod, err)
 	}
 
+	// TODO: initialize did map (or mock) in all eleventy billion tests
```
troyronda: ?

pkg/didcomm/protocol/didexchange/states.go
Moopli: I'll remove the TODO - this is part of issue #1004

pkg/didcomm/protocol/didexchange/states.go
Moopli: purged this check now.
framework initializes this properly, and we don't put checks in to catch broken tests.

pkg/didcomm/packager/packager.go
Moopli: This is done now.

PR #939, opened 2019-12-06
feat: A cache of custom credential schemas of Verifiable Credential

### LINKED ISSUE #185: A cache of custom credential schemas of Verifiable Credential
As a follow-up issue of #172 , we need to cache custom schemas of Verifiable Credential-s.

kdimak:
closes #185

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 package verifiable
 
 import (
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -13,6 +14,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/VictoriaMetrics/fastcache"
 	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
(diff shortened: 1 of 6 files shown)


DRK3:
@kdimak Just a general question - can you elaborate on the need for custom caching of custom schemas? I'm still new to this so I'm just trying to understand.

pkg/doc/verifiable/credential.go
```diff
@@ -619,8 +667,28 @@ func getSchemaLoader(schemas []TypedID, opts *credentialOpts) (gojsonschema.JSON
...
+		return cachedBytes, nil
+	}
+
+	schemaBytes, err := loadJSONSchema(url, client)
+	if err != nil {
+		return nil, err
+	}
+
+	// Put the loaded schema into cache
+	jsonSchemaCache.put(url, schemaBytes)
```
sandrask: Are you supposed to put it into cache even though useCache is false?

pkg/doc/verifiable/credential.go
```diff
@@ -188,7 +189,41 @@ const jsonSchema2018Type = "JsonSchemaValidator2018"
 const vpType = "VerifiablePresentation"
 
 //nolint:gochecknoglobals
-var defaultSchemaLoader = gojsonschema.NewStringLoader(defaultSchema)
+var (
+	defaultSchemaLoader = gojsonschema.NewStringLoader(defaultSchema)
+
+	// jsonSchemaCache keeps cache of JSON schema by URL
+	jsonSchemaCache = schemaCache{schemaBytes: make(map[string][]byte)}
```
sandrask: @fqutishat  Should we use different implementation of cache?

fqutishat: @kdimak  use https://github.com/VictoriaMetrics/fastcache

fqutishat: @troyronda

pkg/doc/verifiable/credential.go
DRK3: One more question: Would we ever _not_ want to use the cache?

pkg/doc/verifiable/credential_test.go
```diff
@@ -678,18 +687,47 @@ func TestDownloadCustomSchema(t *testing.T) {
...
+			require.NoError(t, err)
+			loadsCount++
+		}))
+		defer func() { testServer.Close() }()
+
+		customSchema, err := getJSONSchema(testServer.URL, true, &http.Client{})
+		require.NoError(t, err)
+		require.Equal(t, []byte("custom schema"), customSchema)
+
+		// Check that schema was downloaded only once - i.e. the cache was uses second time
```
DRK3: Typo - change "uses" to "used"

kdimak:
> @kdimak Just a general question - can you elaborate on the need for custom caching of custom schemas? I'm still new to this so I'm just trying to understand. Is it just for improving performance?

Right.

pkg/doc/verifiable/credential.go
kdimak: > Are you supposed to put it into cache even though useCache is false?

It's checked at the beginning of the function:
```
	if !useCache {
		return loadJSONSchema(url, client)
	}
```

kdimak: > One more question: Would we ever _not_ want to use the cache?

I don't know actually :)
By default it's disabled, so maybe it's worth to revert this.
@troyronda what do you think?

DRK3: @sandrask ah, looks like the function returns, so line 686 won't happen.

sandrask: You are right :)

pkg/doc/verifiable/credential.go
kdimak: @sandrask @fqutishat @troyronda I checked `aries` and `fabric`, did not find any usage of an external cache library like https://github.com/VictoriaMetrics/fastcache.
And we have `memStore` struct at pkg/storage/mem/mem_store.go which I looked at and reworked for schemas cache here.

I don't know https://github.com/VictoriaMetrics/fastcache library. If you recommend it, why not to make use of it?

fqutishat: @kdimak we don't use any caching tools yet your PR is the first one

fqutishat: @kdimak  https://github.com/hyperledger/fabric/commit/3a8dae88b9b676856da71d2a5af20ce97212c553

kdimak:
Action items after talking to @troyronda :
- Use https://github.com/VictoriaMetrics/fastcache
- Disable schema caching by default
- Enable it when a user defines caching parameters (e.g. default expiration policy and default cache size limit).

troyronda:
We could potentially enable by default once we have some sane parameters (later).

kdimak:
> We could potentially enable by default once we have some sane default parameters (followup).

I added default size=32Mb (low size bound in **fastcache**) and 1 hour expiration.
@troyronda @fqutishat

pkg/doc/verifiable/credential.go
```diff
@@ -188,7 +190,52 @@ const jsonSchema2018Type = "JsonSchemaValidator2018"
 const vpType = "VerifiablePresentation"
 
 //nolint:gochecknoglobals
-var defaultSchemaLoader = gojsonschema.NewStringLoader(defaultSchema)
+var (
```
troyronda: This package should be refactored to remove globals.

pkg/doc/verifiable/credential.go
```diff
@@ -319,6 +367,32 @@ func WithPublicKeyFetcher(fetcher PublicKeyFetcher) CredentialOpt {
...
+}
+
+// JSONSchemaCacheParam defines JSON schema cache parameters like expiration and cache size limit.
+type JSONSchemaCacheParam struct {
+	Expiration time.Duration
+	Size       int
+}
+
+// ConfigureCustomJSONSchemasCache defines JSON schema cache parameters.
+func ConfigureCustomJSONSchemasCache(param JSONSchemaCacheParam) {
```
troyronda: this function looks dangerous. Mutations without locks.

it would be best to refactor to remove globals. at that point, an instance could be constructed and I doubt you would need a configure function.

kdimak: The schema cache is expected to be used by multiple invocations of `verifiable.NewCredential()`. There is no sense to keep the cache inside `verifiable.Credential` struct as we use downloaded schema content only once while decoding VC. Please advise @troyronda  how to avoid globals in this case.

kdimak: Globals are removed. Thanks @troyronda !

DRK3: @kdimak I still see a "noglobals" in one of the files. Not sure if that one should still be there - just wanted to give you a heads up to double check :)

pkg/doc/verifiable/credential.go
```diff
@@ -319,6 +359,22 @@ func WithPublicKeyFetcher(fetcher PublicKeyFetcher) CredentialOpt {
 	}
 }
 
+// WithJSONSchemaCacheParam option is used to apply JSON Schema caching.
+func WithJSONSchemaCacheParam(param *JSONSchemaCacheParam) CredentialOpt {
```
troyronda: The option should accept a constructed cache so that caches can be shared.

troyronda: It could also make sense to also have the schema loader as well so that there are no globals. In this case, The object passed-in would be a wrapper of cache and schema loader.

kdimak: Makes sense @troyronda , I missed this point, shame on me.
I pushed a new version, no globals, the cache is configurable.

PR #946, opened 2019-12-09
feat: Introduce - add To structure to the proposal messages

soluchok:
* ability to fill To structure for Proposal messages
* tests were refactored

### DIFF
--- a/pkg/didcomm/protocol/introduce/service_test.go
+++ b/pkg/didcomm/protocol/introduce/service_test.go
@@ -4,7 +4,7 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package introduce
+package introduce_test
 
 import (
 	"encoding/json"
@@ -21,8 +21,10 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	dispatcherMocks "github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher/gomocks"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
-	mocks "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce/gomocks"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
+	introduceMocks "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce/gomocks"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 	storageMocks "github.com/hyperledger/aries-framework-go/pkg/storage/gomocks"
 )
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -126,8 +126,8 @@ comma:= ,
 semicolon:= ;
 
 define create_mock
-  mkdir -p $(1)/gomocks && rm -rf $(1)/gomocks/*
-  $(MOCKGEN) -destination $(1)/gomocks/mocks.go -self_package mocks -package mocks $(PROJECT_ROOT)/$(1) $(subst $(semicolon),$(comma),$(2))
+  rm -rf $(1)/gen_mocks.go
+  $(MOCKGEN) -destination $(1)/gen_mocks.go -self_package $(PROJECT_ROOT)/$(1) -package $(2) $(PROJECT_ROOT)/$(1) $(subst $(semicolon),$(comma),$(3))
```
troyronda: Mocks do not belong in the same package as the implementation.

pkg/didcomm/protocol/introduce/states_test.go
```diff
@@ -1,508 +0,0 @@
...
-
-	"github.com/golang/mock/gomock"
-	"github.com/stretchr/testify/require"
-
-	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
-	dispatcherMocks "github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher/gomocks"
-	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
-	mocks "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce/gomocks"
-)
-
```
troyronda: what happened to these tests?

soluchok: Reverted!

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -305,27 +314,31 @@ func (s *abandoning) CanTransitionTo(next state) bool {
 	return next.Name() == stateNameDone
 }
 
-func fillMissingDestination(destinations []*service.Destination) []*service.Destination {
-	if len(destinations) <= 1 {
-		return append([]*service.Destination{getInboundDestination()}, destinations...)
+func fillMissingDestination(recipients []*Recipient) []*Recipient {
```
troyronda: should this func be renamed?

soluchok: No, it is fine. The function sets the destination for the inbound message.

troyronda: It seems to be filling in the missing recipient ... that happens to have a destination.

soluchok: so, this function adds recipient(with destination) or sets the destination (if the recipient doesn't have it). But anyway it is related because of the missing destination.

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -138,26 +139,32 @@ func (s *arranging) CanTransitionTo(next state) bool {
...
+
 	if approve, ok := getApproveFromMsg(m.Msg); ok && !approve {
 		return &abandoning{}, nil
 	}
 
-	destinations := m.dependency.Destinations()
+	recipients := fillMissingDestination(m.dependency.Recipients())
 
-	var destination *service.Destination
+	recipient := recipients[len(recipients)-1]
```
troyronda: this could use a comment.

troyronda: along with L158

pkg/didcomm/protocol/introduce/service_test.go
```diff
@@ -367,81 +425,24 @@ func TestService_Accept(t *testing.T) {
...
 
 	require.False(t, svc.Accept(""))
-	require.True(t, svc.Accept(ProposalMsgType))
-	require.True(t, svc.Accept(RequestMsgType))
-	require.True(t, svc.Accept(ResponseMsgType))
-	require.True(t, svc.Accept(AckMsgType))
-	require.True(t, svc.Accept(ProblemReportMsgType))
-}
-
-func Test_stateFromName(t *testing.T) {
```
troyronda: did these tests get moved?

soluchok: Reverted! FYI: This test will not increase coverage. This is why I thought it was redundant. But anyway better to have more tests :)

pkg/didcomm/protocol/introduce/states.go
soluchok: Refactored!
https://github.com/hyperledger/aries-framework-go/pull/946/files#diff-6c72066e0cf828585a690ba919b45ad7R157-R161
https://github.com/hyperledger/aries-framework-go/pull/946/files#diff-6c72066e0cf828585a690ba919b45ad7R234-R235

PR #955, opened 2019-12-11
feat: Introduce - forward an invitation & notify the introduce

### LINKED ISSUE #954: Introduce - forward an invitation & notify the introduce after received an invitation
Forward an invitation can be done by using `OutboundDispatcher().Send` function.
After receiving an invitation message (didexchange) notify the introduce protocol about it. The invitation structure (model) should be extended with `*decorator.Thread`. Thread filed should contain `pthid`. Notification must be done by using `InvitationReceived` function from the introduce protocol.

soluchok:
* listening `digexchange` events
* forward an invitation is done by using OutboundDispatcher().Send

Closes #954 

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -38,6 +38,8 @@ const (
 	AckMsgType = IntroduceSpec + "ack"
 	// ProblemReportMsgType defines the introduce problem-report message type.
 	ProblemReportMsgType = IntroduceSpec + "problem-report"
+	// stateInvited the didexchange protocol state name to determine specific event
+	stateInvited = "invited"
 )
 
 const (
@@ -104,13 +106,15 @@ type record struct {
 type Service struct {
 	service.Action
 	service.Message
-	store       storage.Store
-	callbacks   chan *metaData
-	ctx         internalContext
-	wg          sync.WaitGroup
-	stop        chan struct{}
-	closedMutex sync.Mutex
-	closed      bool
+	store           storage.Store
+	callbacks       chan *metaData
+	didEvent        chan service.StateMsg
+	didEventService service.Event
+	ctx             internalContext
+	wg              sync.WaitGroup
+	stop            chan struct{}
+	closedMutex     sync.Mutex
+	closed          bool
 }
 
 // Provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context()
@@ -118,12 +122,6 @@ type Provider interface {
 	OutboundDispatcher() dispatcher.Outbound
 	StorageProvider() storage.Provider
 	Service(id string) (interface{}, error)
-}
-
-// Forwarder provides the possibility to forward an invitation
-type Forwarder interface {
-	// method should be implemented in didexchange service
-	SendInvitation(pthID string, inv *didexchange.Invitation, dest *service.Destination) error
 }
 
 // New returns introduce service
@@ -135,22 +133,27 @@ func New(p Provider) (*Service, error) {
 
 	didSvc, err := p.Service(didexchange.DIDExchange)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("load the DIDExchange service: %w", err)
 	}
 
-	forwarderSvc, ok := didSvc.(Forwarder)
+	didService, ok := didSvc.(service.Event)
 	if !ok {
-		return nil, errors.New("cast service to forwarder service failed")
+		return nil, fmt.Errorf("cast service to service.Event")
 	}
 
 	svc := &Service{
 		ctx: internalContext{
-			Outbound:  p.OutboundDispatcher(),
-			Forwarder: forwarderSvc,
+			Outbound: p.OutboundDispatcher(),
 		},
-		store:     store,
-		callbacks: make(chan *metaData),
-		stop:      make(chan struct{}),
+		store:           store,
+		didEventService: didService,
+		callbacks:       make(chan *metaData),
+		didEvent:        make(chan service.StateMsg),
+		stop:            make(chan struct{}),
+	}
+
+	if err = svc.didEventService.RegisterMsgEvent(svc.didEvent); err != nil {
+		return nil, fmt.Errorf("did register msg event: %w", err)
 	}
 
 	// start the listener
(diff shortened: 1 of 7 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -78,11 +82,42 @@ type Service struct {
...
 type context struct {
+	IntroduceNotifier
 	outboundDispatcher dispatcher.Outbound
 	signer             kms.Signer
 	connectionStore    *ConnectionRecorder
 	vdriRegistry       vdriapi.Registry
+	service            func(id string) (interface{}, error)
+}
+
+func (ctx *context) loadIntroduceNotifier() error {
```
troyronda: This seems backwards.

Why wouldn't Introduce register itself with DID Exchange?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -175,6 +175,21 @@ func (s *requested) ExecuteInbound(msg *stateMachineMsg, thid string, ctx *conte
 			return nil, nil, nil, fmt.Errorf("JSON unmarshalling of invitation: %w", err)
 		}
 
+		// Invitation received
+		// TODO: For now, PID can be set only by the introduce protocol.
+		//  if Thread.PID is set, we need to notify the introduce protocol about this action.
+		//  This logic should be changed with some generic solution. What if another protocol sets the PID?
+		//  How to determine which protocol provides PID? Which action we should take?
+		if invitation.Thread != nil && invitation.Thread.PID != "" {
+			if err = ctx.loadIntroduceNotifier(); err != nil {
```
troyronda: The notifier hook is being modified in the middle of the state machine?

pkg/didcomm/protocol/didexchange/service.go
soluchok: Do you mean just create a callback function?

soluchok: Yeah, it makes sense.

pkg/didcomm/protocol/didexchange/states.go
soluchok: Done!

pkg/didcomm/protocol/didexchange/service.go
soluchok: Done!

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -174,6 +174,8 @@ func (s *requested) ExecuteInbound(msg *stateMachineMsg, thid string, ctx *conte
 		if err != nil {
 			return nil, nil, nil, fmt.Errorf("JSON unmarshalling of invitation: %w", err)
 		}
+		// sets invitation
+		msg.invitation = invitation
```
llorllale: modifying input parameters is generally bad practice.

soluchok: I did this to improve performance. Otherwise, we need to unmarshal the message every time.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -125,6 +129,13 @@ func New(prov provider) (*Service, error) {
 	return svc, nil
 }
 
+// RegisterInvitationReceivedCallback registers callback function
+func (s *Service) RegisterInvitationReceivedCallback(fn func(*Invitation)) {
```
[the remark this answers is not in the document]
troyronda: Also - I'm curious why a callback function rather than a channel?

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -291,6 +307,15 @@ func (s *Service) handle(msg *message, aEvent chan<- service.DIDCommAction) erro
 		}
 	}
 
+	// notifies the introduce protocol that invitation was received
+	if msg.Msg.Header.Type == InvitationMsgType && invitation != nil {
+		s.inrcMutex.RLock()
+		if s.invitationReceivedCallback != nil {
+			go s.invitationReceivedCallback(invitation)
```
troyronda: This seems better modelled as a go channel, rather than spinning up a goroutine.

pkg/didcomm/protocol/didexchange/service.go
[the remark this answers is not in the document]
soluchok: `Will we need to consider unregister?` - yeah, it would be nice. I'll add it.
Will we need to eventually handle multiple registrations? - no, it seems to be redundant. 

I can replace it with a channel.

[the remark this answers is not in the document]
soluchok: Actually, it looks like I can use `RegisterMsgEvent`.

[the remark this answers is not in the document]
troyronda: Yes - Using the existing event mechanism is even better!

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -248,22 +257,30 @@ func (s *Service) doHandle(msg *service.DIDCommMsg, outbound bool) (*metaData, e
...
 	}
 
-	msg, err := service.NewDIDCommMsg(payload)
+	didMsg, err := service.NewDIDCommMsg(payload)
 	if err != nil {
 		return fmt.Errorf("invitation received new DIDComm msg: %w", err)
 	}
 
-	_, err = s.HandleInbound(msg)
+	_, err = s.HandleInbound(didMsg)
```
troyronda: i assume the service will silently ignore unrelated threads?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -248,22 +257,30 @@ func (s *Service) doHandle(msg *service.DIDCommMsg, outbound bool) (*metaData, e
 
 // InvitationReceived is used to finish the state machine
 // the function should be called by didexchange after receiving an invitation
-func (s *Service) InvitationReceived(thID string) error {
+func (s *Service) InvitationReceived(msg service.StateMsg) error {
+	if msg.StateID != "invited" || msg.Type != service.PostState {
```
troyronda: invited should be a const.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -248,22 +259,30 @@ func (s *Service) doHandle(msg *service.DIDCommMsg, outbound bool) (*metaData, e
 
 // InvitationReceived is used to finish the state machine
 // the function should be called by didexchange after receiving an invitation
-func (s *Service) InvitationReceived(thID string) error {
+func (s *Service) InvitationReceived(msg service.StateMsg) error {
+	if msg.StateID != stateInvited || msg.Type != service.PostState {
```
llorllale: I'd rather you use `didexchange.InvitationMsgType` rather than this

PR #958, opened 2019-12-12
feature: Add Service Provider Interfaces (SPIs) and default Crypto Provider implementation

Baha-sk:
This commit is a first plugins introductions to Aries-Framework-Go
That includes interface definitions for Crypto, KMS and JWE SPIs

a default Crypto provider implementation is added and may need to be updated (probably for key derivation operations)

The plugins will be hooked into the framework once they're tested with
at least a tested implementaiton for each.


pkg/plugins/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+type Provider interface {
+}
+
+// Creator method to create new pluggable Crypto service
+type Creator func(prov Provider) (AriesCrypto, error)
+
+// AriesCrypto is an interface defining a pluggable Crypto operations provider to Aries agents.
+// It's only responsible for pure crypto operations and nothing more. Envelope building and keys management are managed
+// in other interfaces
+type AriesCrypto interface {
```
troyronda: The Aries prefix isn't needed.

pkg/plugins/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+	// Encrypt will encrypt msg and aad using a matching AEAD primitive in kh
+	// returns:
+	// 		cipherText in []byte
+	//		nonce in []byte
+	//		error in case of errors during encryption
+	Encrypt(msg, aad []byte, kh *keyset.Handle) ([]byte, []byte, error)
```
troyronda: Is `keyset.Handle` extendable by 3rd party crypto implementations?

pkg/plugins/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+	// Decrypt will decrypt cipher with aad and given nonce using a matching AEAD primitive in kh
+	// returns:
+	//		plainText in []byte
+	//		error in case of errors
+	Decrypt(cipher, add, nonce []byte, kh *keyset.Handle) ([]byte, error)
```
troyronda: typo? `add`

pkg/plugins/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
+
+import "github.com/google/tink/go/keyset"
+
+// Provider interface for Crypto ctx
+type Provider interface {
```
troyronda: empty interface?

pkg/plugins/crypto/impl/crypto.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package impl provides the default implementation of pkg/plugins/crypto.Crypto interface
+//
+// It uses github.com/tink/go crypto primitives
+package impl
```
troyronda: This is not a useful package name.

You should restructure - e.g, have a default or impl parent folder and put a crypto inside it.

pkg/plugins/crypto/impl/crypto.go
```diff
@@ -0,0 +1,98 @@
...
+	ivSize := getNonceSize(ps)
+	cipherText := ct[len(ps.Primary.Prefix)+ivSize:]
+	nonce := ct[len(ps.Primary.Prefix) : ivSize-1]
+
+	return cipherText, nonce, nil
+}
+
+func getNonceSize(ps *primitiveset.PrimitiveSet) int {
+	ivSize := aeadsubtle.AESGCMIVSize
+	// TODO double check if the right primitive from PrimitiveSet is Primary, or do we have to search for something else
```
troyronda: ?

pkg/plugins/crypto/impl/crypto.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	ivSize := getNonceSize(ps)
+	cipherText := ct[len(ps.Primary.Prefix)+ivSize:]
+	nonce := ct[len(ps.Primary.Prefix) : ivSize-1]
+
+	return cipherText, nonce, nil
+}
+
+func getNonceSize(ps *primitiveset.PrimitiveSet) int {
```
troyronda: this method doesn't seem to `get` anything.

pkg/plugins/crypto/impl/crypto.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return pt, nil
+}
+
+// Sign will sign msg using the implementation's corresponding signing key
+func (t *DefaultCrypto) Sign(msg []byte, kh *keyset.Handle) ([]byte, error) {
+	return nil, errors.New("not implemented yet")
```
troyronda: is this blocked or just not done yet?

pkg/plugins/crypto/main/example.go
```diff
@@ -0,0 +1,33 @@
+package main
```
troyronda: what's the purpose of the package?

if it is useful, should this be in some path under cmd?

pkg/plugins/jwe/api.go
```diff
@@ -0,0 +1,34 @@
...
+// AriesJWE is an interface defining a pluggable JWE builder operations provider to Aries agents.
+// It's responsible for building JWE structures for either 1: sealing them or 2: open serialized ones. It depends on
+// on the AriesCrypto to encrypt/decrypt content and AriesKMS to get key handles needed for crypto operations.
+type AriesJWE interface {
+	// Seal will encrypt and wrap msg for the sender key represented by skh and the recipients keys represented in rkh
+	//  into a JWE envelope then serializes it
+	// returns:
+	//		[]byte base64UrlEncoding of the serialized JWE envelope
+	//		error in case of an error
+	Seal(msg []byte, skh *keyset.Handle, rkh []*keyset.Handle) ([]byte, error)
```
troyronda: Where is the Seal and Open terminology coming from?

pkg/plugins/jwe/api.go
```diff
@@ -0,0 +1,34 @@
...
+type Provider interface {
+}
+
+// Creator method to create new pluggable JWE service
+type Creator func(prov Provider) (AriesJWE, error)
+
+// AriesJWE is an interface defining a pluggable JWE builder operations provider to Aries agents.
+// It's responsible for building JWE structures for either 1: sealing them or 2: open serialized ones. It depends on
+// on the AriesCrypto to encrypt/decrypt content and AriesKMS to get key handles needed for crypto operations.
+type AriesJWE interface {
```
troyronda: Same comment as earlier - We shouldn't add prefixes.

pkg/plugins/kms/AriesKMS.go
```diff
@@ -0,0 +1,27 @@
...
+// Provider interface for Crypto ctx
+type Provider interface {
+}
+
+// Creator method to create new pluggable KMS service
+type Creator func(prov Provider) (AriesKMS, error)
+
+// AriesKMS is an interface defining a pluggable KMS operations provider to Aries agents.
+// It's responsible for managing keys and provide KMS specific crypto operations.
+type AriesKMS interface {
```
troyronda: Same comment as earlier - We shouldn't add prefixes.

pkg/plugins/crypto/api.go
```diff
@@ -0,0 +1,42 @@
+/*
```
troyronda: it seems weird that the interface definitions are under “plugins”. they are neither plugins nor implementations.

these seem to be service provider interfaces for the framework. as is vdri, storage, etc.

pkg/plugins/crypto/api.go
Baha-sk: I wanted to differentiate it from the currently used Crypto interface. I will remove it..

pkg/plugins/crypto/api.go
Baha-sk: from the doc comment:
```
// Handle provides access to a Keyset protobuf, to limit the exposure of actual protocol
// buffers that hold sensitive key material.
```
Tink serializes all keys as protobufs, the Handle has the primitive info (cipher suite) needed for each of the enabled keys in the keyset.

pkg/plugins/crypto/api.go
Baha-sk: right, I'll rename to `aad`

pkg/plugins/crypto/impl/crypto.go
Baha-sk: I wanted to separate the interface api from the default implementation... will move up default implementation

pkg/plugins/crypto/impl/crypto.go
Baha-sk: it returns the nonce size of the primitives we want to support, either AESGCM nonce size or Xchacha20Poly1305 nonce size

pkg/plugins/crypto/impl/crypto.go
Baha-sk: since a PrimitiveSet has a `Primary` Entry and a `set` of entries with corresponding key info, we need to find out if using `Primary` primitive is sufficient to deduct the nonce size... 
at the time of setting `a` at line 33 above, the key and primitive to be used are known, not sure if corresponds to Primary key or not, need find out.

pkg/plugins/crypto/impl/crypto.go
Baha-sk: it's not blocked.. it will be added when it's needed..

pkg/plugins/crypto/main/example.go
Baha-sk: just a Tink example without a KMS set.. will remove

pkg/plugins/jwe/api.go
Baha-sk: will remove

pkg/plugins/jwe/api.go
Baha-sk: it's for a JWE `envelope` which is sealed by encryption and serialization and then opened by deserialization and decryption.

pkg/plugins/kms/AriesKMS.go
Baha-sk: will remove prefix

pkg/plugins/crypto/impl/crypto.go
troyronda: right - this method should not have a get prefix.

pkg/plugins/crypto/api.go
troyronda: suggest, at least, putting these interface folders under a `spi` directory for now (not under plugins).

perhaps:
`pkg/framework/spi/...`

or

`pkg/common/spi/...`

@rolsonquadras @fqutishat @llorllale

pkg/plugins/crypto/api.go
Baha-sk: prefix removed

pkg/plugins/crypto/api.go
Baha-sk: I was thinking to use `*interface{}` type but `keyset.Handle` is a struct

pkg/plugins/crypto/api.go
Baha-sk: renamed

pkg/plugins/crypto/impl/crypto.go
Baha-sk: removed `get` prefix

pkg/plugins/jwe/api.go
Baha-sk: prefix removed

pkg/plugins/kms/AriesKMS.go
Baha-sk: prefix removed

pkg/plugins/crypto/api.go
Baha-sk: sure, will move to `pkg/framework/spi/...`

Baha-sk: moved plugins interfaces to pkg/framework/spi/...

pkg/plugins/crypto/main/example.go
Baha-sk: removed

pkg/plugins/crypto/impl/crypto.go
Baha-sk: moved up... and removed impl

pkg/framework/spi/jwe/api.go
```diff
@@ -0,0 +1,34 @@
...
+import "github.com/google/tink/go/keyset"
+
+// Provider interface for Crypto ctx
+type Provider interface {
+}
+
+// Creator method to create new pluggable JWE service
+type Creator func(prov Provider) (JWE, error)
+
+// JWE is an interface defining a pluggable JWE builder operations provider to Aries agents.
```
troyronda: How does one call Anoncrypt vs Authcrypt?

pkg/plugins/crypto/crypto.go
```diff
@@ -0,0 +1,99 @@
+/*
```
troyronda: @Baha-sk I might need to pick on the term `plugins` now. There are "Go Plugins" and these are not. I think it could be confusing to re-use this term.

These packages are actually the default framework packages for the SPIs.

pkg/plugins/crypto/crypto.go
```diff
@@ -0,0 +1,99 @@
...
+
+	cryptoapi "github.com/hyperledger/aries-framework-go/pkg/framework/spi/crypto"
+)
+
+// Crypto is the default Crypto implementation
+type Crypto struct {
+}
+
+// Assert that Crypto implements the Crypto interface.
+var _ cryptoapi.Crypto = (*Crypto)(nil)
```
troyronda: This assertion should not be in a code file. This is a test.

troyronda: Also odd how the linter missed this global variable. Please avoid globals :).

pkg/framework/spi/jwe/api.go
Baha-sk: these would be separate implementations building their unique JWE envelope

pkg/plugins/crypto/crypto.go
Baha-sk: I didn't think much about the `plugin` package name.

now that we have pkg/framework/spi for service plugin interfaces... we probably need a proper location for the implementations, including this default one. They should be under pkg/framework as these are Aries pluggable services.

pkg/framework/spi/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Creator method to create new pluggable Crypto service
+type Creator func(prov Provider) (Crypto, error)
+
+// Crypto is an interface defining a pluggable Crypto operations provider to Aries agents.
+// It's only responsible for pure crypto operations and nothing more. Envelope building and keys management are managed
+// in other interfaces
+type Crypto interface {
+	// Encrypt will encrypt msg and aad using a matching AEAD primitive in kh
```
rolsonquadras: I'm bit confused with the wording "using a matching AEAD primitive". This is an interface definition - `"using"` word infers implementation. Shouldn't it be up-to the implementors to decide ?  If we want to force it to the implementation, then we `"must"` would be a better word. 

Same comment for other interface fn godocs.

pkg/plugins/crypto/crypto.go
Baha-sk: this statement enforces the implementation to follow the interface functions and will provide a compilation error if it doesn't.. it's useful during refactoring.

the variable name is omitted here :) ... it's just the type that's defined to enforce the type checking, so there is no global variable defined in this line

troyronda: Please move to _test.go.

Our automated tests ensure that this has compliance. It should not be in the code.

pkg/framework/spi/kms/api.go
```diff
@@ -0,0 +1,33 @@
...
+// KMS is an interface defining a pluggable KMS operations provider to Aries agents.
+// It's responsible for managing keys and provide KMS specific crypto operations.
+type KMS interface {
+	// Encrypt will encrypt msg and aad using a matching AEAD primitive in kh
+	// returns:
+	// 		cipherText in []byte
+	//		nonce in []byte
+	//		error in case of errors during encryption
+	// TODO  to be defined according to existing cloud KMSs and local mobile keystores..
+	Encrypt(msg, aad []byte, keyHandle *keyset.Handle) ([]byte, []byte, error)
```
rolsonquadras: The KMS Encrypt/Decrypt functions here has [same](https://github.com/hyperledger/aries-framework-go/pull/958/files#diff-6d57c757cc93b93f22de0729f28d5a52R27) definition as Crypto interface. Why we have same functionality with 2 interfaces ? Is KMS subset of Crypto ?

pkg/plugins/crypto/crypto.go
troyronda: should be under pkg/framework/foo/package (rather than directly under framework), so that we match the spi structure.

pkg/plugins/crypto/crypto.go
```diff
@@ -0,0 +1,99 @@
...
+
+	ps, err := kh.Primitives()
+	if err != nil {
+		return nil, err
+	}
+
+	ct := make([]byte, 0, len(ps.Primary.Prefix)+len(nonce))
+	ct = append(ct, ps.Primary.Prefix...)
+	ct = append(ct, nonce...)
+	ct = append(ct, cipher...)
```
soluchok: looks like you missed to count cipher length as well e.g `len(ps.Primary.Prefix)+len(nonce)+len(cipher)`

Baha-sk: good point, although append will expand the array's capacity on the fly.. I'll add it anyway

pkg/plugins/crypto/crypto.go
troyronda: Not impl - we should make it clear that is only defaults.

pkg/framework/spi/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+	// returns:
+	// 		cipherText in []byte
+	//		nonce in []byte
+	//		error in case of errors during encryption
+	Encrypt(msg, aad []byte, kh *keyset.Handle) ([]byte, []byte, error)
```
llorllale: We lose flexibility by exposing tink to implementations via the API. Plugin architectures are generally meant to be flexible.

pkg/framework/spi/jwe/api.go
```diff
@@ -0,0 +1,34 @@
...
+// It's responsible for building JWE structures for either 1: sealing them or 2: open serialized ones. It depends on
+// on the AriesCrypto to encrypt/decrypt content and AriesKMS to get key handles needed for crypto operations.
+type JWE interface {
+	// Seal will encrypt and wrap msg for the sender key represented by skh and the recipients keys represented in rkh
+	//  into a JWE envelope then serializes it
+	// returns:
+	//		[]byte base64UrlEncoding of the serialized JWE envelope
+	//		error in case of an error
+	Seal(msg []byte, skh *keyset.Handle, rkh []*keyset.Handle) ([]byte, error)
```
llorllale: "Seal" is not an operation I would associate with JWEs as a whole. 

These two operations, Seal and Open, are behaviors of sodium's cryptobox, not of JWEs.

pkg/framework/spi/crypto/api.go
llorllale: I agree - the wording should be removed from the godocs.

pkg/framework/spi/crypto/api.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Provider interface for Crypto ctx
+type Provider interface {
+}
+
+// Creator method to create new pluggable Crypto service
+type Creator func(prov Provider) (Crypto, error)
+
+// Crypto is an interface defining a pluggable Crypto operations provider to Aries agents.
+// It's only responsible for pure crypto operations and nothing more. Envelope building and keys management are managed
```
llorllale: This line is superfluous.

pkg/framework/spi/kms/api.go
llorllale: I agree - if Crypto is a subset of KMS, then KMS *is a* Crypto. The method definitions shouldn't be duplicated.

pkg/plugins/crypto/crypto.go
troyronda: perhaps a parent folder name like `provider` to match service **provider** interface?

pkg/framework/spi/jwe/api.go
troyronda: Agree. I don't find Seal and Open intuitive.

pkg/framework/spi/crypto/api.go
troyronda: Note: here is the definition of keyset.Handle: https://github.com/google/tink/blob/090075d031a55da56bf0551b2e1495d19e38c000/go/keyset/handle.go#L33

pkg/framework/spi/crypto/api.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
```
troyronda: I'm not seeing any functions that actually supply a Keyset handle. How would I use this API?

pkg/framework/spi/crypto/api.go
troyronda: This implementation almost suggests that tink, itself, is being used as the SPI.

i.e., this feels like a 50/50 solution right now. You didn't directly use tink as the Crypto SPI but used some definitions (and particularly a Key Handle - the essence of a Crypto SPI, I would think).

llorllale: "50/50 solution" is spot on.

Baha-sk: yes.. the intention is not to use Tink as an SPI, but to use its services in our pluggable Crypto and KMS APIs

pkg/framework/spi/jwe/api.go
Baha-sk: we usually `Seal` an envelope prior to sending it and `Open` it when we receive it.. just like real world mail envelopes

pkg/framework/spi/crypto/api.go
Baha-sk: that would be in the KMS implementation.. for now unit tests of the crypto implementation shows an example of how a key handle can be obtained (from Tink)

pkg/framework/spi/crypto/api.go
Baha-sk: it's to highlight the limited scope of this interface, ie it's not responsible for getting keys or building jwes

pkg/framework/spi/kms/api.go
Baha-sk: I don't want Crypto to be defined as a subset of KMS or vice versa. They should be treated as two different services providing their own operations. weather KMS offers some crypto operations, this is up to the KMS implementation, for example current KMSs in Tink don't offer Sign/Verify functions but do offer Encrypt/Decrypt. 

So for both interfaces, although they may have the same functions, should define them separately.

pkg/plugins/crypto/crypto.go
Baha-sk: so default implementations go in `pkg/framework/provider/` and interfaces stay in `pkg/framework/spi` ?

pkg/framework/spi/crypto/api.go
Baha-sk: will remove

pkg/plugins/crypto/crypto.go
troyronda: Yes. 

Note: I assume packages in these folders are adapters. Major components like JWE should be in their own package elsewhere in the repo. These provider packages should import these major components.

pkg/framework/spi/crypto/api.go
troyronda: If you are abstracting tink, you would need interfaces in the SPI.

pkg/framework/spi/jwe/api.go
troyronda: This is not common terminology for JWE.

pkg/plugins/crypto/api.go
troyronda: I am not sure I understand the problem. You can wrap a struct and meet an interface in the wrapper.

pkg/framework/spi/jwe/api.go
llorllale: I like really like Baha's phrasing. In that case I would suggest rename JWE -> Envelope.

But then again I believe there is intent to reuse this elsewhere in contexts where JWEs are explicit and center-class?

pkg/plugins/crypto/crypto.go
Baha-sk: JWE package is one of these providers, just like crypto and KMS.

as depicted in the [use case](https://wiki.hyperledger.org/display/ARIES/Crypto+and+KMS+Plugins) diagram, JWE will use Crypto and KMS. 

In the end all three providers are services available to the framework, this is why they're all part of SPI/providers.. I don't see JWE as a major (separate) component that differs in nature from other providers/SPIs.

pkg/framework/spi/crypto/api.go
Baha-sk: replaced interface arguments of `*keyset.Handle` to `interface{}` in all interfaces functions

Baha-sk: done

pkg/framework/spi/crypto/api.go
Baha-sk: done

pkg/plugins/crypto/crypto.go
Baha-sk: done

pkg/framework/spi/kms/api.go
llorllale: Well, your model strongly suggests there is directional IS-A relationship between KMS and Crypto. If not all implementations of KMS will offer Crypto operations then those should be removed from the KMS.

pkg/framework/spi/jwe/api.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package jwe
+
+import "github.com/google/tink/go/keyset"
```
troyronda: I assume you are removing these imports from the rest of the SPI interfaces?

Baha-sk: they're now all removed.

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,135 @@
...
+	ivSize := nonceSize(ps)
+	prefixLength := len(ps.Primary.Prefix)
+	cipherText := ct[prefixLength+ivSize:]
+	nonce := ct[prefixLength : prefixLength+ivSize]
+
+	return cipherText, nonce, nil
+}
+
+func nonceSize(ps *primitiveset.PrimitiveSet) int {
+	ivSize := aeadsubtle.AESGCMIVSize
```
troyronda: this should be part of a switch / case.

pkg/framework/spi/kms/api.go
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+package kms
+
+// Provider interface for Crypto ctx
+type Provider interface {
+}
+
+// Creator method to create new pluggable KMS service
+type Creator func(prov Provider) (KMS, error)
```
troyronda: usage is not clear.

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	ps, err := keyHandle.Primitives()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	ivSize := nonceSize(ps)
```
troyronda: a comment would be nice on why nonce is built this way.

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ps, err := keyHandle.Primitives()
+	if err != nil {
+		return nil, err
+	}
+
+	ct := make([]byte, 0, len(ps.Primary.Prefix)+len(nonce)+len(cipher))
```
troyronda: a comment would be nice on why ct is being built this way.

pkg/framework/provider/crypto/crypto_test.go
```diff
@@ -0,0 +1,139 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
+
+import (
+	"testing"
+
+	chacha "golang.org/x/crypto/chacha20poly1305"
+
```
troyronda: extra space.

pkg/framework/provider/crypto/crypto_test.go
```diff
@@ -0,0 +1,139 @@
...
+	cryptoapi "github.com/hyperledger/aries-framework-go/pkg/framework/spi/crypto"
+)
+
+func TestCrypto(t *testing.T) {
+	// Assert that Crypto implements the Crypto interface.
+	var _ cryptoapi.Crypto = (*Crypto)(nil)
+}
+
+func TestCrypto_EncryptDecrypt(t *testing.T) {
+	// tests XChacha20Poly1305
```
troyronda: Using subtests for each case would be clearer.

pkg/framework/spi/jwe/api.go
troyronda: I suppose we will see how this works in a followup.

pkg/framework/spi/jwe/api.go
troyronda: The intent is to separate the common JWE implementation into own package.

(This common JWE implementation would be imported by the default provider implementations).

pkg/framework/spi/kms/api.go
Baha-sk: the intention is this creator function will be used by the framework provider to add a KMS.. it will be used when plugging in these interfaces into the framework at agent start-up

pkg/framework/spi/jwe/api.go
Baha-sk: as discussed today... JWE should not be an SPI... I will remove this interface and consolidate Crypto and KMS into 1 SPI

pkg/framework/provider/crypto/crypto.go
Baha-sk: sure.. but this is how Tink assembles its cipher text for Decrypt() call.. this implementation is supposed to handle Tink specific format from the interface user where they only need to pass in the raw cipher text, nonce and aad used as is in the JWE envelope.

pkg/framework/provider/crypto/crypto_test.go
Baha-sk: done

pkg/framework/provider/crypto/crypto_test.go
Baha-sk: removed

pkg/common/crypto/api.go
```diff
@@ -0,0 +1,12 @@
+package crypto
+
+import "github.com/hyperledger/aries-framework-go/pkg/framework/spi/crypto"
+
+// package Crypto contains the consolidated Crypto interface to be used by the framework.
+// it will be created via Options creation of pkg/framework/provider/crypto (implementation of
+// pkg/framework/spi/crypto)
+
+// FrameworkCrypto interface provides all crypto operations needed in the Aries framework
+type FrameworkCrypto interface {
```
troyronda: We don't need the prefix `Framework`. This is implied and this prefix hasn't been used in other interfaces.

pkg/common/crypto/default.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"errors"
+
+	"github.com/hyperledger/aries-framework-go/pkg/framework/provider/crypto"
+	spicrypto "github.com/hyperledger/aries-framework-go/pkg/framework/spi/crypto"
+)
+
+// MainCrypto is the main crypto operations provider to the Aries Framework
+type MainCrypto struct {
```
troyronda: Please remove the prefix `Main`.

pkg/common/crypto/default.go
```diff
@@ -0,0 +1,47 @@
...
+	c *crypto.TinkCrypto
+}
+
+// New return new instance of KMS implementation
+func New(cr spicrypto.Crypto) (FrameworkCrypto, error) {
+	tc, ok := (cr).(*crypto.TinkCrypto)
+	if !ok {
+		return nil, errors.New("bad Crypto type")
+	}
```
soluchok: why do we need a specific type?

soluchok:
```
func New(cr spicrypto.Crypto) FrameworkCrypto {
	return &MainCrypto{c: tc}
}
```

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+
+	err = verifier.Verify(sig, msg)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
soluchok: `return verifier.Verify(sig, msg)` or wrap the error

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	s, err := signer.Sign(msg)
+	if err != nil {
+		return nil, err
+	}
+
+	return s, nil
```
soluchok: same `return signer.Sign(msg)` or wrap an error

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,140 @@
...
+	ct = append(ct, ps.Primary.Prefix...)
+	ct = append(ct, nonce...)
+	ct = append(ct, cipher...)
+
+	pt, err := a.Decrypt(ct, aad)
+	if err != nil {
+		return nil, err
+	}
+
+	return pt, nil
```
soluchok: same `return a.Decrypt(ct, aad)` or wrap an error

pkg/common/crypto/default_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/stretchr/testify/require"
+	chacha "golang.org/x/crypto/chacha20poly1305"
+
+	tinkcrypto "github.com/hyperledger/aries-framework-go/pkg/framework/provider/crypto"
+)
+
+func TestCrypto(t *testing.T) {
+	// Assert that MainCrypto implements the FrameworkCrypto interface.
+	var _ FrameworkCrypto = (*MainCrypto)(nil)
+}
```
soluchok: nit: use the global var and disable a linter `var _ FrameworkCrypto = (*MainCrypto)(nil)`

pkg/framework/provider/crypto/crypto_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/stretchr/testify/require"
+	chacha "golang.org/x/crypto/chacha20poly1305"
+
+	cryptoapi "github.com/hyperledger/aries-framework-go/pkg/framework/spi/crypto"
+)
+
+func TestCrypto(t *testing.T) {
+	// Assert that TinkCrypto implements the TinkCrypto interface.
+	var _ cryptoapi.Crypto = (*TinkCrypto)(nil)
+}
```
soluchok: same, it is compiler check not a test case

Baha-sk: removed test function

pkg/common/crypto/default_test.go
Baha-sk: removed test function

pkg/framework/provider/crypto/crypto.go
Baha-sk: done

pkg/framework/provider/crypto/crypto.go
Baha-sk: done

pkg/framework/provider/crypto/crypto.go
Baha-sk: done

pkg/framework/provider/crypto/crypto.go
Baha-sk: done

pkg/framework/provider/crypto/crypto.go
Baha-sk: done

pkg/common/crypto/default.go
Baha-sk: done

pkg/common/crypto/api.go
Baha-sk: done

pkg/common/crypto/default.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
```
troyronda: i do not think this package makes sense.

- the default packages are held under framework.
- we shouldn’t need a second wrapper of tink.

pkg/framework/provider/crypto/crypto.go
troyronda: yes - this means two case statements.

pkg/common/crypto/default.go
Baha-sk: this Package includes common crypto operations to be used by the framework, the crypto under `provider` are for specific operations like encrypt/decrypt or sign/verify and are used to build an instance of this package.

Baha-sk: as discussed, I'll remove this package and add provider context option for Crypto in the Aries Framework.

Baha-sk: removed

pkg/common/crypto/default.go
Baha-sk: default implementation has been removed.. Crypto is created via context creating in the framework which makes this implementation obsolete.

pkg/framework/provider/crypto/crypto.go
Baha-sk: done

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,134 @@
...
+	return cipherText, nonce, nil
+}
+
+func nonceSize(ps *primitiveset.PrimitiveSet) int {
+	var ivSize int
+	// AESGCM and XChacha20Poly1305 nonce sizes supported only for now
+	switch ps.Primary.Primitive.(type) {
+	case *aeadsubtle.XChaCha20Poly1305:
+		ivSize = chacha20poly1305.NonceSizeX
+	default:
```
troyronda: I still prefer having `case aeadsubtle.AESGCM...` explicitly.

pkg/common/api/crypto/api.go
```diff
@@ -0,0 +1,23 @@
...
+// it will be created via Options creation of pkg/framework/provider/crypto (implementation of
+// pkg/framework/spi/crypto)
+
+// Crypto interface provides all crypto operations needed in the Aries framework
+type Crypto interface {
+	crypto.Crypto
+}
+
+// Creator method to create a new Crypto service
+type Creator func() (Crypto, error)
```
troyronda: I don't understand the purpose of this method. Packages will obtain the crypto service from the context.

pkg/framework/aries/framework.go
```diff
@@ -43,6 +44,8 @@ type Aries struct {
 	inboundTransport       transport.InboundTransport
 	kmsCreator             api.KMSCreator
 	kms                    api.CloseableKMS
+	cryptoCreator          crypto.Creator
```
troyronda: Why is the framework using an interface defined in API rather than SPI? This doesn't make sense.

pkg/framework/provider/crypto/crypto.go
Baha-sk: sure, I'll add it ..

pkg/common/api/crypto/api.go
Baha-sk: this is to allow the creation of an api.Crypto instance using any implementation (SPI.Crypto implementation is one the possiblities)

pkg/framework/aries/framework.go
Baha-sk: the framework uses api.Crypto to invoke crypto operations.. and can be injected with SPI.Crypto to build the instance used by the framework

pkg/common/api/crypto/api.go
troyronda: I don't understand what this means to packages using crypto.

Baha-sk: you're, this doesn't belong in common... I will move this to spi crypto package and make the framework create an instance of spi Crypto (instead of common Crypto) to be added in the context

pkg/framework/aries/framework.go
Baha-sk: I will flip the dependency type around, this should be an SPI crypto type, not `common/api/crypto`

pkg/framework/provider/crypto/crypto.go
```diff
@@ -0,0 +1,136 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package crypto provides the default implementation of pkg/framework/spi/crypto.Crypto interface
+//
+// It uses github.com/tink/go crypto primitives
+package crypto
```
troyronda: Wouldn't a concrete implementation of crypto for tink be `tinkcrypto`?

pkg/framework/spi/crypto/api.go
```diff
@@ -0,0 +1,17 @@
...
```
troyronda: I wonder if we really need an extra folder under spi?

Baha-sk: moved file to spi.go under framework ... the SPIs interfaces are internal to the framework and should not be visible outside of it.

pkg/framework/provider/tinkcrypto/crypto.go
```diff
@@ -0,0 +1,137 @@
...
+// Sign will sign msg using the implementation's corresponding signing key referenced by kh
+func (t *Crypto) Sign(msg []byte, kh interface{}) ([]byte, error) {
+	keyHandle, ok := kh.(*keyset.Handle)
+	if !ok {
+		return nil, errors.New("bad key handle format")
+	}
+
+	signer, err := signature.NewSigner(keyHandle)
+	if err != nil {
+		return nil, err
```
troyronda: any reason why errors are not being wrapped?

pkg/framework/provider/crypto/crypto.go
Baha-sk: renamed package to provider/tinkcrypto instead

pkg/framework/aries/spi.go
```diff
@@ -0,0 +1,14 @@
+/*
```
troyronda: delete this file?

pkg/framework/provider/tinkcrypto/crypto.go
Baha-sk: do they need to be wrapped?

pkg/framework/context/context.go
```diff
@@ -27,6 +28,7 @@ type Provider struct {
 	storeProvider            storage.Provider
 	transientStoreProvider   storage.Provider
 	kms                      kms.KMS
```
troyronda: Seems like this field should be called crypto to match the rest.

Baha-sk: sure will rename to crypto..

Baha-sk: done

pkg/framework/aries/spi.go
Baha-sk: done

pkg/framework/provider/tinkcrypto/crypto.go
Baha-sk: wrapped all errors in this file

pkg/framework/aries/framework.go
Baha-sk: done

pkg/common/api/crypto/api.go
Baha-sk: done

llorllale:
@Baha-sk need to update the wiki

Baha-sk:
> @Baha-sk need to update the wiki

done.. see [here](https://wiki.hyperledger.org/display/ARIES/Crypto+and+KMS+as+pluggable+Service+Provider+Interfaces) for updates

PR #964, opened 2019-12-12
refactor: Introduce - send message logic

### LINKED ISSUE #963: Introduce - need to correct send message logic
Need to use `SendToDID(msg interface{}, myDID, theirDID string) error` function for sending a message. If protocol starts from inbound message use the right DIDs according to `HandleInbound(msg *service.DIDCommMsg, myDID, theirDID string)`

soluchok:
* messages are sent by using `SendToDID(msg interface{}, myDID, theirDID string) error` function
* Handler interface was changed to: 
```
type Handler interface {
	HandleInbound(msg *DIDCommMsg, myDID string, theirDID string) (string, error)
	HandleOutbound(msg *DIDCommMsg, myDID, theirDID string) error
}
```

Closes #963  

### DIFF
--- a/pkg/didcomm/dispatcher/outbound.go
+++ b/pkg/didcomm/dispatcher/outbound.go
@@ -40,6 +40,11 @@ func NewOutbound(prov provider) *OutboundDispatcher {
 	}
 }
 
+// SendToDID msg
+func (o *OutboundDispatcher) SendToDID(msg interface{}, myDID, theirDID string) error {
+	return nil
+}
+
 // Send msg
 func (o *OutboundDispatcher) Send(msg interface{}, senderVerKey string, des *service.Destination) error {
 	for _, v := range o.outboundTransports {
--- a/pkg/framework/context/context.go
+++ b/pkg/framework/context/context.go
@@ -113,7 +113,7 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
 		// find the service which accepts the message type
 		for _, svc := range p.services {
 			if svc.Accept(msg.Header.Type) {
-				_, err = svc.HandleInbound(msg)
+				_, err = svc.HandleInbound(msg, "", "")
 				return err
 			}
 		}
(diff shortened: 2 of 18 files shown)


pkg/didcomm/dispatcher/outbound.go
```diff
@@ -40,6 +40,11 @@ func NewOutbound(prov provider) *OutboundDispatcher {
 	}
 }
 
+// SendToDID msg
+func (o *OutboundDispatcher) SendToDID(msg interface{}, myDID, theirDID string) error {
+	return nil
```
troyronda: This will need to be implemented before BDD tests will work.

soluchok: Yeah, I know. I just added this interface to correct the logic

pkg/framework/context/context.go
```diff
@@ -113,7 +113,7 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
 		// find the service which accepts the message type
 		for _, svc := range p.services {
 			if svc.Accept(msg.Header.Type) {
-				_, err = svc.HandleInbound(msg)
+				_, err = svc.HandleInbound(msg, "", "")
```
llorllale: Should `transport.InboundMessageHandler` be refactored as well?

soluchok: It should be done here https://github.com/hyperledger/aries-framework-go/pull/938
Also, all files which are not related to the introduce protocol will be reverted according to PR above.

soluchok: Except for changes that are related to `HandleOutbound` function. I have changed `HandleOutbound` function signature.

soluchok: @Moopli  Could you please confirm?

soluchok: found! https://github.com/hyperledger/aries-framework-go/pull/938/files#diff-9988f5820bf05228e5657f2842fa071eR114

Moopli: Given how you're using HandleOutbound, it seems like a reasonable change to the interface.

HandleOutbound isn't used in DIDExchange at all, so we can make it use DIDs, which makes things easier for other protocols.

PR #993, opened 2019-12-17
refactor: Moved generic utility functions out of DIDExchange

Moopli:
Functions moved out are for DID doc parsing and destination struct construction. These functions are needed for other protocols besides DIDExchange.

Part of #725


pkg/didcomm/common/service/destination.go
```diff
@@ -0,0 +1,50 @@
...
+	didDoc, err := vdr.Resolve(did)
+	if err != nil {
+		return nil, err
+	}
+
+	return MakeDestination(didDoc, serviceType, keyType)
+}
+
+// MakeDestination makes a Destination object from a DID Doc, following the given parameters.
+func MakeDestination(didDoc *diddoc.Doc, serviceType, keyType string) (*Destination, error) {
```
troyronda: Should be CreateDestination

pkg/doc/did/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+*/
+
+package did
+
+import (
+	"fmt"
+)
+
+// GetDIDCommService returns the service from the given DIDDoc matching the given service type.
+func GetDIDCommService(didDoc *Doc, serviceType string) (*Service, error) {
```
troyronda: These functions don't seem to "Get" anything.

pkg/doc/did/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+*/
+
+package did
+
+import (
+	"fmt"
+)
+
+// LookupService returns the service from the given DIDDoc matching the given service type.
+func LookupService(didDoc *Doc, serviceType string) (*Service, error) {
```
troyronda: did you intend to have lookup functions with an error return instead of bool?

pkg/doc/did/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if index == notFound {
+		return nil, fmt.Errorf("service not found in DID document: %s", serviceType)
+	}
+
+	return &didDoc.Service[index], nil
+}
+
+// LookupRecipientKeys gets the recipient keys from the did doc which match the given parameters.
+func LookupRecipientKeys(didDoc *Doc, serviceType, keyType string) ([]string, error) {
```
troyronda: (later) we may want to think about refactoring these static accessors and the existing instance fields.

pkg/didcomm/common/service/destination.go
```diff
@@ -0,0 +1,50 @@
...
+	didDoc, err := vdr.Resolve(did)
+	if err != nil {
+		return nil, err
+	}
+
+	return CreateDestination(didDoc, serviceType, keyType)
+}
+
+// CreateDestination makes a Destination object from a DID Doc, following the given parameters.
+func CreateDestination(didDoc *diddoc.Doc, serviceType, keyType string) (*Destination, error) {
```
llorllale: `serviceType` should be removed. It's always set to `did-communication`.

llorllale: `keyType` should also be removed as an arg.

The keyType was ensured to be `Ed25519VerificationKey2018` in the `didexchange` package because of the crypto primitives we were willing to support in the framework. Callers of this function should not be allowed to specify their own keyTypes.

pkg/doc/did/helpers.go
Moopli: I'll make them return bool today.

pkg/didcomm/common/service/destination.go
Moopli: > It's always set to did-communication.
Always, for any communication using a Destination created from a DID doc?
Sounds right but just want to be sure.

keyType will need to be refactored away when we have a mechanism for selecting key type. Should I just enforce it to be that constant for now, in here?

llorllale: > Always, for any communication using a Destination created from a DID doc?

Yes. See [Service Conventions](https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions).

> keyType will need to be refactored away when we have a mechanism for selecting key type. Should I just enforce it to be that constant for now, in here?

Moving the enforcement to this centralized location makes sense to me.

pkg/doc/did/helpers.go
Moopli: done

pkg/didcomm/common/service/destination.go
Moopli: done

PR #997, opened 2019-12-19
feat: VC context options

### LINKED ISSUE #992: VC Context and Type Options
Enable the VC user to specify the expected Context and Types as `options`. 

**Modes**

This option has a few variants:
* Allow any additional `@context` and type (in addition to the base values required by the spec).
  * i.e., the current behaviour of the framework.
* Allow only the specified values.
  * Allow only the base values. i.e., `@context` contains only `https://www.w3.org/2018/credentials/v1` and type contains only `VerifiableCredential`.
  * Allow _specified_ values (in addition to the base values).
* JSON-LD validation (when available and enabled).
* Potential followup (TBD): 
  * Ignore `@context` and/or type values.

**Defaults**

*JSON only mode*
`Allow any additional context and type.` 

Note: We should double check the VC specification about handling of unknown contexts and types. We may need to default to `Allow only the base values`.

*JSON-LD mode*
`JSON-LD validation`

**Test Suite**
Without a JSON-LD mode, we will need to use options to pass the remaining issues in the test suite. We should enable the test suite to pass and also enforce passing tests in the CI.

kdimak:
closes #992

A number of VC Test Suite failures:
- Before: 2
- After: 0

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -183,9 +183,35 @@ const defaultSchema = `{
 }
 `
 
+// https://www.w3.org/TR/vc-data-model/#data-schemas
 const jsonSchema2018Type = "JsonSchemaValidator2018"
 
-const vpType = "VerifiablePresentation"
+const (
+	// https://www.w3.org/TR/vc-data-model/#base-context
+	baseContext = "https://www.w3.org/2018/credentials/v1"
+
+	// https://www.w3.org/TR/vc-data-model/#types
+	vcType = "VerifiableCredential"
+
+	// https://www.w3.org/TR/vc-data-model/#presentations-0
+	vpType = "VerifiablePresentation"
+)
+
+// vcModelValidationMode defines constraint put on context and type of VC.
+type vcModelValidationMode int
+
+const (
+	// baseContextValidation when defined it's validated that only the fields and values (when applicable)
+	// are present in the document. No extra fields are allowed (outside of credentialSubject).
+	baseContextValidation vcModelValidationMode = iota
+
+	// baseContextExtendedValidation when set it's validated that fields that are specified in base context are
+	// as specified. Additional fields are allowed.
+	baseContextExtendedValidation
+
+	// jsonldValidation Use the JSON LD parser for validation.
+	jsonldValidation
+)
 
 // SchemaCache defines a cache of credential schemas.
 type SchemaCache interface {
@@ -397,6 +423,9 @@ type credentialOpts struct {
 	issuerPublicKeyFetcher PublicKeyFetcher
 	disabledCustomSchema   bool
 	schemaLoader           *CredentialSchemaLoader
+	modelValidationMode    vcModelValidationMode
+	allowedCustomContexts  map[string]bool
+	allowedCustomTypes     map[string]bool
 }
 
 // CredentialOpt is the Verifiable Credential decoding option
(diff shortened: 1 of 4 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -185,7 +185,29 @@ const defaultSchema = `{
 
 const jsonSchema2018Type = "JsonSchemaValidator2018"
 
-const vpType = "VerifiablePresentation"
+const (
+	baseContext = "https://www.w3.org/2018/credentials/v1"
```
troyronda: It would be good to add a comment linking to the spec.
https://www.w3.org/TR/vc-data-model/#base-context

pkg/doc/verifiable/credential.go
```diff
@@ -426,6 +451,40 @@ func WithCredentialSchemaLoader(loader *CredentialSchemaLoader) CredentialOpt {
 	}
 }
 
+// WithAnyContextAndType any additional @context and type of VC to decode.
+func WithAnyContextAndType() CredentialOpt {
```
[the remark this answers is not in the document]
troyronda: We should rename these methods and constants to reflect that we are really talking about JSON-LD validation.

Also, if you are exporting the constants anyways, you might not need a matching option method for each constant... you could have potentially pass the context into an option.

pkg/doc/verifiable/credential.go
kdimak: Added the links.

pkg/doc/verifiable/credential.go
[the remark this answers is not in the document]
kdimak: Done - I hid the constants but left the methods. The reason is in `WithBaseContextAndTypeExtendedValidation()` where there are two arguments. Having one single method and a need to pass `nil`-s  for the two remaining validation modes does not look good for me.

pkg/doc/verifiable/credential.go
```diff
@@ -426,6 +455,41 @@ func WithCredentialSchemaLoader(loader *CredentialSchemaLoader) CredentialOpt {
...
+// WithJSONLDValidation uses the JSON LD parser for validation.
+func WithJSONLDValidation() CredentialOpt {
+	return func(opts *credentialOpts) {
+		opts.modelValidationMode = jsonldValidation
+	}
+}
+
+// WithBaseContextAndType validates that only the fields and values (when applicable) are present
+// in the document. No extra fields are allowed (outside of credentialSubject).
+func WithBaseContextAndType() CredentialOpt {
```
troyronda: Let's remove "AndType". Keep things succinct.

(Also elsewhere in this PR).

pkg/doc/verifiable/credential.go
```diff
@@ -426,6 +455,41 @@ func WithCredentialSchemaLoader(loader *CredentialSchemaLoader) CredentialOpt {
...
 }
 
+// WithJSONLDValidation uses the JSON LD parser for validation.
+func WithJSONLDValidation() CredentialOpt {
+	return func(opts *credentialOpts) {
+		opts.modelValidationMode = jsonldValidation
+	}
+}
+
+// WithBaseContextAndType validates that only the fields and values (when applicable) are present
```
troyronda: The other options have a suffix of "Validation" and this one doesn't?

pkg/doc/verifiable/credential.go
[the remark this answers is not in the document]
troyronda: @kdimak I also do not want to pass nil.

pkg/doc/verifiable/credential.go
kdimak: Yeah... I just forgot to change the name here. Fixed.

pkg/doc/verifiable/credential.go
kdimak: Fixed.

pkg/doc/verifiable/credential.go
```diff
@@ -545,9 +609,61 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, []byte, e
...
+		return nil
+
+	case baseContextValidation:
+		return validateBaseOnlyContextType(vc)
+
+	case baseContextExtendedValidation:
+		return validateCustomContextType(vc, vcOpts)
+
+	default:
+		logger.Panicf("unsupported vcModelValidationMode: %s", vcOpts.modelValidationMode)
```
troyronda: We cannot have a Panic.

PR #1020, opened 2019-12-23
refactor: Introduce - client stops the protocol without passing an error

soluchok:
* prevented possible bug

### DIFF
--- a/pkg/didcomm/protocol/introduce/states.go
+++ b/pkg/didcomm/protocol/introduce/states.go
@@ -38,10 +38,6 @@ const (
 	stateNameWaiting    = "waiting"
 )
 
-type internalContext struct {
-	dispatcher.Outbound
-}
-
 // The introduce protocol's state.
 type state interface {
 	// Name of this state.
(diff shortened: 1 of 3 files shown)


test/bdd/pkg/introduce/introduce_sdk_steps.go
```diff
@@ -219,7 +219,7 @@ func (a *SDKSteps) checkAndStop(agentID, introduceeID string) error {
 			return fmt.Errorf("%q doesn't want to know %q but got %q", agentID, introduceeID, proposal.To.Name)
 		}
 
-		e.Stop(errors.New("stop the protocol"))
+		e.Stop(nil)
```
troyronda: I’m not a fan of nil arguments.

soluchok: This is a way to hide an error in the BDD tests. Stopping the protocol without passing an error prevents logging an error you asked me before.

soluchok: If you don't like it, it's ok. I can create a user error type. It will be much better.

soluchok: Done!

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -93,11 +89,11 @@ func (s *start) CanTransitionTo(next state) bool {
 	return false
 }
 
-func (s *start) ExecuteInbound(ctx internalContext, _ *metaData) (state, error) {
+func (s *start) ExecuteInbound(_ dispatcher.Outbound, _ *metaData) (state, error) {
 	return nil, errors.New("start ExecuteInbound: not implemented yet")
 }
```
troyronda: will these ever be implemented?

soluchok: I am not sure, we do not use it for now. But better to return an error. It will help us to understand where an error happens.

troyronda: sure. i just think the error text is misleading if this is not going to be implemented.

soluchok: Any specific suggestions?

Issue #1025, opened 2019-12-27
Need to implement the threading of the messages


soluchok:
We need to create `ReplyTo(inboundMsg *Header, outboundMsg interface{}, myDID, theirDID string)` function.  Chatting between agents should be achieved by using this function. To be more specific we need to use `ReplyTo` when the agent receives an inbound message. `sender_order` counter must be save to the DB under the following key: thid_theirDID.

NOTE: This issue doesn't include n-wise communication

PR #1026, opened 2019-12-30
refactor: Separate common fields into Header structure

soluchok:
* common fields were separated into Header structure

```
// Header helper structure which keeps reusable fields
type Header struct {
	ID     string            `json:"@id,omitempty"`
	Thread *decorator.Thread `json:"~thread,omitempty"`
	Type   string            `json:"@type,omitempty"`
}
```

Also, we have a helper functions such as:
* `MsgID() string`
* `MsgThread() *decorator.Thread`
* `MsgType() string`
* `ThreadID() (string, error)`

Part of #1025 

### DIFF
--- a/pkg/didcomm/common/service/service.go
+++ b/pkg/didcomm/common/service/service.go
@@ -31,9 +31,64 @@ type DIDComm interface {
 
 // Header helper structure which keeps reusable fields
 type Header struct {
-	ID     string           `json:"@id"`
-	Thread decorator.Thread `json:"~thread"`
-	Type   string           `json:"@type"`
+	ID     string            `json:"@id,omitempty"`
+	Thread *decorator.Thread `json:"~thread,omitempty"`
+	Type   string            `json:"@type,omitempty"`
+}
+
+// MsgID returns message ID
+func (h *Header) MsgID() string {
+	if h != nil {
+		return h.ID
+	}
+
+	return ""
+}
+
+// MsgThread returns message thread decorator
+func (h *Header) MsgThread() *decorator.Thread {
+	if h != nil {
+		return h.Thread
+	}
+
+	return nil
+}
+
+// MsgType returns message type
+func (h *Header) MsgType() string {
+	if h != nil {
+		return h.Type
+	}
+
+	return ""
+}
+
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+func (h *Header) ThreadID() (string, error) {
+	// we need to return it only if there is no ~thread.thid
+	if h.MsgThread() == nil {
+		if len(h.MsgID()) > 0 {
+			return h.MsgID(), nil
+		}
+
+		return "", ErrInvalidMessage
+	}
+
+	// if message has ~thread.thid but @id is absent this is invalid message
+	if len(h.MsgThread().ID) > 0 && h.MsgID() == "" {
+		return "", ErrInvalidMessage
+	}
+
+	if len(h.MsgThread().ID) > 0 {
+		return h.MsgThread().ID, nil
+	}
+
+	if len(h.MsgID()) > 0 {
+		return h.MsgID(), nil
+	}
+
+	return "", ErrThreadIDNotFound
 }
 
 func (h *Header) clone() *Header {
@@ -41,9 +96,17 @@ func (h *Header) clone() *Header {
 		return nil
 	}
 
+	if h.MsgThread() == nil {
+		return &Header{
+			ID:     h.MsgID(),
+			Thread: h.MsgThread(),
+			Type:   h.MsgType(),
+		}
+	}
+
 	return &Header{
-		ID: h.ID,
-		Thread: decorator.Thread{
+		ID: h.MsgID(),
+		Thread: &decorator.Thread{
 			ID:          h.Thread.ID,
 			PID:         h.Thread.PID,
 			SenderOrder: h.Thread.SenderOrder,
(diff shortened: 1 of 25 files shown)


pkg/didcomm/common/service/service.go
```diff
@@ -21,6 +21,13 @@ type Handler interface {
 	HandleOutbound(msg *DIDCommMsg, myDID, theirDID string) error
 }
 
+// Headerable provides common methods for the message
+type Headerable interface {
+	GetID() string
```
troyronda: Simple accessors shouldn’t be prefixed with “Get”.

Also I think it should be a single interface method MsgHeader() or Header() that returns the Header struct.

soluchok: ok, MsgHeader is fine.

troyronda: Are you planning to directly use the header data fields directly or still call the accessor methods?

soluchok: I am going to use `MsgHeader()` function

troyronda: If the latter, I guess the multiple method interface could make sense with a better prefix.

soluchok: MsgID() ? MsgType?

troyronda: Msg as a prefix is okay.

soluchok: Are you ok with `Headerable` name?

pkg/didcomm/common/service/service.go
```diff
@@ -21,6 +21,13 @@ type Handler interface {
 	HandleOutbound(msg *DIDCommMsg, myDID, theirDID string) error
 }
 
+// Headerable provides common methods for the message
+type Headerable interface {
+	MsgID() string
```
troyronda: I don't actually see these methods being used?

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -47,7 +47,7 @@ func NewOutbound(prov provider) *OutboundDispatcher {
 }
 
 // SendToDID sends a message from myDID to the agent who owns theirDID
-func (o *OutboundDispatcher) SendToDID(msg interface{}, myDID, theirDID string) error {
+func (o *OutboundDispatcher) SendToDID(msg service.Headerable, myDID, theirDID string) error {
```
troyronda: The actual usage of msg is https://github.com/hyperledger/aries-framework-go/pull/1026/files#diff-150ad9be6e6f80a3a4239dcaa00d6d5bR78 in the Send function, where it calls marshal on the underlying struct.

(i.e., cares about the annotated struct rather than the accessor methods you created in service.Headerable).

pkg/didcomm/common/service/service.go
soluchok: Nothing is being used for now. My point is to generalize all messages in the framework. `MsgID` will be used for incrementing a counter (`sender_order`), but not in this PR

pkg/didcomm/dispatcher/outbound.go
troyronda: Or even more specifically, the method currently cares about being able to marshal the message into JSON. It should also care more that the mandatory fields are present (and from those fields, marshalled).

soluchok: Accessor methods don't affect marshaling. This header was created to easily access common fields. I am going to change `Send` function signature as well.

troyronda: Note: `service.Headerable` is acting the same as `interface{}` in the current usage.

troyronda: You shouldn't only be changing the function signature. The method doesn't actually use anything from the interface.

pkg/didcomm/common/service/service.go
troyronda: As you can see in my latter comments, I cannot tell how these methods will be used.

To your question, generally I would expect a `able` interface to have only one or two methods with the matching action. E.g., `Closeable` has a `Close()`. I would generally expect a `Headerable` to have a `Header()` or `MsgHeader()`.

pkg/didcomm/common/service/service.go
troyronda: An accessor should not have side effects. You should use a different method for a counter.

troyronda: I think we shouldn't add an interface until they have a clear usage. I would move this change to another PR, where it is used, so we can evaluate it together.

E.g., You need a counting mechanism. This seems somewhat separate from what has been defined.

pkg/framework/aries/framework_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
```
troyronda: ordered wrong.

troyronda:
I think the main issue is that we do not have layer separation.

- Lower layer DIDComm messaging should handle DIDComm semantics including threading. The higher layer protocols should not be aware of these details. The lower layer package should be useable independently - as if higher layer protocols do not even exist.
- Higher layer protocols should concern themselves with:
  - their own semantics.
  - declaring their type.
  - calling appropriate lower-layer functions (e.g., sending a message, replying on a thread, replying on a parent thread, etc).

This PR has difficulty because the responsibilities are currently inverted. The higher layer protocol embeds the lower layer "header". It then calls Send which marshals whatever is provided (effectively still an empty interface despite changing the type). Instead the higher layer should call Send with a payload and the lower-layer injects into a DIDComm message structure which is then marshalled.

PR #1033, opened 2020-01-06
feat: generic inbound handler

### LINKED ISSUE #1032: generic inbound message handler
Provide feature to register generic inbound message service to handle messages matching given type/field (any key in message).

generic message service lookup should happen when none of the service accepts given message type.

sudeshrshetty:
- generic message handler when no registered service handles
given message type
- current Accept logic is based on message header.
- closes #1032

### DIFF
--- a/pkg/didcomm/common/service/service.go
+++ b/pkg/didcomm/common/service/service.go
@@ -13,12 +13,22 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 )
 
-// Handler provides protocol service handle api.
-type Handler interface {
+// InboundHandler is handler for inbound messages
+type InboundHandler interface {
 	// HandleInbound handles inbound messages.
-	HandleInbound(msg *DIDCommMsg, myDID string, theirDID string) (string, error)
+	HandleInbound(msg *DIDCommMsg, myDID, theirDID string) (string, error)
+}
+
+// OutboundHandler is handler for outbound messages
+type OutboundHandler interface {
 	// HandleOutbound handles outbound messages.
 	HandleOutbound(msg *DIDCommMsg, myDID, theirDID string) error
+}
+
+// Handler provides protocol service handle api.
+type Handler interface {
+	InboundHandler
+	OutboundHandler
 }
 
 // DIDComm defines service APIs.
@@ -34,6 +44,8 @@ type Header struct {
 	ID     string           `json:"@id"`
 	Thread decorator.Thread `json:"~thread"`
 	Type   string           `json:"@type"`
+	// TODO revisit ~purpose, should be generic map [Issue #1037]
+	Purpose []string `json:"~purpose"`
 }
 
 func (h *Header) clone() *Header {
--- a/pkg/didcomm/dispatcher/api.go
+++ b/pkg/didcomm/dispatcher/api.go
@@ -17,6 +17,13 @@ type Service interface {
 	Name() string
 }
 
+// MessageService is service for handling generic messages
+// matching accept criteria
+type MessageService interface {
+	service.InboundHandler
+	Accept(header *service.Header) bool
+}
+
 // Outbound interface
 type Outbound interface {
 	// Sends the message after packing with the sender key and recipient keys.
--- a/pkg/framework/context/context.go
+++ b/pkg/framework/context/context.go
@@ -23,6 +23,7 @@ import (
 // Provider supplies the framework configuration to client objects.
 type Provider struct {
 	services                 []dispatcher.Service
+	messageServices          []dispatcher.MessageService
 	storeProvider            storage.Provider
 	transientStoreProvider   storage.Provider
 	kms                      kms.KMS
@@ -117,6 +118,16 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
 				return err
 			}
 		}
+
+		// in case of no services are registered for given message type,
+		// find generic inbound services registered for given message header
+		for _, svc := range p.messageServices {
+			if svc.Accept(msg.Header) {
+				_, err = svc.HandleInbound(msg, myDID, theirDID)
+				return err
+			}
+		}
+
 		return fmt.Errorf("no message handlers found for the message type: %s", msg.Header.Type)
 	}
 }
@@ -177,6 +188,14 @@ func WithTransportReturnRoute(transportReturnRoute string) ProviderOption {
 func WithProtocolServices(services ...dispatcher.Service) ProviderOption {
 	return func(opts *Provider) error {
 		opts.services = services
+		return nil
+	}
+}
+
+// WithMessageServices injects services for handling generic messages.
+func WithMessageServices(services ...dispatcher.MessageService) ProviderOption {
+	return func(opts *Provider) error {
+		opts.messageServices = services
 		return nil
 	}
 }
--- a/pkg/didcomm/common/service/service_test.go
+++ b/pkg/didcomm/common/service/service_test.go
@@ -137,3 +137,51 @@ func TestDIDCommMsg_Clone(t *testing.T) {
 	cloned = didMsg.Clone()
 	require.Equal(t, didMsg, cloned)
 }
+
+func TestDIDCommMsg_Purpose(t *testing.T) {
+	tests := []struct {
+		name    string
+		id      string
+		typeVal string
+		purpose []string
+		payload []byte
+		err     string
+	}{{
+		name: "empty payload test",
+		err:  "invalid payload data format: unexpected end of JSON input",
+	}, {
+		name:    "valid payload test",
+		id:      "ID",
+		purpose: []string{"team:01453", "delivery", "geotag", "cred"},
+		typeVal: "sample-type",
+		payload: []byte(`{"@id":"ID", "@type":"sample-type", "~purpose":["team:01453", "delivery", "geotag", "cred"], 
+"id":"ID", "data-type":"data-type-01"}`),
+		err: "",
+	}, {
+		name:    "invalid payload",
+		payload: []byte(`[]`),
+		err:     `invalid payload data format: json: cannot unmarshal array into Go value of type service.Header`,
+	}}
+
+	t.Parallel()
+
+	for _, test := range tests {
+		tc := test
+		t.Run(tc.name, func(t *testing.T) {
+			val, err := NewDIDCommMsg(tc.payload)
+
+			if tc.err != "" {
+				require.Error(t, err)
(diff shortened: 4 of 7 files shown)


pkg/didcomm/dispatcher/api.go
```diff
@@ -17,6 +17,13 @@ type Service interface {
 	Name() string
 }
 
+// GenericInboundService is service for handling generic inbound messages
+// matching accept criteria
+type GenericInboundService interface {
+	service.InboundHandler
+	Accept(message []byte) bool
```
sudeshrshetty: @troyronda rightnow Accept gets whole did comm msg, should we filter what we pass to `Accept()`?
Some interface/struct containing subset of DIDCommMsg?

pkg/didcomm/common/service/service.go
```diff
@@ -13,14 +13,24 @@ import (
...
+type InboundHandler interface {
 	// HandleInbound handles inbound messages.
-	HandleInbound(msg *DIDCommMsg, myDID string, theirDID string) (string, error)
-	// HandleOutbound handles outbound messages.
+	HandleInbound(msg *DIDCommMsg, myDID, theirDID string) (string, error)
+}
+
+// OutboundHandler is handler for outbound messages
+type OutboundHandler interface {
+	// HandleInbound handles outbound messages.
```
troyronda: HandleOutbound

pkg/didcomm/dispatcher/api.go
troyronda: I think we should pass a structure.

pkg/didcomm/common/service/service.go
sudeshrshetty: fixed

pkg/framework/context/context.go
```diff
@@ -188,6 +199,14 @@ func WithProtocolServices(services ...dispatcher.Service) ProviderOption {
 	}
 }
 
+// WithGenericInboundServices injects services for handling generic inbound messages.
+func WithGenericInboundServices(services ...dispatcher.GenericInboundService) ProviderOption {
```
troyronda: I am guessing that this `GenericInbound` is not the most intuitive. I am guessing you are trying to differentiate between a full protocol service and a message service?

pkg/framework/context/context.go
```diff
@@ -188,6 +199,14 @@ func WithProtocolServices(services ...dispatcher.Service) ProviderOption {
...
+	return func(opts *Provider) error {
+		opts.genericInboundServices = services
+		return nil
+	}
+}
+
```
troyronda: What is the larger design? We also need:
- exposed mechanism for external code / services to access a messenger (so they can send message, reply, etc).

I assume we will also want to plan:
- a dispatcher implementation of a messenger service that can handle dynamic registrations.
- REST binding of this dispatcher.
- REST binding for the messenger.

pkg/didcomm/dispatcher/api.go
sudeshrshetty: Containing only fields of interest.
eg: ~purpose, data-type, @type

pkg/framework/context/context.go
sudeshrshetty: - For sending message, still thinking about `OutboundHandler` for sending dynamic outbound message from handlers directly or having dedicated send message service.
- Yes, we need dynamic registration of dispatcher & REST bindings. These are still under design.

pkg/didcomm/dispatcher/api.go
troyronda: I would assume it would be the DIDCommMsg `Header` (plus some extra fields). If purpose becomes the only additional field of interest, then extra fields could mean decorators.

pkg/framework/context/context.go
troyronda: `WithMessageServices`, `WithMessageHandlers`, ...?

sudeshrshetty: `MessageHandler` is wildly used, and may be confusing with https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/transport/transport_interface.go#L33

Will go go `MessageService` for now,

troyronda: ok

pkg/framework/context/context.go
troyronda: You mean `Outbound` interface (or what is calling being called messenger) rather than `OutboundHandler`, correct?
https://github.com/hyperledger/aries-framework-go/blob/9da989cdfa11953add749c1959b2c8bce2630a80/pkg/didcomm/dispatcher/api.go#L21

sudeshrshetty: Yes

pkg/framework/context/context.go
sudeshrshetty: done

pkg/didcomm/dispatcher/api.go
sudeshrshetty: done

pkg/didcomm/common/service/service.go
```diff
@@ -122,3 +132,22 @@ func (m *DIDCommMsg) ThreadID() (string, error) {
 
 	return "", ErrThreadIDNotFound
 }
+
+// MessageHeader is header of the generic message
+type MessageHeader struct {
```
troyronda: Do we need a separate struct? @fqutishat @sudeshrshetty

pkg/didcomm/common/service/service.go
```diff
@@ -122,3 +132,22 @@ func (m *DIDCommMsg) ThreadID() (string, error) {
...
+	*Header
+	Purpose  string `json:"purpose"`
+	DataType string `json:"data-type"`
```
troyronda: I would remove.

pkg/didcomm/common/service/service.go
```diff
@@ -122,3 +132,22 @@ func (m *DIDCommMsg) ThreadID() (string, error) {
...
+}
+
+// NewMessageHeader returns MessageHeader from given payload
+func NewMessageHeader(payload []byte) (*MessageHeader, error) {
+	msg := &MessageHeader{}
+
+	err := json.Unmarshal(payload, &msg)
```
troyronda: Don't want more Unmarshals.

pkg/didcomm/common/service/service.go
```diff
@@ -122,3 +132,22 @@ func (m *DIDCommMsg) ThreadID() (string, error) {
 
 	return "", ErrThreadIDNotFound
 }
+
+// MessageHeader is header of the generic message
+type MessageHeader struct {
+	*Header
+	Purpose  string `json:"purpose"`
```
troyronda: purpose is more likely be a decorator (~purpose).

pkg/didcomm/common/service/service.go
sudeshrshetty: @troyronda DIDComm Header may not have all fields of interest.

Either we pass whole DIDComm Msg or Portion of it as a struct

pkg/didcomm/common/service/service.go
```diff
@@ -122,3 +132,22 @@ func (m *DIDCommMsg) ThreadID() (string, error) {
 
 	return "", ErrThreadIDNotFound
 }
+
+// MessageHeader is header of the generic message
+type MessageHeader struct {
+	*Header
```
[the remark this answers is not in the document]
sudeshrshetty: I was trying to pass whole DID Comm msg payload to parse portion of it to MessageHeader

pkg/didcomm/common/service/service.go
sudeshrshetty: Thought of updating it later after RFC completion.

pkg/didcomm/common/service/service.go
sudeshrshetty: It is Same as existing function in same file NewDIDCommMsg
https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/common/service/service.go#L90

pkg/didcomm/common/service/service.go
sudeshrshetty: You mean moving them to different struct?

pkg/didcomm/common/service/service.go
troyronda: I know.

The problem is that InboundMessageHandler (https://github.com/hyperledger/aries-framework-go/pull/1033/files#diff-9988f5820bf05228e5657f2842fa071eR107) will now unmarshal twice in these cases.

pkg/didcomm/common/service/service.go
troyronda: I am unsure that `data-type` will exist, so it seems premature to add it :).

pkg/didcomm/common/service/service.go
sudeshrshetty: As discussed, moved `~purpose` to header and Accept() criteria will use DID Comm msg header.

pkg/didcomm/common/service/service.go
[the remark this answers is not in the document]
sudeshrshetty: Removed `MessageHeader`

pkg/didcomm/common/service/service.go
sudeshrshetty: Removed it

pkg/didcomm/common/service/service.go
sudeshrshetty: done

pkg/didcomm/common/service/service.go
```diff
@@ -31,9 +41,10 @@ type DIDComm interface {
 
 // Header helper structure which keeps reusable fields
 type Header struct {
-	ID     string           `json:"@id"`
-	Thread decorator.Thread `json:"~thread"`
-	Type   string           `json:"@type"`
+	ID      string           `json:"@id"`
+	Thread  decorator.Thread `json:"~thread"`
+	Type    string           `json:"@type"`
+	Purpose string           `json:"~purpose"`
```
troyronda: This will need to be revisited.

pkg/framework/context/context.go
```diff
@@ -23,6 +23,7 @@ import (
 // Provider supplies the framework configuration to client objects.
 type Provider struct {
 	services                 []dispatcher.Service
```
troyronda: probably should followup by renaming Service and services too.

pkg/didcomm/common/service/service.go
sudeshrshetty: changed it to `[]string` as per RFC (https://github.com/hyperledger/aries-rfcs/pull/351) 
Are you sure it should be a map? Still create issue #1037 and added TODO

PR #1036, opened 2020-01-07
feat: Add route client

### LINKED ISSUE #1031: Add route client

fqutishat:
closes #1031

### DIFF
--- a/pkg/client/route/client.go
+++ b/pkg/client/route/client.go
@@ -1,0 +1,85 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package route
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
+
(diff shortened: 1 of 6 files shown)


pkg/client/route/client.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	routeSvc, ok := svc.(protocolService)
+	if !ok {
+		return nil, errors.New("cast service to route service failed")
+	}
+
+	connectionStore, err := connectionstore.NewConnectionRecorder(ctx)
```
troyronda: Do you need R/W access to connections in this protocol?
(or read only.)

fqutishat: just read only

troyronda: Also L32

fqutishat: change it to connection lookup

pkg/client/route/client.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return &Client{
+		routeSvc:         routeSvc,
+		connectionLookup: connectionLookup,
+	}, nil
+}
+
+// SendRequest send route request
+func (c *Client) SendRequest(connectionID string) (string, error) {
```
troyronda: We may need to enhance the naming... `SendRequest` seems very generic.

PR #1038, opened 2020-01-08
feat: Define the Client interface for messenger

soluchok:
Added Client interface for messenger.

Closes https://github.com/hyperledger/aries-framework-go/issues/1039

### DIFF
--- a/pkg/didcomm/common/messenger/messenger.go
+++ b/pkg/didcomm/common/messenger/messenger.go
@@ -1,0 +1,32 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package messenger
+
+import "github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+
+// The messenger package is responsible for the handling of communication between agents.
+// It takes care of threading, timing, etc.
+// Each message that we are going to send should be DIDCommMsg type.
+// e.g we have type model.Ack{Type: "type", ID: "id"}
+// it should be converted to: map[@id:id @type:type]
+// NOTE: The package modifies message data by JSON tag name according to aries-rfcs.
+//       Fields like @id, ~thread , etc. are redundant and may be rewritten.
+
+// Messenger provides methods for the communication
+type Messenger interface {
+	// ReplyTo replies to the message by given msgID.
+	// Keeps threadID in the *decorator.Thread.
+	// Using this function means that communication will be on the same thread.
+	ReplyTo(msgID string, msg service.DIDCommMsg)
+
+	// Send sends the message by starting a new thread.
+	Send(msg service.DIDCommMsg, myDID, theirDID string)
+
+	// ReplyToNested sends the message by starting a new thread.
+	// Keeps parent threadID in the *decorator.Thread
+	ReplyToNested(threadID string, msg service.DIDCommMsg)
+}


cmd/aries-agent-rest/go.sum
```diff
@@ -81,6 +81,7 @@ github.com/pquerna/cachecontrol v0.0.0-20180517163645-1555304b9b35/go.mod h1:prY
 github.com/rs/cors v1.7.0 h1:+88SsELBHx5r+hZ8TCkggzSstaWNbDvThkVK8H6f9ik=
 github.com/rs/cors v1.7.0/go.mod h1:gFx+x8UowdsKA9AchylcLynDq+nNFfI8FkUZdN/jGCU=
 github.com/russross/blackfriday v1.5.2/go.mod h1:JO/DiYxRf+HjHt06OyowR9PTA263kcR/rfWxYHBV53g=
+github.com/soluchok/mapstructure v1.1.3/go.mod h1:6u2xI+dOJ31VigNoBbRGDz7imkS+LpWLIQkRAYEnE90=
```
troyronda: we shouldn't depend on personal forks.

pkg/messenger/messenger.go
```diff
@@ -0,0 +1,34 @@
...
+// e.g we have type model.Ack{Type: "type", ID: "id"}
+// it should be converted to: map[@id:id @type:type]
+// NOTE: The package modifies message data by JSON tag name according to aries-rfcs.
+//       Fields like @id, ~thread , etc. are redundant and may be rewritten.
+
+// Client provides methods for the communication
+type Client interface {
+	// HandleMessage prepares data needed for the communication according to the options.
+	// Options should be provided by the service.
+	HandleMessage(msg map[string]interface{}, opts ...interface{})
```
troyronda: Is it possible to at least have structure for the DIDComm header fields?

cmd/aries-agent-rest/go.sum
soluchok: an example will be removed

pkg/messenger/messenger.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package messenger
```
troyronda: where do you plan to put the implementation? perhaps the interface should be in `pkg/common/api/messenger` or similar?

llorllale:
@soluchok please include the issue number in the PR's description

pkg/messenger/messenger.go
```diff
@@ -0,0 +1,34 @@
...
+// Client provides methods for the communication
+type Client interface {
+	// HandleMessage prepares data needed for the communication according to the options.
+	// Options should be provided by the service.
+	HandleMessage(msg map[string]interface{}, opts ...interface{})
+
+	// ReplyTo replies to the message by given msgID.
+	// Keeps threadID in the *decorator.Thread.
+	// Using this function means that communication will be on the same thread.
+	ReplyTo(msgID string, msg map[string]interface{}, myDID, theirDID string)
```
llorllale: Seems like the inclusion of `myDID` and `theirDID` are superfluous if a user has the `msgID`.

soluchok: Agree!

pkg/messenger/messenger.go
soluchok: implementation - `pkg/didcomm/messenger`
interface - pkg/didcomm/common/messenger

@troyronda @llorllale  Are you ok with that?

pkg/messenger/messenger.go
soluchok: also, I have changed `ReplyToNested(msgID string, msg map[string]interface{})`

pkg/didcomm/common/messenger/messenger.go
```diff
@@ -0,0 +1,34 @@
...
+// e.g we have type model.Ack{Type: "type", ID: "id"}
+// it should be converted to: map[@id:id @type:type]
+// NOTE: The package modifies message data by JSON tag name according to aries-rfcs.
+//       Fields like @id, ~thread , etc. are redundant and may be rewritten.
+
+// Client provides methods for the communication
+type Client interface {
+	// HandleMessage receives incoming message and prepares data needed for the communication according to the options.
+	// Options should be provided by the service.
+	HandleMessage(msg map[string]interface{}, myDID, theirDID string, opts ...interface{})
```
troyronda: I assume you are doing the type alias?

pkg/didcomm/common/messenger/messenger.go
```diff
@@ -0,0 +1,34 @@
...
+// The messenger package is responsible for the handling of communication between agents.
+// It takes care of threading, timing, etc.
+// Each message that we are going to send should be map[string]interface{} type.
+// e.g we have type model.Ack{Type: "type", ID: "id"}
+// it should be converted to: map[@id:id @type:type]
+// NOTE: The package modifies message data by JSON tag name according to aries-rfcs.
+//       Fields like @id, ~thread , etc. are redundant and may be rewritten.
+
+// Client provides methods for the communication
+type Client interface {
```
llorllale: Who are the intended consumers of this interface?

soluchok: Any service/protocol may use it. `aries.context` will provide it. We will be able to get a messenger from the context.

pkg/didcomm/common/messenger/messenger.go
```diff
@@ -0,0 +1,36 @@
...
+	// Keeps threadID in the *decorator.Thread.
+	// Using this function means that communication will be on the same thread.
+	ReplyTo(msgID string, msg service.DIDCommMsg)
+
+	// Send sends the message by starting a new thread.
+	Send(msg service.DIDCommMsg, myDID, theirDID string)
+
+	// ReplyToNested sends the message by starting a new thread.
+	// Keeps parent threadID in the *decorator.Thread
+	ReplyToNested(msgID string, msg service.DIDCommMsg)
```
llorllale: What is the use case for this?

Nested threads typically start a new protocol. In those situations, this method's signature implies that a protocol service will know how to build messages of another protocol.

Please clarify.

pkg/didcomm/common/messenger/messenger.go
```diff
@@ -0,0 +1,36 @@
...
+// e.g we have type model.Ack{Type: "type", ID: "id"}
+// it should be converted to: map[@id:id @type:type]
+// NOTE: The package modifies message data by JSON tag name according to aries-rfcs.
+//       Fields like @id, ~thread , etc. are redundant and may be rewritten.
+
+// Client provides methods for the communication
+type Client interface {
+	// HandleMessage receives incoming message and prepares data needed for the communication according to the options.
+	// Options should be provided by the service.
+	HandleMessage(msg service.DIDCommMsg, myDID, theirDID string, opts ...interface{})
```
llorllale: What are the use cases for this method?

soluchok: HandleMessage will receive all incoming messages. According to `opts ...interface{}` which may be provided by any service, the behavior of this function may be different.

The function is responsible for saving some data which will be retrieved later by ReplyTo, Send, ReplyToNested functions. 

HandleMessage should provide correct sate for  ReplyTo, Send, ReplyToNested functions.
Actually, this function is responsible for counting incoming messages (probably there will be more logic), also it will provide us the possibility to get the threadID by msgID.

It is a wrapper function that will be executed by the dispatcher when we received an incoming message.

llorllale: I still don't understand.

The "messaging service" is a separate component from the protocol services. It will act as the "glue" between protocol services. Right?

Questions:

* What use is HandleMessage when this client can accept message via the other methods?
* How does a protocol service trigger another protocol using this interface?

pkg/didcomm/common/messenger/messenger.go
soluchok: Correct, nested starts a new protocol. But also it requires parent thread id. Instead of threadID we are using messageID. So, no matter which service starts a new protocol(nested) we should pass the messageID which causes this action. Then, we will get the correct thread ID by messageID .

pkg/didcomm/common/messenger/messenger.go
soluchok: The `messenger` is a wrapper over dispatcher.  Instead of using `Outbound` you may use `messenger`. The `messenger` responsibility are:
* add message ID if it is needed
* take care of threading (thid,pthid,sender_order,received_orders)
* possibility to modify message payload(generic) automatically 

e.g You do not need to count sender_order or received_orders for the thread decorator.

sudeshrshetty: - if you plan to expose this interface through context, then `HandleMessage()` should be moved out of this interface since it doesn't make sense to expose it. Services/Clients shouldn't call this function explicitly.
My suggestion to split this package and expose only Send/Reply in client interface.
- Name `Client` doesn't look good if it is going to be exposed through context
- Should get rid of current outbound dispatcher once this interface is available in context

pkg/messenger/messenger.go
llorllale: @soluchok yes

pkg/didcomm/common/messenger/messenger.go
llorllale: @soluchok my issue with this is that protocol services should not know the details of the messages of other protocols.

soluchok: ok, then I will change it to threadID

pkg/didcomm/common/messenger/messenger.go
soluchok: I will remove HandleMessage from the interface. What about the name? Suggestions?

troyronda: @sudeshrshetty @soluchok @rolsonquadras  what's the status on this comment?

sudeshrshetty: `Messenger` should be the right name for this interface, but `messenger.Messenger` stutters.
Should we think about moving this interface to `dispatcher` package?

pkg/didcomm/common/messenger/messenger.go
soluchok: Done!

pkg/didcomm/common/messenger/messenger.go
soluchok: Done!

pkg/didcomm/common/messenger/messenger.go
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package messenger
```
troyronda: I assume this is a new package since these definitions are used beyond services. Correct?

- Shouldn't this package be more generic (e.g., message / msg).
- Shouldn't the DIDCommMsg interface be moved into this same package (in your next PR)?

PR #1042, opened 2020-01-09
refactor: null values in optional fields of Verifiable Credential

### LINKED ISSUE #1041: null values in optional fields of Verifiable Credential
Allow the presence of null values in optional fields of Verifiable Credential when decoding (from JSON).

kdimak:
closes #1041

Additionally, support the case when `RefreshService` is defined as an array.

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -115,18 +115,28 @@ const defaultSchema = `{
       "format": "date-time"
     },
     "proof": {
-      "type": "object",
-      "required": [
-        "type"
-      ],
-      "properties": {
-        "type": {
-          "type": "string"
+      "anyOf": [
+        {
+          "type": "object",
+          "required": [
+            "type"
+          ],
+          "properties": {
+            "type": {
+              "type": "string"
+            }
+          }
+        },
+        {
+          "type": "null"
         }
-      }
+      ]
     },
     "expirationDate": {
-      "type": "string",
+      "type": [
+        "string",
+        "null"
+      ],
       "format": "date-time"
     },
     "credentialStatus": {
@@ -144,30 +154,37 @@ const defaultSchema = `{
   },
   "definitions": {
     "typedID": {
-      "type": "object",
-      "required": [
-        "id",
-        "type"
-      ],
-      "properties": {
-        "id": {
-          "type": "string",
-          "format": "uri"
+      "anyOf": [
+        {
+          "type": "null"
         },
-        "type": {
-          "anyOf": [
-            {
-              "type": "string"
+        {
+          "type": "object",
+          "required": [
+            "id",
+            "type"
+          ],
+          "properties": {
+            "id": {
+              "type": "string",
+              "format": "uri"
             },
-            {
-              "type": "array",
-              "items": {
-                "type": "string"
-              }
+            "type": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "array",
+                  "items": {
+                    "type": "string"
+                  }
+                }
+              ]
             }
-          ]
+          }
         }
-      }
+      ]
     },
     "typedIDs": {
       "anyOf": [
@@ -179,6 +196,9 @@ const defaultSchema = `{
           "items": {
             "$ref": "#/definitions/typedID"
           }
+        },
+        {
+          "type": "null"
         }
       ]
     }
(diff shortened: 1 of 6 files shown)


kdimak:
> The context right now takes only array of string with 2 mandatory value. As per the spec [10:43 AM]
> one (base) is mandatory - https://www.w3.org/2018/credentials/v1
> Other one is optional.But looks like both are mandatory

No, only the base one (https://www.w3.org/2018/credentials/v1) is mandatory. We are using a lot of the following [credential document](https://github.com/hyperledger/aries-framework-go/blob/d23287c6981e6f387f7492610fa9335b66f3c0b0/pkg/doc/verifiable/support_test.go#L25) in VC tests.

pkg/doc/verifiable/credential.go
```diff
@@ -796,12 +826,34 @@ func newCredential(raw *rawCredential) (*Credential, error) {
...
+	if err == nil {
+		return []TypedID{singleTypedID}, nil
+	}
+
+	var composedTypedID []TypedID
+
+	err = json.Unmarshal(bytes, &composedTypedID)
+	if err == nil {
+		return composedTypedID, nil
+	}
```
DRK3: Just wondering what's going on here (where you unmarshal to a TypedID and if that fails, you try unmarshalling to an array of TypedID). Is that because you don't know if the JSON will be an array or a single value?

pkg/doc/verifiable/credential.go
```diff
@@ -1062,9 +1124,25 @@ func contextToSerialize(context []string, cContext []interface{}) interface{} {
 	return context
 }
 
+func typedIdsToSerialize(typedIds []TypedID) ([]byte, error) {
```
DRK3: ID should be capitalized

DRK3: (also the variables - typedIDs)

DRK3: Do you mean "ToSerialize**d**" instead of "ToSerialize"? As in a conversion of []TypedID into []byte?

pkg/doc/verifiable/credential.go
kdimak: Yes, exactly. It can be defined either as a single value or as an array. It's a common thing for VC data model (unfortunately).

pkg/doc/verifiable/credential.go
kdimak: Naming is not good, I will think of some better names here...

DRK3: ok :)

pkg/doc/verifiable/credential.go
DRK3: Gotcha - just wanted to confirm :)

pkg/doc/verifiable/credential.go
kdimak: It's now `typedIDsToRaw()`. I also renamed functions similar to this one.

talwinder50:
> credential document

Ok i re run the test for me it was failing. Will verify

pkg/doc/verifiable/credential.go
```diff
@@ -369,26 +389,26 @@ type Credential struct {
...
+	Type           interface{}     `json:"type,omitempty"`
+	Subject        Subject         `json:"credentialSubject,omitempty"`
+	Issued         *time.Time      `json:"issuanceDate,omitempty"`
+	Expired        *time.Time      `json:"expirationDate,omitempty"`
+	Proof          *Proof          `json:"proof,omitempty"`
+	Status         *TypedID        `json:"credentialStatus,omitempty"`
+	Issuer         interface{}     `json:"issuer,omitempty"`
+	Schema         interface{}     `json:"credentialSchema,omitempty"`
+	Evidence       *Evidence       `json:"evidence,omitempty"`
+	TermsOfUse     json.RawMessage `json:"termsOfUse,omitempty"`
```
troyronda: why was the change from TypedID to RawMessage needed?

kdimak: Because we don't know is it defined as a single object or an array. That's why we postpone the parsing.

pkg/doc/verifiable/credential.go
```diff
@@ -369,26 +389,26 @@ type Credential struct {
...
-	Issuer         interface{} `json:"issuer,omitempty"`
-	Schema         interface{} `json:"credentialSchema,omitempty"`
-	Evidence       *Evidence   `json:"evidence,omitempty"`
-	TermsOfUse     []TypedID   `json:"termsOfUse,omitempty"`
-	RefreshService *TypedID    `json:"refreshService,omitempty"`
+	Context        interface{}     `json:"@context,omitempty"`
+	ID             string          `json:"id,omitempty"`
+	Type           interface{}     `json:"type,omitempty"`
+	Subject        Subject         `json:"credentialSubject,omitempty"`
+	Issued         *time.Time      `json:"issuanceDate,omitempty"`
```
talwinder50: Dima, from you user perspective, 	
issueDate := time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day(),
		time.Now().Hour(), time.Now().Minute(), time.Now().Second(), 0, time.UTC) Validation only passed when i was entering this structure of time in the vc. Is that correct ?

kdimak: @talwinder50 Correct.

PR #1050, opened 2020-01-09
refactor: removed provider directory and moved tinkcrypto pkg up

Baha-sk:


pkg/framework/aries/default.go
```diff
@@ -21,8 +21,8 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
 	arieshttp "github.com/hyperledger/aries-framework-go/pkg/didcomm/transport/http"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
-	"github.com/hyperledger/aries-framework-go/pkg/framework/provider/tinkcrypto"
 	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/tinkcrypto"
```
troyronda: tinkcrypto does not belong at the top level.

Baha-sk: I just pushed another patch where I moved `pkg/common/api/crypto` to `pkg/crypto` and `pkg/framework/provider/tinkcrypto` to `pkg/crypto/tinkcrypto`

this way the interface and implementation are both under `pkg/crypto`

Issue #1039, opened 2020-01-08
Define the Client interface for messenger


soluchok:
We need to define the interface for the communication between agents.

Issue #1040, opened 2020-01-08
service.Header.Purpose?


llorllale:
https://github.com/hyperledger/aries-framework-go/blob/4d6d2333de088df2e2f555065ad99f4090a0be5d/pkg/didcomm/common/service/service.go#L42-L49

My mental model of `service.Header.Purpose` was that of a collection of common attributes that all DIDComm protocols share: `@id`, `@type`, `~thread`, etc. Which is why the addition of `~purpose` does not make sense as it is defined by a very specific protocol.

Should the handling of `~purpose` be contained within its specific protocol service package?

sudeshrshetty:
We have a plan to move all the decorators out to separate struct, I am not sure if we already have a in-progress issue for this.

PR #1054, opened 2020-01-10
feat: Replace DIDCommMsg type with map[string]interface{}

soluchok:
The intention of this PR is to generalize all messages in the system/framework. Because our messages are JSON based messages the structure map[string]interface is enough to make messages with the same type. The benefit of this can be:
* we cannot send an array/string etc.
* the new message type provides common functions ID(),Type() etc.
* message can be modified directly by key (without knowing the structure)

Since the message is a map, it can be modified by some other package/client. To prevent this we need to expose only the message interface. The interface provides Type and Scan functions. Having these functions any package/client may convert a message into a specific structure.

```
type DIDMsg interface {
	ID() string
	Type() string

	ThreadID() (string, error)
	Scan(v interface{}) error
}
```

The usage:
In our current logic, we send events to the client/user. DIDCommAction and StateMsg accordingly. In that case, we cannot expose the message to the client. Because we are using DIDCommMsg message in our services (background handling e.g Continue function).  So, instead of exposing DIDCommMsg we should expose an interface DIDMsg.

Closes https://github.com/hyperledger/aries-framework-go/issues/1065 https://github.com/hyperledger/aries-framework-go/issues/1039

### DIFF
--- a/pkg/didcomm/common/service/did_comm_msg.go
+++ b/pkg/didcomm/common/service/did_comm_msg.go
@@ -1,0 +1,138 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package service
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+)
+
+const (
+	jsonID       = "@id"
+	jsonType     = "@type"
+	jsonThread   = "~thread"
+	jsonThreadID = "thid"
+)
+
+// Header helper structure which keeps reusable fields
+// Deprecated: Use DIDCommMsg instead of Header
+// TODO: Remove deprecated Header structure https://github.com/hyperledger/aries-framework-go/issues/1075
+type Header struct {
+	ID     string           `json:"@id"`
+	Thread decorator.Thread `json:"~thread"`
+	Type   string           `json:"@type"`
+	// TODO revisit ~purpose, should be generic map [Issue #1037]
+	Purpose []string `json:"~purpose"`
+}
+
+// DIDCommMsgMap did comm msg
+type DIDCommMsgMap map[string]interface{}
+
+// NewDIDCommMsg returns DIDCommMsg with Header
+func NewDIDCommMsg(payload []byte) (DIDCommMsgMap, error) {
+	var msg DIDCommMsgMap
+
+	err := json.Unmarshal(payload, &msg)
+	if err != nil {
+		return nil, fmt.Errorf("invalid payload data format: %w", err)
+	}
+
+	return msg, nil
+}
+
+// ThreadID returns msg ~thread.thid if there is no ~thread.thid returns msg @id
+// message is invalid if ~thread.thid exist and @id is absent
+func (m DIDCommMsgMap) ThreadID() (string, error) {
+	if m == nil {
+		return "", ErrInvalidMessage
+	}
+
+	msgID := m.ID()
+	thread, ok := m[jsonThread].(map[string]interface{})
+
+	if ok && thread[jsonThreadID] != nil {
+		var thID string
+		if v, ok := thread[jsonThreadID].(string); ok {
+			thID = v
+		}
+
+		// if message has ~thread.thid but @id is absent this is invalid message
+		if len(thID) > 0 && msgID == "" {
+			return "", ErrInvalidMessage
+		}
+
+		if len(thID) > 0 {
+			return thID, nil
+		}
+	}
+
+	// we need to return it only if there is no ~thread.thid
+	if len(msgID) > 0 {
+		return msgID, nil
+	}
+
+	return "", ErrThreadIDNotFound
+}
+
+// Type returns the message type
+func (m DIDCommMsgMap) Type() string {
+	if m == nil || m[jsonType] == nil {
+		return ""
+	}
+
+	res, ok := m[jsonType].(string)
+	if !ok {
+		return ""
+	}
+
+	return res
+}
+
+// ID returns the message id
+func (m DIDCommMsgMap) ID() string {
+	if m == nil || m[jsonID] == nil {
+		return ""
+	}
+
+	res, ok := m[jsonID].(string)
+	if !ok {
+		return ""
+	}
+
+	return res
+}
+
+// Decode converts message to  struct
+func (m DIDCommMsgMap) Decode(v interface{}) error {
+	decoder, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+		DecodeHook: func(rt1 reflect.Type, rt2 reflect.Type, v interface{}) (interface{}, error) {
+			if rt1.Kind() == reflect.String && rt2 == reflect.TypeOf(time.Time{}) {
+				return time.Parse(time.RFC3339, v.(string))
+			}
+
+			if rt1.Kind() == reflect.String && rt2.Kind() == reflect.Slice {
+				return base64.StdEncoding.DecodeString(v.(string))
+			}
+
+			return v, nil
+		},
+		WeaklyTypedInput: true,
+		Result:           v,
+		TagName:          "json",
+	})
+	if err != nil {
+		return err
+	}
+
+	return decoder.Decode(m)
+}
--- a/pkg/didcomm/common/service/event.go
+++ b/pkg/didcomm/common/service/event.go
@@ -33,7 +33,7 @@ type StateMsg struct {
 	StateID string
 
 	// DIDComm message along with message type
-	Msg *DIDCommMsg
+	Msg DIDCommMsg
 
 	// Properties contains value based on specific protocol. The consumers need to call the protocol client
 	// functions to get the data.
@@ -52,7 +52,7 @@ type DIDCommAction struct {
 	ProtocolName string
 
 	// DIDComm message
-	Message *DIDCommMsg
+	Message DIDCommMsg
 
 	// Continue function to be called by the consumer for further processing the message.
 	Continue func(args interface{})
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -42,7 +42,7 @@ const (
 
 // message type to store data for eventing. This is retrieved during callback.
 type message struct {
-	Msg           *service.DIDCommMsg
+	Msg           service.DIDCommMsgMap
 	ThreadID      string
 	Options       *options
 	NextStateName string
@@ -64,8 +64,7 @@ type provider interface {
 
 // stateMachineMsg is an internal struct used to pass data to state machine.
 type stateMachineMsg struct {
-	header     *service.Header
-	payload    []byte
+	service.DIDCommMsg
 	connRecord *connection.Record
 	options    *options
 }
@@ -121,8 +120,8 @@ func New(prov provider) (*Service, error) {
 }
 
 // HandleInbound handles inbound didexchange messages.
-func (s *Service) HandleInbound(msg *service.DIDCommMsg, myDID, theirDID string) (string, error) {
-	logger.Debugf("receive inbound message : %s", msg.Payload)
+func (s *Service) HandleInbound(msg service.DIDCommMsg, myDID, theirDID string) (string, error) {
+	logger.Debugf("receive inbound message : %s", msg)
 
 	// fetch the thread id
 	thID, err := threadID(msg)
@@ -131,7 +130,7 @@ func (s *Service) HandleInbound(msg *service.DIDCommMsg, myDID, theirDID string)
 	}
 
 	// valid state transition and get the next state
-	next, err := s.nextState(msg.Header.Type, thID)
+	next, err := s.nextState(msg.Type(), thID)
 	if err != nil {
 		return "", fmt.Errorf("handle inbound - next state : %w", err)
 	}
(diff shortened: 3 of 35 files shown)


rolsonquadras:
@soluchok I would be nice if we can describe the reason for this change. Currently, I don't see any in PR or Issue description.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// we need to return it only if there is no ~thread.thid
+	if len(msgID) > 0 {
+		return msgID, nil
+	}
+
+	return "", ErrThreadIDNotFound
+}
+
+// Purpose returns list of elements
```
sudeshrshetty: returns message purpose?

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,309 @@
...
+	if !ok {
+		return ""
+	}
+
+	return res
+}
+
+// ID returns the message id
+func (m DIDCommMsg) ID() string {
+	if m == nil || m[jsonID] == nil {
```
sudeshrshetty: what if ID is ""?

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,309 @@
...
+	res, ok := m[jsonID].(string)
+	if !ok {
+		return ""
+	}
+
+	return res
+}
+
+// ToStruct converts msh message to  struct
+func (m DIDCommMsg) ToStruct(v interface{}) (err error) {
```
sudeshrshetty: Just curious, why can't we use "https://github.com/mitchellh/mapstructure"?

pkg/framework/context/context.go
```diff
@@ -129,13 +129,19 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
...
 		// find generic inbound services registered for given message header
 		for _, svc := range p.msgSvcProvider.Services() {
-			if svc.Accept(msg.Header) {
+			h := &service.Header{}
+			err = msg.ToStruct(h)
+			if err != nil {
+				return err
+			}
+
+			if svc.Accept(h) {
```
sudeshrshetty: Passing whole message may be risky to check acceptance criteria, for now add a TODO to pass subset of DIDComm msg map.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,309 @@
...
+	// we need to return it only if there is no ~thread.thid
+	if len(msgID) > 0 {
+		return msgID, nil
+	}
+
+	return "", ErrThreadIDNotFound
+}
+
+// Purpose returns list of elements
+func (m DIDCommMsg) Purpose() []string {
```
llorllale: Please remove this method. `Purpose` will be removed from the header (see #1040 and #1058).

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,309 @@
...
+// Type returns the message type
+func (m DIDCommMsg) Type() string {
+	if m == nil || m[jsonType] == nil {
+		return ""
+	}
+
+	res, ok := m[jsonType].(string)
+	if !ok {
+		return ""
+	}
```
llorllale: We're hiding an error condition here. Likewise in `ID()`.

pkg/didcomm/common/service/did_comm_msg.go
llorllale: Consider renaming to [Scan](https://golang.org/pkg/database/sql/#Rows.Scan).

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -275,7 +273,7 @@ func (s *Service) handle(msg *message, aEvent chan<- service.DIDCommAction) erro
 		s.sendMsgEvents(&service.StateMsg{
 			ProtocolName: DIDExchange,
 			Type:         service.PostState,
-			Msg:          msg.Msg.Clone(),
```
llorllale: It's interesting that we now lose this aspect (no longer exporting a copy).

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -275,14 +275,19 @@ func (s *Service) InvitationReceived(msg service.StateMsg) error {
 		return nil
 	}
 
-	if msg.Msg.Header == nil || msg.Msg.Header.Thread.PID == "" {
+	var h = service.Header{}
+	if err := msg.Msg.ToStruct(&h); err != nil {
```
llorllale: I really think `Scan` would be a better name. `ToStruct` implies the whole message will be scanned into the input struct.

llorllale:
@soluchok we lost the `Clone()` method. Rather than recreating it, I propose we make `DIDCommMsg` a struct and hide the map inside it. Leave `ID()`, `ThreadID()`, and `ToStruct()/Scan()` as they are. We get immutability for free this way + clients can scan to their specific struct with `ToStruct()/Scan()`

pkg/didcomm/common/service/did_comm_msg.go
soluchok: mapstructure not maintainable anymore https://github.com/mitchellh/mapstructure/pull/145#issuecomment-572053192

pkg/framework/context/context.go
soluchok: Accept may use an interface e.g
```
type Typer interface {
   Type() string
}
```

pkg/didcomm/protocol/didexchange/service.go
llorllale: An interface or a struct is fine by me personally. Gophers would probably prefer it be a struct though.

pkg/didcomm/common/service/did_comm_msg.go
llorllale: I still think it's better than rolling our own code until we have a good reason to.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: I think the best way is to return a default value.

pkg/didcomm/common/service/did_comm_msg.go
troyronda: @sudeshrshetty thoughts?

pkg/didcomm/common/service/did_comm_msg.go
troyronda: @soluchok @llorllale I believe the issue is that we need omit:empty and that feature hasn't been merged into mapstructure. correct?

pkg/didcomm/protocol/didexchange/service.go
troyronda: An interface should only be returned if that is need to satisfy the signature of another interface.
Otherwise return a struct. The receiver should use an interface, of course.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,309 @@
...
+const (
+	jsonID       = "@id"
+	jsonType     = "@type"
+	jsonPurpose  = "~purpose"
+	jsonThread   = "~thread"
+	jsonThreadID = "thid"
+)
+
+// Header helper structure which keeps reusable fields
+// Deprecated: Use DIDCommMsg instead of Header
```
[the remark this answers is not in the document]
soluchok: I can add a TODO, we can remove it on another PR. This one is too big.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: The ID function will return the default value. Default is "".

pkg/didcomm/common/service/did_comm_msg.go
sudeshrshetty: @llorllale We are just removing it from header, may be replace this from key based decorator lookup?

pkg/didcomm/common/service/did_comm_msg.go
sudeshrshetty: Shouldn't it be invalid error if ID is empty?

pkg/framework/context/context.go
sudeshrshetty: We are also going to check it based on `~purpose`.

pkg/didcomm/common/service/did_comm_msg.go
llorllale: @sudeshrshetty by "key based decorator lookup" do you mean leave it in the map? Sure...

pkg/didcomm/common/service/did_comm_msg.go
llorllale: `NewDidCommMsg()` should guarantee these base values exist.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: yeah, but I think it is not a big problem

pkg/didcomm/protocol/didexchange/service.go
llorllale: Two public artifacts is overkill in this case. Either make it a public struct or a public interface + private struct. It's immutable and the method set [ID(), Type(), ThreadID(), Scan()] to me strikes the ideal balance between size, cohesion, and generality/usefulness.

pkg/didcomm/common/service/did_comm_msg.go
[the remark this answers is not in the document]
soluchok: Added TODO

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -217,7 +216,7 @@ func (s *Service) handle(msg *message, aEvent chan<- service.DIDCommAction) erro
 		s.sendMsgEvents(&service.StateMsg{
 			ProtocolName: DIDExchange,
 			Type:         service.PreState,
-			Msg:          msg.Msg.Clone(),
+			Msg:          msg.Msg,
```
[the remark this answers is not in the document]
soluchok: Instead of copy, we are providing an interface.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Since the message is generic, it doesn't make sense to return an error. The empty type should be returned. We may provide the IsValid function to check the message if it is needed. For the functions which have some logic, I agree there should be an error. e.g ThreadID function. When we just want to get the direct value, it does not make sense to return an error.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: https://github.com/hyperledger/aries-framework-go/pull/1054#discussion_r367974155

pkg/framework/context/context.go
troyronda: @Moopli ^^^

pkg/didcomm/protocol/introduce/models.go
```diff
@@ -64,7 +64,8 @@ type Content struct {
 // PleaseIntroduceTo includes all field from To structure
 // also it has Discovered the field which should be provided by help-me-discover protocol
 type PleaseIntroduceTo struct {
-	To
+	// nolint: staticcheck
```
troyronda: what was the linter issue?

soluchok: JSON does not have `squash` e.g `json:",squash"`
Squash needs to be used when we inherit a type.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,138 @@
...
+type Header struct {
+	ID     string           `json:"@id"`
+	Thread decorator.Thread `json:"~thread"`
+	Type   string           `json:"@type"`
+	// TODO revisit ~purpose, should be generic map [Issue #1037]
+	Purpose []string `json:"~purpose"`
+}
+
+// DIDCommMsg did comm msg
+type DIDCommMsg map[string]interface{}
```
troyronda: This should probably not be in common/service anymore.

It is an implementation rather than a common interface being consumed by other packages.

pkg/didcomm/common/service/event.go
```diff
@@ -17,6 +17,15 @@ const (
 	PostState
 )
 
+// DIDMsg describes DIDCommMsg interface
+type DIDMsg interface {
```
troyronda: Assuming you move the implementation to a more appropriate package, would this be named `DIDCommMsg`?

soluchok: Sure!

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Give me a hint of where to move it. `pkg/common/didcommmsg` ?

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -0,0 +1,138 @@
...
+	res, ok := m[jsonID].(string)
+	if !ok {
+		return ""
+	}
+
+	return res
+}
+
+// Scan converts message to  struct
+func (m DIDCommMsg) Scan(v interface{}) error {
```
Baha-sk: What does `Scan` mean for the user of this API? 

Personally, I prefer `DecodeTo` as portrayed by the function logic or `Convert` as per the doc comment.

or perhaps could it be renamed to `Unmarshal(v)` or `CopyTo(v)` ?  

The function name should be easy to guess and accurate [ref](https://talks.golang.org/2014/names.slide#3), `Scan` can mean different things to different people.

Baha-sk: From the code usage refactored in other files below, this function is used exactly like json.Unmarshal() ... so probably `Unmarshal()` makes more sense.

troyronda: I also had a bit trouble reconciling the word `Scan` in this usage.

Also note the underlying `mapstructure` library seems to call it Decode (https://godoc.org/github.com/mitchellh/mapstructure#Decode)

PR #1060, opened 2020-01-13
feat: Outbound dispatcher for message handler

### LINKED ISSUE #1059: MessageHandler - Register outbound dispatcher to send messages

sudeshrshetty:
- `framework.Messenger()` to returns messenger to support outbound messaging feature 
from agent framework
- closes #1059 

### DIFF
--- a/pkg/didcomm/common/service/service.go
+++ b/pkg/didcomm/common/service/service.go
@@ -45,6 +45,7 @@ type Header struct {
 	Thread decorator.Thread `json:"~thread"`
 	Type   string           `json:"@type"`
 	// TODO revisit ~purpose, should be generic map [Issue #1037]
+	// TODO ~purpose should be out of header [Issue #1058, #1040]
 	Purpose []string `json:"~purpose"`
 }
 
--- a/pkg/didcomm/dispatcher/api.go
+++ b/pkg/didcomm/dispatcher/api.go
@@ -10,30 +10,20 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 )
 
-// Service protocol service
-type Service interface {
+// ProtocolService is service interface for protocol services available in framework
+// for matching acceptance criteria based on message type
+type ProtocolService interface {
 	service.Handler
 	Accept(msgType string) bool
 	Name() string
 }
 
 // MessageService is service for handling generic messages
-// matching accept criteria
+// matching accept criteria based on message header
 type MessageService interface {
 	service.InboundHandler
 	Accept(header *service.Header) bool
 	Name() string
-}
-
-// MessageHandler maintains registered message services
-// and it allows dynamic registration of message services
-type MessageHandler interface {
-	// Services returns list of available message services in this message handler
-	Services() []MessageService
-	// Register registers given message services to this message handler
-	Register(msgSvcs ...MessageService) error
-	// Unregister unregisters message service with given name from this message handler
-	Unregister(name string) error
 }
 
 // Outbound interface
--- a/pkg/didcomm/msghandler/msghandler.go
+++ b/pkg/didcomm/msghandler/msghandler.go
@@ -38,7 +38,10 @@ func (m *Registrar) Services() []dispatcher.MessageService {
 	m.lock.RLock()
 	defer m.lock.RUnlock()
 
-	return m.services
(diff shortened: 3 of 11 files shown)


pkg/didcomm/common/service/service.go
```diff
@@ -45,6 +45,7 @@ type Header struct {
 	Thread decorator.Thread `json:"~thread"`
 	Type   string           `json:"@type"`
 	// TODO revisit ~purpose, should be generic map [Issue #1037]
+	// TODO ~purpose should be out of header [Issue #1058]
```
llorllale: See #1040

sudeshrshetty: updated TODO

pkg/didcomm/dispatcher/api.go
```diff
@@ -30,6 +30,9 @@ type MessageService interface {
 type MessageHandler interface {
 	// Services returns list of available message services in this message handler
 	Services() []MessageService
+	// SetOutbound sets outbound dispatcher to this provider
+	// TODO should use dedicated messenger interface instead of Outbound dispatcher [Issue #1058]
+	SetOutbound(Outbound) error
```
llorllale: We don't use setters in this project. This can be injected when instantiating.

pkg/didcomm/msghandler/msghandler.go
```diff
@@ -12,13 +12,15 @@ package msghandler
...
 	"sync"
+	"sync/atomic"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 )
 
 const (
 	errAlreadyRegistered = "registration failed, message service with name `%s` already registered"
 	errNeverRegistered   = "failed to unregister, unable to find registered message service with name `%s`"
+	outboundAlreadySet   = "outbound is already set"
```
llorllale: missing `err` prefix (like the two above)

sudeshrshetty: fixed

pkg/didcomm/dispatcher/api.go
sudeshrshetty: this `SetOutbound()` is to set Outbound from framework to injected handler.

sudeshrshetty: Or we can call it `InitOutbound(...)` to avoid confusion

llorllale: Is there an expectation to change the `outbound` after instantiation?

sudeshrshetty: external code cant use context. This way, I was injecting framework's `outbound` to registrar. I am going to change this.

sudeshrshetty: removed `SetOutbound`

pkg/framework/aries/framework.go
```diff
@@ -251,6 +251,12 @@ func (a *Aries) Context() (*context.Provider, error) {
 	)
 }
 
+// Messenger returns messenger for sending messages through this agent framework
+// TODO should use dedicated messenger interface instead of Outbound dispatcher [Issue #1058]
+func (a *Aries) Messenger() dispatcher.Outbound {
+	return a.outboundDispatcher
+}
+
```
llorllale: If I heard correctly - the messenger should not be exposed through the framework. It should be an internal implementation detail (for protocol services to communicate with each other).

sudeshrshetty: No, this is for external code to send message through already running agent without using context

pkg/didcomm/dispatcher/api.go
```diff
@@ -10,8 +10,8 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 )
 
-// Service protocol service
-type Service interface {
+// ProtocolService is service interface for protocol services
+type ProtocolService interface {
```
rolsonquadras: A short description about difference between protocol services v message services would be helpful.

sudeshrshetty: done

pkg/framework/aries/framework.go
troyronda: We should see if there is a better idea as a followup.

PR #1071, opened 2020-01-15
feat: Register agent with Router

### LINKED ISSUE #1072: Register agent with Router
Provide a function in Route service to register the agent with the router. Once registration successful, the connectionID of the Router need to be stored, so that it can be used later on to lookup destination details of the router.

rolsonquadras:
- New function in Router Service to register the agent with the router

closes #1072 


pkg/didcomm/protocol/route/service.go
```diff
@@ -296,6 +328,80 @@ func (s *Service) handleForward(msg *service.DIDCommMsg) error {
...
+		return fmt.Errorf("fetch router connection id : %w", err)
+	}
+
+	if routerConnID != "" {
+		return errors.New("router is already registered")
+	}
+
+	// get the connection record for the ID to fetch DID information
+	conn, err := s.connectionLookup.GetConnectionRecord(connectionID)
+	if err != nil {
```
sudeshrshetty: Any reasons for separating these data not found error? Any handling in the caller?

pkg/didcomm/protocol/route/service.go
```diff
@@ -296,6 +328,80 @@ func (s *Service) handleForward(msg *service.DIDCommMsg) error {
...
+
+	// generate message ID
+	msgID := uuid.New().String()
+
+	// register chan for callback processing
+	grantCh := make(chan Grant)
+
+	s.routeRegistrationMapLock.Lock()
+	s.routeRegistrationMap[msgID] = grantCh
+	s.routeRegistrationMapLock.Unlock()
```
sudeshrshetty: Move set/get to different function, just to avoid accidents in future

pkg/didcomm/protocol/route/service.go
```diff
@@ -194,14 +216,24 @@ func (s *Service) handleRequest(msg *service.DIDCommMsg, myDID, theirDID string)
...
+	err := json.Unmarshal(msg.Payload, grantMsg)
 	if err != nil {
 		return fmt.Errorf("route grant message unmarshal : %w", err)
 	}
 
-	// TODO https://github.com/hyperledger/aries-framework-go/issues/948 integrate with framework components
+	// check if there are any channels registered for the message ID
+	s.routeRegistrationMapLock.RLock()
+	grantCh := s.routeRegistrationMap[grantMsg.ID]
+	s.routeRegistrationMapLock.RUnlock()
```
sudeshrshetty: Move get to different function just to avoid accidents in future

pkg/didcomm/protocol/route/service.go
rolsonquadras: yeah. we need to handle these cases differently in the caller ie. If router connection is present and that connection is not found, then user should be able to add a new router connectionID.

pkg/didcomm/protocol/route/service.go
rolsonquadras: [done](https://github.com/hyperledger/aries-framework-go/compare/4d6cf6d2117e03d7fa5fe3e7f5623c5d317dc7df..f2c71755a81cac382569f3cb14fae87828d66b2c#diff-7cc3df019152fe2c557826a312cc0a12R227).

pkg/didcomm/protocol/route/service.go
rolsonquadras: [done](https://github.com/hyperledger/aries-framework-go/compare/4d6cf6d2117e03d7fa5fe3e7f5623c5d317dc7df..f2c71755a81cac382569f3cb14fae87828d66b2c#diff-7cc3df019152fe2c557826a312cc0a12R361).

sudeshrshetty (approved):

pkg/didcomm/protocol/route/service.go
```diff
@@ -77,11 +89,14 @@ type provider interface {
...
-	endpoint   string
-	kms        legacykms.KeyManager
-	vdRegistry vdri.Registry
+	routeStore               storage.Store
+	connectionLookup         *connection.Lookup
+	outbound                 dispatcher.Outbound
+	endpoint                 string
+	kms                      legacykms.KeyManager
+	vdRegistry               vdri.Registry
+	routeRegistrationMap     map[string]chan Grant
```
llorllale: Why not storage?

pkg/didcomm/protocol/route/service.go
```diff
@@ -296,6 +326,91 @@ func (s *Service) handleForward(msg *service.DIDCommMsg) error {
...
+func (s *Service) Register(connectionID string) error {
+	// check if router is already registered
+	routerConnID, err := s.getRouterConnectionID()
+	if err != nil && !errors.Is(err, storage.ErrDataNotFound) {
+		return fmt.Errorf("fetch router connection id : %w", err)
+	}
+
+	if routerConnID != "" {
+		return errors.New("router is already registered")
+	}
```
llorllale: I'm not quite sure I understand this: if the data is not found, we return an error. If the data is found (?) we also return an error. 

saveRouterConnectionId() is only called from this method and is passed in the input connectionID as parameter.

I would have assumed that if getRouterConnectionID() returns ErrDataNotFound then the agent has not registered a route and therefore we can proceed.

rolsonquadras: We are not returning error if the data is not found. The condition [!errors.Is(err, storage.ErrDataNotFound)](https://github.com/hyperledger/aries-framework-go/pull/1071/files#diff-7cc3df019152fe2c557826a312cc0a12R338) makes sure that we continue even if there is No data.

pkg/didcomm/protocol/route/service.go
```diff
@@ -296,6 +326,91 @@ func (s *Service) handleForward(msg *service.DIDCommMsg) error {
 	return s.outbound.Forward(forward.Msg, dest)
 }
 
+// Register registers the agent with a router with the on connectionID passed in. The agent is
+// registered with the router and retrieves the router endpoint and routing keys. This function
+// throws an error if the agent is already registered against a router. This function internally
+// works as a sync function and works in-conjunction with handleGrant function. ie, The request
+// did-comm message is sent to the router and it waits for the grant response from the router
+// till timeout.
```
llorllale: We should be able to register against multiple routers.

Also, suggest some modifications to the text:

```
// Register this agent with the router on the other end of the connection identified by connectionID.
// This method blocks until a response is received from the router. 
```

pkg/didcomm/protocol/route/service.go
llorllale: D'Oh! My bad.

pkg/didcomm/protocol/route/service.go
```diff
@@ -194,14 +216,22 @@ func (s *Service) handleRequest(msg *service.DIDCommMsg, myDID, theirDID string)
...
 		return fmt.Errorf("route grant message unmarshal : %w", err)
 	}
 
-	// TODO https://github.com/hyperledger/aries-framework-go/issues/948 integrate with framework components
+	// check if there are any channels registered for the message ID
+	grantCh := s.getRouteRegistrationCh(grantMsg.ID)
+
+	if grantCh != nil {
+		// invoke the channel for the incoming message
+		go func(grantCh chan Grant, grantMsg Grant) {
```
troyronda: It doesn't seem correct to spin a goroutine to send on a channel. Why isn't the channel buffered for this case?

pkg/didcomm/protocol/route/service.go
rolsonquadras: Created an issue https://github.com/hyperledger/aries-framework-go/issues/1076 and added a TODO to register against multiple routers. Also, updated the goDoc.

pkg/didcomm/protocol/route/service.go
rolsonquadras: Agree. We don't need a go routine here as this logic is already running inside a goroutine. Updated the [code](https://github.com/hyperledger/aries-framework-go/compare/f2c71755a81cac382569f3cb14fae87828d66b2c..03621c455d43be8f44255077dc50427b5667fe6a#diff-7cc3df019152fe2c557826a312cc0a12R231).

go.sum
```diff
@@ -3,12 +3,11 @@ cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMT
 cloud.google.com/go v0.38.0 h1:ROfEUZz+Gh5pa62DJWXSaonyu3StP6EA6lPEXPI6mCo=
 cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSRU=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/VictoriaMetrics/fastcache v1.5.4 h1:0BaXbRH01RycJk79OOBwMCXlNryko9z4yEf6RqbP+Xo=
-github.com/VictoriaMetrics/fastcache v1.5.4/go.mod h1:ptDBkNMQI4RtmVo8VS/XwRY6RoTu1dAWCbrk+6WsEM8=
 github.com/VictoriaMetrics/fastcache v1.5.5 h1:HsBlzPgzKG0566YOl1mmfyz8SCU0zLKfbl9RDLsiLD8=
 github.com/VictoriaMetrics/fastcache v1.5.5/go.mod h1:ptDBkNMQI4RtmVo8VS/XwRY6RoTu1dAWCbrk+6WsEM8=
 github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412 h1:w1UutsfOrms1J05zt7ISrnJIXKzwaspym5BTKGx93EI=
 github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412/go.mod h1:WPjqKcmVOxf0XSf3YxCJs6N6AOSrOx3obionmG7T0y0=
+github.com/allegro/bigcache v1.2.1-0.20190218064605-e24eb225f156 h1:eMwmnE/GDgah4HI848JfFxHt+iPb26b4zyfspmqY0/8=
```
troyronda: @bstasyszyn @fqutishat @sandrask suggestions for a cache package?

rolsonquadras: @troyronda No new dependency was added in this PR and this got added as part of `go mod tidy`.

troyronda: Ah - I guess we we use github.com/bluele/gcache elsewhere and I'm curious why we changed in this project?

bstasyszyn: We've been using gcache - mostly because it has a lazy-loading feature. Fabric has beed using VictoriaMetrics/fastcache for their state DB.

pkg/didcomm/protocol/route/service.go
```diff
@@ -194,14 +216,20 @@ func (s *Service) handleRequest(msg *service.DIDCommMsg, myDID, theirDID string)
...
+	err := json.Unmarshal(msg.Payload, grantMsg)
 	if err != nil {
 		return fmt.Errorf("route grant message unmarshal : %w", err)
 	}
 
-	// TODO https://github.com/hyperledger/aries-framework-go/issues/948 integrate with framework components
+	// check if there are any channels registered for the message ID
+	grantCh := s.getRouteRegistrationCh(grantMsg.ID)
+
+	if grantCh != nil {
```
troyronda: when does nil happen?

go.sum
troyronda: Yeh I also see fastcache in this project. I wonder why we need both.

troyronda (approved):

go.sum
rolsonquadras: @troyronda It looks like VictoriaMetrics/fastcache internally depends on [`github.com/allegro/bigcache`]( https://github.com/VictoriaMetrics/fastcache/blob/master/go.mod#L4).

pkg/didcomm/protocol/route/service.go
rolsonquadras: If the channel is not registered for the msgID. basically when the agent doesn't recognize the message ID.

go.sum
troyronda: Ah - that's where it came from. cool, thanks.

PR #1074, opened 2020-01-16
feat: REST binding for message service registration and handling

### LINKED ISSUE #973: Generic Message Service: Controller API
Design : https://wiki.hyperledger.org/pages/viewpage.action?pageId=24778806

sudeshrshetty:
- added below endpoints for message service register, unregister and
registered service list

`POST /message/register-service`
`POST /message/unregister-service`
`GET /message/services`

- closes #973

### DIFF
--- a/pkg/restapi/operation/common/models.go
+++ b/pkg/restapi/operation/common/models.go
@@ -42,3 +42,63 @@ type CreatePublicDIDResponse struct {
 	// TODO return base64-encoded raw bytes of the DID doc [Issue: #855]
 	DID *did.Doc `json:"did"`
 }
+
+// RegisterMessageServiceRequest model
+//
+// This is used for operation to register a message service to message handler
+//
+// swagger:parameters registerMsgSvc
+type RegisterMessageServiceRequest struct {
+	// Params for registering message service
+	//
+	// in: body
+	Params *RegisterMsgSvcParams
+}
+
+// RegisterMsgSvcParams contains parameters for registering a message service to message handler
+type RegisterMsgSvcParams struct {
+	// Name of the message service
+	// required: true
+	Name string `json:"name"`
+
+	// Acceptance criteria for message service based on message purpose
+	// in case of multiple purposes, message will be dispatched if any one of the purpose matches
+	// with the purpose of incoming message.
+	// Can be provided in conjunction with other acceptance criteria.
+	Purpose []string `json:"purpose"`
+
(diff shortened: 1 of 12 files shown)


pkg/restapi/operation/common/models.go
```diff
@@ -42,3 +42,63 @@ type CreatePublicDIDResponse struct {
...
+type RegisterMsgSvcParams struct {
+	// Name of the message service
+	// required: true
+	Name string `json:"name"`
+
+	// Acceptance criteria for message service based on message purpose
+	// in case of multiple purposes, message will be dispatched if any one of the purpose matches
+	// with the purpose of incoming message.
+	// Can be provided in conjunction with other acceptance criteria.
+	Purpose []string `json:"purpose"`
```
troyronda: @sudeshrshetty @Moopli @llorllale I think Filip is updating the design to no longer need purpose?

pkg/restapi/operation/common/models.go
```diff
@@ -42,3 +42,63 @@ type CreatePublicDIDResponse struct {
...
+}
+
+// RegisterMsgSvcParams contains parameters for registering a message service to message handler
+type RegisterMsgSvcParams struct {
+	// Name of the message service
+	// required: true
+	Name string `json:"name"`
+
+	// Acceptance criteria for message service based on message purpose
+	// in case of multiple purposes, message will be dispatched if any one of the purpose matches
```
troyronda: *if any one of the purpose matches* @llorllale @sudeshrshetty @Moopli is this correct?

sudeshrshetty: Yes, if someone can register single handler for set of purposes he is interested in. Instead of registering one for each purpose which is too tiresome.

pkg/restapi/operation/common/models.go
sudeshrshetty: Will talk to Filip and create an issue for it.

pkg/restapi/operation/common/models.go
Moopli: I think I will introduce a priority mechanism, in which case we'll update this with priority ordering.

pkg/restapi/operation/common/models.go
Moopli: I'm not sure which is better - keeping vs removing. I think it's better to keep, since it means each protocol can use the same filtering mechanism.

Issue #1064, opened 2020-01-14
JS Worker Binding (with WASM)


troyronda:
* Enable JS application to control an Aries Agent running in the browser sandbox.
* Build will produce aries-worker.js and aries-worker.wasm. The WASM enables the Aries Framework and the JS is loadable in the page as a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). 
* The Web Worker (and WASM) enables:
  * API surface for the Controller pattern (similar to REST binding).
    * DID Exchange, Message Handler, ...
  * API surface for agent stores.
    * Connections (similar to REST binding).
    * VC.
  * API surface for DIDs and VC encoding and validation.
* Create a [Service Worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API)
  * Provides a template for building an Agent application.
  * Cache the agent resources (example: Web Worker and WASM).
  * Enables BDD tests (and a simple agent for development/testing purposes).
* Will need the ability to plug implementations and specify options. Examples:
  * Pluggable KMS and Crypto - to enable remote (when available).
  * Pluggable storage to enable remote storage (when available) and sync with local storage.
  * Should consider a built-in optional remote capability as an option (particularly if pluggable becomes difficult with WASMs, but could be a good idea anyways).

-----
Tasks

- [x] JS Worker Scaffolding #1069
- [x] Use "REST API" Operations (as generic channel/event interface) #1070

PR #1081, opened 2020-01-18
feat: initial scaffolding for JS web worker

llorllale:
Part of #1064 

* WASM callable with 'handleMsg'
* WASM posts results to 'handleResult'
* aries-worker.js: simple Web Worker wrapper around WASM
* aries.js: exposes user-facing APIs
* excluded `/gomocks` from the license check. Gomocks are auto-generated and are now being committed to the repo with the merge of #1068

The overall design is similar to that of the service events pattern: user-facing APIs submit a msg with an ID to the WASM and listen for a result referencing that same ID.

### DIFF
--- a/cmd/aries-js-worker/aries.js
+++ b/cmd/aries-js-worker/aries.js
@@ -1,0 +1,65 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// TODO not all browsers support private members of classes
+/* @class Aries provides Aries SSI-agent functions. */
+const Aries = new function() {
+    // TODO synchronize access on this map?
+    this._pending = new Map()
+    this._worker = new Worker('aries-worker.js')
+    this._worker.onmessage = e => {
+        const result = e.data
+        const cb = this._pending.get(result.id)
+        this._pending.delete(result.id)
+        cb(result)
+    }
+
+    /**
+     * Test methods.
+     * TODO - remove. Used for testing.
+     */
+    this._test = new function() {
+
+        /**
+         * Returns the input text prepended with "echo: ".
+         * TODO - remove.
+         * @param text
+         * @returns {Promise<unknown>}
+         * @private
+         */
+        super.this._echo = async function(text) {
+            return Aries._invoke("test", "echo", text, "timeout while accepting invitation")
+        }
+    }
+
+    this._newMsg = function(pkg, fn, payload) {
+        return {
+            // TODO there are several approaches to generate random strings:
+            // - which should we implement? do we need cryptographic-grade randomness for this?
+            // - alternatively, should the generator be provided by the client?
+            id: Math.random().toString(36).slice(2),
+            pkg: pkg,
+            fn: fn,
+            payload: payload
+        }
+    }
+
+    this._invoke = async function(pkg, fn, arg, msgTimeout) {
+        return new Promise((resolve, reject) => {
+            const timer = setTimeout(_ => reject(new Error(msgTimeout)), 5000)
+            const msg = Aries._newMsg(pkg, fn, arg)
+            Aries._pending.set(msg.id, result => {
+                clearTimeout(timer)
+                if (result.isErr) {
+                    reject(new Error(result.errMsg))
+                }
+                resolve(result.payload)
+            })
+            Aries._worker.postMessage(msg)
+        })
+    }
+}
+
--- a/cmd/aries-js-worker/index.html
+++ b/cmd/aries-js-worker/index.html
@@ -1,0 +1,36 @@
+<!--
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+ -->
+
+<!--
+This is a test page for developers to test WASM integration.
+-->
+
+<html>
+<head>
+    <meta charset="utf-8"/>
+    <script src="aries.js"></script>
+    <script>
+        async function handleInput() {
+            const invite = document.querySelector("#test-input").value
+            const connID = await Aries._test._echo(invite)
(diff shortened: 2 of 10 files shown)


cmd/aries-js-worker/aries.js
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// TODO not all browsers support private members of classes
+/* @class Aries provides Aries SSI-agent functions. */
+const Aries = new function() {
```
troyronda: In the npm package followup PR, we may need exports.

cmd/aries-js-worker/main.go
```diff
@@ -0,0 +1,138 @@
...
+	go pipe(input, output)
+
+	go sendTo(output)
+
+	js.Global().Set("handleMsg", js.FuncOf(takeFrom(input)))
+
+	select {}
+}
+
+func takeFrom(in chan *command) func(js.Value, []js.Value) interface{} {
```
troyronda: are there any tests possible?

cmd/aries-js-worker/aries.js
```diff
@@ -0,0 +1,52 @@
...
+    // TODO synchronize access on this map?
+    this._pending = new Map()
+    this._WORKER = new Worker('aries-worker.js')
```
troyronda: what does the all caps convention represent?

cmd/aries-js-worker/main.go
```diff
@@ -0,0 +1,138 @@
...
+	ID      string `json:"id"`
+	IsErr   bool   `json:"isErr"`
+	ErrMsg  string `json:"errMsg"`
+	Payload string `json:"payload"`
+}
+
+// All supported command handlers.
+func handlers() map[string]map[string]func(*command) *result {
+	return map[string]map[string]func(*command) *result{
+		"test": {
```
troyronda: I assume we will load handlers from the responsible package in a follow-up PR?

cmd/aries-js-worker/main.go
```diff
@@ -0,0 +1,138 @@
...
+	input := make(chan *command)
+	output := make(chan *result)
+
+	go pipe(input, output)
+
+	go sendTo(output)
+
+	js.Global().Set("handleMsg", js.FuncOf(takeFrom(input)))
+
+	select {}
```
troyronda: We may need to consider a stop signal in a future PR.

cmd/aries-js-worker/index.html
```diff
@@ -0,0 +1,32 @@
+<!--
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+ -->
+
+<html>
```
troyronda: I assume this will become a test page where we run headless browser tests against?

cmd/aries-js-worker/aries.js
llorllale: Sorry, I was not consistent. What is the proper convention?

cmd/aries-js-worker/main.go
llorllale: Will think about it more in the follow-up PR.

cmd/aries-js-worker/main.go
llorllale: Yes, we may.

cmd/aries-js-worker/index.html
llorllale: Yes, this is just a test page. I'll add a comment on it.

It is not being included in the build.

cmd/aries-js-worker/aries.js
```diff
@@ -0,0 +1,52 @@
...
+     * TODO - remove.
+     * Used for testing.
+     * @param text
+     * @returns {Promise<String>}
+     */
+    this.echo = function(text) {
+        return new Promise((resolve, reject) => {
+            setTimeout(_ => reject(new Error("timeout while accepting invitation")), 5000)
+            const msg = this._newMsg("test", "echo", text)
+            this._pending.set(msg.id, result => {
```
troyronda: this bit of code looks like it could be generic.

cmd/aries-js-worker/aries.js
```diff
@@ -0,0 +1,52 @@
...
+        cb(result)
+    }
+
+    /**
+     * TODO - remove.
+     * Used for testing.
+     * @param text
+     * @returns {Promise<String>}
+     */
+    this.echo = function(text) {
```
troyronda: to show the expected pattern, we would see this.test.echo.

llorllale: Ok - you really want that pattern? `Aries.<namespace>.<method>()`?

cmd/aries-js-worker/aries.js
troyronda: I think camel case would be fine. I think `_` prefix for private fields is a good convention (while browser support for true private fields is lacking).

cmd/aries-js-worker/aries.js
troyronda: Yes.

cmd/aries-js-worker/main.go
llorllale: Added integration tests with wasmbrowsertest.

cmd/aries-js-worker/aries.js
llorllale: done

cmd/aries-js-worker/aries.js
llorllale: done

cmd/aries-js-worker/aries.js
llorllale: Done.

However, autocomplete is not working for `_echo` in my IDE (Goland)....

PR #1091, opened 2020-01-20
feat: REST Binding - send new message and reply message

### LINKED ISSUE #1090: Messenger Service REST Binding : Send New Message & Send Reply Message

sudeshrshetty:
- added 2 endpoints for sending new messages and sending reply messages
- closes #1090

### DIFF
--- a/pkg/restapi/operation/common/models.go
+++ b/pkg/restapi/operation/common/models.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package common
 
 import (
+	"encoding/json"
+
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 )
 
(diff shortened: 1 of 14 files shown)


pkg/restapi/operation/common/models.go
```diff
@@ -102,3 +104,74 @@ type RegisteredServicesResponse struct {
...
+//	3. ServiceEndpoint (With recipient Keys, endpoint and optional routing keys)
+type SendNewMessageParams struct {
+
+	// Connection ID of the message destination
+	// This parameter takes precedence over all the other destination parameters.
+	ConnectionID string `json:"connection_ID,omitempty"`
+
+	// DID of the destination.
+	// This parameter takes precedence over `ServiceEndpoint` destination parameter.
+	TheirDID string `json:"their_did,omitempty"`
```
rolsonquadras: Is this public DID ?

pkg/restapi/operation/common/models.go
```diff
@@ -102,3 +104,74 @@ type RegisteredServicesResponse struct {
...
+
+	// ServiceEndpointDestination service endpoint destination.
+	// This param can be used to send messages outside connection.
```
rolsonquadras: If there is no connection/didexchange, does this protocol internally do didexchange and then send message ?

sudeshrshetty: No this is to send message outside existing connection,

pkg/restapi/operation/common/models.go
sudeshrshetty: TheirDID to lookup connection

pkg/restapi/operation/common/models.go
```diff
@@ -102,3 +104,74 @@ type RegisteredServicesResponse struct {
...
+// SendNewMessageRequest model
+//
+// This is used for operation to send new message
+//
+// swagger:parameters sendNewMessage
+type SendNewMessageRequest struct {
+	// Params for sending new message
+	//
+	// in: body
+	Params *SendNewMessageParams
```
troyronda: Perhaps you could explain this reason for nesting a `SendNewMessageParams` within a `SendNewMessageRequest`. Is this some swagger pattern?

pkg/restapi/operation/common/models.go
```diff
@@ -102,3 +104,74 @@ type RegisteredServicesResponse struct {
...
+
+// SendNewMessageParams contains parameters for sending new message
+// with three destination options
+//	1. ConnectionID
+//	2. TheirDID
+//	3. ServiceEndpoint (With recipient Keys, endpoint and optional routing keys)
+type SendNewMessageParams struct {
+
+	// Connection ID of the message destination
+	// This parameter takes precedence over all the other destination parameters.
```
troyronda: Can you add more comments detailing when parameters are optional. E.g., ConnectionID is not needed when ...

pkg/restapi/operation/common/models.go
```diff
@@ -102,3 +104,74 @@ type RegisteredServicesResponse struct {
...
+// SendReplyMessageRequest model
+//
+// This is used for operation to send reply to message
+//
+// swagger:parameters sendReplyMessage
+type SendReplyMessageRequest struct {
+	// Params for sending message reply
+	//
+	// in: body
+	Params *SendReplyMessageParams
```
troyronda: Same question as above (https://github.com/hyperledger/aries-framework-go/pull/1091/files#r369234172)

pkg/restapi/operation/common/models.go
sudeshrshetty: Since this is POST message and all params belongs in request body as a single JSON.
Combined them to single struct for 'in:body' annotation

sudeshrshetty: if we annote them all 'in:body' separately then, (validation fails)

`- operation "sendNewMessage" has more than 1 body param: ["connection_ID" "message_body" "service_endpoint" "their_did"]` 

we we make only `MessageBody` `in:body` then POST request will look like this,
`curl -X POST "http://localhost:8082/message/send-new-message?connection_ID=asdasd&their_did=asdas" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"msg\": \"msg text is here\"}"`
few params in URL and few in body will make it hard to maintain in future.

pkg/restapi/operation/common/models.go
sudeshrshetty: updated

pkg/restapi/operation/common/operation.go
```diff
@@ -37,10 +41,21 @@ const (
 	registerMsgService    = msgServiceOperationID + "/register-service"
 	unregisterMsgService  = msgServiceOperationID + "/unregister-service"
 	msgServiceList        = msgServiceOperationID + "/services"
+	sendNewMsg            = msgServiceOperationID + "/send-new-message"
```
troyronda: should we be more succinct? e.g., send and reply?

sudeshrshetty: send-message & reply-message?

sudeshrshetty: done

Issue #1093, opened 2020-01-21
Support "jws" Linked Data proof


kdimak:
Currently, we support the verification of Linked Data proofs which takes value to verify from `proofValue` element of the `proof` of JSON-LD document (e.g. Verifiable Credential). For example:
```
"proof": {
  "type": "Ed25519Signature2018",
  "created": "2020-01-21T16:44:53+02:00",
  "proofValue": "eyJhbGciOiJSUzI1NiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..kTCYt5XsITJX1CxPCT8yAV-TVIw5WEuts01mq-pQy7UJiN5mgREEMGlv50aqzpqh4Qq_PbChOMqsLfRoPsnsgxD-WUcX16dUOqV0G_zS245-kronKb78cPktb3rk-BuQy72IFLN25DYuNzVBAh4vGHSrQyHUGlcTwLtjPAnKb78"
}
```

Next, we need to support proof values kept in `jws` element, for example:
```
"proof": {
  "type": "Ed25519Signature2018",
  "created": "2020-01-20T16:11:36Z",
  "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..67TTULBvibJaJ2oZf3tGYhxZqxYS89qGQykL5hfCoh-MF0vrwQqzciZhjNrAGTAgHtDZsnSQVwJ8bO_7Sc0ECw",
  "proofPurpose": "assertionMethod"
}
```

The reference implementation could be https://github.com/digitalbazaar/jsonld-signatures

To be used by Verifiable Credentials.

PR #1094, opened 2020-01-21
feat: Linked Data proof of Verifiable Credential

### LINKED ISSUE #174: Linked Data proof of Verifiable Credential
Support [Linked Data Proof](https://w3c.github.io/vc-data-model/#linked-data-proofs) of Verifiable Credential.

kdimak:
closes #174

Support of Linked Data proofs with `Ed25519Signature2018 ` type and proof value inside `proofValue` element of the `proof`.
The follow-up PR to support JWS-based proof values is #1093 .

### DIFF
--- a/pkg/doc/signature/proof/utils.go
+++ b/pkg/doc/signature/proof/utils.go
@@ -20,9 +20,14 @@ func GetProofs(jsonLdObject map[string]interface{}) ([]*Proof, error) {
 		return nil, ErrProofNotFound
 	}
 
-	typedEntry, ok := entry.([]interface{})
-	if !ok {
-		return nil, errors.New("expecting []interface{}, got something else")
+	var typedEntry []interface{}
+	switch te := entry.(type) {
+	case []interface{}:
+		typedEntry = te
+	case map[string]interface{}:
+		typedEntry = []interface{}{te}
+	default:
+		return nil, errors.New("expecting []interface{} or map[string]interface{}, got something else")
 	}
 
 	var result []*Proof
@@ -51,11 +56,11 @@ func AddProof(jsonLdObject map[string]interface{}, proof *Proof) error {
 	entry, exists := jsonLdObject[jsonldProof]
 
 	if exists {
-		var ok bool
-
-		proofs, ok = entry.([]interface{})
-		if !ok {
-			return errors.New("expecting []interface{}, got something else")
+		switch p := entry.(type) {
+		case []interface{}:
+			proofs = p
+		default:
+			proofs = []interface{}{p}
 		}
 	}
 
--- a/pkg/doc/verifiable/credential_ldp.go
+++ b/pkg/doc/verifiable/credential_ldp.go
@@ -1,0 +1,22 @@
+package verifiable
+
+import (
+	"fmt"
+)
+
+// AddLinkedDataProof appends proof to the Verifiable Credential.
+func (vc *Credential) AddLinkedDataProof(context *LinkedDataProofContext) error {
+	vcBytes, err := vc.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("add linked data proof to VC: %w", err)
+	}
+
+	proofs, err := addLinkedDataProof(context, vcBytes)
+	if err != nil {
+		return err
+	}
+
+	vc.Proofs = proofs
+
+	return nil
+}
--- a/pkg/doc/verifiable/linked_data_proof.go
+++ b/pkg/doc/verifiable/linked_data_proof.go
@@ -1,0 +1,130 @@
+package verifiable
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/signer"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
+)
+
+// signatureSuite encapsulates signature suite methods required for signing documents
+type signatureSuite interface {
+
+	// GetCanonicalDocument will return normalized/canonical version of the document
+	GetCanonicalDocument(doc map[string]interface{}) ([]byte, error)
+
+	// GetDigest returns document digest
+	GetDigest(doc []byte) []byte
+
+	// Accept registers this signature suite with the given signature type
+	Accept(signatureType string) bool
+}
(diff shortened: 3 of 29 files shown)


pkg/doc/signature/proof/utils.go
```diff
@@ -20,9 +20,14 @@ func GetProofs(jsonLdObject map[string]interface{}) ([]*Proof, error) {
...
-	if !ok {
-		return nil, errors.New("expecting []interface{}, got something else")
+	var typedEntry []interface{}
+	switch te := entry.(type) {
+	case []interface{}:
+		typedEntry = te
+	case map[string]interface{}:
+		typedEntry = []interface{}{te}
+	default:
+		return nil, errors.New("expecting []interface{} or map[string]interface{}, got something else")
```
troyronda: can be more succinct :). don't need ", got something else".

pkg/doc/verifiable/linked_data_proof.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return pubKeyBytes, nil
+}
+
+// LinkedDataProofContext holds options needed to build a Linked Data Proof.
+// todo refactor, do not pass privateKey (https://github.com/hyperledger/aries-framework-go/issues/339)
+type LinkedDataProofContext struct {
+	SignatureType string               // required
+	Suite         signerSignatureSuite // required
+	PrivateKey    []byte               // required
```
sandrask: This should be addressed soon

pkg/doc/verifiable/credential_ldp.go
```diff
@@ -0,0 +1,22 @@
+package verifiable
+
+import (
+	"fmt"
+)
+
+// AddLinkedDataProof appends proof to the Verifiable Credential.
+func (vc *Credential) AddLinkedDataProof(context *LinkedDataProofContext) error {
```
troyronda: Is there a reason why the linked data proof is being added after the VC is instantiated?

pkg/store/verifiable/store_test.go
```diff
@@ -47,14 +47,6 @@ var udCredential = `
 
   "issuanceDate": "2010-01-01T19:23:24Z",
 
-  "proof": {
```
troyronda: are these proofs being removed from a test because they are no longer valid?

pkg/doc/verifiable/presentation_ldp.go
```diff
@@ -0,0 +1,22 @@
+package verifiable
+
+import (
+	"fmt"
+)
+
+// AddLinkedDataProof appends proof to the Verifiable Presentation.
+func (vp *Presentation) AddLinkedDataProof(context *LinkedDataProofContext) error {
```
troyronda: Similar question as earlier - is there a reason that linked data proofs are being added after instantiation?

pkg/doc/verifiable/presentation.go
```diff
@@ -189,7 +194,7 @@ func (vp *Presentation) Credentials() []interface{} {
 func (vp *Presentation) SetCredentials(creds ...interface{}) error {
```
troyronda: I didn't notice this method from an earlier PR - why do credentials need to be set after instantiation?

pkg/doc/verifiable/linked_data_proof.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return nil
+}
+
+type rawProof struct {
+	Proof json.RawMessage `json:"proof,omitempty"`
+}
+
+func addLinkedDataProof(context *LinkedDataProofContext, jsonldBytes []byte) ([]Proof, error) {
```
troyronda: seems like `addLinkedDataProof` actually creates a signature/proof AND adds it to the document?

pkg/doc/verifiable/linked_data_proof.go
troyronda: yes - we need these crypto operations & key holding to handled by Crypto/KMS.

pkg/doc/verifiable/presentation_ldp.go
sandrask: Yes, if you would like to sign them. This is convenience method for signing VC.

pkg/doc/verifiable/credential_ldp.go
troyronda: It seems like `AddLinkedDataProof ` actually means create signature and proof AND append the proof.

sandrask: Yes, that is what it means.

pkg/doc/signature/proof/utils.go
kdimak: Agree :) I took this pattern from @sandrask's  `signature` package - one of the cleanest and clearest packages i've seen 👍

pkg/doc/verifiable/credential_ldp.go
kdimak: Yes, that's right. For example, the Issuer wants to issue a new VC. It can be created directly by filling `verifiable.Credential` structure. To make it verifiable, she can use `Credential::AddLinkedDataProof()`. Finally, the credential is marshaled, e.g. to JSON or JWS (this adds another verification layer).

pkg/store/verifiable/store_test.go
kdimak: The test was failing... As we now check embedded proofs when unmarshalling `Credential` and thus we need to pass a way of how to get the public key for signature verification. Also, the removed proof was of "RsaSignature2018" type what is currently not supported in `signature` package (and thus in `verifiable` package as well).

pkg/doc/verifiable/presentation.go
kdimak: It's related to the scenario when a Holder wants to prepare a Verifiable Presentation to send to a Verifier. The Presentation can contain one or more Verifiable Credentials. They could be stored in the Holder's wallet. `SetCredentials()` allows attaching the Credentials to the Presentation.

pkg/doc/verifiable/linked_data_proof.go
kdimak: Correct. The function's name is incorrect, it should be smth like `getLinkedDataProofs()`. I will change this in the following PR related to #1093 .

PR #1096, opened 2020-01-21
feat: Route Service - Retrieve Router configs (endpoint and routingkeys)

### LINKED ISSUE #1095: Support to Fetch end point and routing keys from Route service
The DIDExchange services need to fetch end point and routing keys from [Route service](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/route/service.go).

rolsonquadras:
Closes #1095 

### DIFF
--- a/pkg/didcomm/protocol/route/service.go
+++ b/pkg/didcomm/protocol/route/service.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package route
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sync"
@@ -68,10 +69,16 @@ const (
 const (
 	// data key to store router connection ID
 	routeConnIDDataKey = "route-connID"
+
+	// data key to store router config
+	routeConfigDataKey = "route-config"
 )
 
 // ErrConnectionNotFound connection not found error
 var ErrConnectionNotFound = errors.New("connection not found")
+
+// ErrRouterNotRegistered router not registered error
+var ErrRouterNotRegistered = errors.New("router not registered")
 
 // provider contains dependencies for the Routing protocol and is typically created by using aries.Context()
 type provider interface {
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/route/api.go
```diff
@@ -0,0 +1,13 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package route
+
+// Config provides the router configuration.
```
sudeshrshetty: Do we need this interface?
As I can see route service returns config in one of the exported function, but it can return struct instead. 
Unless we try to inject this config from somewhere.

rolsonquadras: [Removed](https://github.com/hyperledger/aries-framework-go/compare/5234d05aa790ca34b1ede2f5f240e61d339640a2..94b9306c4274e5dbdd72571d2a46a7ff6dbf24e5#diff-829f87faf4556f4197c6a5292929cda5L10) the interface.

sudeshrshetty (approved):

pkg/didcomm/protocol/route/service.go
```diff
@@ -427,11 +441,24 @@ func (s *Service) AddKey(recKey string) error {
...
-	s.setRouteRegistrationCh(msgID, nil)
+	s.setKeyUpdateResponseCh(msgID, nil)
 
 	return nil
 }
 
+// Config fetches the router config - endpoint and routingKeys.
+func (s *Service) Config() (*Config, error) {
+	// check if router is already registered
+	_, err := s.getRouterConnectionID()
```
troyronda: I assume this case cannot be determined by the call to s.getRouterConfig()?

troyronda (approved):

pkg/didcomm/protocol/route/service.go
rolsonquadras: well, we can check if route configs are present or not. here, i feel better to check the source of truth for validation, since all the subsequent actions like AddKey, RemoveKey are dependent on this data.

PR #1098, opened 2020-01-22
feat: Messenger implementation (send message works)

soluchok:
* `ParseDIDCommMsgMap` - converts bytes to DIDCommMsgMap
* `NewDIDCommMsgMap` - converts struct to DIDCommMsgMap


pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -112,6 +112,31 @@ func (m DIDCommMsgMap) ID() string {
 	return res
 }
 
+// Encode converts struct to message
+func (m DIDCommMsgMap) Encode(v interface{}) error {
```
troyronda: This would make more sense as a type of constructor rather than a member function.

soluchok: Sure! It will. Also `NewDIDCommMsgMap` will be renamed to `ParseDIDCommMsg`

pkg/framework/context/context.go
```diff
@@ -130,14 +136,25 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
...
 
+			err = msg.Decode(h)
 			if err != nil {
 				return err
 			}
 
 			if svc.Accept(h) {
+				var opts []interface{}
+				// TODO: create an interface in some common package
+				if svcOpts, ok := svc.(interface{ MessengerOpts() []interface{} }); ok {
```
troyronda: The interface should be explicitly defined & named rather than inline.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Done!

pkg/framework/context/context.go
soluchok: Done!

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -112,6 +122,31 @@ func (m DIDCommMsgMap) ID() string {
...
```
troyronda: Why does this function need to be exported?

soluchok: Done!

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -48,6 +49,11 @@ func NewDIDCommMsgMap(payload []byte) (DIDCommMsgMap, error) {
 	return msg, nil
 }
 
+// NewDIDCommMsgMap returns DIDCommMsg with Header
```
troyronda: The argument to this function should be described in the function comment.

pkg/didcomm/common/service/messenger.go
```diff
@@ -28,12 +28,25 @@ type Messenger interface {
...
+// MessengerHandler includes Messenger interface and Handle function to handle inbound messages
+type MessengerHandler interface {
+	Messenger
+	// HandleInbound handles all inbound messages
+	HandleInbound(msg DIDCommMsgMap, myDID, theirDID string, opts ...interface{}) error
+}
+
+// MessengerOptions allows modifying Messenger behavior
+// NOTE: if service implements it, it will work automatically
+type MessengerOptions interface {
```
troyronda: I'm not sure what this interface does?

pkg/didcomm/messenger/messenger.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// HandleInbound handles all inbound messages
+// function behavior is flexible and can be modified by providing options
+func (*Messenger) HandleInbound(msg service.DIDCommMsgMap, myDID, theirDID string, opts ...interface{}) error {
+	return nil
+}
+
+// Send sends the message by starting a new thread.
+func (m *Messenger) Send(msg service.DIDCommMsgMap, myDID, theirDID string) error {
```
troyronda: I'm confused on why the msg argument in these methods is not of the interface type `DIDCommMsg`.

I also see that calls deeper within SendToDID still use json.Marsal instead of DIDCommMsg.Encode.

pkg/didcomm/common/service/messenger.go
```diff
@@ -28,12 +28,25 @@ type Messenger interface {
 	// ReplyTo replies to the message by given msgID.
 	// Keeps threadID in the *decorator.Thread.
 	// Using this function means that communication will be on the same thread.
-	ReplyTo(msgID string, msg DIDCommMsgMap)
+	ReplyTo(msgID string, msg DIDCommMsgMap) error
 
 	// Send sends the message by starting a new thread.
-	Send(msg DIDCommMsgMap, myDID, theirDID string)
+	Send(msg DIDCommMsgMap, myDID, theirDID string) error
```
troyronda: I'm confused on why the msg argument in these methods is not of the interface type DIDCommMsg.

test/bdd/pkg/messaging/messaging_sdk_steps.go
```diff
@@ -90,16 +92,16 @@ func (d *SDKSteps) sendMessage(fromAgentID, msg, msgType, purpose, toAgentID str
 		}
 	}
 
-	message := &genericInviteMsg{
+	msgMap := service.NewDIDCommMsgMap(&genericInviteMsg{
 		ID:      uuid.New().String(),
```
troyronda: I thought details like `ID` shouldn't need to be handled by higher layer code.

pkg/framework/context/context.go
```diff
@@ -110,19 +116,33 @@ func (p *Provider) InboundTransportEndpoint() string {
 	return p.inboundTransportEndpoint
 }
 
+func (p *Provider) tryToHandle(svc service.InboundHandler, msg service.DIDCommMsgMap, myDID, theirDID string) error {
+	var opts []interface{}
+	if svcOpts, ok := svc.(service.MessengerOptions); ok {
+		opts = svcOpts.MessengerOptions()
```
troyronda: I only see MessengerOptions being used by `HandleInbound`. It's usage is not so clear since I don't see the HandleInbound code.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -134,3 +140,78 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
 
 	return decoder.Decode(m)
 }
+
+func toMap(v interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+
+	rv := reflect.Indirect(reflect.ValueOf(v))
+	for rfv, field := range mapValueStructField(rv) {
+		// TODO: should be snake_case
```
troyronda: ?

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -134,3 +140,78 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
 
 	return decoder.Decode(m)
 }
+
+func toMap(v interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+
+	rv := reflect.Indirect(reflect.ValueOf(v))
```
troyronda: some comments within these functions could be helpful, particularly for those who do not use reflect much.

pkg/didcomm/common/service/messenger.go
sudeshrshetty: @troyronda , may be because, DIDCommMsg doesn't exactly looks like interface for DID Communication message type, 
eg: presence of `Decode(v interface{}) error` in  DIDCommMsg interface forces sender to implement Decode function

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -134,3 +140,78 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
...
+	for rfv, field := range mapValueStructField(rv) {
+		// TODO: should be snake_case
+		name := strings.ToLower(field.Name)
+
+		tags := strings.Split(field.Tag.Get(`json`), ",")
+		if tags[0] != "" {
```
troyronda: does this happen when field.Tag.Get, itself returns an empty string?

pkg/didcomm/common/service/messenger.go
sudeshrshetty: just curious: any option to send without connection(without myDID and theirDID), 
eg: `Destination` in outbound dispatcher interface?

pkg/didcomm/messenger/messenger.go
```diff
@@ -0,0 +1,52 @@
...
+// NewMessenger returns a new instance of the Messenger
+func NewMessenger(ctx Provider) *Messenger {
+	return &Messenger{
+		dispatcher: ctx.OutboundDispatcher(),
+	}
+}
+
+// HandleInbound handles all inbound messages
+// function behavior is flexible and can be modified by providing options
+func (*Messenger) HandleInbound(msg service.DIDCommMsgMap, myDID, theirDID string, opts ...interface{}) error {
```
sudeshrshetty: Why we need this?

pkg/didcomm/messenger/messenger.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Send sends the message by starting a new thread.
+func (m *Messenger) Send(msg service.DIDCommMsgMap, myDID, theirDID string) error {
+	return m.dispatcher.SendToDID(msg, myDID, theirDID)
+}
+
+// ReplyTo replies to the message by given msgID.
+func (*Messenger) ReplyTo(msgID string, msg service.DIDCommMsgMap) error {
+	return nil
```
sudeshrshetty: can we throw error if not implemented, to avoid accidental call to this function.
Since there is not send receipt here, sender will end up thinking he sent a message successfully.

pkg/didcomm/messenger/messenger.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ReplyToNested sends the message by starting a new thread.
+func (*Messenger) ReplyToNested(threadID string, msg service.DIDCommMsgMap, myDID, theirDID string) error {
```
sudeshrshetty: same as above comment

pkg/framework/aries/api/protocol.go
```diff
@@ -23,6 +24,7 @@ var ErrSvcNotFound = errors.New("service not found")
 // Provider interface for protocol ctx
 type Provider interface {
 	OutboundDispatcher() dispatcher.Outbound
```
sudeshrshetty: Do we still need `OutboundDispatcher in context`? can we deprecate it if not needed.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -134,3 +140,78 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
...
+
+		if sf.Tag.Get(`json`) == "-" {
+			continue
+		}
+
+		if sf.PkgPath != "" {
+			continue
+		}
+
+		if sf.Anonymous {
```
troyronda: embedded struct

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -134,3 +140,78 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
...
+	rt := value.Type()
+
+	for i := 0; i < rt.NumField(); i++ {
+		rv, sf := value.Field(i), rt.Field(i)
+
+		if sf.Tag.Get(`json`) == "-" {
+			continue
+		}
+
+		if sf.PkgPath != "" {
```
troyronda: unexported field

pkg/framework/aries/framework.go
```diff
@@ -373,6 +382,19 @@ func createOutboundDispatcher(frameworkOpts *Aries) error {
 	return nil
 }
 
+func createMessenger(frameworkOpts *Aries) error {
+	ctx, err := context.New(
```
sudeshrshetty: Since messenger is something which is built on top of OutboundDispatcher, both shouldn't co-exist in context.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -134,3 +140,78 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
...
+}
+
+func mapValueStructField(value reflect.Value) map[interface{}]reflect.StructField {
+	fields := make(map[interface{}]reflect.StructField)
+	rt := value.Type()
+
+	for i := 0; i < rt.NumField(); i++ {
+		rv, sf := value.Field(i), rt.Field(i)
+
+		if sf.Tag.Get(`json`) == "-" {
```
troyronda: `json:"-"` means ignore.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Changed to ` NewDIDCommMsgMap converts structure(model) to DIDCommMsgMap`

test/bdd/pkg/messaging/messaging_sdk_steps.go
soluchok: Sure, the messenger will add it automatically in case if it absent. But, it will be done in my next PR.

pkg/framework/context/context.go
soluchok: ok, I moved these options to the `HandleInbound` let me know if it`s ok.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: for now, the first one is enough, but I think we will handle `omitempty` as well

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Done!

pkg/didcomm/common/service/did_comm_msg.go
soluchok: added more comments

pkg/didcomm/common/service/did_comm_msg.go
soluchok: I've just added omitempty as well

pkg/didcomm/common/service/messenger.go
soluchok: if for some reason you do want to turn messenger logic off or change the behavior of how it works you can do this by implementing this interface in your service. Messenger will check whether the service implements it and will act according to that. Possible options will be implemented in the messenger. For now, we do not have any options. Implementing this interface in the service is optional.  It intends to provide a possibility to customize the messenger behavior in case you are advanced user :)

pkg/didcomm/messenger/messenger.go
soluchok: We do not need this interface here. The interface should be provided when you are using only read operation. Messenger aims to modify the message. If we will use interface we need to do type assertion under the hood. Which is bad in case if someone decided to provide a different type of message (custom implementation of did comm message interface).

We will use `unmarshal` for inbound messages and `marshal` for outbound messages (e.g see `ParseDIDCommMsgMap` function). I can create Encode function which will execute  `marshal` under the hood to convert message payload to bytes. But it will break current logic, it requires to change OutboundDispatcher send logic which is not simple.

pkg/didcomm/common/service/messenger.go
soluchok: The interface should be provided only when someone wants to do `read` operations. There is no need to implement another did message interface and should not be. Messages across the framework should be `type DIDCommMsgMap map[string]interface{}` which covers any type of possible message.

soluchok: unfortunately, there is no such possibility. Not sure even how to implement it. I do not even know whether we have a destination from where this message came (e.g need for a reply). But I guess, we will come up with something.

pkg/didcomm/messenger/messenger.go
soluchok: it receives all incoming messages to increase order_received value needed for the thread decorator.

pkg/didcomm/messenger/messenger.go

pkg/framework/aries/api/protocol.go
soluchok: It will be deprecated as soon as the messenger will be done!

pkg/framework/aries/framework.go
soluchok: We need both at least for now. As soon as everything in the framework will be replaced with the messenger usage it can be removed.

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -48,6 +49,11 @@ func NewDIDCommMsgMap(payload []byte) (DIDCommMsgMap, error) {
 	return msg, nil
 }
 
+// NewDIDCommMsgMap converts structure(model) to DIDCommMsgMap
+func NewDIDCommMsgMap(v interface{}) DIDCommMsgMap {
+	return toMap(v)
```
llorllale: Can't use mapstructure here as well?

soluchok: I tried it, but unfortunately, it doesn't work as I expected!

pkg/didcomm/common/service/messenger.go
llorllale: So is the intent to delete `service.DIDCommMsg` and have consumers define their dependency?

pkg/didcomm/common/service/messenger.go
llorllale: can we defer this until needed?

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -151,22 +150,17 @@ func (s *arranging) ExecuteInbound(dis dispatcher.Outbound, m *metaData) (state,
 		recipient = m.Recipients[1]
 	}
 
-	return &noOp{}, dis.SendToDID(&Proposal{
+	return &noOp{}, messenger.Send(service.NewDIDCommMsgMap(Proposal{
 		Type:   ProposalMsgType,
 		ID:     uuid.New().String(),
 		To:     recipient.To,
 		Thread: &decorator.Thread{ID: m.ThreadID},
```
llorllale: I would expect the messenger to take care of threading, yes?

pkg/didcomm/common/service/messenger.go
soluchok: no, my assumption is that everything should be a map interface. But it can be easily changed. if you really need to define a custom implementation, then service.DIDCommMsg should provide one more method e.g ToMapInterface() and it will be used at least for the messenger.

soluchok: but my intention was to create a one generic message type

pkg/framework/context/context.go
```diff
@@ -110,19 +116,28 @@ func (p *Provider) InboundTransportEndpoint() string {
 	return p.inboundTransportEndpoint
 }
 
+func (p *Provider) tryToHandle(svc service.InboundHandler, msg service.DIDCommMsgMap, myDID, theirDID string) error {
```
llorllale: is this method temporary until the inbound handler is substituted by the messenger?

pkg/didcomm/protocol/introduce/states.go
soluchok: It will, but it does not work yet. Also, the protocol will use ReplyTo instead of Send. I did it just to check that my custom function which converts struct to map works as expected.

pkg/didcomm/common/service/did_comm_msg.go
llorllale: can you put a note in a comment in there?

pkg/framework/context/context.go
soluchok: actually I did not think about it, I thought it is enough. @troyronda What do you think?

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Done! 
```
	// NOTE: do not try to replace it with mapstructure pkg
	// it doesn't work as expected, at least time.Time won't be converted
```

pkg/didcomm/common/service/messenger.go
soluchok: sure, I can remove this

pkg/framework/aries/api/protocol.go
rolsonquadras: Do we have a TODO/Issue for this ? This looks to be a major refactor has we have been using Outbound dispatcher throughout the project.

pkg/framework/aries/framework.go
```diff
@@ -373,6 +382,19 @@ func createOutboundDispatcher(frameworkOpts *Aries) error {
...
+		context.WithOutboundDispatcher(frameworkOpts.outboundDispatcher),
+	)
+	if err != nil {
+		return fmt.Errorf("context creation failed: %w", err)
+	}
+
+	frameworkOpts.messenger = messenger.NewMessenger(ctx)
```
rolsonquadras: Shouldn't Messenger be configurable ? ie, Pass it as an option while creating the framework and framework provides just a default option.

pkg/framework/context/context.go
```diff
@@ -110,19 +116,28 @@ func (p *Provider) InboundTransportEndpoint() string {
...
+		msg, err := service.ParseDIDCommMsgMap(message)
 		if err != nil {
 			return err
 		}
 
 		// find the service which accepts the message type
 		for _, svc := range p.services {
 			if svc.Accept(msg.Type()) {
-				_, err = svc.HandleInbound(msg, myDID, theirDID)
-				return err
```
troyronda: why is messenger re-processing the msg in each iteration of the loop?

pkg/framework/context/context.go
troyronda: see comment about loop.

pkg/framework/aries/framework.go
troyronda: agree - we should always have configurable components in framework.

note: the whole inbound function that is currently in context.go should be configurable...

pkg/framework/context/context.go
soluchok: messenger does not re-processing it, there is the return statement.

pkg/didcomm/common/service/messenger.go
troyronda: agree - remove until needed.

pkg/framework/context/context.go
troyronda: Ah. Right.

Although logically, I think it is more clear (to me) that messenger is a pre-processor prior to calling services. If we start doing service chains (calling more than one service), I assume messenger should only be called once.

pkg/didcomm/common/service/messenger.go
llorllale: @soluchok can you provide an example of how this msg map will be modified? And by whom? I think I understand that the `Messenger` will be modifying it for low-level didcomm stuff, but I'm curious to know if the higher-level protocols will also be modifying it? Can a protocol add a hypothetical `~payment` decorator?

llorllale: It is required for supporting the `~service` decorator, which is in AIP 1.0 (although there are [lingering concerns](https://github.com/hyperledger/aries-rfcs/issues/358#issuecomment-575873239)).

pkg/didcomm/messenger/messenger.go
llorllale: @troyronda @soluchok is this resolved?

pkg/framework/context/context.go
llorllale: @soluchok I believe this function is *temporarily* handling the msg twice, right? The end-goal is to have only the messenger handle the msg, correct?

pkg/framework/context/context.go
llorllale: I thought the messenger would be the new inbound dispatcher?

pkg/didcomm/common/service/messenger.go
soluchok: Done!

pkg/framework/context/context.go
soluchok: @troyronda  `If we start doing service chains` - it looks like middleware. If we implement it, the messenger should be first in the chain. But I don't know how this implementation will look like. The messenger signature function for handle inbound might be the same as a service handle inbound function. It means that messenger can play a role as a service (and should be first in the chain).

pkg/framework/context/context.go
soluchok: It does not handle the message twice.

pkg/framework/aries/framework.go
soluchok: @troyronda @rolsonquadras  I have added a possibility to pass it as an option. But I can not remove it. We are using OutboundDispatcher for the messenger. Because of that our BBD tests are failing!

rolsonquadras: @soluchok In the new code, the default messenger overrides the passed in option all the time. We can provide the default implementation in https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/aries/default.go#L74.

pkg/framework/aries/api/protocol.go
rolsonquadras: @soluchok fyi - may be you missed this comment.

pkg/framework/aries/framework.go
soluchok: It is good to know, but it does not help. I have added an additional check for that. And the unit test as well. Now you can pass it as an option. https://github.com/hyperledger/aries-framework-go/pull/1098/files#diff-7b2448fdc2181299533bc944239b9744R379-R382

pkg/framework/context/context.go
llorllale: Yes it is.... the message is being handled by the messenger and then by the service.

pkg/framework/aries/framework.go
rolsonquadras: Its created through creator pattern - https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/aries/default.go#L56. just a FYI.

soluchok: The problem is that an outbound dispatcher is needed for the messenger. But we create outbound dispatcher at the end.

rolsonquadras: Yes, we have similar cases in default package - https://github.com/hyperledger/aries-framework-go/blob/master/pkg/framework/aries/default.go.

pkg/framework/aries/framework.go
```diff
@@ -128,6 +137,14 @@ func initializeServices(frameworkOpts *Aries) (*Aries, error) {
 	return frameworkOpts, nil
 }
 
+// WithMessengerHandler injects a messenger handler to the Aries framework
+func WithMessengerHandler(mh service.MessengerHandler) Option {
```
troyronda: I think Handler is not the right term for this.

Issue #1076, opened 2020-01-17
Register agent with multiple routers


rolsonquadras:
Based on review comments 
- Support multiple routers - https://github.com/hyperledger/aries-framework-go/pull/1071#discussion_r367939444.
- Option to chose the router - https://github.com/hyperledger/aries-framework-go/pull/1103#discussion_r370314046

PR #1103, opened 2020-01-23
feat: Integrate Route Service with DIDExchange Request message

### LINKED ISSUE #1100: DIDExchange Request - Add recKey to the Router
Add recKeys generated as part of new DIDDoc(did-communication service) to the router (#1073).

rolsonquadras:
Closes #1100 

### DIFF
--- a/pkg/didcomm/protocol/didexchange/states.go
+++ b/pkg/didcomm/protocol/didexchange/states.go
@@ -22,8 +22,10 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/ed25519signature2018"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
 	connectionstore "github.com/hyperledger/aries-framework-go/pkg/store/connection"
 )
 
@@ -416,10 +418,31 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
 
 	logger.Debugf("creating new '%s' did for connection", didMethod)
 
+	// get the route configs (pass empty service endpoint, as default servie endpoint added in VDRI)
+	serviceEndpoint, routingKeys, err := route.GetRouterConfig(ctx.routeSvc, "")
+	if err != nil {
+		return nil, nil, fmt.Errorf("did doc - fetch router config : %w", err)
+	}
+
 	// by default use peer did
-	newDidDoc, err := ctx.vdriRegistry.Create(didMethod)
+	newDidDoc, err := ctx.vdriRegistry.Create(
+		didMethod,
+		vdri.WithServiceEndpoint(serviceEndpoint),
+		vdri.WithRoutingKeys(routingKeys),
+	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("create %s did: %w", didMethod, err)
+	}
+
+	svc, found := did.LookupService(newDidDoc, didCommServiceType)
+	if found {
+		for _, recKey := range svc.RecipientKeys {
+			// TODO https://github.com/hyperledger/aries-framework-go/issues/1105 Support to Add multiple
+			//  recKeys to the Router
+			if err = route.AddKeyToRouter(ctx.routeSvc, recKey); err != nil {
+				return nil, nil, fmt.Errorf("did doc - add key to the router : %w", err)
+			}
+		}
 	}
 
 	err = ctx.connectionStore.SaveDIDFromDoc(newDidDoc)
--- a/pkg/framework/aries/api/vdri/vdri.go
+++ b/pkg/framework/aries/api/vdri/vdri.go
@@ -91,6 +91,7 @@ type CreateDIDOpts struct {
 	ServiceType     string
 	KeyType         string
 	ServiceEndpoint string
+	RoutingKeys     []string
 	RequestBuilder  func([]byte) (io.Reader, error)
 }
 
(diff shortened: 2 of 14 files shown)


pkg/framework/aries/api/vdri/vdri.go
```diff
@@ -118,6 +119,13 @@ func WithServiceEndpoint(serviceEndpoint string) DocOpts {
 	}
 }
 
+// WithRoutingKeys allows for setting routing keys.
+func WithRoutingKeys(routingKeys []string) DocOpts {
```
troyronda: This will need to be revisited for multiple router support. Also there is the option of not using a router (even when routers are configured).

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -416,12 +418,30 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
...
+	newDidDoc, err := ctx.vdriRegistry.Create(
+		didMethod,
+		vdri.WithServiceEndpoint(serviceEndpoint),
+		vdri.WithRoutingKeys(routingKeys),
+	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("create %s did: %w", didMethod, err)
 	}
 
+	for _, svc := range newDidDoc.Service {
```
llorllale: Must only select the service with type `did-communication`

llorllale: Use `did.LookUpService`

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -416,12 +418,30 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
...
+		for _, recKey := range svc.RecipientKeys {
+			if err = route.AddKeyToRouter(ctx.routeSvc, recKey); err != nil {
```
llorllale: 1. The [example](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md#keylist-update) `keylist_update` message in the RFC suggests you can register several keys in a single batch
2. Need a TODO: it's possible that one or several of those recipient keys can be registered in another router.

pkg/framework/aries/api/vdri/vdri.go
llorllale: The use case of not using a router is fulfilled by not passing this option when creating the DID.

pkg/framework/aries/default.go
```diff
@@ -46,8 +46,9 @@ func defFrameworkOpts(frameworkOpts *Aries) error {
 		frameworkOpts.storeProvider = storeProv
 	}
 
+	// order is important as DIDExchange service depends on Route service and Introduce depends on DIDExchange
 	frameworkOpts.protocolSvcCreators = append(frameworkOpts.protocolSvcCreators,
-		newExchangeSvc(), newIntroduceSvc(), newRouteSvc())
+		newRouteSvc(), newExchangeSvc(), newIntroduceSvc())
```
llorllale: any reason for the reordering here?

pkg/framework/aries/api/vdri/vdri.go
rolsonquadras: Added this to the list #1076 .

pkg/didcomm/protocol/didexchange/states.go
rolsonquadras: Thanks for pointing it out (I wasn't aware of function). [Updated](https://github.com/hyperledger/aries-framework-go/compare/70e04d120f1905e8fee04847f9fa56eaa14711e8..488523d9b32e5efb07b85bcc1dd8b3965fc9a9e1#diff-86ab318ebdf86972bc415c0a51a2f1ceR437) the code.

pkg/didcomm/protocol/didexchange/states.go
rolsonquadras: Yes, missed to create a TODO. Created a new issue and [added](https://github.com/hyperledger/aries-framework-go/compare/70e04d120f1905e8fee04847f9fa56eaa14711e8..488523d9b32e5efb07b85bcc1dd8b3965fc9a9e1#diff-7cc3df019152fe2c557826a312cc0a12R398) a TODO.

pkg/framework/aries/default.go
rolsonquadras: Order is important as DIDExchange service depends on Route service and Introduce depends on DIDExchange. Already have a [comment](https://github.com/hyperledger/aries-framework-go/pull/1103/files#diff-da8e5280b0328cb616f0b42390a35b00R49) for this.

llorllale: Ouch.

llorllale: This needs to be fixed. I understand there are "foundational" protocols like did-exchange and maybe routing, but protocols *must* be composable in general. These protocol services shouldn't be coupled like that.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -416,12 +418,33 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
 
 	logger.Debugf("creating new '%s' did for connection", didMethod)
 
+	// get the route configs (pass empty service endpoint, as default servie endpoint added in VDRI)
+	serviceEndpoint, routingKeys, err := route.GetRouterConfig(ctx.routeSvc, "")
```
llorllale: Based on the conversation we had [here](https://github.com/hyperledger/aries-framework-go/pull/1103/files#r370323241), I strongly suggest we stop coupling protocol services together, starting with this. All didexchange requires is a `NewDID()` function - nothing more. It doesn't care about routing, it just cares about sending *a* DID document.

pkg/framework/aries/default.go
rolsonquadras: We already have an issue for fixing the framework and context relationship https://github.com/hyperledger/aries-framework-go/issues/212. we can handle this issue as part of the larger cleanup.

pkg/didcomm/protocol/didexchange/states.go
rolsonquadras: Is this a blocker for this merge or can we handle it later (a TODO) ?

llorllale: an urgent TODO I guess

rolsonquadras: Created an issue - https://github.com/hyperledger/aries-framework-go/issues/1107.

Issue #1106, opened 2020-01-23
Implement RFC-0335 (HTTP over DIDComm) as new message service


sudeshrshetty:
Implement [RFC-0335 ](https://github.com/hyperledger/aries-rfcs/blob/master/features/0335-http-over-didcomm/README.md)as new message service

Issue #1075, opened 2020-01-17
Remove deprecated Header structure


soluchok:
Header structure is deprecated and should be removed

PR #1108, opened 2020-01-23
feat: msgservice acceptance criteria to accept only type and purpose

sudeshrshetty:
- Accept() of message services to accept only type and purpose
- rearranged packages for message service and handlers
- pushed initial skeleton code for HTTP over DID Comm
- Part of #1106


pkg/framework/context/context.go
```diff
@@ -129,14 +129,16 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
 		// in case of no services are registered for given message type,
 		// find generic inbound services registered for given message header
 		for _, svc := range p.msgSvcProvider.Services() {
-			h := &service.Header{}
```
troyronda: you should remember to delete purpose from header.
https://github.com/hyperledger/aries-framework-go/blob/4de0955d13d7123da31c2f2386f4d3fb25f3106b/pkg/didcomm/common/service/did_comm_msg.go#L35

pkg/didcomm/messaging/service/httpmsgsvc.go
```diff
@@ -0,0 +1,53 @@
+/*
+ *
+ * Copyright SecureKey Technologies Inc. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * /
+ *
+ */
+
+package service
```
troyronda: http message service could be its own package.

pkg/framework/context/context.go
sudeshrshetty: whole `service.header` going to be removed, I removed it's reference from context and message service in this PR.

pkg/didcomm/messaging/service/httpmsgsvc.go
sudeshrshetty: fixed, 
tried to avoid stutter `service.HTTPOverDIDComm` vs 'http.OverDIDComm'

pkg/framework/context/context.go
troyronda: Is the next PR to remove the whole header?

sudeshrshetty: Removed `Purpose` now and either me or @soluchok will work on #1075 to remove deprecated header structure.

PR #1118, opened 2020-01-27
feat: initial scaffolding for NodeJS

llorllale:
Initial scaffolding for NodeJS support.

### DIFF
--- a/cmd/aries-js-worker/app.js
+++ b/cmd/aries-js-worker/app.js
@@ -1,0 +1,21 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// This is a test nodejs app for developers to test WASM integration.
+
+const { Aries } = require('./aries.js')
+
+// TODO add a 'ready' signal from the wasm to let users know when it's loaded
+
+setTimeout(() => {
+    Aries._test._echo("test").then(
+        response => console.log(response),
+        err => console.error(err)
+    )
+}, 1000)
+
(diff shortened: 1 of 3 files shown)


cmd/aries-js-worker/app.js
```diff
@@ -0,0 +1,21 @@
+/*
```
troyronda: should we make the filename more explicit about being a node app?

llorllale: isn't "app" the convention?

cmd/aries-js-worker/aries.js
```diff
@@ -4,18 +4,15 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
+'use strict'
+
+const _isNodeJS = typeof module !== 'undefined' && module.exports
```
troyronda: is this variable name global to the containing page?

cmd/aries-js-worker/app.js
llorllale: https://nodejs.org/en/docs/guides/getting-started-guide/

cmd/aries-js-worker/aries.js
llorllale: done

cmd/aries-js-worker/app.js
troyronda: We are showing both browser and node.js samples aren't we?

troyronda: we should revisit once we have npm packaging. perhaps two subdirectories for the samples would make sense.

PR #1137, opened 2020-01-29
feature: Support "jws" Linked Data proof

### LINKED ISSUE #1093: Support "jws" Linked Data proof

kdimak:
closes #1093

Based on https://github.com/digitalbazaar/jsonld-signatures

### DIFF
--- a/pkg/doc/signature/ed25519signature2018/suite.go
+++ b/pkg/doc/signature/ed25519signature2018/suite.go
@@ -13,7 +13,7 @@ package ed25519signature2018
 
 import (
 	"crypto/ed25519"
-	"crypto/sha512"
+	"crypto/sha256"
 	"errors"
 
 	"github.com/piprate/json-gold/ld"
@@ -74,7 +74,7 @@ func (s *SignatureSuite) GetCanonicalDocument(doc map[string]interface{}) ([]byt
 
 // GetDigest returns document digest
 func (s *SignatureSuite) GetDigest(doc []byte) []byte {
-	digest := sha512.Sum512(doc)
+	digest := sha256.Sum256(doc)
 	return digest[:]
 }
 
--- a/pkg/doc/signature/proof/data.go
+++ b/pkg/doc/signature/proof/data.go
@@ -5,7 +5,10 @@ SPDX-License-Identifier: Apache-2.0
 
 package proof
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+)
 
 const jsonldContext = "@context"
 
@@ -17,6 +20,32 @@ type signatureSuite interface {
 
 	// GetDigest returns document digest
 	GetDigest(doc []byte) []byte
+}
+
+// SignatureRepresentation defines a representation of signature value.
+type SignatureRepresentation int
+
+const (
+	// SignatureProofValue uses "proofValue" field in a Proof to put/read a digital signature.
+	SignatureProofValue SignatureRepresentation = iota
+
+	// SignatureJWS uses "jws" field in a Proof as an element for representation of detached JSON Web Signatures.
+	SignatureJWS
+)
+
+// CreateVerifyData creates data that is used to generate or verify a digital signature.
+// It depends on the signature value holder type.
+// In case of "proofValue", the standard Create Verify Hash algorithm is used.
+// In case of "jws", verify data is built as JSON Web Signature (JWS) with detached payload.
+func CreateVerifyData(suite signatureSuite, jsonldDoc map[string]interface{}, proof *Proof) ([]byte, error) {
+	switch proof.SignatureRepresentation {
+	case SignatureProofValue:
+		return CreateVerifyHash(suite, jsonldDoc, proof.JSONLdObject())
+	case SignatureJWS:
+		return createVerifyJWS(suite, jsonldDoc, proof)
+	}
+
+	return nil, fmt.Errorf("unsupported signature representation: %v", proof.SignatureRepresentation)
 }
 
 // CreateVerifyHash returns data that is used to generate or verify a digital signature
@@ -48,18 +77,13 @@ func CreateVerifyHash(suite signatureSuite, jsonldDoc, proofOptions map[string]i
 }
 
 func prepareCanonicalProofOptions(suite signatureSuite, proofOptions map[string]interface{}) ([]byte, error) {
-	value, ok := proofOptions[jsonldCreator]
-	if !ok || value == nil {
-		return nil, errors.New("creator is missing")
-	}
-
-	value, ok = proofOptions[jsonldCreated]
+	value, ok := proofOptions[jsonldCreated]
 	if !ok || value == nil {
 		return nil, errors.New("created is missing")
 	}
 
 	// copy from the original proof options map without specific keys
-	proofOptionsCopy := make(map[string]interface{})
+	proofOptionsCopy := make(map[string]interface{}, len(proofOptions))
 
 	for key, value := range proofOptions {
 		if excludedKeyFromString(key) == 0 {
@@ -86,14 +110,21 @@ const (
 	proofType excludedKey = iota + 1
 	proofID
 	proofValue
+	jws
+)
+
+//nolint:gochecknoglobals
+var (
+	excludedKeysStr = [...]string{"type", "id", "proofValue", "jws"}
+	excludedKeys    = [...]excludedKey{proofType, proofID, proofValue, jws}
 )
 
 func (ek excludedKey) String() string {
-	return [...]string{"type", "id", "proofValue"}[ek-1]
+	return excludedKeysStr[ek-1]
 }
 
 func excludedKeyFromString(s string) excludedKey {
-	for _, ek := range [...]excludedKey{proofType, proofID, proofValue} {
+	for _, ek := range excludedKeys {
 		if ek.String() == s {
 			return ek
 		}
--- a/pkg/doc/signature/proof/jws.go
+++ b/pkg/doc/signature/proof/jws.go
@@ -1,0 +1,195 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package proof
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"strings"
+
+	"github.com/piprate/json-gold/ld"
+)
+
+const securityContext = "https://w3id.org/security/v2"
+
+const (
+	jwtPartsNumber   = 3
+	jwtHeaderPart    = 0
+	jwtSignaturePart = 2
+)
+
+// CreateDetachedJWTHeader creates detached JWT header.
+func CreateDetachedJWTHeader(p *Proof) string {
+	jwtHeaderMap := map[string]interface{}{
+		"alg":  p.Type,
+		"b64":  false,
+		"crit": []string{"b64"},
+	}
+
+	jwtHeaderBytes, err := json.Marshal(jwtHeaderMap)
+	if err != nil {
+		panic(err)
+	}
+
+	return base64.RawURLEncoding.EncodeToString(jwtHeaderBytes)
+}
+
+// GetJWTSignature returns signature part of JWT
+func GetJWTSignature(jwt string) ([]byte, error) {
+	jwtParts := strings.Split(jwt, ".")
+	if len(jwtParts) != jwtPartsNumber || jwtParts[jwtSignaturePart] == "" {
+		return nil, errors.New("invalid JWT")
+	}
+
+	return base64.RawURLEncoding.DecodeString(jwtParts[jwtSignaturePart])
+}
+
+func getJWTHeader(jwt string) (string, error) {
+	jwtParts := strings.Split(jwt, ".")
+	if len(jwtParts) != jwtPartsNumber { // nolint:gomnd
+		return "", errors.New("invalid JWT")
+	}
+
+	return jwtParts[jwtHeaderPart], nil
+}
+
+// createVerifyJWS creates a data to be used to create/verify a digital signature in the
+// form of JSON Web Signature (JWS) with detached content (https://tools.ietf.org/html/rfc7797).
+// The algorithm of building the payload is similar to conventional  Create Verify Hash algorithm.
+// It differs by using https://w3id.org/security/v2 as context for JSON-LD canonization of both
+// JSON and Signature documents and by preliminary JSON-LD compacting of JSON document.
+// The current implementation is based on the https://github.com/digitalbazaar/jsonld-signatures.
+func createVerifyJWS(suite signatureSuite, jsonldDoc map[string]interface{}, p *Proof) ([]byte, error) {
+	proofOptions := p.JSONLdObject()
+
+	canonicalProofOptions, err := prepareJWSProof(suite, proofOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	proofOptionsDigest := suite.GetDigest(canonicalProofOptions)
+
+	canonicalDoc, err := prepareDocumentForJWS(suite, jsonldDoc)
+	if err != nil {
+		return nil, err
+	}
+
+	docDigest := suite.GetDigest(canonicalDoc)
+
+	verifyData := append(proofOptionsDigest, docDigest...)
+
+	jwtHeader, err := getJWTHeader(p.JWS)
+	if err != nil {
+		return nil, err
+	}
+
+	return append([]byte(jwtHeader+"."), verifyData...), nil
+}
+
+func prepareJWSProof(suite signatureSuite, proofOptions map[string]interface{}) ([]byte, error) {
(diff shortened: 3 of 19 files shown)


pkg/doc/signature/proof/data.go
```diff
@@ -19,6 +22,32 @@ type signatureSuite interface {
 	GetDigest(doc []byte) []byte
 }
 
+// SignatureHolding defines a representation of signature value.
+type SignatureHolding int
```
troyronda: I'm not finding "Holding" to be an intuitive description of this type.

kdimak: It's a sort of "a type of signature value holder". Maybe `SignatureRepresentation` will sound better? Or "SignatureHoldingType"?

troyronda: probably SignatureRepresentation is better. the word holding confuses me for an enumerated type.

kdimak: Thank you! renamed.

pkg/doc/signature/proof/data.go
```diff
@@ -19,6 +22,32 @@ type signatureSuite interface {
...
+// In case of "jws",
+func CreateVerifyData(suite signatureSuite, jsonldDoc map[string]interface{}, proof *Proof) ([]byte, error) {
+	switch proof.SignatureRepresentation {
+	case SignatureProofValue:
+		return CreateVerifyHash(suite, jsonldDoc, proof.JSONLdObject())
+	case SignatureJWS:
+		return createVerifyJWS(suite, jsonldDoc, proof)
+	}
+
+	return nil, fmt.Errorf("unsupported signature value holder: %v", proof.SignatureRepresentation)
```
troyronda: string wasn't updated to representation.

pkg/doc/verifiable/linked_data_proof_test.go
```diff
@@ -41,7 +44,58 @@ func Test_keyResolverAdapter_Resolve(t *testing.T) {
...
+func TestLinkedDataProofVerifier(t *testing.T) {
+	pubKeyStr := base58.Decode("5fYcSLLNTzCzCRTP7dNJiQ4AcUqB9Jf63Y4bPJ6abEFe")
+	pubKey := ed25519.PublicKey(pubKeyStr)
+
+	//nolint:lll
+	vcStr := `
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://transmute-industries.github.io/vc-greeting-card/context/vc-greeting-card-v0.0.jsonld"
```
troyronda: Will the test break if this URL goes away?

kdimak: Yes...

kdimak: Fixed!

pkg/doc/signature/proof/data.go
kdimak: fixed

pkg/doc/signature/proof/data.go
```diff
@@ -19,6 +22,32 @@ type signatureSuite interface {
...
+	SignatureProofValue SignatureRepresentation = iota
+
+	// SignatureJWS uses "jws" field in a Proof as an element for representation of detached JSON Web Signatures.
+	SignatureJWS
+)
+
+// CreateVerifyData creates data that is used to generate or verify a digital signature.
+// It depends on the signature value holder type.
+// In case of "proofValue", the standard Create Verify Hash algorithm is used.
+// In case of "jws",
```
sandrask: unfinished sentence

pkg/doc/signature/proof/jws.go
```diff
@@ -0,0 +1,183 @@
...
+	if len(jwtParts) != 3 || jwtParts[2] == "" {
+		return nil, errors.New("invalid JWT")
+	}
+
+	return base64.RawURLEncoding.DecodeString(jwtParts[2])
+}
+
+func getJWTHeader(jwt string) (string, error) {
+	jwtParts := strings.Split(jwt, ".")
+	if len(jwtParts) != 3 { // nolint:gomnd
```
sandrask: define 3 as constant e.g. jwtPartNumber instead of using magic numbers

pkg/doc/signature/proof/jws.go
```diff
@@ -0,0 +1,183 @@
...
+		return "", errors.New("invalid JWT")
+	}
+
+	return jwtParts[0], nil
```
sandrask: also create constants for token parts such as 0 should be jwtHeaderPart, 2 should be jwtSignaturePart etc.

pkg/doc/signature/proof/jws.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// createVerifyJWS creates a data to be used to create/verify a digital signature.
+// By the time of development, no specification was found with the algorithm description.
+// Current implementation is based on the https://github.com/digitalbazaar/jsonld-signatures.
```
sandrask: Did you have a chance to look at this spec: https://tools.ietf.org/html/rfc7797?

pkg/doc/signature/proof/data.go
```diff
@@ -19,6 +22,32 @@ type signatureSuite interface {
...
+func CreateVerifyData(suite signatureSuite, jsonldDoc map[string]interface{}, proof *Proof) ([]byte, error) {
```
sandrask: CreateVerifyData and CreateVerifyHash have same meaning to create data for signature (creating this data involves hashing thus CreateVerifyHash). We should probably have only one function,  let's discuss.

pkg/doc/signature/proof/jws.go
kdimak: Yes. I will add a reference to this spec. At https://github.com/digitalbazaar/jsonld-signatures, they use several heuristics which I did not found in any spec:
1) Use https://w3id.org/security/v2 as a `@context` for JSON-LD canonization of both JSON and Signature documents.
2) Apply preliminary JSON-LD compacting of JSON document using the same https://w3id.org/security/v2 `@context`.
3) Remove only `jws` and `proofValue` elements only from the proof document before canonization. At `signature` package (and what is stated in https://w3c-dvcg.github.io/ld-signatures/#create-verify-hash-algorithm), we remove `id`, `type`, `proofValue` and `jws` elements.

pkg/doc/signature/proof/data.go
kdimak: Yes, agree, it makes sense. The reason I added a separate method is that CreateVerifyHash is different in some steps in the case of JWS. I described it above https://github.com/hyperledger/aries-framework-go/pull/1137#discussion_r374120358.

pkg/doc/verifiable/linked_data_proof_test.go
```diff
@@ -11,8 +11,14 @@ import (
 	"crypto/rsa"
 	"errors"
 	"testing"
+	"time"
 
+	"github.com/btcsuite/btcutil/base58"
 	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/ed25519signature2018"
+
```
soluchok: extra line

pkg/doc/verifiable/example_credential_test.go
```diff
@@ -342,11 +342,11 @@ func ExampleCredential_AddLinkedDataProof() {
 	}
 
 	err = vc.AddLinkedDataProof(&verifiable.LinkedDataProofContext{
-		Created:       &issued,
-		Creator:       "Example University",
-		SignatureType: "Ed25519Signature2018",
-		Suite:         ed25519signature2018.New(),
-		PrivateKey:    privIssuerKey,
+		Created:                 &issued,
+		SignatureType:           "Ed25519Signature2018",
```
soluchok: better to create a constant for `"Ed25519Signature2018"`

pkg/doc/signature/proof/jws.go
```diff
@@ -0,0 +1,192 @@
...
+
+// CreateDetachedJWTHeader creates detached JWT header.
+func CreateDetachedJWTHeader(p *Proof) string {
+	jwtHeaderMap := map[string]interface{}{
+		"alg":  p.Type,
+		"b64":  false,
+		"crit": []string{"b64"},
+	}
+
+	jwtHeaderBytes, _ := json.Marshal(jwtHeaderMap) //nolint:errcheck
```
soluchok: If you want to ignore error handling, better to use panic.

```
jwtHeaderBytes, err := json.Marshal(jwtHeaderMap)
if err != nil {
    panic(err)
}
```

pkg/doc/verifiable/example_credential_test.go
kdimak: I think it's OK to use the constant as is in the test.

pkg/doc/signature/proof/jws.go
kdimak: I remember @troyronda advised not to **panic** in production code. @troyronda what's the best approach here? `jwtHeaderMap` is a constant JSON document defined in the same file (`jws.go`).

soluchok: In this case, panic is better. At least you can see the problem if it happens.

pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -52,7 +52,7 @@ func (s *SignatureSuite) GetCanonicalDocument(doc map[string]interface{}) ([]byt
 
 // GetDigest returns document digest
 func (s *SignatureSuite) GetDigest(doc []byte) []byte {
-	digest := sha512.Sum512(doc)
+	digest := sha256.Sum256(doc)
```
llorllale: What is the reason for this change? Where is the test that validates this change?

pkg/doc/signature/proof/jws.go
troyronda: @kdimak It is true that as a library we must avoid panic.

- https://github.com/golang/go/wiki/CodeReviewComments#handle-errors
- https://golang.org/doc/effective_go.html#panic
- https://github.com/uber-go/guide/blob/master/style.md#dont-panic

For error situations that are supposed to be impossible, this is an exceptional situation - something must be seriously wrong with the execution environment (or something seriously wrong with an update to some library or code we are using).

pkg/doc/signature/proof/jws.go
troyronda: Agree- shouldn't nolint this one.

pkg/doc/signature/ed25519signature2018/suite.go
kdimak: Please see https://w3c-dvcg.github.io/ld-signatures/#create-verify-hash-algorithm, ISSUE 8:
> The Ed25519Signature2018 algorithm also does not perform this last step -- and, in fact, uses SHA-512 internally. In short, this last step should better communicate that the 64 bytes produced from concatenating the SHA-256 of the canonicalized options with the SHA-256 of the canonicalized document are passed into the signing algorithm with a presumption that the signing algorithm will include hashing of its own.

We will call GetDigest() twice and concatenate the result, so it's worth to have it 32+32 bytes.

kdimak: Here is how it's implemented in https://github.com/digitalbazaar/jsonld-signatures:

https://github.com/digitalbazaar/jsonld-signatures/blob/f1e78c2674a786c3cdc8d310dbf228ed8a1e1b8e/lib/suites/LinkedDataSignature.js#L213

pkg/doc/signature/proof/jws.go
kdimak: @troyronda thank you for the really detailed explanation! So I will put a panic() here as @soluchok suggested, thanks Andrew!

pkg/doc/verifiable/linked_data_proof_test.go
kdimak: Fixed

pkg/doc/signature/proof/jws.go
kdimak: Corrected.

pkg/doc/signature/proof/data.go
kdimak: fixed

pkg/doc/signature/proof/jws.go
kdimak: fixed

pkg/doc/signature/proof/jws.go
kdimak: fixed

pkg/doc/signature/proof/data.go
```diff
@@ -86,14 +110,15 @@ const (
 	proofType excludedKey = iota + 1
 	proofID
 	proofValue
+	jws
 )
 
 func (ek excludedKey) String() string {
-	return [...]string{"type", "id", "proofValue"}[ek-1]
+	return [...]string{"type", "id", "proofValue", "jws"}[ek-1]
```
soluchok: I would make it a global vav  [...]string{"type", "id", "proofValue", "jws"}

pkg/doc/signature/proof/data.go
```diff
@@ -86,14 +110,15 @@ const (
...
 }
 
 func excludedKeyFromString(s string) excludedKey {
-	for _, ek := range [...]excludedKey{proofType, proofID, proofValue} {
+	for _, ek := range [...]excludedKey{proofType, proofID, proofValue, jws} {
```
soluchok: And this one as well under the const definition `[...]excludedKey{proofType, proofID, proofValue, jws} `

pkg/doc/signature/proof/jws.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, err
+	}
+
+	return append([]byte(jwtHeader+"."), verifyData...), nil
+}
+
+func prepareJWSProof(suite signatureSuite, proofOptions map[string]interface{}) ([]byte, error) {
+	proofOptions[jsonldContext] = securityContext
+
+	proofOptionsCopy := make(map[string]interface{})
```
soluchok: Capacity can be added e.g make(map[string]interface{},len(proofOptions))

pkg/doc/verifiable/linked_data_proof_test.go
```diff
@@ -41,7 +46,107 @@ func Test_keyResolverAdapter_Resolve(t *testing.T) {
...
 	})
 }
+
+type dummyKeyResolver struct {
+	pubKey []byte
+}
+
+func (kr *dummyKeyResolver) Resolve(string) ([]byte, error) {
+	return kr.pubKey, nil
+}
```
soluchok: Should be cheaper and simpler e.g usage `dummyKeyResolver(bytes)`
```
type dummyKeyResolver []byte

func (kr dummyKeyResolver) Resolve(string) ([]byte, error) {
	return kr, nil
}
```

pkg/doc/signature/proof/data.go
kdimak: It's a matter of taste, I think. But in this particular case, I will create two constants for `[...]string{"type", "id", "proofValue"}` and `[...]excludedKey{proofType, proofID, proofValue, jws} ` so that they will be placed together and it will be easier to refine them in future.

pkg/doc/verifiable/linked_data_proof_test.go
kdimak: Yes, I agree with that. On one hand, it's a struct definition in test code. But we need to strive to have test code even better than production one. :)

PR #1138, opened 2020-01-29
feat: Messenger - implement ReplyTo and Send functions

soluchok:
Following functions were implemented:
* Send
* ReplyTo
* ReplyToNested

NOTE: This implementation allows us to use messenger. The next changes related to the messenger would be done under the hood and should not affect usage.

### DIFF
--- a/pkg/client/introduce/example_test.go
+++ b/pkg/client/introduce/example_test.go
@@ -13,15 +13,26 @@ import (
 
 	"github.com/golang/mock/gomock"
 
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/messenger"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
 	clientIntroduceMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/client/introduce"
 	serviceMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/common/service"
+	dispatcherMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/dispatcher"
+	messengerMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/messenger"
 	protocolIntroduceMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/protocol/introduce"
 	storageMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
+
+// TODO: remove init function [Issue #1158]
+// nolint: gochecknoinits
+func init() {
+	// turns warnings off
+	log.SetLevel("aries-framework/didcomm/messenger", log.ERROR)
+}
 
 const (
 	Alice = "Alice"
@@ -80,17 +91,26 @@ func provider(agent string, transport map[string]chan transportMsg) (Provider, c
 		Return(fakeStore(ctrl), nil).
 		AnyTimes()
 
-	// creates messenger
-	messenger := serviceMocks.NewMockMessenger(ctrl)
-	messenger.EXPECT().Send(gomock.Any(), gomock.Any(), gomock.Any()).
-		Do(func(msg interface{}, myDID string, theirDID string) error {
+	storageProvider.EXPECT().OpenStore(gomock.Any()).Return(fakeStore(ctrl), nil)
+	dispatcherOutbound := dispatcherMocks.NewMockOutbound(ctrl)
+	messengerProvider := messengerMocks.NewMockProvider(ctrl)
+	messengerProvider.EXPECT().StorageProvider().Return(storageProvider)
+	messengerProvider.EXPECT().OutboundDispatcher().Return(dispatcherOutbound)
+	// use real messenger
+	msgr, err := messenger.NewMessenger(messengerProvider)
+	if err != nil {
+		fmt.Println("messenger.NewMessenger", err)
+	}
+
+	dispatcherOutbound.EXPECT().
+		SendToDID(gomock.Any(), gomock.Any(), gomock.Any()).
+		Do(func(msg service.DIDCommMsg, myDID string, theirDID string) error {
 			// sends a message
 			transport[theirDID] <- transportMsg{
 				msg:      msg.(service.DIDCommMsgMap),
 				myDID:    myDID,
 				theirDID: theirDID,
 			}
-
 			return nil
 		}).AnyTimes()
 
@@ -107,7 +127,7 @@ func provider(agent string, transport map[string]chan transportMsg) (Provider, c
 		Return(storageProvider)
 	introduceProvider.EXPECT().
 		Messenger().
-		Return(messenger)
+		Return(msgr)
 	introduceProvider.EXPECT().
 		Service(didexchange.DIDExchange).
 		Return(didexchangeService, nil)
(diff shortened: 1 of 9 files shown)


pkg/client/introduce/example_test.go
```diff
@@ -13,16 +13,27 @@ import (
...
 	clientIntroduceMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/client/introduce"
 	serviceMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/common/service"
+	dispatcherMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/dispatcher"
+	messengerMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/messenger"
 	protocolIntroduceMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/didcomm/protocol/introduce"
 	storageMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
+// TODO: remove this after correct implementation of the introduce protocol
```
troyronda: Is this the next PR?

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -277,9 +280,19 @@ func (s *confirming) ExecuteInbound(messenger service.Messenger, m *metaData) (s
 
 	msgMap := service.NewDIDCommMsgMap(model.Ack{
 		Type:   AckMsgType,
-		ID:     uuid.New().String(),
 		Thread: &decorator.Thread{ID: m.ThreadID},
 	})
+
+	// TODO: Should be replaced with ReplyTo. Currently, it is hard to do that.
```
troyronda: Issue number?

pkg/didcomm/protocol/introduce/states.go
```diff
@@ -202,13 +209,11 @@ func getApproveFromMsg(msg service.DIDCommMsg) (bool, bool) {
...
 
 func sendProblemReport(messenger service.Messenger, m *metaData, recipients []*Recipient) (state, error) {
-	problem := service.NewDIDCommMsgMap(model.ProblemReport{
-		Type: ProblemReportMsgType,
-		ID:   m.ThreadID,
-	})
-
 	for _, recipient := range recipients {
-		if err := messenger.Send(problem, recipient.MyDID, recipient.TheirDID); err != nil {
+		// TODO: description.code is required for the ProblemReport
```
troyronda: issue number?

pkg/didcomm/messenger/messenger.go
```diff
@@ -7,46 +7,155 @@ SPDX-License-Identifier: Apache-2.0
...
+	// saves message payload
+	return m.saveRecord(msg.ID(), record{MyDID: myDID, TheirDID: theirDID, ThreadID: thID})
 }
 
 // Send sends the message by starting a new thread.
 func (m *Messenger) Send(msg service.DIDCommMsgMap, myDID, theirDID string) error {
+	fixIfMalformed(msg)
+
+	if msg[jsonThread] != nil {
+		logger.Warnf("do not pass message %s, it will be ignored with the next change", jsonThread)
```
troyronda: message?

pkg/client/introduce/example_test.go
rolsonquadras: what does "correct implementation of the introduce protocol" mean here ? may be one line description of exact issue would be good.

pkg/didcomm/messenger/messenger.go
```diff
@@ -7,46 +7,155 @@ SPDX-License-Identifier: Apache-2.0
...
+	}
+
+	// sets parent threadID
+	msg[jsonThread] = map[string]interface{}{jsonParentThreadID: threadID}
+
+	return m.dispatcher.SendToDID(msg, myDID, theirDID)
+}
+
+// fixIfMalformed populates message with common fields such as ID
+func fixIfMalformed(msg service.DIDCommMsgMap) {
```
rolsonquadras: "fixIfMalformed" function name gives wrong impression like passed in request is incorrect. May be "populateDefault" or something like makes more sense.

pkg/didcomm/messenger/messenger.go
```diff
@@ -7,46 +7,155 @@ SPDX-License-Identifier: Apache-2.0
...
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+const (
+	messengerStore = "messenger_store"
+
+	jsonID             = "@id"
+	jsonThread         = "~thread"
```
sudeshrshetty: can't we export these constants from `pkg/didcomm/common/service/did_comm_msg.go`?

pkg/didcomm/messenger/messenger.go
sudeshrshetty: @troyronda --> he meant to warn `do not pass message ~thread, it will be ignored with the next change`

@soluchok : It would be nicer if you can provide detailed comment on this function, something about using decorators, dos and don'ts would be nice. Just to avoid incorrect use of this function while replying.

And, are you checking data integrity here? something like, what if someone tries to send same message with same thread ID, sender order, ID collision etc? if not, then please mention it in comments.

pkg/didcomm/messenger/messenger.go
```diff
@@ -7,46 +7,155 @@ SPDX-License-Identifier: Apache-2.0
...
+		logger.Warnf("do not pass message %s, it will be ignored with the next change", jsonThread)
+	}
+
 	return m.dispatcher.SendToDID(msg, myDID, theirDID)
 }
 
 // ReplyTo replies to the message by given msgID.
-func (*Messenger) ReplyTo(msgID string, msg service.DIDCommMsgMap) error {
-	return errors.New("does not implemented yet")
+func (m *Messenger) ReplyTo(msgID string, msg service.DIDCommMsgMap) error {
```
sudeshrshetty: dos and don't in comments please

pkg/didcomm/messenger/messenger.go
```diff
@@ -7,46 +7,155 @@ SPDX-License-Identifier: Apache-2.0
...
+	}
+
+	if msg[jsonThread] != nil {
+		logger.Warnf("do not pass message %s, the package will rewrite it", jsonThread)
+	}
+
+	// sets threadID
+	msg[jsonThread] = map[string]interface{}{jsonThreadID: rec.ThreadID}
+
+	return m.dispatcher.SendToDID(msg, rec.MyDID, rec.TheirDID)
```
sudeshrshetty: who updates `thid, pthid, sender_order, receiver_order`?

pkg/client/introduce/example_test.go
soluchok: Done! @rolsonquadras see https://github.com/hyperledger/aries-framework-go/issues/1157

pkg/didcomm/protocol/introduce/states.go
soluchok: Done!

pkg/didcomm/protocol/introduce/states.go
soluchok: Done!

pkg/didcomm/messenger/messenger.go
soluchok: Changed to `do not pass message with ~thread decorator`

pkg/didcomm/messenger/messenger.go

pkg/didcomm/messenger/messenger.go
soluchok: I'll keep it in mind. I don't want to export it. It will look like a mess. In the messenger, I am gonna use a lot of JSON tags. And import two tags from another package looks awful to me. Better to keep it together. For now, it will be separated. But, I will come up with how to merge them together. It might be related to moving the current DidCommMsg implementation under the messenger folder.

pkg/didcomm/messenger/messenger.go

pkg/didcomm/messenger/messenger.go
soluchok: It is not implemented yet! But this implementation allows you to use pkg already.

PR #1154, opened 2020-01-30
Secretlock API + mock

Baha-sk:
This API's main purpose is to encrypt (wrap) and decrypt (unwrap) keys written to/from the KMS

### DIFF
--- a/pkg/secretlock/api.go
+++ b/pkg/secretlock/api.go
@@ -1,0 +1,17 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package secretlock
+
+// package secretlock contains secret lock services to secure keys used by the Aries agent
+// and more specifically used by the KMS service.
+
+// Service provides crypto service used internally by the KMS
+type Service interface {
+	// Encrypt req for master key in keyURI
+	Encrypt(keyURI string, req *EncryptRequest) (*EncryptResponse, error)
+	// Decrypt req for master key in keyURI
+	Decrypt(keyURI string, req *DecryptRequest) (*DecryptResponse, error)
+}
--- a/pkg/secretlock/secretlock.go
+++ b/pkg/secretlock/secretlock.go
@@ -1,0 +1,28 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package secretlock
+
+// EncryptRequest for encrypting remote kms requests
+type EncryptRequest struct {
+	Plaintext                   string
+	AdditionalAuthenticatedData string
+}
+
+// DecryptRequest for decrypting remote kms requests
+type DecryptRequest struct {
+	Ciphertext                  string
+	AdditionalAuthenticatedData string
+}
+
+// EncryptResponse for receiving encryption response from remote kms requests
+type EncryptResponse struct {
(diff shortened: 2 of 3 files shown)


troyronda:
needs rebase @Baha-sk

Baha-sk:
> needs rebase @Baha-sk

done

pkg/secretlock/api.go
```diff
@@ -0,0 +1,17 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package secretlock
+
+// package secretlock contains secret lock services to secure keys used by the Aries agent
```
troyronda: Package

pkg/secretlock/api.go
```diff
@@ -0,0 +1,17 @@
...
+// and more specifically used by the KMS service.
+
+// Service provides crypto service used internally by the KMS
```
troyronda: This could probably have more description.

pkg/secretlock/secretlock.go
```diff
@@ -0,0 +1,28 @@
+/*
```
troyronda: just curious why this is a second file?

pkg/secretlock/secretlock.go
```diff
@@ -0,0 +1,28 @@
...
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package secretlock
+
+// EncryptRequest for encrypting remote kms requests
+type EncryptRequest struct {
+	Plaintext                   string
+	AdditionalAuthenticatedData string
```
llorllale: minor: suggest rename to AEAD to avoid the mouthful

pkg/secretlock/secretlock.go
```diff
@@ -0,0 +1,28 @@
...
+
+// DecryptRequest for decrypting remote kms requests
+type DecryptRequest struct {
+	Ciphertext                  string
+	AdditionalAuthenticatedData string
+}
+
+// EncryptResponse for receiving encryption response from remote kms requests
+type EncryptResponse struct {
+	Ciphertext string
```
rolsonquadras: Are we expecting more field in this ? If not, directly using the field would be more useful as we can avoid working with structs.

pkg/secretlock/api.go
Baha-sk: will update in next PR..

pkg/secretlock/secretlock.go
Baha-sk: to avoid mixing structs, vars or constants with API interfaces...

pkg/secretlock/api.go
Baha-sk: I will add more description for this service..

pkg/secretlock/secretlock.go
Baha-sk: I didn't wanna diverge from struct names in Tink.. they have similar names

pkg/secretlock/secretlock.go
Baha-sk: I followed Tink's AEADEnvelope encrypt/decrypt requests used to talk with external secret locks. Technically these are cloud requests/responses. In the case of this PR and the next one, we will mimic cloud calls for our default (local) secret lock but this will be eventually used by GCP/AWS/Hashicorp as well, so it's safer to match field names.

PR #1165, opened 2020-02-03
feat: Add Metadata field to the thread decorator

### LINKED ISSUE #1164: Add Metadata field to the thread decorator
The messenger will not be responsible for connecting a few threads into one protocol context. It should be on the protocol. The protocol will have metadata and may do whatever it needs. Messenger is responsible for saving it and provide it for the incoming message (by threadID). 

```
// Thread thread data
type Thread struct {
   ...
   // Metadata may contain additional payload for the protocol. It might be populated by the client/protocol
   // for outbound messages. If metadata were populated, the messenger will automatically add it to the incoming
   // messages by the threadID. If Metadata is <nil> in the outbound message the previous payload
   // will be added to the incoming message. Otherwise, the payload will be rewritten.
   // NOTE: Metadata is not a part of the JSON message. The payload will not be sent to another agent.
   Metadata map[string]interface{} `json:"metadata"`
}
```

soluchok:
* Added Metadata to the ~thread decorator
* Messenger saves and provides data

Closes #1164 

### DIFF
--- a/pkg/didcomm/common/service/did_comm_msg.go
+++ b/pkg/didcomm/common/service/did_comm_msg.go
@@ -23,7 +23,21 @@ const (
 	jsonType     = "@type"
 	jsonThread   = "~thread"
 	jsonThreadID = "thid"
+	jsonMetadata = "_internal_metadata"
 )
+
+// Metadata may contain additional payload for the protocol. It might be populated by the client/protocol
+// for outbound messages. If metadata were populated, the messenger will automatically add it to the incoming
+// messages by the threadID. If Metadata is <nil> in the outbound message the previous payload
+// will be added to the incoming message. Otherwise, the payload will be rewritten.
+// NOTE: Metadata is not a part of the JSON message. The payload will not be sent to another agent.
+// Metadata should be used by embedding it to the model structure. e.g
+// 	type A struct {
+// 		Metadata `json:",squash"`
+// 	}
+type Metadata struct {
+	Payload map[string]interface{} `json:"_internal_metadata,omitempty"`
+}
 
 // Header helper structure which keeps reusable fields
 // Deprecated: Use DIDCommMsg instead of Header
@@ -88,6 +102,20 @@ func (m DIDCommMsgMap) ThreadID() (string, error) {
 	}
 
 	return "", ErrThreadIDNotFound
+}
+
+// Metadata returns message metadata
+func (m DIDCommMsgMap) Metadata() map[string]interface{} {
+	if m[jsonMetadata] == nil {
+		return nil
+	}
+
+	metadata, ok := m[jsonMetadata].(map[string]interface{})
+	if !ok {
+		return nil
+	}
+
+	return metadata
 }
 
 // Type returns the message type
--- a/pkg/didcomm/messenger/messenger.go
+++ b/pkg/didcomm/messenger/messenger.go
@@ -22,10 +22,13 @@ import (
 const (
 	messengerStore = "messenger_store"
 
+	metadataKey = "metadata_%s"
+
 	jsonID             = "@id"
 	jsonThread         = "~thread"
 	jsonThreadID       = "thid"
 	jsonParentThreadID = "pthid"
+	jsonMetadata       = "_internal_metadata"
 )
 
 var logger = log.New("aries-framework/didcomm/messenger")
@@ -35,6 +38,7 @@ type record struct {
 	MyDID    string
 	TheirDID string
 	ThreadID string
+	Metadata map[string]interface{}
 }
 
 // Provider contains dependencies for the Messenger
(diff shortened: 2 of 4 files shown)


pkg/didcomm/messenger/messenger.go
```diff
@@ -77,16 +81,84 @@ func (m *Messenger) HandleInbound(msg service.DIDCommMsgMap, myDID, theirDID str
...
+
+	// if the thread had only metadata we remove it from the message as well
+	if len(thread) == 0 {
+		delete(msg, jsonThread)
+	}
+
+	return m.saveRecord(fmt.Sprintf(metadataKey, thID), record{Metadata: metadata})
+}
+
+func (m *Messenger) withMetadata(thID string, msg service.DIDCommMsgMap) error {
```
troyronda: I'd prefer to use the prefix "with" to indicate an Option.

Could we rename this one?

soluchok: Sure

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -25,6 +25,12 @@ type Thread struct {
 	PID            string         `json:"pthid,omitempty"`
 	SenderOrder    int            `json:"sender_order,omitempty"`
 	ReceivedOrders map[string]int `json:"received_orders,omitempty"`
+	// Metadata may contain additional payload for the protocol. It might be populated by the client/protocol
+	// for outbound messages. If metadata were populated, the messenger will automatically add it to the incoming
+	// messages by the threadID. If Metadata is <nil> in the outbound message the previous payload
+	// will be added to the incoming message. Otherwise, the payload will be rewritten.
+	// NOTE: Metadata is not a part of the JSON message. The payload will not be sent to another agent.
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
```
troyronda: I wonder if it would make sense to have a wrapper struct embedding an anonymous Thread struct + Metadata to more clearly delineate on-the-wire fields from fields kept within the agent.

(also if a protocol doesn't care about metadata then they can continue to use the undecorated Thread structure.)

soluchok: I am not sure, e.g when we send `msg1` we have this wrapper. When we receive a message `msg2` we cannot inject metadata. Because it does not have this wrapper structure.

rolsonquadras: agree with @troyronda. By adding these, we are diluting the Thread struct(which is used as part of DIDComm message).

pkg/didcomm/messenger/messenger.go
soluchok: I've changed it to `populateMetadata`

pkg/didcomm/messenger/messenger.go
```diff
@@ -77,16 +81,84 @@ func (m *Messenger) HandleInbound(msg service.DIDCommMsgMap, myDID, theirDID str
...
+	if msg[jsonThread] == nil {
+		return nil
+	}
+
+	thread, ok := msg[jsonThread].(map[string]interface{})
+	if !ok || thread == nil {
+		return nil
+	}
+
+	metadata, ok := thread[jsonMetadata].(map[string]interface{})
```
llorllale: If protocols need access to this metadata then `service.DIDCommMsgMap` should have an accesor for it.

pkg/didcomm/messenger/messenger.go
```diff
@@ -77,16 +81,84 @@ func (m *Messenger) HandleInbound(msg service.DIDCommMsgMap, myDID, theirDID str
...
+	if err := m.populateMetadata(thID, msg); err != nil {
+		return fmt.Errorf("with metadata: %w", err)
+	}
+
 	// saves message payload
 	return m.saveRecord(msg.ID(), record{MyDID: myDID, TheirDID: theirDID, ThreadID: thID})
 }
 
+func (m *Messenger) saveMetadata(msg service.DIDCommMsgMap) error {
+	if msg[jsonThread] == nil {
```
llorllale: I would assume a protocol may add metadata to a message when initiating a protocol. The `~thread` decorator is not normally inserted in those cases. This suggests to me that the metadata should be contained outside the `~thread`.

pkg/didcomm/messenger/messenger.go
```diff
@@ -77,16 +81,84 @@ func (m *Messenger) HandleInbound(msg service.DIDCommMsgMap, myDID, theirDID str
...
+	if err != nil {
+		return fmt.Errorf("get record: %w", err)
+	}
+
+	if rec.Metadata == nil {
+		return nil
+	}
+
+	if msg[jsonThread] == nil {
+		msg[jsonThread] = map[string]interface{}{jsonMetadata: rec.Metadata}
```
llorllale: Suggest placing the metadata outside the `~thread`.

pkg/didcomm/protocol/decorator/decorator.go
llorllale: I would place the metadata entirely outside the `~thread` decorator.

pkg/didcomm/messenger/messenger.go
soluchok: Done! I have added a comment about how to use it.
```
type MetaData struct {
	Metadata map[string]interface{} `json:"metadata,omitempty"`
}
```

pkg/didcomm/protocol/decorator/decorator.go
soluchok: Done!

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -25,6 +25,19 @@ const (
...
+// MetaData may contain additional payload for the protocol. It might be populated by the client/protocol
+// for outbound messages. If metadata were populated, the messenger will automatically add it to the incoming
+// messages by the threadID. If Metadata is <nil> in the outbound message the previous payload
+// will be added to the incoming message. Otherwise, the payload will be rewritten.
+// NOTE: Metadata is not a part of the JSON message. The payload will not be sent to another agent.
+// Metadata should be used by embedding it to the model structure. e.g
+// 	type A struct {
+// 		MetaData `json:",squash"`
+// 	}
+type MetaData struct {
```
troyronda: I assume this should be `Metadata`?

soluchok: If you have some suggestions about naming, let me know! The problem is type and field name should be different.
```
type MetaData struct {
	Metadata map[string]interface{} `json:"metadata,omitempty"`
}
```

troyronda: Seems like a structure related to a Msg or Messenger?

Examples (there are probably better ideas):
- MsgInfo
- MessengerData

soluchok: Actually, the messenger doesn't need this. It is a protocol-specific payload. For me, metadata is the best name.  I've modified it. Also, I have added an accesor for it.

```
type Metadata struct {
	Payload map[string]interface{} `json:"metadata,omitempty"`
}

```

pkg/didcomm/messenger/messenger.go

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -23,8 +23,22 @@ const (
 	jsonType     = "@type"
 	jsonThread   = "~thread"
 	jsonThreadID = "thid"
+	jsonMetadata = "metadata"
```
troyronda: This should be a prefixed field to avoid collision with a real field named metadata.

soluchok: I've changed `metadata` to `internal_metadata`

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -23,8 +23,22 @@ const (
 	jsonType     = "@type"
 	jsonThread   = "~thread"
 	jsonThreadID = "thid"
+	jsonMetadata = "internal_metadata"
```
troyronda: Nit: maybe also put a leading symbol like an _

soluchok: Done! `_internal_metadata`

PR #1162, opened 2020-01-31
feat: Support for Multiple Inbound Transports

rolsonquadras:
Closes #1124 

Signed-off-by: Rolson Quadras <rolson.quadras@securekey.com>

PR #1170, opened 2020-02-03
refactor: Update context.InboundTransportEndpoint to context.ServiceEndpoint

rolsonquadras:
Support for multiple inbound transports was added as part of #1162. A new context method was added to provide Router Service endpoint.  This change updates the context method name which provides the service end point to be in sync with Router service endpoint.

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -35,7 +35,7 @@ var ErrConnectionNotFound = errors.New("connection not found")
 type provider interface {
 	Service(id string) (interface{}, error)
 	LegacyKMS() legacykms.KeyManager
-	InboundTransportEndpoint() string
+	ServiceEndpoint() string
 	StorageProvider() storage.Provider
 	TransientStorageProvider() storage.Provider
 }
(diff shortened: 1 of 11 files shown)


pkg/client/didexchange/client.go
```diff
@@ -35,19 +35,19 @@ var ErrConnectionNotFound = errors.New("connection not found")
 type provider interface {
 	Service(id string) (interface{}, error)
 	LegacyKMS() legacykms.KeyManager
-	InboundTransportEndpoint() string
```
troyronda: It would be nice to add a comment to `InboundTransportEndpoint` regarding when to use it (and I assume to prefer ServiceEndpoint?).

rolsonquadras: [Added](https://github.com/hyperledger/aries-framework-go/compare/1845bf271e498ea6bda19e69e21961bf0c0bf0d1..cdfef7229d58c9c2bcc938d50f1e7c2d1c9bb964#diff-9988f5820bf05228e5657f2842fa071eR115).

PR #1172, opened 2020-02-03
feat: Add (local) SecretLock implementation

### LINKED ISSUE #1146: Add default SecretLock implementation
part of #985

Baha-sk:
This change includes reading a masterkey
            from 2 sources:
            1. masterKey loaded from env variable
            2. masterKey loaded from local path
    
    This change also includes password protection lock to encrypt
    the master key for additional protection of keys.
    
closes #1146

### DIFF
--- a/pkg/framework/aries/default.go
+++ b/pkg/framework/aries/default.go
@@ -122,6 +122,8 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
 		frameworkOpts.msgSvcProvider = &noOpMessageServiceProvider{}
 	}
 
+	// TODO add SecretLock creation here.. #1150
+
 	return nil
 }
 
--- a/pkg/framework/aries/framework.go
+++ b/pkg/framework/aries/framework.go
@@ -25,6 +25,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
 	vdriapi "github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/context"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/vdri"
 	"github.com/hyperledger/aries-framework-go/pkg/vdri/peer"
@@ -50,6 +51,7 @@ type Aries struct {
 	inboundTransports      []transport.InboundTransport
 	kmsCreator             api.KMSCreator
 	kms                    api.CloseableKMS
+	secretLock             secretlock.Service
 	crypto                 crypto.Crypto
 	packagerCreator        packager.Creator
 	packager               commontransport.Packager
@@ -211,6 +213,14 @@ func WithLegacyKMS(k api.KMSCreator) Option {
 	}
 }
 
+// WithSecretLock injects a SecretLock service to the Aries framework
+func WithSecretLock(s secretlock.Service) Option {
+	return func(opts *Aries) error {
+		opts.secretLock = s
+		return nil
+	}
+}
+
 // WithCrypto injects a crypto service to the Aries framework
 func WithCrypto(c crypto.Crypto) Option {
 	return func(opts *Aries) error {
@@ -257,6 +267,7 @@ func (a *Aries) Context() (*context.Provider, error) {
 		context.WithOutboundTransports(a.outboundTransports...),
 		context.WithProtocolServices(a.services...),
 		context.WithLegacyKMS(a.kms),
+		context.WithSecretLock(a.secretLock),
 		context.WithCrypto(a.crypto),
 		context.WithServiceEndpoint(serviceEndpoint(a)),
 		context.WithRouterEndpoint(routingEndpoint(a)),
--- a/pkg/framework/aries/framework_test.go
+++ b/pkg/framework/aries/framework_test.go
@@ -9,6 +9,8 @@ SPDX-License-Identifier: Apache-2.0
 package aries
 
 import (
+	"crypto/rand"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -19,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/google/tink/go/subtle/random"
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
@@ -38,6 +41,9 @@ import (
 	mockkms "github.com/hyperledger/aries-framework-go/pkg/mock/kms/legacykms"
 	"github.com/hyperledger/aries-framework-go/pkg/mock/storage"
 	mockvdri "github.com/hyperledger/aries-framework-go/pkg/mock/vdri"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
+	locallock "github.com/hyperledger/aries-framework-go/pkg/secretlock/local"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock/local/masterlock/hkdf"
 	"github.com/hyperledger/aries-framework-go/pkg/storage/leveldb"
 	"github.com/hyperledger/aries-framework-go/pkg/vdri/peer"
 )
@@ -380,14 +386,73 @@ func TestFramework(t *testing.T) {
 		err = aries.Close()
 		require.NoError(t, err)
 	})
-	t.Run("test error from kms svc", func(t *testing.T) {
-		// with custom kms
+
+	t.Run("test error from legacy kms svc", func(t *testing.T) {
+		// with custom legacy kms
 		_, err := New(WithInboundTransport(&mockInboundTransport{}),
 			WithLegacyKMS(func(ctx api.Provider) (api.CloseableKMS, error) {
 				return nil, fmt.Errorf("error from kms")
 			}))
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "error from kms")
+	})
+
+	t.Run("test new with custom kms svc and secret lock svc", func(t *testing.T) {
+		// pre steps (preparation), create a protected master key and store it in a local file
+		masterKeyFilePath := "masterKey_aries.txt"
+		tmpfile, err := ioutil.TempFile("", masterKeyFilePath)
+		require.NoError(t, err)
+
+		defer func() {
+			// close file
+			require.NoError(t, tmpfile.Close())
+			// clean up
+			require.NoError(t, os.Remove(tmpfile.Name()))
+		}()
+
+		passphrase := "testPassword"
+		keySize := sha256.New().Size()
+
+		salt := make([]byte, keySize)
+		_, err = rand.Read(salt)
+		require.NoError(t, err)
+
+		// create a master lock to protect the master key (salt is optional)
+		masterLock, err := hkdf.NewMasterLock(passphrase, sha256.New, salt)
+		require.NoError(t, err)
+		require.NotEmpty(t, masterLock)
+
+		// generate a random master key
+		masterKeyContent := random.GetRandomBytes(uint32(keySize))
+		require.NotEmpty(t, masterKeyContent)
+
+		// encrypt it
+		masterKeyEnc, err := masterLock.Encrypt("", &secretlock.EncryptRequest{
+			Plaintext: string(masterKeyContent)})
+		require.NoError(t, err)
+		require.NotEmpty(t, masterKeyEnc)
+
+		// store encrypted content to file
+		n, err := tmpfile.Write([]byte(masterKeyEnc.Ciphertext))
+		require.NoError(t, err)
+		require.Equal(t, len(masterKeyEnc.Ciphertext), n)
+
+		// create a reader for this file
+		r, err := locallock.MasterKeyFromPath(tmpfile.Name())
+		require.NoError(t, err)
+		require.NotEmpty(t, r)
+
+		// normal steps - once a protected master key is stored in a local path (or env variable)
+		// 		a user will first create a secretLock using a masterLock (same as above) and
+		//		a reader to the local path (or env using locallock.MasterKeyFromEnv())
+		s, err := locallock.NewService(r, masterLock)
+		require.NoError(t, err)
+		require.NotEmpty(t, s)
+
+		// final step, create the Aries agent with this secret lock
+		a, err := New(WithSecretLock(s))
+		require.NoError(t, err)
+		require.NotEmpty(t, a)
 	})
 
 	t.Run("test transient store - with user provided transient store", func(t *testing.T) {
--- a/pkg/framework/context/context.go
+++ b/pkg/framework/context/context.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
 	vdriapi "github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
 	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
@@ -28,6 +29,7 @@ type Provider struct {
 	storeProvider          storage.Provider
 	transientStoreProvider storage.Provider
 	kms                    legacykms.KMS
+	secretLock             secretlock.Service
 	crypto                 crypto.Crypto
 	packager               commontransport.Packager
 	primaryPacker          packer.Packer
@@ -80,6 +82,11 @@ func (p *Provider) Service(id string) (interface{}, error) {
 // LegacyKMS returns a kms service.
 func (p *Provider) LegacyKMS() legacykms.KeyManager {
 	return p.kms
+}
+
+// SecretLock returns a secret lock service
+func (p *Provider) SecretLock() secretlock.Service {
+	return p.secretLock
 }
 
 // Crypto returns the Crypto service
@@ -244,6 +251,14 @@ func WithProtocolServices(services ...dispatcher.ProtocolService) ProviderOption
 func WithLegacyKMS(w legacykms.KMS) ProviderOption {
 	return func(opts *Provider) error {
 		opts.kms = w
+		return nil
+	}
+}
+
+// WithSecretLock injects a secret lock service into the context
+func WithSecretLock(s secretlock.Service) ProviderOption {
+	return func(opts *Provider) error {
+		opts.secretLock = s
 		return nil
 	}
 }
--- a/pkg/secretlock/local/local_secret_lock.go
+++ b/pkg/secretlock/local/local_secret_lock.go
@@ -1,0 +1,147 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package local
+
+import (
+	"crypto/cipher"
+	"encoding/base64"
+	"fmt"
+	"io"
+
+	"github.com/google/tink/go/subtle/random"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
+	cipherutil "github.com/hyperledger/aries-framework-go/pkg/secretlock/local/internal/cipher"
+)
+
+// package local provides a local secret lock service. The user must create a master key and store it
+// in a local file or an environment variable prior to using this service.
+//
+// The user has the option to encrypt the master key using hkdf.NewMasterLock(passphrase, hash func(), salt)
+// found in the sub package masterlock/hkdf.
+//
+// The master key must be stored (encrypted with a MasterLock or not encrypted) either in a file or in
+// an environment variable.
+//
+// The user can then call either:
+//		MasterKeyFromPath(path) or
+//		MasterKeyFromEnv(envPrefix, keyURI)
+// to get an io.Reader instance needed to read the master key and create a keys Lock service.
+//
+// It is recommended for the content of the reader to be base64URL encoded (by masterlock if protected or manually
+// if not). This is particularly true when setting a master key in an environment variable as some OSs may
+// reject setting env variables with binary data as value. The service will attempt to base64URL decode the content of
+// reader first and if it fails, will try to create the service with the raw (binary) content.
+//
+// To get the lock service, call:
+// 		NewService(reader, secLock)
+// where reader is the instance returned from one of the MasterKeyFrom..() functions mentioned above
+// and secLock which is the masterKey lock used to encrypt/decrypt the master key. If secLock is nil
+// then the masterKey content in reader will be used as-is without being decrypted. The keys however are always
+// encrypted using the read masterKey.
+
+var logger = log.New("aries-framework/lock")
+
+const masterKeyLen = 512
+
+// Lock is a secret lock service responsible for encrypting keys using a master key
+type Lock struct {
+	aead cipher.AEAD
+}
+
+// NewService creates a new instance of local secret lock service using a master key in masterKeyReader.
+// If the masterKey is not protected (secLock=nil) this function will attempt to base64 URL Decode the
+// content of masterKeyReader and if it fails, then will attempt to create a secret lock cipher with the raw key as is.
+func NewService(masterKeyReader io.Reader, secLock secretlock.Service) (secretlock.Service, error) {
+	masterKeyData := make([]byte, masterKeyLen)
+
+	if masterKeyReader == nil {
+		return nil, fmt.Errorf("masterKeyReader is nil")
+	}
+
+	n, err := masterKeyReader.Read(masterKeyData)
+	if err != nil {
+		if err != io.EOF && err != io.ErrUnexpectedEOF {
+			return nil, err
+		}
+	}
+
+	if n == 0 {
+		return nil, fmt.Errorf("masterKeyReader is empty")
+	}
+
+	var masterKey []byte
+
+	// if secLock not empty, then masterKeyData is encrypted (protected), let's decrypt it first.
+	if secLock != nil {
+		decResponse, e := secLock.Decrypt("", &secretlock.DecryptRequest{
+			Ciphertext: string(masterKeyData[:n])})
+		if e != nil {
+			return nil, e
+		}
+
+		masterKey = []byte(decResponse.Plaintext)
+	} else {
+		// masterKeyData is not encrypted, base64URL decode it
+		masterKey, err = base64.URLEncoding.DecodeString(string(masterKeyData[:n]))
+		if err != nil {
+			logger.Warnf("base64URL.Decode of unprotected master key failed. " +
+				"Will attempt to create a service using the key content from reader as is.")
+
+			masterKey = make([]byte, n)
+
+			// copy masterKey read from reader directly
+			copy(masterKey, masterKeyData)
+		}
+	}
+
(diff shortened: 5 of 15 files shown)


pkg/secretlock/env/env_secret_lock.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// Lock is a secret lock service responsible for encrypting keys using a master key stored as environment variable
+type Lock struct {
+	aead cipher.AEAD
+}
+
+// NewService creates a new instance of a local secret lock service
+func NewService(keyURI string) (secretlock.Service, error) {
+	mk := os.Getenv("LOCAL_" + strings.ReplaceAll(keyURI, "/", "_"))
```
troyronda: the prefix should be an argument - normally representing the end application name.

troyronda: also i do not understand why keyURI is passed in here.

shouldn’t the env key be constructed in the other functions based on keyURI?

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	mkData = mkData[0:n]
+
+	masterKey, err := base64.URLEncoding.DecodeString(string(mkData))
+	if err != nil {
+		return nil, err
+	}
+
+	aead, err := chacha20poly1305.NewX(masterKey)
```
troyronda: cannot hard-code to chacha.

why not AES GCM?

pkg/secretlock/env/env_secret_lock.go
```diff
@@ -0,0 +1,75 @@
...
+	if mk == "" {
+		return nil, fmt.Errorf("masterKey not set")
+	}
+
+	masterKey, err := base64.URLEncoding.DecodeString(mk)
+	if err != nil {
+		return nil, err
+	}
+
+	aead, err := chacha20poly1305.NewX(masterKey)
```
troyronda: cannot hard-code to chacha.

why not AES GCM?

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
+)
+
+// Lock is a secret lock service responsible for encrypting keys using a master key stored as environment variable
+type Lock struct {
+	aead cipher.AEAD
+}
+
+// NewService creates a new instance of a local secret lock service for a master key in path
+func NewService(path string) (secretlock.Service, error) {
```
troyronda: - should have more generic base version that uses io.writer.

- need to allow for optional passphrase option to protect key files. use secret expansion algorithm.
- might be sensible to chain secret locks together... have a passphrase instance nested on top of io.writer instance.

pkg/secretlock/env/env_secret_lock.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &Lock{aead: aead}, nil
+}
+
+// Encrypt req using key in the local secret lock service (keyURI is used for remote locks, it is ignored by this
+// implementation)
```
troyronda: I see keyURI is always being ignored in these functions. Instead it is being passed into the constructor and then ignored.

 Can you elaborate on this design?

pkg/secretlock/env/env_secret_lock.go
Baha-sk: sure, I will make it a prefix argument

pkg/secretlock/env/env_secret_lock.go
Baha-sk: this is internal to framework, the user should not offered to select the key wrapping algorithm for local storage, or do you think otherwise?

chacha is selected for faster/shorter key wrapping ops

pkg/secretlock/env/env_secret_lock.go
Baha-sk: to support cloud storage (even Android keystore in Java/iOS Keychain in objectiveC), master keys are referenced by a keyURI. This is an implementation of secretlock.Service

for local/env storage, keyURI is meaningless as it is fetched from a local path/env variable.

pkg/secretlock/local/local_secret_lock.go
Baha-sk: it's probably good for a passphrase implementation, we're using pure crypto here which requires a real crypto key

again, secret locks are not meant to be exposed to the user, it's an internal service (used by KMS primarily)

pkg/secretlock/local/local_secret_lock.go
Baha-sk: Secret locks are not meant to be exposed to the user, it's an internal service (used by KMS primarily), so the user will not choose this algorithm.

chacha is used for faster/shorter crypto operations

Moopli: It's not about user choice, it's about ability to meet a NIST-only requirement - it should be possible to build an agent that only uses AES for key wrapping.

pkg/secretlock/local/local_secret_lock.go
troyronda: @Baha-sk there are other places a key can come from other than the disk. Better to use io.Writer as the base - I don't want to deal with each variation individually.

troyronda: @Baha-sk I think you also missed the chaining comment. There are deployments where there is an encrypted key in *some* storage and that key needs to be decrypted by a passphrase. The decrypted secret is then used.

Baha-sk: I'll add a variation that takes an io.Reader as option

pkg/secretlock/env/env_secret_lock.go
Baha-sk: done

pkg/secretlock/local/local_secret_lock.go
Baha-sk: switched

pkg/secretlock/env/env_secret_lock.go
Baha-sk: switched to AES GCM

pkg/secretlock/local/local_secret_lock.go
Baha-sk: added

pkg/secretlock/env/env_secret_lock_test.go
```diff
@@ -0,0 +1,161 @@
...
+			require.NoError(t, e)
+		}()
+
+		// test the creation of a local secret lock service with an invalid master key
+		s, err := NewService(envPrefix, testKeyURI)
+		require.Error(t, err)
+		require.Empty(t, s)
+	})
+
+	t.Run("error case - master key length greater than chacha20poly1305.KeySize", func(t *testing.T) {
```
Moopli: AES, not chacha

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return NewServiceFromReader(masterKeyFile)
+}
+
+// NewServiceFromReader creates a new instance of local secret lock service for a master key reader.
+// the master key should be an AES key with size either 16, 24, or 32 bytes to select AES-128, AES-192, or AES-256.
+func NewServiceFromReader(masterKeyReader io.Reader) (secretlock.Service, error) {
```
Moopli: Could we enforce key size? That is, return an error if the key provided is too short for the security policy?

pkg/framework/aries/framework.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/messenger"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
```
soluchok: Wrong order! Should be `build-in, 3rd party and local`

pkg/framework/aries/framework_test.go
```diff
@@ -390,6 +394,39 @@ func TestFramework(t *testing.T) {
...
 
+	t.Run("test new with custom kms svc", func(t *testing.T) {
+		masterKeyFilePath := "masterKey_aries.txt"
+		tmpfile, err := ioutil.TempFile("", masterKeyFilePath)
+		require.NoError(t, err)
+
+		defer func() {
+			er := os.Remove(tmpfile.Name()) // clean up
+			require.NoError(t, er)
+		}()
```
soluchok: I really don't like it when an error variable is not `err`. Pretty of unusual. I would replace it with `defer func() { require.NoError(t, os.Remove(tmpfile.Name())) }()`

pkg/secretlock/env/env_secret_lock_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/google/tink/go/subtle/random"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
+)
+
+func TestSecretLockService(t *testing.T) {
+	// verify Lock implements secretlock.Service
+	var _ secretlock.Service = (*Lock)(nil)
```
soluchok: Move it outside the body function

pkg/framework/aries/framework.go
Baha-sk: it's not wrong order unless we separate them by empty lines... this is how `go imports` work and linter doesn't fail with this

troyronda: @Baha-sk you should use `goimports -local github.com/hyperledger/aries-framework-go` or manually sort as @soluchok  mentioned. The linter is supposed to be looking for this too.

`goimports --help`
"-local string put imports beginning with this string after 3rd-party packages"

pkg/framework/aries/default.go
```diff
@@ -122,6 +122,8 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
 		frameworkOpts.msgSvcProvider = &noOpMessageServiceProvider{}
 	}
 
+	// TODO add SecretLock creation here..
```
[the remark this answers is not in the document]
Baha-sk: this will be updated when the new local kms change is pushed

pkg/secretlock/env/env_secret_lock_test.go
Baha-sk: chacha removed

pkg/framework/aries/framework.go
Baha-sk: that was the IDE adding the import at the wrong place... it's now fixed

pkg/secretlock/env/env_secret_lock_test.go
Baha-sk: removed assignment statement

pkg/framework/aries/framework_test.go
Baha-sk: removed variable

pkg/secretlock/local/local_secret_lock.go
Baha-sk: the new API generates the key according the cipher... the only thing the user is required to pass is a passphrase. There's no need to expose a key size option to the user.

pkg/framework/aries/default.go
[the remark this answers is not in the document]
troyronda: Issue number?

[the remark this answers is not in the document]
Baha-sk: added issue #

pkg/secretlock/local/standard_secret_lock.go
```diff
@@ -0,0 +1,140 @@
...
+
+var logger = log.New("aries-framework/lock")
+
+// Lock is a secret lock service responsible for encrypting keys using a master key stored as environment variable.
+type Lock struct {
+	aead cipher.AEAD
+}
+
+// InitFromPath creates a new instance of a local secret lock Reader for an encrypted master key in path.
+func InitFromPath(passphrase, path string) (io.Reader, error) {
```
llorllale: These two "Init" functions could be renamed to something that alludes to their purpose - get the master key.

I suggest something like "MasterKeyFromPath"/"MasterKeyFromEnv" or something similar.

pkg/secretlock/local/standard_secret_lock.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	return newReader(passphrase, base58.Decode(mk))
+}
+
+// NewServiceFromReader creates a new instance of local secret lock service for a master key reader.
+// Even though the reader returned by newReader() in this package does authenticate passphrase,
+// it is required here to authenticate masterKeyReader (if passing in a reader not returned by newReader())
+func NewServiceFromReader(passphrase string, masterKeyReader io.Reader) (secretlock.Service, error) {
+	masterKey := make([]byte, 32)
```
llorllale: should we fix the key's size? Shouldn't the key type be parameterized?

pkg/secretlock/local/standard_secret_lock.go
```diff
@@ -0,0 +1,140 @@
...
+func InitFromEnv(passphrase, envPrefix, keyURI string) (io.Reader, error) {
+	mk := os.Getenv(envPrefix + strings.ReplaceAll(keyURI, "/", "_"))
+	if mk == "" {
+		return nil, fmt.Errorf("masterKey not set")
+	}
+
+	return newReader(passphrase, base58.Decode(mk))
+}
+
+// NewServiceFromReader creates a new instance of local secret lock service for a master key reader.
```
llorllale: s/for a master key reader/from a master key reader/

pkg/secretlock/local/masterkey_reader.go
```diff
@@ -0,0 +1,61 @@
...
+
+package local
+
+import (
+	"fmt"
+	"io"
+)
+
+// newReader will create a new master key reader using passphrase
+func newReader(passphrase string, encMasterKey []byte) (*Reader, error) {
```
llorllale: nit pick - you could just return `io.Reader`

pkg/secretlock/local/standard_secret_lock.go
Baha-sk: Init for initializing the SecretLock, which happens to be fetching a reader of the master key to be used. If `MasterKey` makes more sense than `Init`, I'll rename the functions.

pkg/secretlock/local/standard_secret_lock.go
Baha-sk: I'll s/secret lock service for a master key reader/secret lock service using a master key in reader/

pkg/secretlock/local/standard_secret_lock.go
Baha-sk: Reader must return a SHA256 string to be verified against passphrase which makes it a fixed size.

 see `func generateMasterKey(passphrase string)` in passphrase_util.go above

this can be updated in the future if SHA256 protection is updated to something else

however, the size here can be increased, to be large enough to support any key, say 1024. For now, it is not needed unless the passphrase protection

pkg/secretlock/local/masterkey_reader.go
Baha-sk: good idea

pkg/secretlock/local/standard_secret_lock.go
troyronda: SHA256 doesn't offer "protection".

pkg/secretlock/local/passphrase_util.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func generateMasterKey(passphrase string) []byte {
+	// TODO consider using bcrypt: https://github.com/golang/crypto/tree/master/bcrypt
+	// mk, err := bcrypt.GenerateFromPassword([]byte(passphrase), bcrypt.MaxCost)
+	// if err != nil {
+	//  	return nil, err
+	// }
+	// instead of sha256 below - this means above logic must validate key length
+	masterKey := sha256.Sum256([]byte(passphrase))
```
troyronda: This should be some form of secret expansion - simply hashing the passphrase is generally insufficient.

pkg/secretlock/local/passphrase_util.go
```diff
@@ -0,0 +1,85 @@
...
+func CreateMasterKey(passphrase string) (string, error) {
+	masterKey := generateMasterKey(passphrase)
+
+	aeadCipher, err := createAESCipher(masterKey)
+	if err != nil {
+		return "", err
+	}
+
+	nonce := random.GetRandomBytes(uint32(aeadCipher.NonceSize()))
+	ct := aeadCipher.Seal(nil, nonce, masterKey, []byte(passphrase))
```
troyronda: why is passphrase being used as additional data?
it's supposed to be encrypting the master key.

Baha-sk: `masterKey` is being used to encrypt itself with `passphrase` as aad.

for decryption, we regenerate `masterKey` using `passphrase` and try to decrypt the the cipher text with `passphrase` as aad, if successful, it means `masterKey` is valid.

pkg/secretlock/local/passphrase_util.go
Baha-sk: bcrypt in the comment above is an option, or we can use HKDF: https://godoc.org/golang.org/x/crypto/hkdf#ex-package--Usage

the aim, is this can be done in a separate PR

pkg/secretlock/local/standard_secret_lock.go
llorllale: `MasterKey` makes more sense to me.

Baha-sk: the functions are now renamed to `MasterKeyFromPath` and `MasterKeyFromEnv`

pkg/secretlock/local/standard_secret_lock.go
Baha-sk: the logic has now been changed, master keys are protected by key extraction using HKDF.. the key size must match the hash size as we use aead for encryption/decryption

pkg/secretlock/local/masterkey_reader.go
Baha-sk: the reader file has been removed.. there is no more reader implementation in this change.

pkg/secretlock/local/passphrase_util.go
Baha-sk: done in master_secret.go .. we now use HKDF secret extraction

pkg/secretlock/local/passphrase_util.go
Baha-sk: this file and its logic has been removed.

pkg/secretlock/local/standard_secret_lock.go
Baha-sk: this function was refactored

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+	mk := os.Getenv(envPrefix + strings.ReplaceAll(keyURI, "/", "_"))
+	if mk == "" {
+		return nil, fmt.Errorf("masterKey not set")
+	}
+
+	return bytes.NewReader([]byte(mk)), nil
+}
+
+// NewService creates a new instance of local secret lock service using a master key in masterKeyReader.
+func NewService(masterKeyReader io.Reader, secLock secretlock.Service) (secretlock.Service, error) {
```
troyronda: should use an option instead of relying on nil.

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+	masterKeyData := make([]byte, 1024)
```
troyronda: why 1024?

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+	ct = append(nonce, ct...)
+
+	return &secretlock.EncryptResponse{
+		Ciphertext: base64.URLEncoding.EncodeToString(ct),
+	}, nil
+}
+
+// Decrypt a key in req using master key in the local secret lock service
+// (keyURI is used for remote locks, it is ignored by this implementation)
+// nolint:dupl
```
troyronda: what is being duplicated?

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	defer func() {
+		err = masterKeyFile.Close()
+		if err != nil {
+			logger.Warnf("failed to close file: %w", err)
+		}
+	}()
+
+	mkData := make([]byte, 1024)
```
troyronda: why 1024. also should be a const.

pkg/secretlock/local/masterkey_secret.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// NewMasterLock is responsible for encrypting/decrypting a master key
+// for the passed in `passphrase`, hash function `h`, `salt`.
+// The size of a master key passed to Encrypt() must match `h()`.Size() since the key used for
+// encryption/decryption is expanded from passphrase using HKDF.
+// The salt is optional and can be set to nil.
+// This implementation must not be used directly in Aries framework. It should be passed in
+// as the second argument of NewService(masterKeyReader io.Reader, secLock secretlock.Service) of this local package.
+func NewMasterLock(passphrase string, h func() hash.Hash, salt []byte) (secretlock.Service, error) {
```
troyronda: Shouldn't the function name reflect that this is a HKDF or passphrase?

pkg/secretlock/local/local_secret_lock.go
Baha-sk: a large enough array, it can be half this size as we won't have keys larger than 32 bytes, encrypted and encoded, it should around < 100 bytes

I'll make it a const

pkg/secretlock/local/local_secret_lock.go
Baha-sk: same here..  I'll make it a const

pkg/secretlock/local/local_secret_lock.go
Baha-sk: this implementation function has the same logic in `Lock` and `masterLockHKDF`, although both have their own instance of aead cipher.. the linter still complains the functions are duplicates.

pkg/secretlock/local/masterkey_secret.go
Baha-sk: technically this function is the constructor of `masterLockHKDF`

pkg/framework/context/context.go
```diff
@@ -28,6 +29,7 @@ type Provider struct {
 	storeProvider          storage.Provider
 	transientStoreProvider storage.Provider
 	kms                    legacykms.KMS
+	secretLock             secretlock.Service
```
llorllale: Q: why isn't the secret lock injected into the KMS?

pkg/secretlock/local/local_secret_lock.go
llorllale: RSA keys are definitely bigger. TODO I guess.

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+func NewService(masterKeyReader io.Reader, secLock secretlock.Service) (secretlock.Service, error) {
+	masterKeyData := make([]byte, 1024)
+
+	if masterKeyReader == nil {
+		return nil, fmt.Errorf("masterKeyReader is nil")
+	}
+
+	n, err := masterKeyReader.Read(masterKeyData)
+	if err != nil {
+		if err != io.EOF && err != io.ErrUnexpectedEOF {
```
llorllale: not sure why io.ErrUnexpectedEOF is not a true error condition

pkg/framework/context/context.go
Baha-sk: the new kms is not created yet

pkg/secretlock/local/local_secret_lock.go
Baha-sk: HKDF and AEAD don't work with RSA keys directly.. smaller keys are derived then used

pkg/secretlock/local/local_secret_lock.go
Baha-sk: will update

pkg/secretlock/local/local_secret_lock.go
Baha-sk: Apparently a reader that reads all its contents in one Read() call throws io.ErrUnexpectedEOF, for larger contents, multiple Read() calls are needed and a normal io.EOF is returned.

pkg/secretlock/local/local_secret_lock.go
Baha-sk: I'll probably move masterLockHKDF into a sub package

llorllale: TODO - extract common code into a function

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+			logger.Warnf("failed to close file: %w", err)
+		}
+	}()
+
+	mkData := make([]byte, 1024)
+
+	n, err := io.ReadFull(masterKeyFile, mkData)
+	if err != io.ErrUnexpectedEOF {
+		return nil, err
+	}
```
soluchok: Are you sure it works?
What if err is <nil> ? <nil> is not equal to io.ErrUnexpectedEOF
The function will return return nil, nil

Baha-sk: good catch.. missing `if err != nil` condition here..

pkg/secretlock/local/local_secret_lock.go
```diff
@@ -0,0 +1,176 @@
...
+
+	aead, err := cipher.NewGCM(cipherBlock)
+	if err != nil {
+		return nil, err
+	}
+
+	return aead, nil
+}
```
soluchok: Can be simplified or wrap an error
```
func createAESCipher(masterKey []byte) (cipher.AEAD, error) {
	cipherBlock, err := aes.NewCipher(masterKey)
	if err != nil {
		return nil, fmt.Errorf("new cipher",err)
	}

	return cipher.NewGCM(cipherBlock)
}
```

pkg/secretlock/local/local_secret_lock.go
Baha-sk: The issue with option is we have to set the masterKey as a `Lock` attribute which is not necessary since we only need the cipher

pkg/secretlock/local/local_secret_lock.go
Baha-sk: done

pkg/secretlock/local/local_secret_lock.go
Baha-sk: added `if err != nil` condition

pkg/secretlock/local/local_secret_lock.go
Baha-sk: so io.EOF and io.ErrUnexpectedEOF should not be treated as read errors.

In the case where the reader did not read anything (empty content), n is 0, the code checks this condition a few lines below.

pkg/secretlock/local/masterkey_secret.go
Baha-sk: master key lock is now in it's own sub package, so this function would be called as follows:
`hdkf.NewMasterLock()`

pkg/secretlock/local/local_secret_lock.go
Baha-sk: added const and set it to 512

pkg/secretlock/local/local_secret_lock.go
Baha-sk: Decrypt() functions are in different packages now.. removed the `nolint:dupl` comment

pkg/secretlock/local/local_secret_lock.go
Baha-sk: added const and set it to 512

PR #1177, opened 2020-02-04
feat: vdri and messaging REST operation to use controller commands

### LINKED ISSUE #1175: Move messaging & vdri REST operation logic to command controller

sudeshrshetty:
- removed existing `pkg/restapi` operations for `messaging` and `vdri`,
and added very thin layer of REST operation(`controller/restapi`) built
on top of messaging and vdri commands. All open API annotations will be
confined to `controller/restapi` package.
- closes #1175

TODO:
- migrate DID Exchange REST operation to controller
- rest handlers to be added from controller restapi.
- after all migration is complete, remove `pkg/restapi/...


troyronda:
@sudeshrshetty which files are merely moved or copied?

sudeshrshetty:
Things you can safely ignore (the ones copied from `pkg/restapi`)

   1.) All unit-tests, Same set of unit tests copied, REST version will use http request/response and commands will use io.writer/reader
2.) pkg/controller/mocks - Same as existing REST api mocks, once migration done then old ones will be deleted
3.)pkg/controller/webhook - Same as existing REST api webhook, once migration done then old ones will be deleted
4.) pkg/controller/restapi/errors - Same as existing REST api error handler, once migration done then old ones will be deleted

pkg/controller/restapi/operation/messaging/models.go
```diff
@@ -0,0 +1,84 @@
+/*
```
troyronda: should double check on whether or not anonymous embedded model structs should be pointers.

pkg/controller/restapi/operation/messaging/models.go
```diff
@@ -0,0 +1,84 @@
...
+	// in: body
+	Params *messaging.RegisterHTTPMsgSvcArgs
+}
+
+// RegisteredServicesResponse model
+//
+// This is used for returning list of registered service names
+//
+// swagger:response registeredServicesResponse
+type RegisteredServicesResponse struct {
```
troyronda: Do these structs actually get used somewhere?

sudeshrshetty: These structs wrap actual model objects so that swagger export them in spec. If we are ok to have swagger annotations on command models (inside command pkg), we can remove these. What you think?

troyronda: If these structs aren't used it would be good to unexport them and/or put them in a separate file with a build constraint (assuming that swagger can still generate from them).

pkg/controller/restapi/operation/messaging/models.go
sudeshrshetty: removed pointers, it didn't hurt spec.

pkg/controller/restapi/operation/messaging/models.go
sudeshrshetty: unexported them.

pkg/controller/restapi/operation/vdri/models.go
```diff
@@ -0,0 +1,37 @@
...
+	// in: path
+	*vdricommand.CreatePublicDIDArgs
+}
+
+// createPublicDIDResponse model
+//
+// This is used for returning public DID created
+//
+// swagger:response createPublicDIDResponse
+type createPublicDIDResponse struct {
```
troyronda: Why is this struct used?

pkg/controller/restapi/operation/vdri/models.go
```diff
@@ -0,0 +1,37 @@
...
+// createPublicDIDRequest model
+//
+// This is used for operation to create public DID
+//
+// swagger:parameters createPublicDID
+type createPublicDIDRequest struct { // nolint: unused,deadcode
+	// Params for creating public DID
+	//
+	// in: path
+	*vdricommand.CreatePublicDIDArgs
```
troyronda: Pointer?

pkg/controller/restapi/operation/vdri/models.go
sudeshrshetty: it is used in unit tests, so didn't require that tag.

pkg/controller/restapi/operation/vdri/models.go
sudeshrshetty: fixed it.

pkg/controller/restapi/operation/messaging/operation.go
```diff
@@ -0,0 +1,155 @@
...
+// New returns new common operations rest client instance
+func New(ctx provider, registrar operation.MessageHandler, notifier webhook.Notifier) (*Operation, error) {
+	msgcmd, err := messaging.New(ctx, registrar, notifier)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create messaging controller command: %w", err)
+	}
+
+	o := &Operation{command: msgcmd}
+
+	defer o.registerHandler()
```
troyronda: Why defer?

pkg/controller/restapi/operation/vdri/operation.go
```diff
@@ -0,0 +1,97 @@
...
+type Operation struct {
+	handlers []operation.Handler
+	command  *vdri.Command
+}
+
+// New returns new common operations rest client instance
+func New(ctx provider) (*Operation, error) {
+	o := &Operation{command: vdri.New(ctx)}
+
+	defer o.registerHandler()
```
troyronda: Why defer?

pkg/controller/restapi/operation/messaging/models.go
```diff
@@ -0,0 +1,84 @@
+/*
+ *
+ * Copyright SecureKey Technologies Inc. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * /
+ *
+ */
+
+package messaging
```
rolsonquadras: @sudeshrshetty @troyronda - just wondering if we need operation folder in between restapi and feature(messaging). These are already placed inside controller folder. ie. `pkg/controller/restapi/messaging` would make more sense.

troyronda: good point.

sudeshrshetty: Yes we need operation folder for now, rest api initializer, opts and other APIs from ‘pkg/restapi’ will move under ‘pkg/controller/restapi’.. we can revisit folder structure once I move all REST packages.

pkg/controller/restapi/operation/messaging/operation.go
sudeshrshetty: we don't need it anymore, will remove it.

pkg/controller/restapi/operation/vdri/operation.go
sudeshrshetty: we don't need it anymore, will remove it.

PR #1178, opened 2020-02-04
feat: VC With DID Resolution

### LINKED ISSUE #1155: VC With DID Resolution
Resolve Public Key from DID URL without unmarshaling VC twice.

kdimak:

#1155

To close #1155, there will be one more PR to support the use case with Linked Data Proofs.

### DIFF
--- a/pkg/doc/verifiable/common.go
+++ b/pkg/doc/verifiable/common.go
@@ -20,6 +20,8 @@ import (
 
 	"github.com/square/go-jose/v3"
 	"github.com/xeipuuv/gojsonschema"
+
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
 )
 
 // TODO https://github.com/square/go-jose/issues/263 support ES256K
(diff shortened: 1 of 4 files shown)


pkg/doc/verifiable/credential_jwt_proof_test.go
```diff
@@ -222,7 +226,59 @@ func createKeyFetcher(t *testing.T) func(issuerID string, keyID string) (interfa
...
+		ResolveValue: didDoc,
+	}
+
+	resolver := NewDIDKeyResolver(v)
+	require.NotNil(t, resolver)
+
+	return func(issuerID, keyID string) (i interface{}, err error) {
+		pKey, keyErr := resolver.resolvePublicKey(issuerID, keyID)
+		if keyErr != nil {
+			return nil, err
```
sandrask: return nil, keyErr

pkg/doc/verifiable/common.go
```diff
@@ -59,6 +61,47 @@ func SingleKey(pubKey interface{}) PublicKeyFetcher {
...
+	return &DIDKeyResolver{vdriRegistry: vdriRegistry}
+}
+
+func (r *DIDKeyResolver) resolvePublicKey(issuerDID, keyID string) (interface{}, error) {
+	doc, err := r.vdriRegistry.Resolve(issuerDID)
+	if err != nil {
+		return nil, fmt.Errorf("resolve DID %s: %w", issuerDID, err)
+	}
+
+	if len(doc.PublicKey) == 0 {
```
sandrask: You probably don't need this check since this should be enforced during did creation. Anyway, the result would be not found error.

sandrask: Why do we default to first key?

pkg/doc/verifiable/credential_jwt_proof_test.go
kdimak: ah! thanks!

pkg/doc/verifiable/common.go
kdimak: I see. Well, maybe we should return the first key if it's the only key and raise an error otherwise? @sandrask

troyronda: No - there are keys with different purposes in a DID document. We need to be able to select the appropriate key. @kdimak  @sandrask

sandrask: For now I would keep it simple with no defaults. User has to provide key id and you return not found if the key is not in the document. Later on we can add functions/implementations that will select appropriate type of key.

kdimak: OK, got it! I will change this soon.

pkg/doc/verifiable/credential_jwt_proof_test.go
kdimak: fixed

pkg/doc/verifiable/common.go
kdimak: fixed

PR #1180, opened 2020-02-04
chore: Expose mocks files

### LINKED ISSUE #1179: Expose mocks files

fqutishat:
closes #1179


rolsonquadras:
The mocks were moved inside the internal folder so that consumers wouldn't use it in actual code. As @Moopli suggested, may be we can move these mocks to `_test.go` file, so that only consumer test code can access it.

sandrask:
Now that we start using aries components, we should expose mocks for them. Maybe we don't need to expose all of them but certainly we need mocks for core services such as VDRI, KMS etc. Otherwise you have to implement mocks in every single project. We can definitely discuss which approach to take.

troyronda:
@fqutishat @sandrask @rolsonquadras do we really need to expose every mock? maybe there is a smaller subset?

Issue #1070, opened 2020-01-15
Use REST API Operations (as generic channel/event interface)


troyronda:
The REST API currently includes "Operation" wrappers for protocols. Example: [DID Exchange](https://github.com/hyperledger/aries-framework-go/tree/590aaff0b52a5c24754655f12aa9116f4d5bdc9d/pkg/restapi/operation/didexchange)

We want to share these Operations between bindings.
- These Operations should support an event API where requests and responses can be passed over a generic Go channel. See [Channel-based Event API](https://github.com/hyperledger/aries-framework-go/blob/90804c45b6245d19fb4805b40ebad17d46debe85/cmd/aries-js-worker/main.go#L56).  Side-note: This channel-based event API should be moved into a package to be re-usable by other bindings too.
- The REST API should use the Operation to HTTP request / response.
- The JS Worker API should convert the Go channel events to JS messages (sent via the Worker).
- (future: presumably a "C" language binding could also leverage the event interface as well.)

Notes:
- Current calls from the REST API (outside Operation) directly to protocol services also need to be moved to this new layer.
- Connection store access should also be split in the lower layer Go Channel binding.

Part of #1064

troyronda:
@sudeshrshetty

PR #1181, opened 2020-02-04
feat: messaging controller commands

sudeshrshetty:
- created controller command packages which contains business logic and
models needed for execution of all controller API commands, including
REST endpoints and upcoming JS Worker commands (#1070).

- added command errors, which can be used by REST and JS Worker(WASM)
for handling/sending errors.

- Part of #1175


pkg/controller/command/errors.go
```diff
@@ -0,0 +1,82 @@
...
+// Note: recommended to use [0-9]*000 pattern for any new entries
+// Example: 2000, 3000, 4000 ...... 25000
+type Group int32
+
+const (
+	// Common error group for general command errors
+	Common Group = 1000
+
+	// DIDExchange error group for DID exchange protocol command errors
+	DIDExchange Group = 2000
```
troyronda: We shouldn't be referencing specific protocols.

sudeshrshetty: these are specific to command types, I will fix the comment.

sudeshrshetty: fixed

troyronda: My point is actually that DIDExchange nor any other specific protocol/message should not appear in this file.

troyronda: I guess this is existing code, but I don't really understand why we didn't use a namespace string defined by individual command packages.

Issue #989, opened 2019-12-16
DIDComm Router/Mediator - REST API Binding


rolsonquadras:
DIDComm Router/Mediator - REST API Binding

PR #1193, opened 2020-02-06
feat: Route - Controller Command

rolsonquadras:
- Add command method for Route registration for different bindings
- The command uses route client

part of #989 

### DIFF
--- a/pkg/controller/command/route/command.go
+++ b/pkg/controller/command/route/command.go
@@ -1,0 +1,86 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package route
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/hyperledger/aries-framework-go/pkg/client/route"
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/controller/command"
+)
+
+var logger = log.New("aries-framework/command/route")
+
+// Error codes
+const (
+	// InvalidRequestErrorCode for invalid requests
+	InvalidRequestErrorCode = command.Code(iota + command.ROUTE)
+
+	// ResponseWriteErrorCode for response write error
+	RegisterMissingConnIDCode
+
+	// RegisterRouterErrorCode for register router error
+	RegisterRouterErrorCode
+)
+
+// provider contains dependencies for the route protocol and is typically created by using aries.Context().
+type provider interface {
+	Service(id string) (interface{}, error)
+}
+
+// Command contains command operations provided by route controller.
+type Command struct {
+	routeClient *route.Client
+}
+
+// New returns new route controller command instance.
+func New(ctx provider) (*Command, error) {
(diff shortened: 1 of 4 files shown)


pkg/controller/command/route/command.go
```diff
@@ -0,0 +1,92 @@
...
+
+var logger = log.New("aries-framework/command/route")
+
+// Error codes
+const (
+	// InvalidRequestErrorCode for invalid requests
+	InvalidRequestErrorCode = command.Code(iota + command.ROUTE)
+
+	// ResponseWriteErrorCode for response write error
+	ResponseWriteErrorCode
```
sudeshrshetty: You shouldn't throw error response write, it **may** go in infinite loop. Since error handling again tries to write it to on same writer/response.

pkg/controller/command/route/command.go
```diff
@@ -0,0 +1,92 @@
...
+	return writeResponse(rw, nil)
+}
+
+// writeResponse writes interface value to response
+// TODO https://github.com/hyperledger/aries-framework-go/issues/1176 - Move these to common
+//  along with same instances in other packages
+func writeResponse(rw io.Writer, v interface{}) command.Error {
+	if err := json.NewEncoder(rw).Encode(v); err != nil {
+		logger.Errorf("Unable to send error response, %s", err)
+		return command.NewExecuteError(ResponseWriteErrorCode, err)
```
sudeshrshetty: Shouldn't return error.

pkg/controller/command/route/command.go
```diff
@@ -0,0 +1,92 @@
...
+	err = o.routeClient.Register(request.ConnectionID)
+	if err != nil {
+		return command.NewExecuteError(RegisterRouterErrorCode, err)
+	}
+
+	return writeResponse(rw, nil)
+}
+
+// writeResponse writes interface value to response
+// TODO https://github.com/hyperledger/aries-framework-go/issues/1176 - Move these to common
```
sudeshrshetty: Tried to move this part of separate packages, looked ugly :)
this function is nothing but a json encoder logic. For the sake of improving code coverage this one is moved to separate function in other commands

pkg/controller/command/route/command.go
rolsonquadras: [Removed](https://github.com/hyperledger/aries-framework-go/compare/e3d2bf8f0eff2b46016f1ba077db6cb3aefac866..2f1b5db8a74a7b7659c94b4d4c346b41e71d5c51#diff-d08f92ab0cfb3a36ef618e710ffc3fedL27)

pkg/controller/command/route/command.go
rolsonquadras: [Refactored](https://github.com/hyperledger/aries-framework-go/compare/e3d2bf8f0eff2b46016f1ba077db6cb3aefac866..2f1b5db8a74a7b7659c94b4d4c346b41e71d5c51#diff-d08f92ab0cfb3a36ef618e710ffc3fedL88)

pkg/controller/command/route/command.go
rolsonquadras: Removed this TODO.

sudeshrshetty (approved):

troyronda (approved):

PR #1209, opened 2020-02-08
feat: command handlers for JS Worker

### LINKED ISSUE #1208: controller command handlers for aries-js-worker
controller  to provide command handler list to be used by aries-js-worker input handlers.

sudeshrshetty:
- controller package to expose list command handlers it provides so that
it can be used to build command list in `aries-js-worker`
- controller rest api also refactored to provide similar features.

controller.GetRESTHandlers() to be used by aries-agent-rest
controller.GetCommandHandlers() to be used by aries-js-worker

- also refactored controller rest and command packages to look symmetric
- closes #1208


pkg/controller/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	defaultLabel string
+	autoAccept   bool
+	msgHandler   command.MessageHandler
+}
+
+// Opt represents a controller option.
+type Opt func(opts *allOpts)
+
+// WithWebhookURLs is an option for setting up a webhook dispatcher which will notify clients of events
+func WithWebhookURLs(webhookURLs ...string) Opt {
```
troyronda: How does this work for the JavaScript (it should get events directly rather than Webhooks.

pkg/controller/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	vcmd := vdricmd.New(ctx)
+
+	// messaging command operation
+	msgcmd, err := messagingcmd.New(ctx, cmdOpts.msgHandler, webhook.NewHTTPNotifier(cmdOpts.webhookURLs))
+	if err != nil {
+		return nil, err
+	}
+
+	var allHandlers []command.Handler
+	allHandlers = append(allHandlers, didexcmd.GetHandlers()...)
```
troyronda: Also route right? @rolsonquadras  @sudeshrshetty

sudeshrshetty: Yes, I rebased and pushing it shortly after tests.

pkg/controller/controller.go
sudeshrshetty: This one is used in case of http notifier, JS version will use callback func version.

aries-agent-rest gets webhook from cli, for javascript it should be in wasm initialization params, by default I am thinking of javascript callback.

troyronda: Take a look at generators.
https://modernweb.com/replacing-callbacks-with-es6-generators/

troyronda: async generators: http://thecodebarbarian.com/async-generator-functions-in-javascript.html

pkg/controller/controller.go
sudeshrshetty: done

PR #1220, opened 2020-02-10
fix: Ignore Transport Return Route decorator for forward message

### LINKED ISSUE #1218: Ignore Transport Return Route decorator for forward message
Currently, in case of forward message the framework adds transport route decorator in the destination message. The decorator shouldn't be added in this case.

rolsonquadras:
Closes #1218 


rolsonquadras:
> tests coming later?

[Done](https://github.com/hyperledger/aries-framework-go/compare/3fd5514c73c6879d90a2164b6c3e17df4c0817ca..5b0fc5f2b3645fe34a18a471a3574cc7c19bd9b7).

PR #1223, opened 2020-02-10
fix: Invoke websocket.Ping() function to keep the connection alive

### LINKED ISSUE #1214: Websocket is timing out
After 1-2 minutes the WebSocket from an Agent to the Router seems to timeout. 

[aries-framework/ws] 2020/02/09 13:16:37 UTC - ws.(*connPool).listener -> ERROR Error reading request message: failed to get reader: failed to read frame header: EOF
 [aries-framework/ws] 2020/02/09 13:16:37 UTC - ws.(*connPool).close -> ERROR connection close error

I need to retrieve a new invitation from the agent to trigger it to restart the WebSocket. Then you get:
 [aries-framework/ws] 2020/02/09 13:20:08 UTC - ws.acceptRecipient -> INFO failed to ping to the connection for key=failed to ping: failed to write control frame opPing: websocket closed: failed to read frame header: EOF err=%!v(MISSING)

And then the agent starts working again.

If I periodically send a basic message, the connection stays active.

Once I let it idle for 1 minute, the connection drops again.

rolsonquadras:
The web server, load balancer, network routers between the client and server closes the TCP keepalives connection. Added a new function to invoke Ping requests frequently to keep the connection alive

closes #1214 

### DIFF
--- a/pkg/didcomm/transport/ws/pool.go
+++ b/pkg/didcomm/transport/ws/pool.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"sync"
+	"time"
 
 	"github.com/btcsuite/btcutil/base58"
 	"nhooyr.io/websocket"
@@ -16,6 +17,11 @@ import (
 	commtransport "github.com/hyperledger/aries-framework-go/pkg/didcomm/common/transport"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+)
+
+const (
+	// TODO configure ping request frequency
+	pingFrequency = 30 * time.Second
 )
 
 type connPool struct {
@@ -63,10 +69,12 @@ func (d *connPool) remove(verKey string) {
 	delete(d.connMap, verKey)
 }
 
-func (d *connPool) listener(conn *websocket.Conn) {
+func (d *connPool) listener(conn *websocket.Conn, outbound bool) {
 	verKeys := []string{}
 
 	defer d.close(conn, verKeys)
+
+	go keepConnAlive(conn, outbound, pingFrequency)
 
 	for {
 		_, message, err := conn.Read(context.Background())
--- a/pkg/didcomm/transport/ws/upgrade_js_wasm.go
+++ b/pkg/didcomm/transport/ws/upgrade_js_wasm.go
@@ -9,6 +9,7 @@ package ws
 import (
 	"errors"
 	"net/http"
+	"time"
 
 	"nhooyr.io/websocket"
 )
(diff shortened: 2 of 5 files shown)


pkg/didcomm/transport/ws/pool.go
```diff
@@ -18,6 +19,11 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
 )
 
+const (
+	// TODO configure ping request frequency
+	pingFrequency = 20 * time.Second
```
troyronda: Longer frequency didn’t work?

pkg/didcomm/transport/ws/upgrade_js_wasm.go
```diff
@@ -30,3 +31,7 @@ func acceptRecipient(pool *connPool, keys []string) bool {
 
 	return false
 }
+
+func keepConnAlive(conn *websocket.Conn, outbound bool, frequency time.Duration) {
+	// TODO make sure connection is alive (conn.Ping() doesn't work with JS/WASM build)
```
troyronda: So the WASM won’t work?

pkg/didcomm/transport/ws/upgrade_srv.go
```diff
@@ -42,3 +43,20 @@ func acceptRecipient(pool *connPool, keys []string) bool {
...
+func keepConnAlive(conn *websocket.Conn, outbound bool, frequency time.Duration) {
+	if outbound {
+		for {
+			if err := conn.Ping(context.Background()); err != nil {
+				logger.Errorf("websocket ping error : %v", err)
+
+				break
+			}
+
+			time.Sleep(frequency)
```
troyronda: Shouldn’t we be using a timer/select pattern?

rolsonquadras: [Updated](https://github.com/hyperledger/aries-framework-go/compare/ab242869e92bb67c3cd0fd5bfbf474d1dd3db97a..873d9d4b0278aae40c2dad5a609270145a0332c1#diff-e3dec0159ab0c461e6c5bd749080659dR52) to for-range pattern.

pkg/didcomm/transport/ws/pool.go
rolsonquadras: We should be able to configure this frequency (TODO).

pkg/didcomm/transport/ws/upgrade_js_wasm.go
rolsonquadras: I'm wondering if this needs to be handled at the proxy server/load balancer.  Lots of browser based applications use websocket and don't think they would need to reconnect every min. I'll do more reading on this.

troyronda: Ok

PR #1233, opened 2020-02-11
fix: aries-js-worker - camelCase

llorllale:
* Exposed as a constructor that accepts options
* Refactored to use camelCase
* Unexported the worker and the pending task map

### DIFF
--- a/cmd/aries-js-worker/src/aries.js
+++ b/cmd/aries-js-worker/src/aries.js
@@ -25,12 +25,22 @@ __webpack_public_path__ = __publicPath()
 
 const { _getWorker } = require("worker_loader")
 
+// TODO synchronize access on this map?
+const PENDING = new Map()
+const WORKER = _getWorker(PENDING)
+
+// TODO Aries should be a singleton
+
+// TODO implement Aries options
+
 // TODO not all browsers support private members of classes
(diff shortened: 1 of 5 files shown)


cmd/aries-js-worker/vue-aries-framework-go/src/main.js
```diff
@@ -7,6 +7,9 @@ SPDX-License-Identifier: Apache-2.0
 import Vue from 'vue'
 import App from './App.vue'
 
+import { Aries } from "@hyperledger/aries-framework-go"
+Object.defineProperty(Vue.prototype, "$aries", { value: new Aries({}) })
```
troyronda: Might need to come back to this and see if this is the idiomatic way to do this. I had thought to use props.

cmd/aries-js-worker/src/aries.js
```diff
@@ -25,12 +25,22 @@ __webpack_public_path__ = __publicPath()
 
 const { _getWorker } = require("worker_loader")
 
+// TODO synchronize access on this map?
+const PENDING = new Map()
+const WORKER = _getWorker(PENDING)
+
+// TODO Aries should be a singleton
+
+// TODO implement Aries options
```
rolsonquadras: Do we have any issues for these TODOs ?

PR #1240, opened 2020-02-12
feat : Aries JS Worker Start/Stop and Notifier

sudeshrshetty:
- Aries.Start(opts) can be used to start aries in JS worker
sample :
`{"agent-default-label":"dem-js-agent","http-resolver-url":"","auto-accept":true,"outbound-transport":["ws","http"],"transport-return-route":"all","notifier-func-name":"sample-topic"}`
- Aries.Stop() can be used to stop aries
- Command controller can only be called between start and stop
- Register notifier using `Aries.waitForNotification("<topic-name>")`, make sure you use same `notifier-func-name` in aries opts.

### DIFF
--- a/cmd/aries-js-worker/main.go
+++ b/cmd/aries-js-worker/main.go
@@ -12,12 +12,18 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"syscall/js"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/hyperledger/aries-framework-go/pkg/controller"
 	cmdctrl "github.com/hyperledger/aries-framework-go/pkg/controller/command"
+	"github.com/hyperledger/aries-framework-go/pkg/controller/webhook"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/messaging/msghandler"
+	arieshttp "github.com/hyperledger/aries-framework-go/pkg/didcomm/transport/http"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport/ws"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/aries-js-worker/src/aries.js
```diff
@@ -34,6 +34,28 @@ export const Aries = new function() {
...
     }
 
+    this.Start = async function(opts) {
+        return Aries._invoke("aries", "Start", opts, "timeout while starting aries")
+    }
+
+
+    this.waitForNotification = async function(topic) {
+        return new Promise((resolve, reject) => {
+            const timer = setTimeout(_ => reject(new Error("timed out waiting for notification")), 10000)
```
troyronda: what happens after the timeout?

cmd/aries-js-worker/src/aries.js
```diff
@@ -126,7 +148,8 @@ export const Aries = new function() {
     }
 
     this.getWorker = () => {
-        return _getWorker(this._pending)
+        console.log(" >> get worker called !!!", this._pending)
```
troyronda: what's this console log for?

sudeshrshetty: removing it

cmd/aries-js-worker/src/aries.js
sudeshrshetty: current yield returns with no incoming message. Generator will be called again.

check `startNotifier` in index.html

cmd/aries-js-worker/main.go
```diff
@@ -230,3 +191,159 @@ func defaultHandlers() map[string]map[string]func(*command) *result {
...
+	return options, nil
+}
+
+// jsNotifier notifies incoming events once registered
+type jsNotifier struct {
+	topic string
+}
+
+// Notify is mock implementation of webhook notifier Notify()
+func (n *jsNotifier) Notify(topic string, message []byte) error {
```
troyronda: why is the topic argument not sent to Javascript?

cmd/aries-js-worker/src/worker-loader-node.js
```diff
@@ -10,7 +10,7 @@ import wasmJS from "./wasm_exec.js"
 import wasm from "./aries-js-worker.wasm.gz"
 import workerJS from "./worker-impl-node"
 
-export function _getWorker(pending) {
+export function _getWorker(pending, notifications) {
```
troyronda: This needs to be completed for Node.

sudeshrshetty: will fix it in upcoming PR

cmd/aries-js-worker/main.go
sudeshrshetty: Notifier follows REST pattern, it is initialized with URLs instead of Notifier implementation. I added new option in this PR `WithNotifier()` to support commands. REST doesn't have selective topic subscription feature, and also DIDComm msg handling fails when there isn't any webhook URL available. We may need to revisit webhook logic.

I can pass array of topic names from Javascript to notify by topic(protocol)

PR #1243, opened 2020-02-12
feat: Create a new Storage type with Delete() function

### LINKED ISSUE #1147: Create a new Storage type with Delete() function
part of #985

Baha-sk:
In order to support rotating keys from the storage (or deleting data in general),
there should be a storage service with Delete() function available

closes #1147

### DIFF
--- a/pkg/storage/jsindexeddb/jsindexeddb.go
+++ b/pkg/storage/jsindexeddb/jsindexeddb.go
@@ -127,6 +127,22 @@ func (s *store) Iterator(start, limit string) storage.StoreIterator {
 	return newIterator(batch, err)
 }
 
+// Delete will delete record with k key
+func (s *store) Delete(k string) error {
+	if k == "" {
+		return errors.New("key is mandatory")
+	}
+
+	req := s.db.Call("transaction", s.name, "readwrite").Call("objectStore", s.name).Call("delete", k)
+
+	_, err := getResult(req)
+	if err != nil {
+		return fmt.Errorf("failed to delete data with key: %s - error: %w", k, err)
+	}
+
+	return nil
+}
+
 type iterator struct {
 	batch *js.Value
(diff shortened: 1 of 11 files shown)


pkg/storage/jsindexeddb/jsindexeddb.go
```diff
@@ -44,6 +44,16 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
 	return &store{name: name, db: db}, nil
 }
 
+// OpenStoreWithDelete open store with Delete() capability
+func (p *Provider) OpenStoreWithDelete(name string) (storage.StoreWithDelete, error) {
```
fqutishat: why we need open store with delete

Baha-sk: I'll remove it..

Baha-sk:
> Implementation looks good, though I'd also suggest to add Delete to the store interface instead of creating a second interface.
>
yup, that's what I'm currently doing, I'll push another patch in a bit..
 
> I'm also thinking we might someday want the iterator to allow deleting the entry it points to, but that's an optimization question for the future.
yeah, this can be added in the future (if needed)

pkg/storage/jsindexeddb/jsindexeddb.go
llorllale: Keys need to be rotated even in the browser. Why would we remove this capability?

Baha-sk: a store should support Delete() function by default, there's no need to create a `different` store type with Delete() function.

a new patch has been pushed to amend the existing `Store`.

llorllale: Ok - I understand now that the comment was about this particular method

Issue #1157, opened 2020-01-31
Introduce - the protocol should not depend on thread decorator


soluchok:
Currently, the protocol depends on `threadID`. What does it mean? When protocol communicates with three participants it uses the same `threadID` per communication. e.g Alice wants to introduce Carol to Bob. Messages in the protocol will look like:
* Alice sends message to Bob `{"@id":"111","~thread":{"thid":"000"}}`
* Alice sends message to Carol `{"@id":"222","~thread":{"thid":"000"}}`

The example above clearly says that Alice-Bob and Alice-Carol have the same threadID `000`.It was done in that way to identify the context in the protocol. The protocol relies on that thread. It is how we understand that the introduction is related to three persons. Otherwise, there is no way to understand that participants are in the same context.
According to our RFCs which describe JSON-based communication there should not be a message with the same ID (messageID is unique). Also, RFCs say `threadID` is the ID of the first message. Which makes impossible to have two communication with the same `threadID`. It means that the introduce protocol does not follow logic like it is described in RFC.

Let's imagine the use case when Alice wants to introduce Carol to Bob. Alice sends two proposal messages(simultaneously) through the introduce `Client`. 
Messages might look like:
* Alice sends message to Bob `{"@id":"111"}`
* Alice sends message to Carol `{"@id":"222"}`

According to the example above threads would be `~thread":{"thid":"111"}` and `~thread":{"thid":"222"}`. Currently, our HandleOutbound function allows us to send only one message. This means that we will send two proposals independently. The protocol will not know that messages are related to each other(same context). It is an issue.

Possible solutions:
* allow HandleOutbound function to send more than one message `HandleOutbound(msgs ..DidMsgWithDestination)`. (a bit complicated solution)
* add an additional field to the Proposal message e.g `ContextID`

An example of how Proposal might look like for the internal usage.
```
type Proposal struct {
	....
       // ContextID is an internal field that allows the protocol to identify and connect independent 
       // threads into the same context. The field will be ignored in the JSON message.
	ContextID     string            `json:"-"`
}
```

NOTE: Having a `contextID`, allows us to create a connection map e.g `map[thid]=contextID`. By `contextID` we can get a list of the threads e.g `map[contextID]=[]thids`.

troyronda:
@soluchok Adding correlatable IDs to messages is not a desirable solution. i.e, no contextID.

troyronda:
[...]

PR #1252, opened 2020-02-13
refactor: Messenger and DIDCommMsgMap improvements (part 1)

soluchok:
* added ParentThreadID() function
*  added Clone() function

Part of #1157 

### DIFF
--- a/pkg/didcomm/common/service/did_comm_msg.go
+++ b/pkg/didcomm/common/service/did_comm_msg.go
@@ -19,11 +19,12 @@ import (
 )
 
 const (
-	jsonID       = "@id"
-	jsonType     = "@type"
-	jsonThread   = "~thread"
-	jsonThreadID = "thid"
-	jsonMetadata = "_internal_metadata"
+	jsonID             = "@id"
+	jsonType           = "@type"
+	jsonThread         = "~thread"
+	jsonThreadID       = "thid"
+	jsonParentThreadID = "pthid"
+	jsonMetadata       = "_internal_metadata"
 )
 
 // Metadata may contain additional payload for the protocol. It might be populated by the client/protocol
@@ -60,6 +61,9 @@ func ParseDIDCommMsgMap(payload []byte) (DIDCommMsgMap, error) {
 		return nil, fmt.Errorf("invalid payload data format: %w", err)
 	}
 
+	// sets empty metadata
+	msg[jsonMetadata] = map[string]interface{}{}
+
 	return msg, nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/didcomm/common/service/did_comm_msg_test.go
```diff
@@ -104,6 +100,66 @@ func TestDIDCommMsgMap_Type(t *testing.T) {
 	}
 }
 
+func TestDIDCommMsgMap_Clone(t *testing.T) {
+	tests := []struct {
+		name     string
```
Baha-sk: not really important but `name` is not used, unless you use it in a test run like: t.Run(`tc.name`, func(t *testng.T){... })

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -132,6 +135,21 @@ func (m DIDCommMsgMap) Type() string {
 	return res
 }
 
+// ParentThreadID returns the message parentT hreadID
```
troyronda: typo "parentT hreadID"

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -107,12 +108,14 @@ func (m DIDCommMsgMap) ThreadID() (string, error) {
 // Metadata returns message metadata
 func (m DIDCommMsgMap) Metadata() map[string]interface{} {
 	if m[jsonMetadata] == nil {
-		return nil
+		m[jsonMetadata] = map[string]interface{}{}
```
troyronda: seems odd to mutate during an accessor.

soluchok: Yeah, but I was trying to avoid another method e.g SetMetadata. I do not think it is good to set metadata in the protocol directly e.g `didmsg[jsonmetadatakey] =map[]interfase{...}`. So I decided that It would be better if didCommmsgMap will always return the map. Actually, there are a few possible ways how to solve this. 
1. Add additional function e.g. SetMetadata
2. Always create an empty map when we use functions like ParseDIDCommMsgMap or NewDIDCommMsgMap
3. The protocol sets metadata to the message (bad idea)

Let me know what do you like more. Suggestions are welcome.

pkg/didcomm/common/service/did_comm_msg_test.go
soluchok: I will fix it

pkg/didcomm/common/service/did_comm_msg.go
troyronda: I guess I had been thinking about constructor cases (i.e., 2).

troyronda: so updating to 2 then?

soluchok: sure, I will do that

pkg/didcomm/common/service/did_comm_msg_test.go
soluchok: Done!

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Done!

pkg/didcomm/common/service/did_comm_msg.go
soluchok: Done!

PR #1253, opened 2020-02-13
refactor: Introduce - removes the improper dependency on the thread (part 2)

soluchok:
* service was fully rewritten

Part of #1157 


pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -132,6 +135,21 @@ func (m DIDCommMsgMap) Type() string {
 	return res
 }
 
+// ParentThreadID returns the message parentT hreadID
+func (m DIDCommMsgMap) ParentThreadID() string {
```
Baha-sk: I see code duplication with [1252](https://github.com/hyperledger/aries-framework-go/pull/1252)

can this function be extracted into a helper function?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -221,65 +155,74 @@ func (s *Service) startInternalListener() {
...
+
 	thID, err := msg.ThreadID()
+	if errors.Is(err, service.ErrThreadIDNotFound) {
+		msg.(service.DIDCommMsgMap)["@id"] = uuid.New().String()
+		return msg.(service.DIDCommMsgMap)["@id"].(string)
+	}
+
 	if err != nil {
-		return nil, err
+		panic(err)
```
troyronda: Are you saying this is an impossible case?

soluchok: It can happen only when the message is invalid. The invalid message is a message with missing id. But it will be filtered by the messenger on the lower level. So, it will never happen. But, I also think that the `ThreadID()` function should not return an error. I would like to rewrite it. Instead of an error, an empty value will be returned. It will reduce this error handling and improve readability.

pkg/didcomm/common/service/did_comm_msg.go
soluchok: It is duplicated because this PR contains code from the previous PR. You will not see these changes after the first one will be merged and this one rebased.

pkg/didcomm/messenger/messenger.go
```diff
@@ -138,6 +138,8 @@ func (m *Messenger) Send(msg service.DIDCommMsgMap, myDID, theirDID string) erro
 		return fmt.Errorf("save metadata: %w", err)
 	}
 
+	delete(msg, jsonThread)
```
troyronda: Why?

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,81 @@
...
+	metaInvitation   = "invitation"
+	metaRecipients   = "recipients"
+)
+
+// Opt describes option signature for the Continue function
+type Opt func(m map[string]interface{})
+
+// WithInvitation is used when introducee wants to provide invitation.
+// NOTE: Introducee can provide invitation only after receiving ProposalMsgType
+// USAGE: event.Continue(WithInvitation(inv))
```
troyronda: How would it be called from an async client like REST or WASM that cannot call event.Continue?

pkg/didcomm/messenger/messenger.go
soluchok: The thread should not be in the message. The messenger is responsible for that. The function Send starts a new thread. Which means that the message cannot contain a custom thread.

pkg/didcomm/protocol/introduce/metadata.go
troyronda: ie, there needs to be a method on Introduce that can continue the state machine based on a serialized identifier (presumably the thread ID). we cannot rely on the event object instance as the continuation might be called on the REST/WASM api, after restart or even on a different cluster node.

pkg/didcomm/protocol/introduce/service_test.go
```diff
@@ -32,2640 +32,1293 @@ import (
...
-	provider.EXPECT().Service(didexchange.DIDExchange).Return(didService, nil)
-
-	svc, err := introduce.New(provider)
-	require.NoError(t, err)
-
-	require.NoError(t, svc.Stop())
-	require.EqualError(t, svc.Stop(), "server was already stopped")
-}
-
-func TestService_Name(t *testing.T) {
```
troyronda: i see the tests are mainly deleted. are they not needed anymore?

pkg/client/introduce/client.go
```diff
@@ -1,226 +0,0 @@
-/*
```
rolsonquadras: Don't we need this file ?

troyronda: I see it coming back in the next PR.

rolsonquadras:
@soluchok @troyronda @llorllale @fqutishat @Baha-sk May be these refactor need to be done in steps (may be API by API - supporting new/old simultaneously or something). I'm but skeptical of deleting the client and BDD test files as a whole and then adding it. It's hard to see what has changed. This refactoring looks similar to one we are having with KMS.

troyronda:
@soluchok maybe it would be better to disable files in this PR with a build constraint rather than deleting in this one and re-creating in the next one?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -478,7 +398,7 @@ func stateFromName(name string) state {
 // canTriggerActionEvents checks if the incoming message can trigger an action event
 func canTriggerActionEvents(msg service.DIDCommMsg) bool {
 	switch msg.Type() {
-	case ProposalMsgType, ResponseMsgType, RequestMsgType:
+	case ProposalMsgType, RequestMsgType:
```
rolsonquadras: Is this change part of refactor ? This looks like logic change.

pkg/didcomm/protocol/introduce/service_test.go
soluchok: tests I have deleted are useless. Some functions were removed from the service. e.g Stop.

pkg/client/introduce/client.go
soluchok: I cannot use the previous client. The service was changed. Logic now is different. New The client is implemented e in this PR  https://github.com/hyperledger/aries-framework-go/pull/1254

pkg/didcomm/protocol/introduce/service.go
soluchok: The logic was changed. It looks like ResponseMsgType is redundant and no needed. It only makes service and usage complicated. I don't see any benefits of this action. @llorllale Could you please confirm that this action is redundant?

test/bdd/features/introduce.feature
```diff
@@ -1,294 +0,0 @@
-
```
troyronda: Why delete?

pkg/internal/gomocks/client/introduce/mocks.go
```diff
@@ -1,63 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
troyronda: Why delete

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,81 @@
...
+
+package introduce
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
+)
+
+const (
+	metaContextID    = "context_id"
```
troyronda: ContextID is a fairly generic term (and the term is already used in other situations). Perhaps rename or at least comment on its usage (in the code).

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -490,113 +410,194 @@ func isNoOp(s state) bool {
...
+	inv := &didexchange.Invitation{}
 
-	for !isNoOp(current) {
-		next, err := s.execute(current, msg)
-		if err != nil {
-			return fmt.Errorf("execute: %w", err)
+	switch v := msg.Metadata()[metaInvitation].(type) {
+	case service.DIDCommMsgMap:
+		if err := v.Decode(inv); err != nil {
+			// should never happen, otherwise, the protocol logic is broken
```
troyronda: OR the data model changed between versions and we didn't handle versioning properly.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -490,113 +410,194 @@ func isNoOp(s state) bool {
...
+			return err
+		}
+	}
 
 	return nil
 }
 
-func (s *Service) handle(msg *metaData) error {
-	current := stateFromName(msg.StateName)
+func contextInvitation(msg service.DIDCommMsg) *didexchange.Invitation {
```
troyronda: where is this function called?

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Done! I have exposed the `Continue` function for that.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -415,35 +400,42 @@ func nextState(msg service.DIDCommMsg, rec *record, outbound bool) (state, error
...
+		return fmt.Errorf("marshal transited: %w", err)
 	}
 
-	var r *record
-	if err := json.Unmarshal(src, &r); err != nil {
-		return nil, err
+	return s.store.Put(transitedKey+id, src)
+}
+
+func (s *Service) getTransited(id string) (*Transited, error) {
```
troyronda: I am have trouble groking  the term `transited`

pkg/didcomm/protocol/introduce/metadata.go
troyronda: I believe this is the identifier of a particular instance (or execution) of the state machine?

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,81 @@
...
+	metaSkipProposal = "skip_proposal"
+	metaInvitation   = "invitation"
+	metaRecipients   = "recipients"
```
troyronda: When we start putting more real data in the messenger's storage space, I start wondering about data lifecycle ... and if we should be namespacing according to the protocol (aka in the protocol's storage space).

test/bdd/features/introduce.feature
soluchok: There's no way that it will work. I cannot ignore this file as I did with go files.

pkg/didcomm/protocol/introduce/service.go
llorllale: @soluchok I think so. However, is this the only state for which the framework is skipping action events (including did-exchange)? If it is, it will stick out like a sore thumb

pkg/didcomm/protocol/introduce/service.go
soluchok: Changed it to `TransitionalPayload`

pkg/didcomm/protocol/introduce/service.go
soluchok: Done! I have removed panic

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Added prefix!

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Yes, it is. Context ID is an identifier witch helps protocol to determine that few participants are `on the same page`.

pkg/didcomm/protocol/introduce/service.go
soluchok: Not sure about did-exchange, but there is no reason to execute the Continue function. Currently, the Continue function is executed when:
* need to accept the proposal (introducee may provide invitation)
* need to accept the request (introducer should provide invitation or send a proposal to another introducee)

It doesn't make sense to reject or accept the `response`.

pkg/internal/gomocks/client/introduce/mocks.go
soluchok: we do not use it anymore, I can't even generate it.

pkg/didcomm/protocol/introduce/metadata.go
troyronda: We also use "Context" as the name of dependencies injected into protocols and golang has a context package.

It would be nice to use some other name.

pkg/didcomm/protocol/introduce/metadata.go
troyronda: Actually my point was more about should the service be storing (some of) these fields to its own storage provider.

pkg/didcomm/protocol/introduce/service.go
troyronda: As in a payload that will be used between states?

(transitioning between states?)

test/bdd/features/introduce.feature
rolsonquadras: Can you try following steps and see ?
1. Remove [all](https://github.com/hyperledger/aries-framework-go/blob/master/test/bdd/features/introduce.feature#L8) tag from the feature file.
2. Ignore [introduce](https://github.com/hyperledger/aries-framework-go/blob/master/test/bdd/bddtests_test.go#L41) in the bdd test tags.ie, `tags := "all,~introduce"`

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -348,43 +305,76 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
+
+			s.processCallback(md)
 		},
-		Stop: func(err error) { actionStop(&customError{error: err}) },
+		Stop: func(err error) { actionStop(customError{error: err}) },
 	}
 }
 
+// Continue allows proceeding with the action by the context ID
+func (s *Service) Continue(ctxID string, opt Opt) error {
```
rolsonquadras: How can the client get the ctxID in case of async? Do we have any API where we can see all the introduce message which require the user/client action ?

test/bdd/features/introduce.feature
soluchok: Thanks, it works! I was enough to remove tag `all`

pkg/didcomm/protocol/introduce/service.go
soluchok: No, we do not have any API. ctxID for actions is equal to the threadID. Also functions HandleOutbound and HandleInbound will return the context-id.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -87,35 +61,29 @@ type Recipient struct {
...
-	disapprove bool
-	inbound    bool
-	myDID      string
-	theirDID   string
+	TransitionalPayload
+	state        state
+	msgClone     service.DIDCommMsg
+	participants []participant
+	contextID    string
+	disapprove   bool
```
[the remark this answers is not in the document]
soluchok: Changed to `rejected`

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,81 @@
...
+		m[metaRecipients] = []interface{}{
+			&Recipient{To: to}, recipient,
+		}
+	}
+}
+
+// WrapWithMetadataContextID wraps message with metadata.
+// The function is used by the introduce client to define that a few messages are related to each other.
+// e.g When two proposals are sent simultaneously context_id helps the protocol to determine that messages are related.
+func WrapWithMetadataContextID(msgMap service.DIDCommMsgMap, ctxID string) service.DIDCommMsgMap {
```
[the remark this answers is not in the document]
troyronda: i am also wondering about this.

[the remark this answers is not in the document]
soluchok: No, the client does not need to remember context-id. I have changed the function signature. The client does not provide ID anymore.

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Changed it to `psmID` (protocol state machine identifier)

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Currently, the messenger stores and provides it. If service stores it than storing metadata in the messenger is redundant.
The actual logic is that message metadata is stored under the thread-ID key. And metadata is automatically populated when a new message comes (with the same thread). Metadata is only available for one agent (who provided metadata). Actually,  metadata is a thread based key-value storage. Only one record is being stored per thread.

`Should the service be storing some of these fields to its own storage provider` - I do not see the reason for that. Only if we remove this common metadata logic. About protocol namespace,  metadata is thread-base storage. And the thread can't be duplicated. Looks like we do not need this namespace. What we can do it is to open a new store for metadata. But even for that no reason.

pkg/didcomm/protocol/introduce/service.go
llorllale: I can't really argue one way or the other at this point in time

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -161,48 +126,25 @@ func New(p Provider) (*Service, error) {
...
-
 // startInternalListener listens to messages in gochannel for callback messages from clients.
 func (s *Service) startInternalListener() {
 	for {
 		select {
 		case msg := <-s.callbacks:
-			// if no error - do handle or it was disapproved
-			if msg.err == nil || msg.disapprove {
+			// if no error or it was rejected do handle
+			if msg.err == nil || msg.rejected {
```
rolsonquadras: Should the service continue to handle the message in case of rejected ? ie `rejected=true`. I would have imagined the service to handle message when the consumer sets rejected to false or not set it at all (default).

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -348,43 +306,76 @@ func (s *Service) sendMsgEvents(msg *service.StateMsg) {
...
+
+			s.processCallback(md)
 		},
-		Stop: func(err error) { actionStop(&customError{error: err}) },
+		Stop: func(err error) { actionStop(customError{error: err}) },
 	}
 }
 
+// Continue allows proceeding with the action by the psmID
+func (s *Service) Continue(psmID string, opt Opt) error {
```
rolsonquadras: Are we planning to have explicit APIs in client package? As a consumer of the framework it would be hard to understand what Continue does in async case (am I continuing from state a -> state b or state c -> state d). I feel it would be better to have explicit/clear APIs for such cases. Internally we can use this continue function.

pkg/didcomm/protocol/introduce/service.go
rolsonquadras: I feel the consumer/clients shouldn't be aware of protocol layer elements such as threadID. They should work with application ids (in case of DIDExchange it's connectionID).

rolsonquadras: And we would need to provide some APIs to lookup the introduce messages by id or state.

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
+)
+
+const (
+	// protocol state machine identifier
+	metaPsmID        = Introduce + "_psm_id"
```
troyronda: a play on this. PIID -> protocol instance ID

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,84 @@
...
+			&Recipient{To: to}, recipient,
+		}
+	}
+}
+
+// WrapWithMetadataPsmID wraps message with metadata.
+// The function is used by the introduce client to define that a few messages are related to each other.
+// e.g When two proposals are sent simultaneously psmID helps the protocol to determine that messages are related.
+func WrapWithMetadataPsmID(msgMap ...service.DIDCommMsg) {
+	var ctxID = uuid.New().String()
```
troyronda: no longer ctx.

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,84 @@
...
+// USAGE: event.Continue(WithRecipients(to, recipient))
+func WithRecipients(to *To, recipient *Recipient) Opt {
+	return func(m map[string]interface{}) {
+		m[metaRecipients] = []interface{}{
+			&Recipient{To: to}, recipient,
+		}
+	}
+}
+
+// WrapWithMetadataPsmID wraps message with metadata.
```
troyronda: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/didcomm/protocol/introduce/service.go
soluchok: `rejected` is an internal field. User has two possible options:
* Continue(...) 
* Stop()
When user executes `Stop` function protocol sets `rejected=true`. Then we send a `response` message with approve false.

pkg/didcomm/protocol/introduce/service.go
soluchok: For now, we do not have an API. The introduce client might provide these functions based on the `Continue` function.

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Done!

pkg/didcomm/protocol/introduce/metadata.go
soluchok: Done!

pkg/didcomm/protocol/introduce/service.go
soluchok: @rolsonquadras  Why do we need to lookup the introduce messages by id or state?

pkg/didcomm/common/service/service.go
```diff
@@ -15,7 +15,7 @@ type InboundHandler interface {
 // OutboundHandler is handler for outbound messages
 type OutboundHandler interface {
 	// HandleOutbound handles outbound messages.
-	HandleOutbound(msg DIDCommMsg, myDID, theirDID string) error
+	HandleOutbound(msg DIDCommMsg, myDID, theirDID string) (string, error)
```
llorllale: would help to document the return values on this function. I see all implementations just returning an empty string?

pkg/didcomm/messenger/messenger.go
llorllale: @soluchok sorry - I'm not following here. I don't see where a `~thread` decorator will be included in the message finally sent over the wire.

pkg/didcomm/protocol/introduce/metadata.go
```diff
@@ -0,0 +1,84 @@
...
+		m[metaRecipients] = []interface{}{
+			&Recipient{To: to}, recipient,
+		}
+	}
+}
+
+// WrapWithMetadataPIID wraps message with metadata.
+// The function is used by the introduce client to define that a few messages are related to each other.
+// e.g When two proposals are sent simultaneously piID helps the protocol to determine that messages are related.
+func WrapWithMetadataPIID(msgMap ...service.DIDCommMsg) {
```
llorllale: I don't understand why the client package needs to do this in order for this to work. Why can't this package guarantee this is done?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -222,66 +164,76 @@ func (s *Service) startInternalListener() {
...
-	// sets the next state name
-	rec.StateName = next.Name()
-
 	return &metaData{
-		record:   *rec,
-		Msg:      msg,
-		copyMsg:  msg.Clone(),
-		ThreadID: thID,
+		TransitionalPayload: TransitionalPayload{
+			StateName: next.Name(),
```
llorllale: Any need to duplicate state between `metaData` and `TransitionalPayload`?

pkg/didcomm/messenger/messenger.go
troyronda: I also am not seeing thread handling in the codepath.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -222,66 +164,76 @@ func (s *Service) startInternalListener() {
...
+
+func threadID(msg service.DIDCommMsg) (string, error) {
+	if pthID := msg.ParentThreadID(); pthID != "" {
+		return pthID, nil
+	}
+
 	thID, err := msg.ThreadID()
+	if errors.Is(err, service.ErrThreadIDNotFound) {
+		msg.(service.DIDCommMsgMap)["@id"] = uuid.New().String()
+		return msg.(service.DIDCommMsgMap)["@id"].(string), nil
```
llorllale: breaking the abstraction here

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -290,8 +242,8 @@ func (s *Service) InvitationReceived(msg service.StateMsg) error {
 	_, err := s.HandleInbound(service.NewDIDCommMsgMap(&model.Ack{
 		Type:   AckMsgType,
 		ID:     uuid.New().String(),
-		Thread: &decorator.Thread{ID: h.Thread.PID},
-	}), "", "")
+		Thread: &decorator.Thread{ID: msg.Msg.ParentThreadID()},
+	}), "internal", "internal")
```
llorllale: internal?

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -490,113 +482,195 @@ func isNoOp(s state) bool {
...
+	if err != nil {
+		return fmt.Errorf("threadID: %w", err)
+	}
+
+	// TODO: code below might cause an issue! [ issues-1284 ]
+	//  Since the protocol support handling messages simultaneously,
+	//  there is a possibility when two requests might be executed at the same time.
+	//  It might cause a data race. As a result, some of the participants might be lost.
+	//  Possible solutions:
+	//  1. Get and save participants must be done under the lock or in the same transaction.
```
troyronda: we cannot depend on locks. we can be running in a cluster against a distributed database.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -490,113 +482,195 @@ func isNoOp(s state) bool {
...
+	//  2. If the protocol did not receive the answer (e.g Ack) another message should be sent.
+	//  NOTE: The issue might happen when two parallel executions (with the same piID)
+	//  are trying to execute saveParticipants function.
+	md.participants, err = s.getParticipants(md.piID)
```
troyronda: instead of getParticipants and appending, we should instead use separate records to avoid data loss.

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -490,113 +482,195 @@ func isNoOp(s state) bool {
...
 	}
 
-	// sets an invitation to which will be forwarded later
-	msg.Invitation = resp.Invitation
+	if err := s.saveStateName(md.piID, stateName); err != nil {
+		return fmt.Errorf("failed to persist state %s: %w", stateName, err)
+	}
+
+	for _, action := range actions {
+		if action == nil {
```
llorllale: why would action be nil?

troyronda: agree - it is odd that action would be nil at this point (or that we didn't filter it out earlier).

pkg/didcomm/protocol/introduce/service.go
soluchok: We do not need it. Instead of empty values, I have added `internal` identifier. It may help in the case of debugging. Instead of looking for why values are empty you will receive it `internal`.

pkg/didcomm/protocol/introduce/service.go
soluchok: Done! https://github.com/hyperledger/aries-framework-go/pull/1254

pkg/didcomm/protocol/introduce/service.go
soluchok: Done! https://github.com/hyperledger/aries-framework-go/pull/1254

PR #1254, opened 2020-02-13
refactor: Introduce - adds client and BDD tests (part 3)

### LINKED ISSUE #1157: Introduce - the protocol should not depend on thread decorator

soluchok:
* added Introduce Client
* added BDD tests

Closes #1157


rolsonquadras:
@soluchok FYI - branch has conflicts.

pkg/client/introduce/client.go
```diff
@@ -53,176 +38,63 @@ func New(ctx Provider, inv *didexchange.Invitation) (*Client, error) {
...
 func (c *Client) SendProposalWithInvitation(inv *didexchange.Invitation, recipient *introduce.Recipient) error {
-	return c.sendProposal(InvitationEnvelope{
-		Inv:   inv,
-		Recps: []*introduce.Recipient{recipient},
-	})
+	proposal := introduce.CreateProposal(recipient.To)
+
+	introduce.WrapWithMetadataPublicInvitation(proposal, inv)
+
+	_, err := c.service.HandleOutbound(proposal, recipient.MyDID, recipient.TheirDID)
```
troyronda: what is the ignored return value?

pkg/client/introduce/client.go
```diff
@@ -53,176 +38,63 @@ func New(ctx Provider, inv *didexchange.Invitation) (*Client, error) {
...
-	return o, nil
-}
-
-func invitationEnvelopeKey(thID string) string {
-	return invitationEnvelopePrefix + thID
-}
+// WithInvitation is used when introducee wants to provide invitation.
+// NOTE: Introducee can provide invitation only after receiving ProposalMsgType
+// USAGE: event.Continue(WithInvitation(inv))
+var WithInvitation = introduce.WithInvitation // nolint: gochecknoglobals
```
troyronda: why is there a global exported variable?

test/bdd/features/introduce.feature
```diff
@@ -10,284 +10,177 @@ Feature: Introduce protocol
...
 
     When   "Alice" sends introduce proposal back to the requester with pub invitation
-      And   "Bob" doesn't want to know "Carol" and sends introduce response
-
-    Then   "Alice" continue with the introduce protocol
-
-    Then   "Alice" checks the history of introduce protocol events "arranging,arranging,arranging,arranging,delivering,delivering,abandoning,abandoning,done,done"
-      And   "Bob" checks the history of introduce protocol events "requesting,requesting,deciding,deciding,abandoning,abandoning,done,done"
-
-  @skip_proposal_introducer_stop
```
troyronda: is this case no longer relevant?

test/bdd/features/introduce.feature
```diff
@@ -10,284 +10,177 @@ Feature: Introduce protocol
...
-      And   "Alice" sends introduce proposal to the "Bob" with "Carol" invitation
-
-    When   "Bob" wants to know "Carol" and sends introduce response with approve
-      And   "Alice" stops the introduce protocol
-
-    Then   "Alice" checks the history of introduce protocol events "arranging,arranging,abandoning,abandoning,done,done"
-      And   "Bob" checks the history of introduce protocol events "deciding,deciding,waiting,waiting,abandoning,abandoning,done,done"
+    And   "Bob" doesn't want to know "Carol" and sends introduce response
 
-  @skip_proposal_introducer_stop_with_request
```
troyronda: is this case no longer relevant?

test/bdd/features/introduce.feature
```diff
@@ -10,284 +10,177 @@ Feature: Introduce protocol
...
+    Then   "Alice" sends introduce proposal back to the "Bob" and requested introduce
+    And   "Bob" wants to know "Carol" and sends introduce response with approve
+    And   "Carol" doesn't want to know "Bob" and sends introduce response
 
-    Then   "Alice" checks the history of introduce protocol events "arranging,arranging,arranging,arranging,delivering,delivering,abandoning,abandoning,done,done"
+    Then   "Alice" checks the history of introduce protocol events "arranging,arranging,arranging,arranging,arranging,arranging,abandoning,abandoning,done,done"
     And   "Bob" checks the history of introduce protocol events "requesting,requesting,deciding,deciding,waiting,waiting,abandoning,abandoning,done,done"
-    And   "Carol" checks the history of introduce protocol events "deciding,deciding,abandoning,abandoning,done,done"
-
-  @proposal_introducer_stop
```
troyronda: is this case no longer relevant?

test/bdd/features/introduce.feature
```diff
@@ -10,284 +10,177 @@ Feature: Introduce protocol
...
-    When   "Bob" creates introduce client with an invitation
-      And   "Bob" sends introduce request to the "Alice" asking about "Carol"
-      And   "Alice" sends introduce proposal back to the "Bob" and requested introduce
-      And   "Bob" wants to know "Carol" and sends introduce response with approve
-      And   "Alice" stops the introduce protocol
-
-    Then   "Alice" checks the history of introduce protocol events "arranging,arranging,abandoning,abandoning,done,done"
-    And   "Bob" checks the history of introduce protocol events "requesting,requesting,deciding,deciding,waiting,waiting,abandoning,abandoning,done,done"
-
-  @proposal_introducer_second_stop
```
troyronda: is this case no longer relevant?

pkg/client/introduce/client.go
```diff
@@ -9,40 +7,27 @@ SPDX-License-Identifier: Apache-2.0
...
-	service    service.DIDComm
-	store      storage.Store
-	defaultInv *didexchange.Invitation
-	newUUID    func() string
+	service service.DIDComm
 }
 
 // New return new instance of introduce client
-func New(ctx Provider, inv *didexchange.Invitation) (*Client, error) {
+func New(ctx Provider) (*Client, error) {
```
troyronda: how does the client async continue the protocol without event.Continue?

pkg/client/introduce/client.go
soluchok: It is PIID, I did it similar to did-exchange (connectionID). But we do not use it. Not even sure whether it will be used.

troyronda: ok

pkg/client/introduce/client.go
soluchok: It is a function alias. Instead of importing these functions from the service user can import them from the client.

troyronda: It is mutable - shouldn't have this.

test/bdd/features/introduce.feature
soluchok: No, since I have deleted the action event after receiving a response it is not possible anymore.

test/bdd/features/introduce.feature
soluchok: No, since I have deleted the action event after receiving a response it is not possible anymore.

test/bdd/features/introduce.feature
soluchok: No, since I have deleted the action event after receiving a response it is not possible anymore.

test/bdd/features/introduce.feature
soluchok: Same. Related to the tests which contains `introducer..stop`

pkg/client/introduce/client.go
```diff
@@ -53,176 +38,63 @@ func New(ctx Provider, inv *didexchange.Invitation) (*Client, error) {
...
-//  - HandleRequestWithInvitation
-// usage: e.Continue(c.InvitationEnvelope(threadID))
-func (c *Client) InvitationEnvelope(thID string) *InvitationEnvelope {
-	opts, err := c.getInvitationEnvelope(thID)
-
-	if errors.Is(err, storage.ErrDataNotFound) {
-		return &InvitationEnvelope{Inv: c.defaultInv}
-	}
+// SendProposal sends a proposal to the introducees (the client does not have a public Invitation).
+func (c *Client) SendProposal(recipient1, recipient2 *introduce.Recipient) error {
```
rolsonquadras: The introduce service dependency(`introduce.Recipient`) shouldn't be exposed to the users of this client. We can wrap the service models in client package similar to [DIDExchange](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/client/didexchange/models.go#L52).

pkg/client/introduce/client.go
```diff
@@ -53,176 +38,63 @@ func New(ctx Provider, inv *didexchange.Invitation) (*Client, error) {
...
-// SendProposal sends a proposal to the introducees (the client does not have a public Invitation).
-func (c *Client) SendProposal(recipient1, recipient2 *introduce.Recipient) error {
-	return c.sendProposal(InvitationEnvelope{
-		Recps: []*introduce.Recipient{recipient1, recipient2},
-	})
+	return err
 }
 
 // SendProposalWithInvitation sends a proposal to the introducee (the client has a public Invitation).
 func (c *Client) SendProposalWithInvitation(inv *didexchange.Invitation, recipient *introduce.Recipient) error {
```
rolsonquadras: dependency for Invitation model should be used from DIDExchange client rather than service pkg.

test/bdd/features/introduce.feature
```diff
@@ -10,284 +10,177 @@ Feature: Introduce protocol
   @skip_proposal
   Scenario: Alice has a Carol's public invitation
     Given   "Bob,Carol" exchange DIDs with "Alice"
-      And   "Alice" sends introduce proposal to the "Bob" with "Carol" invitation
-
+    And   "Alice" sends introduce proposal to the "Bob" with "Carol" invitation
```
rolsonquadras: I believe the end goal of Introduce to have a connection established between 2 parties. Do we have any e2e scenario where one agent(alice) introduce 2 parties(bob and carol) and validate that the DIDExchange/connection between them(bob and carol) is complete ?

soluchok: I have done it. Then I received a comment `It is not a part of the introduce BDD tests`. It is related to did exchange. In these tests, I am checking states id e.g done. It is possible to receive done after did-exchange received an invitation. So, this is how we can be sure that the message was delivered. Was the connection established successfully is not a part of the introduce.

rolsonquadras: May be we don't need to add all the DID-Exchange steps, but we need to validate DIDExchange connection is complete post introduction. These are integration tests and basically we need to show post introduce DIDExchange imo.

@troyronda @llorllale @fqutishat @sandrask thoughts ?

troyronda: @rolsonquadras I think this is a good idea.

pkg/client/introduce/client.go

pkg/client/introduce/client.go
soluchok: Done! I have added the `UnfinishedActions` function

pkg/client/introduce/client.go
```diff
@@ -9,220 +7,131 @@ SPDX-License-Identifier: Apache-2.0
...
-
-var logger = log.New("aries-framework/introduce/client")
-
 // Provider contains dependencies for the introduce protocol and is typically created by using aries.Context()
 type Provider interface {
 	Service(id string) (interface{}, error)
-	StorageProvider() storage.Provider
+}
+
+// ProtocolService defines DID Exchange service.
```
troyronda: DID Exchange?

pkg/client/introduce/client.go
```diff
@@ -9,220 +7,131 @@ SPDX-License-Identifier: Apache-2.0
...
-// InvitationEnvelope keeps the information needed for sending a proposal
-type InvitationEnvelope struct {
-	// Invitation must be set when we have a public Invitation
-	Inv *didexchange.Invitation `json:"invitation,omitempty"`
-	// Recipients contain one or two elements
-	// one element - for the public Invitation, otherwise two elements
-	Recps []*introduce.Recipient `json:"recipients,omitempty"`
+// ContinueWithInvitation is used when introducee wants to provide invitation.
+// NOTE: For async usage. Introducee can provide invitation only after receiving ProposalMsgType
+func (c *Client) ContinueWithInvitation(piID string, inv *didexchange.Invitation) error {
```
rolsonquadras: What's the client API which provides piID in async usecase ?

soluchok: See `UnfinishedActions` function

pkg/client/introduce/client.go
```diff
@@ -9,220 +7,131 @@ SPDX-License-Identifier: Apache-2.0
...
+	return c.service.Continue(piID, WithInvitation(inv))
 }
 
-// Invitation returns an Invitation needed for the service (state machine depends on it)
-func (o *InvitationEnvelope) Invitation() *didexchange.Invitation {
-	return o.Inv
+// ContinueWithPublicInvitation is used when introducer wants to provide public invitation.
+// NOTE: For async usage. Introducer can provide invitation only after receiving RequestMsgType
+func (c *Client) ContinueWithPublicInvitation(piID string, inv *didexchange.Invitation, to *introduce.To) error {
```
troyronda: I'm a bit confused. Why are we exposing both `ContinueWithPublicInvitation` and `WithPublicInvitation` from this package? And we haven't exposed a generic `Continue` for `WithPublicInvitation`?

soluchok: ContinueWithPublicInvitation for async. WithPublicInvitation is an option for the Continue function (when we received an event message)

troyronda: This is too confusing.

soluchok: When you received an action event you can call a.Continue(WithPublicInvitation) . It is much cheaper because we avoiding the call to DB. 

When you call API  ContinueWithPublicInvitation(PIID,..) It will get data from the DB (by PIID)

troyronda: I also don't see this happening in DID Exchange client: https://godoc.org/github.com/hyperledger/aries-framework-go/pkg/client/didexchange

troyronda: I think we didn't wrap the event continuation functions - as these are considered advanced usage... instead, we simply created client methods for the more typical async case. 

@rolsonquadras @llorllale @soluchok

soluchok: In DID Exchange it is `AcceptExchangeRequest` function

troyronda: I think to reduce confusion, I see two possibilities:
- Supply individual client methods for the async case but no wrappers for the event continuation functions.
- Supply wrappers for continuation functions and only have a single Continue method in the client.

I think DID Exchange client has the first mechanism.

rolsonquadras: @troyronda agree. The DIDExchange client/service has individual APIs and handle() function is called from these APIs inside service pkg (e.Continue calls handle() as well). We can wrap svc.Continue() function in client package as well. The idea in DIDExchange was to give a clear APIs to the consumers, so that they would know what exact action they are performing.

troyronda: @rolsonquadras ok. so how should this PR be changed?

rolsonquadras: @troyronda I would suggest associating the function names with action/state(stateMachine) rather than input.

- introduce.AcceptProposal(piID string, inv *didexchange.Invitation)
- introduce.AcceptRequestWithPublicInvitation(piID string, inv *didexchange.Invitation, to *To)
- introduce.AcceptRequestWithRecipients(piID string, to *To, recipient *Recipient)
- introduce.Actions()

llorllale: > When you received an action event you can call a.Continue(WithPublicInvitation) . It is much cheaper because we avoiding the call to DB.
> 
> When you call API ContinueWithPublicInvitation(PIID,..) It will get data from the DB (by PIID)

`Continue()` is hitting the store always.

llorllale: @troyronda 

> IMO - this is confusing. Especially since ContinueWithPublicInvitation is basically just a wrapper for Continue + WithPublicInvitation.

Agreed

> I think we didn't wrap the event continuation functions in DID Exchange - as these are considered advanced usage... instead, we simply created client methods for the more typical async case.

I consider the "continue with this invitation" capability required for basic use of the introduce protocol.

> * Supply individual client methods for the async case but no wrappers for the event continuation functions.
> * Supply wrappers for continuation functions and only have a single Continue method in the client.

I think the second option would look more elegant.

troyronda: > I consider the "continue with this invitation" capability required for basic use of the introduce protocol 

It's not basic usage to do so from an event object.

pkg/client/introduce/client.go
troyronda: @soluchok did you see this comment?

test/bdd/bddtests_test.go
```diff
@@ -15,6 +15,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/test/bdd/pkg/introduce"
```
troyronda: imports are not ordered quite right.

test/bdd/features/introduce.feature
rolsonquadras: created an issue to track this - https://github.com/hyperledger/aries-framework-go/issues/1332

Issue #1149, opened 2020-01-29
Add a default (local) KMS implementation with Storage and SecretLock as providers


Baha-sk:
part of #985

PR #1258, opened 2020-02-13
feat: store Reader/Writer for default local KMS implementation

Baha-sk:
To prepare for the default local KMS implementation, a store
Reader and Writer need to be created to help persisting KMS
key sets in a local storage.

This is pre step for the next patch that includes the full
local KMS implementation.

part of #1149

### DIFF
--- a/pkg/kms/localkms/localkms_reader.go
+++ b/pkg/kms/localkms/localkms_reader.go
@@ -1,0 +1,48 @@
+/*
+ Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package localkms
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// newReader will create a new local storage storeReader of a keyset with ID value = keysetID
+// it is used internally by local kms
+func newReader(store storage.Store, keysetID string) *storeReader {
+	return &storeReader{
+		storage:  store,
+		keysetID: keysetID,
+	}
+}
+
+// storeReader struct to load a keyset from a local storage
+type storeReader struct {
+	buf      *bytes.Buffer
+	storage  storage.Store
+	keysetID string
+}
+
+// Read the keyset from local storage into p.
+func (l *storeReader) Read(p []byte) (int, error) {
+	if l.buf == nil {
+		if l.keysetID == "" {
+			return 0, fmt.Errorf("keysetID is not set")
+		}
+
+		data, err := l.storage.Get(l.keysetID)
+		if err != nil {
+			return 0, fmt.Errorf("cannot read data for keysetID %s: %w", l.keysetID, err)
+		}
+
+		l.buf = bytes.NewBuffer(data)
+	}
+
+	return l.buf.Read(p)
+}
--- a/pkg/kms/localkms/localkms_writer.go
+++ b/pkg/kms/localkms/localkms_writer.go
@@ -1,0 +1,74 @@
+/*
+ Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package localkms
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+
+	"github.com/google/tink/go/subtle/random"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// newWriter creates a new instance of local storage key storeWriter in the given store and for masterKeyURI
+func newWriter(kmsStore storage.Store, masterKeyURI string) *storeWriter {
+	mkURI := masterKeyURI
+	if strings.LastIndex(mkURI, "/") < len(mkURI)-1 {
+		mkURI += "/"
+	}
+
+	return &storeWriter{
+		storage:      kmsStore,
+		masterKeyURI: mkURI,
+	}
+}
+
+// storeWriter struct to store a keyset in a local store
+type storeWriter struct {
+	storage      storage.Store
+	masterKeyURI string
+	// KeysetID is set when Write() is called
+	KeysetID string
+}
+
+// Write a marshaled keyset p in localstore with masterKeyURI prefix + randomly generated KeysetID
+func (l *storeWriter) Write(p []byte) (int, error) {
+	if l.masterKeyURI == "" {
+		return 0, fmt.Errorf("master key is not set")
+	}
+
+	const keySetIDLength = 32
+
+	baseID := l.masterKeyURI
+	ksID := ""
+
+	for {
+		// generate random ID prefixed with masterKeyURI
+		ksID = baseID + base64.URLEncoding.EncodeToString(random.GetRandomBytes(keySetIDLength))
+
+		// ensure ksID is not already used
+		_, e := l.storage.Get(ksID)
+		if e != nil {
+			if e == storage.ErrDataNotFound {
+				break
+			}
+
+			return 0, e
+		}
+	}
+
+	err := l.storage.Put(ksID, p)
+	if err != nil {
+		return 0, err
+	}
+
+	l.KeysetID = ksID
+
+	return len(p), nil
+}
--- a/pkg/kms/localkms/localkms_reader_test.go
+++ b/pkg/kms/localkms/localkms_reader_test.go
@@ -1,0 +1,108 @@
+/*
+ Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package localkms
+
+import (
+	"fmt"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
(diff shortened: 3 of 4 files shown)


pkg/kms/localkms/localkms_reader.go
```diff
@@ -0,0 +1,68 @@
...
+// storeReader struct to load a keyset from a local storage
+type storeReader struct {
+	buf      []byte
+	lastRead int
+	storage  storage.Store
+	keysetID string
+}
+
+// Read a marshaled keyset from local storage representing a keyset with ID = l.keysetID and copy it to p
+// the length of the content copied will be len(p)
```
llorllale:
```suggestion
// Read the keyset from local storage into p.
```

pkg/kms/localkms/localkms_reader.go
```diff
@@ -0,0 +1,68 @@
...
+func (l *storeReader) Read(p []byte) (int, error) {
+	if l.keysetID == "" {
+		return 0, fmt.Errorf("keysetID is not set")
+	}
+
+	// populate buf if empty
+	if len(l.buf) == 0 {
+		data, err := l.storage.Get(l.keysetID)
+		if err != nil {
+			return 0, err
```
llorllale: Should wrap this error otherwise consumer will just receive a `ErrNoDataFound`

pkg/kms/localkms/localkms_reader.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		capLengthFactor := 2
```
llorllale: ?

pkg/kms/localkms/localkms_reader.go
```diff
@@ -0,0 +1,68 @@
...
+		l.buf = make([]byte, len(data), capLengthFactor*len(data))
```
llorllale: I think with `l.buf` you can now just create and hold on to a `byte.Buffer` and delegate `Read()` calls to it - avoiding the rest of the code below.

pkg/kms/localkms/localkms_writer.go
```diff
@@ -0,0 +1,69 @@
...
+		masterKeyURI: masterKeyURI,
+	}
+}
+
+// storeWriter struct to store a keyset in a local store
+type storeWriter struct {
+	storage      storage.Store
+	masterKeyURI string
+	// KeysetID is set when Write() is called
+	KeysetID string
```
llorllale: accidental export?

pkg/kms/localkms/localkms_writer.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Write a marshaled keyset p in localstore with masterKeyURI prefix + randomly generated KeysetID
+func (l *storeWriter) Write(p []byte) (int, error) {
+	if l.masterKeyURI == "" {
+		return 0, fmt.Errorf("master key is not set")
+	}
+
+	const keySetIDLength = 32
+
+	baseID := l.masterKeyURI + "/"
```
llorllale: should this uri be normalized?

pkg/kms/localkms/localkms_reader.go
Baha-sk: The function cannot control the length of `p`, it can be either less or more than the content retrieved from the store. The code below is needed if it's smaller than the store's keyset

pkg/kms/localkms/localkms_writer.go
```diff
@@ -0,0 +1,69 @@
...
+	ksID := ""
+
+	for {
+		// generate random ID
+		ksID = baseID + base64.URLEncoding.EncodeToString(random.GetRandomBytes(keySetIDLength))
```
llorllale: Why not just use crypto/rand? crypto/rand guarantees a "cryptographically secure random number generator" so there should be no collisions

pkg/kms/localkms/localkms_reader.go
Baha-sk: will remove cap value, it's unnecessary...

pkg/kms/localkms/localkms_writer.go
Baha-sk: it's not accidental, this is how the writer exposes the KeySet ID after `Write()` is called, the KMS will access it for reference later (for getting or rotating a key)

pkg/kms/localkms/localkms_reader.go
Baha-sk: will replace logic with bytes.Buffer

pkg/kms/localkms/localkms_writer.go
Baha-sk: it's just a prefix for the keys being used with the mastery from secretlock. An agent may want to use different master keys for a different set of keysets  and this baseID can distinguish the keys.

It can be done later when the need becomes prevalent.

pkg/kms/localkms/localkms_writer.go
Baha-sk: random.GetRandomBytes() is a utility function that calls crypto/rand under the hood, it saves the boilerplate code of creating and copying a byte array.

pkg/kms/localkms/localkms_writer.go
Baha-sk: I will move `+ "/"` up in the constructor if this helps..

pkg/kms/localkms/localkms_reader.go
Baha-sk: updated

pkg/kms/localkms/localkms_reader.go
Baha-sk: done

pkg/kms/localkms/localkms_reader.go
Baha-sk: done

pkg/kms/localkms/localkms_reader.go
Baha-sk: done

pkg/kms/localkms/localkms_writer.go
Baha-sk: moved

pkg/kms/localkms/localkms_reader.go
```diff
@@ -0,0 +1,49 @@
...
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// newReader will create a new local storage storeReader of a keyset with ID value = keysetID
+// it is used internally by local kms
+func newReader(store storage.Store, keysetID string) io.Reader {
```
troyronda: Why not return a struct pointer? It is not needed to return a particular interface from a constructor.

also how will this function be used - i see it is not exported from this package.

pkg/kms/localkms/localkms_writer.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/google/tink/go/subtle/random"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// newWriter creates a new instance of local storage key storeWriter in the given store and for masterKeyURI
```
troyronda: Why not return a struct pointer? It is not needed to return a particular interface from a constructor.

also how will this function be used - i see it is not exported from this package.

pkg/kms/localkms/localkms_writer.go
llorllale: it's not just any prefix - it's a URI. I'm just wondering what the behaviour would be with two trailing forward slashes.

pkg/kms/localkms/localkms_reader.go
Baha-sk: this will be only used the by the local kms, this is why it's not exported.

the constructor returns a plain io.Reader.. I'll change to storeReader if it makes more sense

pkg/kms/localkms/localkms_writer.go
Baha-sk: same as reader above

pkg/kms/localkms/localkms_writer.go
Baha-sk: for now, this condition will take care of the trailing `/`:
```
	if strings.LastIndex(mkURI, "/") < len(mkURI)-1 {
 		mkURI += "/"
 	}
```
I'm not too concerned about this URI as it only identify a master key (if an agent ever needs to use more than one master key to manage keys). This is more important for cloud KMSs. 

You will notice the KMS and secret lock interfaces always have a URI as a first argument in their functions. It mainly for cloud KMSes and secret locks.

pkg/kms/localkms/localkms_reader.go
troyronda: returning struct pointers makes more sense than returning an interface UNLESS you are trying to meet another interface. (and that's not the case here).

Baha-sk: sure, I'm replacing it..

Baha-sk: done

pkg/kms/localkms/localkms_writer.go
Baha-sk: set return to struct

PR #1265, opened 2020-02-14
feat: JWT library with flexible signers

### LINKED ISSUE #1264: JWT library with flexible signers
We use https://github.com/square/go-jose to deal with JWT in `verifiable` package.
In order to sign the JWT, `go-jose` requires creation of `Signer` object with the mandatory passing of a private key.  We would like to deal with private keys inside KMS solely and thus this approach does not satisfy us. 

Requirements for the light-weight impl of the JWT:
- `Signer` interface does not require private key.
- Support JWS and unsecured JWT.
- Support detached payloads (to be used by `signature` package.
- Support `b64` header.
- Support RS256 and EdDSA verifications from the box (as they are used in `verifiable`).
- To be opened for extra algorithms (e.g. secp256k1).

kdimak:
closes #1264

This package will be initially used by `verifiable` and `signature` packages.
I already have adopted `verifiable` but do not want to include it to this PR in order to keep the PR as small as possible.

### DIFF
--- a/pkg/doc/jose/jws.go
+++ b/pkg/doc/jose/jws.go
@@ -1,0 +1,388 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package jose
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/square/go-jose/v3"
+	"github.com/square/go-jose/v3/json"
+)
+
+const (
+	jwsPartsCount    = 3
+	jwsHeaderPart    = 0
+	jwsPayloadPart   = 1
+	jwsSignaturePart = 2
+)
+
+// Headers represents JOSE headers.
+type Headers map[string]interface{}
+
+// JWK (JSON Web Key) is a JSON data structure that represents a cryptographic key.
+type JWK jose.JSONWebKey
+
+// JWK gets JWK from JOSE headers.
+func (h Headers) JWK() (*JWK, bool) {
+	jwkRaw, ok := h[HeaderJSONWebKey]
+	if !ok {
+		return nil, false
+	}
+
+	var jwk JWK
+
+	err := convertMapToValue(jwkRaw, &jwk)
+	if err != nil {
+		return nil, false
+	}
+
+	return &jwk, true
+}
+
+// KeyID gets Key ID from JOSE headers.
+func (h Headers) KeyID() (string, bool) {
+	return h.stringValue(HeaderKeyID)
+}
+
+// Algorithm gets Key ID from JOSE headers.
+func (h Headers) Algorithm() (string, bool) {
+	return h.stringValue(HeaderAlgorithm)
+}
+
+func (h Headers) stringValue(key string) (string, bool) {
+	kRaw, ok := h[key]
+	if !ok {
+		return "", false
+	}
+
+	kStr, ok := kRaw.(string)
+
+	return kStr, ok
+}
+
+// JSONWebSignature defines JSON Web Signature (https://tools.ietf.org/html/rfc7515)
+type JSONWebSignature struct {
+	ProtectedHeaders   Headers
+	UnprotectedHeaders Headers
+	Payload            []byte
+
+	signature   []byte
+	joseHeaders Headers
+}
+
+// SignatureVerifier makes verification of JSON Web Signature.
+type SignatureVerifier interface {
+	// Verify verifies JWS based on the signing input.
+	Verify(joseHeaders Headers, payload, signingInput, signature []byte) error
+}
+
+// SignatureVerifierFunc is a function wrapper for SignatureVerifier.
+type SignatureVerifierFunc func(joseHeaders Headers, payload, signingInput, signature []byte) error
+
+// Verify verifies JWS signature.
+func (s SignatureVerifierFunc) Verify(joseHeaders Headers, payload, signingInput, signature []byte) error {
+	return s(joseHeaders, payload, signingInput, signature)
+}
+
+// CompositeAlgSigVerifier defines composite signature verifier based on the algorithm
+// taken from JOSE header alg.
+type CompositeAlgSigVerifier struct {
+	verifierByAlg map[string]SignatureVerifier
+}
+
+// AlgSignatureVerifier defines verifier for particular signature algorithm.
+type AlgSignatureVerifier struct {
+	Alg      string
+	Verifier SignatureVerifier
+}
+
+// NewCompositeAlgSigVerifier creates a new CompositeAlgSigVerifier
+func NewCompositeAlgSigVerifier(v AlgSignatureVerifier, vOther ...AlgSignatureVerifier) *CompositeAlgSigVerifier {
+	verifierByAlg := make(map[string]SignatureVerifier, 1+len(vOther))
+	verifierByAlg[v.Alg] = v.Verifier
+
+	for _, v := range vOther {
+		verifierByAlg[v.Alg] = v.Verifier
+	}
+
+	return &CompositeAlgSigVerifier{
+		verifierByAlg: verifierByAlg,
+	}
+}
+
+// Verify verifiers JWS signature.
+func (v *CompositeAlgSigVerifier) Verify(joseHeaders Headers, payload, signingInput, signature []byte) error {
+	alg, ok := joseHeaders.Algorithm()
+	if !ok {
+		return errors.New("'alg' JOSE header is not present")
+	}
+
+	verifier, ok := v.verifierByAlg[alg]
+	if !ok {
+		return fmt.Errorf("no verifier found for %s algorithm", alg)
+	}
+
+	return verifier.Verify(joseHeaders, payload, signingInput, signature)
+}
+
+// Signer defines JWS Signer interface. It makes signing of data and provides custom JWS headers relevant to the signer.
+type Signer interface {
+	// Sign signs.
+	Sign(data []byte) ([]byte, error)
+
+	// Headers provides JWS headers. "alg" header must be provided (see https://tools.ietf.org/html/rfc7515#section-4.1)
+	Headers() Headers
+}
+
+// NewJWS creates JSON Web Signature.
+func NewJWS(protectedHeaders, unprotectedHeaders Headers, payload []byte, signer Signer) (*JSONWebSignature, error) {
+	jws := &JSONWebSignature{
+		ProtectedHeaders:   protectedHeaders,
+		UnprotectedHeaders: unprotectedHeaders,
+		Payload:            payload,
+		joseHeaders:        mergeHeaders(protectedHeaders, signer.Headers()),
+	}
+
+	signature, err := sign(jws.joseHeaders, payload, signer)
+	if err != nil {
+		return nil, fmt.Errorf("sign JWS: %w", err)
+	}
+
+	jws.signature = signature
+
+	return jws, nil
+}
+
+// SerializeCompact makes JWS Compact Serialization (https://tools.ietf.org/html/rfc7515#section-7.1)
+func (s JSONWebSignature) SerializeCompact(detached bool) (string, error) {
+	byteHeaders, err := json.Marshal(s.joseHeaders)
+	if err != nil {
+		return "", fmt.Errorf("marshal JWS JOSE Headers: %w", err)
+	}
+
+	b64Headers := base64.RawURLEncoding.EncodeToString(byteHeaders)
+
+	b64Payload := ""
+	if !detached {
+		b64Payload = base64.RawURLEncoding.EncodeToString(s.Payload)
+	}
+
+	b64Signature := base64.RawURLEncoding.EncodeToString(s.signature)
+
+	return fmt.Sprintf("%s.%s.%s",
+		b64Headers,
+		b64Payload,
+		b64Signature), nil
+}
+
+// Signature returns a copy of JWS signature.
+func (s JSONWebSignature) Signature() []byte {
+	if s.signature == nil {
+		return nil
+	}
+
+	sCopy := make([]byte, len(s.signature))
+	copy(sCopy, s.signature)
+
+	return sCopy
+}
+
+func mergeHeaders(h1, h2 Headers) Headers {
+	h := make(Headers, len(h1)+len(h2))
+
+	for k, v := range h2 {
+		h[k] = v
+	}
+
+	for k, v := range h1 {
+		h[k] = v
+	}
+
+	return h
+}
+
+func sign(joseHeaders Headers, payload []byte, signer Signer) ([]byte, error) {
+	err := checkJWSHeaders(joseHeaders)
+	if err != nil {
+		return nil, fmt.Errorf("check JOSE headers: %w", err)
+	}
+
+	sigInput, err := signingInput(joseHeaders, payload)
+	if err != nil {
+		return nil, fmt.Errorf("prepare JWS verification data: %w", err)
+	}
+
+	signature, err := signer.Sign(sigInput)
+	if err != nil {
+		return nil, fmt.Errorf("sign JWS verification data: %w", err)
+	}
+
+	return signature, nil
+}
+
+// jwsParseOpts holds options for the JWS Parsing.
+type jwsParseOpts struct {
+	detachedPayload []byte
+}
+
+// JWSParseOpt is the JWS Parser option.
+type JWSParseOpt func(opts *jwsParseOpts)
+
+// WithJWSDetachedPayload option is for definition of JWS detached payload.
+func WithJWSDetachedPayload(payload []byte) JWSParseOpt {
+	return func(opts *jwsParseOpts) {
+		opts.detachedPayload = payload
+	}
+}
+
+// ParseJWS parses serialized JWS. Currently only JWS Compact Serialization parsing is supported.
+func ParseJWS(jws string, verifier SignatureVerifier, opts ...JWSParseOpt) (*JSONWebSignature, error) {
+	pOpts := &jwsParseOpts{}
+
+	for _, opt := range opts {
+		opt(pOpts)
+	}
+
+	if strings.HasPrefix(jws, "{") {
+		// TODO support JWS JSON serialization format
+		//  https://github.com/hyperledger/aries-framework-go/issues/1331
+		return nil, errors.New("JWS JSON serialization is not supported")
+	}
+
+	return parseCompacted(jws, verifier, pOpts)
+}
+
+// IsCompactJWS checks weather input is a compact JWS (based on https://tools.ietf.org/html/rfc7516#section-9)
+func IsCompactJWS(s string) bool {
+	parts := strings.Split(s, ".")
+
+	return len(parts) == jwsPartsCount
+}
+
+func parseCompacted(jwsCompact string, verifier SignatureVerifier, opts *jwsParseOpts) (*JSONWebSignature, error) {
+	parts := strings.Split(jwsCompact, ".")
+	if len(parts) != jwsPartsCount {
+		return nil, errors.New("invalid JWS compact format")
+	}
+
+	joseHeaders, err := parseCompactedHeaders(parts)
+	if err != nil {
+		return nil, err
+	}
+
+	payload, err := parseCompactedPayload(parts[jwsPayloadPart], opts)
+	if err != nil {
+		return nil, err
+	}
+
+	sInput, err := signingInput(joseHeaders, payload)
+	if err != nil {
+		return nil, fmt.Errorf("build signing input: %w", err)
+	}
+
+	signature, err := base64.RawURLEncoding.DecodeString(parts[jwsSignaturePart])
+	if err != nil {
+		return nil, fmt.Errorf("decode base64 signature: %w", err)
+	}
+
+	err = verifier.Verify(joseHeaders, payload, sInput, signature)
+	if err != nil {
+		return nil, err
+	}
+
+	return &JSONWebSignature{
+		ProtectedHeaders: joseHeaders,
+		Payload:          payload,
+		signature:        signature,
+		joseHeaders:      joseHeaders,
+	}, nil
+}
+
+func parseCompactedPayload(jwsPayload string, opts *jwsParseOpts) ([]byte, error) {
+	if len(opts.detachedPayload) > 0 {
+		return opts.detachedPayload, nil
+	}
+
+	payload, err := base64.RawURLEncoding.DecodeString(jwsPayload)
+	if err != nil {
+		return nil, fmt.Errorf("decode base64 payload: %w", err)
+	}
+
+	return payload, nil
+}
+
+func parseCompactedHeaders(parts []string) (Headers, error) {
+	headersBytes, err := base64.RawURLEncoding.DecodeString(parts[jwsHeaderPart])
+	if err != nil {
+		return nil, fmt.Errorf("decode base64 header: %w", err)
+	}
+
+	var joseHeaders Headers
+
+	err = json.Unmarshal(headersBytes, &joseHeaders)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal JSON headers: %w", err)
+	}
+
+	err = checkJWSHeaders(joseHeaders)
+	if err != nil {
+		return nil, err
+	}
+
+	return joseHeaders, nil
+}
+
+func signingInput(headers Headers, payload []byte) ([]byte, error) {
+	headersBytes, err := json.Marshal(headers)
+	if err != nil {
+		return nil, fmt.Errorf("serialize JWS headers: %w", err)
+	}
+
+	hBase64 := true
+
+	if b64, ok := headers[HeaderB64Payload]; ok {
+		if hBase64, ok = b64.(bool); !ok {
+			return nil, errors.New("invalid b64 header")
+		}
+	}
+
+	headersStr := base64.RawURLEncoding.EncodeToString(headersBytes)
+
+	var payloadStr string
+
+	if hBase64 {
+		payloadStr = base64.RawURLEncoding.EncodeToString(payload)
+	} else {
+		payloadStr = string(payload)
+	}
+
+	return []byte(fmt.Sprintf("%s.%s", headersStr, payloadStr)), nil
+}
+
+func checkJWSHeaders(headers Headers) error {
+	if _, ok := headers[HeaderAlgorithm]; !ok {
+		return fmt.Errorf("%s JWS header is not defined", HeaderAlgorithm)
+	}
+
+	return nil
+}
+
+func convertMapToValue(vOriginToBeMap, vDest interface{}) error {
+	if _, ok := vOriginToBeMap.(map[string]interface{}); !ok {
+		return errors.New("expected value to be a map")
+	}
+
+	mBytes, err := json.Marshal(vOriginToBeMap)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(mBytes, vDest)
+}
--- a/pkg/doc/jwt/jwt.go
+++ b/pkg/doc/jwt/jwt.go
@@ -1,0 +1,280 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package jwt
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"reflect"
+
+	"github.com/square/go-jose/v3/json"
+	"github.com/square/go-jose/v3/jwt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
+)
+
+const (
+	// TypeJWT defines JWT type
+	TypeJWT = "JWT"
+
+	// AlgorithmNone used to indicate unsecured JWT
+	AlgorithmNone = "none"
+)
+
+// Claims defines JSON Web Token Claims (https://tools.ietf.org/html/rfc7519#section-4)
+type Claims jwt.Claims
+
+// jwtParseOpts holds options for the JWT parsing.
+type parseOpts struct {
+	detachedPayload []byte
+	sigVerifier     jose.SignatureVerifier
+}
+
+// ParseOpt is the JWT Parser option.
+type ParseOpt func(opts *parseOpts)
+
+// WithJWTDetachedPayload option is for definition of JWT detached payload.
+func WithJWTDetachedPayload(payload []byte) ParseOpt {
+	return func(opts *parseOpts) {
+		opts.detachedPayload = payload
+	}
+}
+
+// WithSignatureVerifier option is for definition of JWT detached payload.
+func WithSignatureVerifier(signatureVerifier jose.SignatureVerifier) ParseOpt {
+	return func(opts *parseOpts) {
+		opts.sigVerifier = signatureVerifier
+	}
+}
+
+type signatureVerifierFunc func(joseHeaders jose.Headers, payload, signingInput, signature []byte) error
+
+func (v signatureVerifierFunc) Verify(joseHeaders jose.Headers, payload, signingInput, signature []byte) error {
+	return v(joseHeaders, payload, signingInput, signature)
+}
+
+func verifyUnsecuredJWT(joseHeaders jose.Headers, _, _, signature []byte) error {
+	alg, ok := joseHeaders.Algorithm()
+	if !ok {
+		return errors.New("alg is not defined")
+	}
+
+	if alg != AlgorithmNone {
+		return errors.New("alg value is not 'none'")
+	}
+
+	if len(signature) > 0 {
+		return errors.New("not empty signature")
+	}
+
+	return nil
+}
+
+// UnsecuredJWTVerifier provides verifier for unsecured JWT.
+func UnsecuredJWTVerifier() jose.SignatureVerifier {
+	return signatureVerifierFunc(verifyUnsecuredJWT)
+}
+
+type unsecuredJWTSigner struct {
+	extraHeaders map[string]interface{}
+}
+
+func (s unsecuredJWTSigner) Sign(_ []byte) ([]byte, error) {
+	return []byte(""), nil
+}
+
+func (s unsecuredJWTSigner) Headers() jose.Headers {
+	jHeaders := map[string]interface{}{
+		jose.HeaderAlgorithm: AlgorithmNone,
+		jose.HeaderType:      TypeJWT,
+	}
+
+	for k, v := range s.extraHeaders {
+		if _, ok := jHeaders[k]; !ok {
+			jHeaders[k] = v
+		}
+	}
+
+	return jHeaders
+}
+
+// JSONWebToken defines JSON Web Token (https://tools.ietf.org/html/rfc7519)
+type JSONWebToken struct {
+	Headers jose.Headers
+
+	Payload map[string]interface{}
+
+	signature []byte
+}
+
+// Parse parses input JWT in serialized form into JSON Web Token.
+// Currently JWS and unsecured JWT is supported.9
+func Parse(jwtSerialized string, opts ...ParseOpt) (*JSONWebToken, error) {
+	if !jose.IsCompactJWS(jwtSerialized) {
+		return nil, errors.New("JWT of compacted JWS form is supported only")
+	}
+
+	pOpts := &parseOpts{}
+
+	for _, opt := range opts {
+		opt(pOpts)
+	}
+
+	return parseJWS(jwtSerialized, pOpts)
+}
+
+// DecodeClaims fills input c with claims of a token.
+func (j *JSONWebToken) DecodeClaims(c interface{}) error {
+	pBytes, err := json.Marshal(j.Payload)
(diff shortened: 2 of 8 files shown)


pkg/doc/jwt/claims.go
```diff
@@ -0,0 +1,123 @@
...
+	return &nd
+}
+
+// MarshalJSON converts NumericDate to JSON bytes.
+func (n NumericDate) MarshalJSON() ([]byte, error) {
+	return []byte(strconv.FormatInt(int64(n), 10)), nil
+}
+
+// UnmarshalJSON defines custom unmarshalling of NumericDate from JSON bytes.
+func (n *NumericDate) UnmarshalJSON(data []byte) error {
```
troyronda: this seems like a constructor rather than a member method.

pkg/doc/jwt/claims.go
```diff
@@ -0,0 +1,123 @@
...
+func (s Audience) MarshalJSON() ([]byte, error) {
+	if len(s) == 1 {
+		return json.Marshal(s[0])
+	}
+
+	return json.Marshal([]string(s))
+}
+
+// UnmarshalJSON defines custom unmarshalling of Audience from JSON bytes.
+func (s *Audience) UnmarshalJSON(b []byte) error {
```
troyronda: this seems like a constructor rather than a member function.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+
+	// AlgorithmNone used to indicate unsecured JWT
+	AlgorithmNone = "none"
+
+	// SignatureEdDSA defines EdDSA alg
+	SignatureEdDSA = "EdDSA"
+
+	// SignatureRS256 defines RS256 alg
+	SignatureRS256 = "RS256"
+)
```
troyronda: we also need ECDSA to support NIST P256 curve.

pkg/doc/jwt/claims.go
kdimak: It implements `json.Unmarshaler` interface so that custom `Audience` unmarshalling is enabled.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+
+// Header makes look up of particular header.
+func (j *jsonWebToken) Header(name string) (interface{}, bool) {
+	v, ok := j.headers[name]
+
+	return v, ok
+}
+
+// StringHeader makes look up of particular header with string value.
+func (j *jsonWebToken) StringHeader(name string) string {
```
troyronda: Maybe HeaderString?

pkg/doc/jwt/jwt.go
kdimak: Are you OK if I will create a separate PR for that? The same for `secp256k1`.

pkg/doc/jwt/claims.go
kdimak: It implements `json.Unmarshaler` interface so that custom `NumericDate` unmarshalling is enabled.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+// Signature returns a copy of JWT signature if defined.
+func (j *jsonWebToken) Signature() []byte {
+	if j.signature == nil {
+		return nil
+	}
+
+	return append(j.signature[:0:0], j.signature...)
+}
+
+// Signer defines JWT Signer interface. It makes signing of data and provides custom JWT headers relevant to the signer.
```
troyronda: we also need pluggable verification so the crypto primitives can be pluggable.

kdimak: `Verifier` interface is defined at separate file `verifier.go`. There is one basic implementation of it created by `NewVerifier()`. It now supports EdDSA and RS256. And currently, there is not way to plug in an extra concrete algorithm verifier. @troyronda  do you want to make it pluggable at this point?

pkg/doc/jwt/jwt.go
troyronda: ok

pkg/doc/jwt/jwt.go
troyronda: we will need to make it pluggable, yes.
(and the default/basic implementations can be overridden.)

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+type Signer interface {
```
troyronda: what is the plan for JWE? I assume followup?

pkg/doc/jwt/jwt.go
kdimak: https://github.com/hyperledger/aries-framework-go/issues/1266

pkg/doc/jwt/jwt.go
kdimak: https://github.com/hyperledger/aries-framework-go/issues/1267

pkg/doc/jwt/jwt.go
kdimak: I did not implement it because it's now needed by `verifiable` package for now.

pkg/doc/jwt/verifier.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Verifier makes verification of JSON Web Token.
+type Verifier interface {
+
+	// Verify verifies JWT.
+	Verify(j JSONWebToken) error
+}
+
+// KeyResolver resolves public key based on what and kid.
+type KeyResolver interface {
```
Baha-sk: the new Crypto interface and default Tink implementation is already in place, see interface here:
https://github.com/hyperledger/aries-framework-go/blob/master/pkg/crypto/api.go

you can take advantage of that to Sign/Verify using KeyHandle (see https://github.com/hyperledger/aries-framework-go/blob/master/pkg/crypto/tinkcrypto/crypto_test.go#L107 for example of signing and verifying)

soon, the KMS implementation will also be pushed so we can take advantage of getting managed keys (without passing in raw key `[]byte`).

pkg/doc/jwt/jwt.go
Baha-sk: JWE is a layer above simple crypto operations, I see it more at the level of Pack/Unpack where Anoncrypt and Authcrypt implement it.

pkg/doc/jwt/verifier.go
troyronda: I think we shouldn't be tightly bound to a KMS interface.

The document packages, JWT, JWE, etc. should take generic signer, encrypter, etc. interfaces and the framework should construct those from KMS.

This way the packages can be used independently from the rest of the framework but still take advantage of framework services like KMS.

Baha-sk: the Crypto api can also be split by operation (eg Signer, Verifier, Encrypter, Decrypter interfaces)

pkg/doc/jwt/support_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func newRS256SingleKeySigner(privKey *rsa.PrivateKey, headers map[string]interface{}) *rs256SingleKeySigner {
+	return &rs256SingleKeySigner{
+		privKey: privKey,
+		headers: PrepareJWSHeaders(headers, SignatureRS256),
+	}
+}
+
+func (s rs256SingleKeySigner) Sign(data []byte) ([]byte, error) {
+	hash := crypto.SHA256.New()
```
troyronda: @Baha-sk does tink also supply hashing primitives?

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+		return nil, b.err
+	}
+
+	return &jsonWebToken{
+		headers: PrepareJWTUnsecuredHeaders(headers),
+		payload: b.claims,
+	}, nil
+}
+
+// New initiates creation of a new JSON Web Tokne. claims could be anything marshallable to the map.
```
troyronda: typo Tokne

troyronda:
@kdimak another consideration: we will want to re-use common parts for our JWM implementation (https://tools.ietf.org/html/draft-looker-jwm-01). Example - JWS and JWE.

pkg/doc/jwt/jwt.go
troyronda: Let’s remember that the JWE package should be useable beyond Pack/Unpack (/JWM).

We also need to share the JWS implementation with JWM (see below)

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+	HeaderString(name string) string
+
+	// Signature returns a signature if defined.
+	Signature() []byte
+
+	// CompactSerialize makes compact serialization of token.
+	CompactSerialize(detached bool) (string, error)
+}
+
+type jsonWebToken struct {
```
troyronda: It would be nice to make this generic between JWM and JWT.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+	// HeaderKeyID defines kid header of string type
+	HeaderKeyID = "kid" // string
+
+	// TypeJWT defines JWT type
+	TypeJWT = "JWT"
+
+	// AlgorithmNone used to indicate unsecured JWT
+	AlgorithmNone = "none"
+
+	// SignatureEdDSA defines EdDSA alg
```
troyronda: Once both signer and verifier are externalized, should these constants be moved to the test files?

(Other than none, of course)

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+func New(claims interface{}) Builder {
+	m, err := toMap(claims)
+
+	return &builder{
+		claims: m,
+		err:    err,
+	}
+}
+
+// FromSigned parses JWS in compact serialization format.
```
troyronda: So the calling function should validate the alg header field?

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+
+	err = json.Unmarshal(payload, &jwtPayload)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal JSON payload: %w", err)
+	}
+
+	return &jsonWebToken{
+		headers:   jwtHeaders,
+		payload:   jwtPayload,
+		signature: nil,
```
troyronda: FYI: Not needed - zero value.

pkg/doc/jwt/claims.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+// Time provides time.Time value of NumericDate.
+func (n NumericDate) Time() time.Time {
+	return time.Unix(int64(n), 0)
+}
+
+// Audience identifies the recipients that the JWT is intended for.
+type Audience []string
```
troyronda: The below code doesn't look "Audience" specific. Why not have a generic type for a String slice that can unmarshal both a single string element or an array of strings?

pkg/doc/jwt/claims.go
```diff
@@ -0,0 +1,123 @@
+/*
+ *
+ *  * Copyright SecureKey Technologies Inc. All Rights Reserved.
```
troyronda: the license header doesn't look normal.

pkg/doc/jwt/claims.go
troyronda: I feel like we shouldn't need to export these marshal/unmarshal methods. i.e., in other packages we have the raw pattern to enable us to hide such details.

pkg/doc/jwt/claims.go
```diff
@@ -0,0 +1,123 @@
...
+	// IssuedAt defines iat claim.
+	IssuedAt *NumericDate `json:"iat,omitempty"`
+
+	// ID defines jti claim.
+	ID string `json:"jti,omitempty"`
+}
+
+// NumericDate is a JSON numeric value representing the number of seconds from
+// 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds.
+type NumericDate int64
```
troyronda: To me, this is the marshalled type... not the data model type (which I assume should be a Time value).

pkg/doc/jwt/claims.go
```diff
@@ -0,0 +1,123 @@
+/*
```
troyronda: These are specifically JWT registered claims (aka JWT's data model). This file doesn't actually handle all claims in the JWT.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+
+	// SignatureEdDSA defines EdDSA alg
+	SignatureEdDSA = "EdDSA"
+
+	// SignatureRS256 defines RS256 alg
+	SignatureRS256 = "RS256"
+)
+
+const (
+	jwtPartsNumber   = 3
```
troyronda: what's a PartsNumber?

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+	jwtHeaderPart    = 0
+	jwtPayloadPart   = 1
+	jwtSignaturePart = 2
+)
+
+// JSONWebToken defines JWT basic operations with JWT headers, payload and signature.
+type JSONWebToken interface {
+
+	// Claims fills input c with claims. It could be structure (e.g. Claims or its extension) or map[string]interface{}.
```
troyronda: I think we have prefixed such methods with `Decode` in recent code. See the messenger code.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+	Claims(c interface{}) error
+
+	// VerificationData build verification data which can be used for signing or verification.
+	VerificationData() ([]byte, error)
+
+	// Header makes look up of particular header.
+	Header(name string) (interface{}, bool)
```
troyronda: This seems like a `Lookup` method.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+type Signer interface {
+	// Sign signs.
+	Sign(data []byte) ([]byte, error)
+
+	// Headers provides JWT headers.
+	Headers() map[string]interface{}
+}
+
+// Builder construct JSON Web Token.
+type Builder interface {
```
troyronda: I'm curious why we don't have a different builder interface for each type?

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+
+	return &builder{
+		claims: m,
+		err:    err,
+	}
+}
+
+// FromSigned parses JWS in compact serialization format.
+// verifier can be nil, in this case signature verification is not made.
+func FromSigned(jws string, verifier Verifier, detachedPayload []byte) (JSONWebToken, error) {
```
troyronda: it might be an idea to separate JWS methods into their own file.

troyronda: The other comment is we will eventually need to have both JWT and JWM become signed as JWS.

pkg/doc/jwt/jwt_test.go
```diff
@@ -0,0 +1,749 @@
...
+	"crypto/rsa"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/square/go-jose/v3/jwt"
```
troyronda: suggest we separate tests against the square library into their own file.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,448 @@
...
+	return append(j.signature[:0:0], j.signature...)
+}
+
+// Signer defines JWT Signer interface. It makes signing of data and provides custom JWT headers relevant to the signer.
+type Signer interface {
+	// Sign signs.
+	Sign(data []byte) ([]byte, error)
+
+	// Headers provides JWT headers.
+	Headers() map[string]interface{}
```
troyronda: which headers need to be provided by the signer?

pkg/doc/jwt/verifier.go
```diff
@@ -0,0 +1,116 @@
...
+	if alg == "" {
+		return errors.New("JWS alg is not defined")
+	}
+
+	verificationData, err := j.VerificationData()
+	if err != nil {
+		return fmt.Errorf("build JWS verification data: %w", err)
+	}
+
+	switch alg {
```
troyronda: TODO: the verifiers should be registered against alg in the constructor (allowing for default implementation).

kdimak:
> @kdimak another consideration: we will want to re-use common parts for our JWM implementation (https://tools.ietf.org/html/draft-looker-jwm-01). Example - JWS and JWE should be mostly common between JWT and JWM.

I need to familiarize myself with JWM, but from I read at JWM overview it's close to the JWT. Let's create a new issue regarding JWM.

pkg/doc/jwt/jwt.go
kdimak: OK, it makes sense.

pkg/doc/jwt/jwt.go
kdimak: They are used in `basicVerifier` implementation of `Verifier`. I think we should move these constants to the `verifier.go`. And probably make them internal and they won't be used externally.

pkg/doc/jwt/jwt.go
kdimak: I think you meant how a caller distinguishes between `FromSigned` and `FromUnsecured`? I created these two methods because for the latter the `Verifier` is not needed. And I think that unsecured JWT is a very special case. go-jose even does not want to support that! So I think the caller does know if it's JWS or unsecured JWT.

pkg/doc/jwt/jwt.go
kdimak: I know :) I just wanted to emphasize this point - that we have no signature in the case of Unsecured JWT.

pkg/doc/jwt/claims.go
kdimak: Makes sense.

pkg/doc/jwt/claims.go
kdimak: Ah. Maybe I placed a wrong template for the license setting in my IDE. Thanks for the catch!

pkg/doc/jwt/claims.go
kdimak: I will check this...

pkg/doc/jwt/claims.go
kdimak: These are the base claims. A user can extend `Claims` (e.g. via struct embedding). Or you mean we need to extend the `Claims` itself with some extra fields? I added the fields based on the https://tools.ietf.org/html/rfc7519#section-4.1.

pkg/doc/jwt/jwt.go
kdimak: It's a number of the compact serialized JWT (separated by ".").

pkg/doc/jwt/jwt.go
kdimak: I will take a look.

pkg/doc/jwt/jwt.go
kdimak: Yes. I will change that.

pkg/doc/jwt/jwt.go
kdimak: Did you mean to have just e.g. `NewSigned()` instead of `New().Signed()`? I think it makes sense (if I understand this correctly).

pkg/doc/jwt/jwt.go
kdimak: Agree.

pkg/doc/jwt/jwt_test.go
kdimak: Good idea. It becomes a sort of integration test. I will separate it.

pkg/doc/jwt/jwt.go
kdimak: OK.

pkg/doc/jwt/verifier.go
kdimak: I was thinking about something similar! Added a reference to your comment to https://github.com/hyperledger/aries-framework-go/issues/1267.

kdimak:
After talking to @troyronda , I think it's worth to re-use claims-related code from go-jose (https://github.com/square/go-jose/blob/master/jwt/claims.go). Our implementation is suspiciously similar to the go-jose one :)

pkg/doc/jwt/jwt.go
troyronda: I think the Number suffix confuses a bit - perhaps Count or just leave it off.

pkg/doc/jwt/jwt.go
troyronda: Ok - maybe add a comment too.

kdimak: I removed `nil` assignment.

pkg/doc/jwt/jwt.go
kdimak: `Claims()` renamed to `DecodeClaims()`.

pkg/doc/jwt/jwt.go
kdimak: `Header()` renamed to `LookupHeader()`, `HeaderString()` renamed to `LookupStringHeader()`.

pkg/doc/jwt/jwt.go
kdimak: Done.

kdimak:
Created a follow-up issue to replace all the Signing/Verification logic with calls to the Crypto package (using Tink underneath) https://github.com/hyperledger/aries-framework-go/issues/1278
@Baha-sk

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,457 @@
...
+)
+
+const (
+	jwtPartsCount    = 3
+	jwtHeaderPart    = 0
+	jwtPayloadPart   = 1
+	jwtSignaturePart = 2
+)
+
+// JSONWebToken defines JWT basic operations with JWT headers, payload and signature.
```
troyronda: I do not understand why we are declaring interfaces and hidden structs in the same package. These interfaces seem unnecessary.

pkg/doc/jwt/jwt.go
troyronda: I don’t understand why we even have a Builder interface.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	return &jsonWebToken{
+		headers: PrepareJWTUnsecuredHeaders(headers),
+		payload: b.claims,
+	}, nil
+}
+
+// New initiates creation of a new JSON Web Token. claims could be anything marshallable to the map.
+func New(claims interface{}) Builder {
```
troyronda: what is the purpose of returning an interface?

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,457 @@
...
+type Builder interface {
+	// Signed creates JWS.
+	Signed(signer Signer) (JSONWebToken, error)
+
+	// Unsecured creates unsecured JWT.
+	Unsecured(headers map[string]interface{}) (JSONWebToken, error)
+}
+
+// builder implements Builder.
+type builder struct {
```
troyronda: this struct isn't acting as a builder pattern. (nor is the associated interface.)

pkg/doc/jwt/jwt.go
kdimak: Fixed

pkg/doc/jwt/jwt.go
kdimak: Changed.

pkg/doc/jwt/jwt.go
kdimak: Corrected.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+
+	if _, ok := headers[HeaderType]; !ok {
+		headers[HeaderType] = TypeJWT
+	}
+
+	return nil
+}
+
+// Unsecured creates unsecured JWT.
```
troyronda: why are we returning interfaces rather than struct pointers?

troyronda:
Some thoughts @kdimak @Baha-sk :
* We should have a `jose` package that is agnostic to JWT and JWM. This package performs JSON signatures, JSON encryption and nested JSON signature and encryption. Crypto primitives and keys are always injected via dependency. Presumably this package operates on a generic payload map (somewhat similar to the DIDComm message structure) plus JWS/JWE header structures.
* We should have a `jwt` and `jwm` packages that structure and validate the standard fields from each spec. We must be able to provide the output of either package to `jose`.
* What parts of the go-jose project should we be using?

pkg/doc/jwt/verifier_test.go
```diff
@@ -0,0 +1,173 @@
...
+import (
+	"crypto"
+	"crypto/ed25519"
+	"crypto/rand"
+	"crypto/rsa"
+	"errors"
+	"testing"
+
+	"github.com/square/go-jose/v3/jwt"
+
```
soluchok: extra line

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,459 @@
...
+	newHeaders[HeaderAlgorithm] = AlgorithmNone
+
+	return newHeaders
+}
+
+// IsCompactJWS checks that input is a JWS in compact form.
+func IsCompactJWS(s string) bool {
+	parts := strings.Split(s, ".")
+
+	return len(parts) == 3 &&
```
soluchok: I assume `3` can be replaced with `jwtPartsCount`

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,459 @@
...
+
+	for k, v := range headers {
+		newHeaders[k] = v
+	}
+
+	newHeaders[HeaderType] = TypeJWT
+	newHeaders[HeaderAlgorithm] = AlgorithmNone
+
+	return newHeaders
+}
```
soluchok: Looks like we do not need this function. Same can be achieved by executing  `PrepareJWSHeaders(headers, AlgorithmNone)`

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,459 @@
...
+
+	if verifier != nil {
+		err = verifier.Verify(jwt)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return jwt, nil
+}
```
soluchok: nit: better to avoid nested conditions
```
if verifier == nil {
 return jwt, nil
}

if err = verifier.Verify(jwt); err != nil {
    return nil, err
}

return jwt, nil
```

pkg/doc/jwt/jwt.go
kdimak: Right!

pkg/doc/jwt/jwt.go
kdimak: Yeah

pkg/doc/jwt/jwt.go
kdimak: agree

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,273 @@
...
+func New(claims interface{}) (*JSONWebToken, error) {
+	m, err := toMap(claims)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshallable claims: %w", err)
+	}
+
+	return &JSONWebToken{
+		Headers:   nil,
+		Payload:   m,
+		signature: nil,
```
soluchok: remove these `nil` definitions

pkg/doc/jwt/verifier.go
```diff
@@ -0,0 +1,150 @@
...
+// VerifyEdDSA verifies EdDSA signature.
+func VerifyEdDSA(pubKey interface{}, message, signature []byte) error {
+	// TODO Use crypto for signing/verification logic
+	//  https://github.com/hyperledger/aries-framework-go/issues/1278
+	pubKeyEdDSA, ok := pubKey.([]byte)
+	if !ok {
+		pubKeyEdDSA, ok = pubKey.(ed25519.PublicKey)
+		if !ok {
+			return errors.New("not []byte or *ed25519.PublicKey public key")
+		}
```
soluchok: The error message is not correct. It says `*ed25519.PublicKey` but you are checking ` pubKey.(ed25519.PublicKey)`

pkg/doc/jose/jws.go
```diff
@@ -0,0 +1,382 @@
...
+
+	for k, v := range h2 {
+		if _, ok := h[k]; !ok {
+			h[k] = v
+		}
+	}
+
+	return h
+}
+
```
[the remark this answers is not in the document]
kdimak: Good idea!

pkg/doc/jwt/jwt.go
kdimak: yeah, done, thanks

pkg/doc/jwt/verifier.go
kdimak: corrected

pkg/doc/jose/jws.go
```diff
@@ -0,0 +1,380 @@
...
+)
+
+// Headers represents JOSE headers.
+type Headers map[string]interface{}
+
+// JWK (JSON Web Key) is a JSON data structure that represents a cryptographic key.
+type JWK jose.JSONWebKey
+
+// GetJWK gets JWK from JOSE headers.
+func (jh Headers) GetJWK() (*JWK, bool) {
```
troyronda: Why do we prefix these methods with Get? We don't seem to be getting anything.

kdimak: Agree. I will rename it to `JWK()`. The same applied for the `KeyID()` and `Algorithm()`.

pkg/doc/jose/jws.go
```diff
@@ -0,0 +1,380 @@
...
+// NewCompositeAlgSignatureVerifier creates a new CompositeAlgSignatureVerifier
+func NewCompositeAlgSignatureVerifier() *CompositeAlgSignatureVerifier {
+	return &CompositeAlgSignatureVerifier{
+		verifierByAlg: make(map[string]SignatureVerifier),
+	}
+}
+
+// AddVerifier adds a new verifier of algorithm.
+// If algorithm verifier is already present, it's get rewritten.
+func (v *CompositeAlgSignatureVerifier) AddVerifier(
```
troyronda: I assume we expect to add to this structure over time instead of just constructing with Verifiers?

pkg/doc/jose/jws.go
```diff
@@ -0,0 +1,380 @@
...
+	return &JSONWebSignature{
+		ProtectedHeaders:   protectedHeaders,
+		UnprotectedHeaders: unprotectedHeaders,
+		Payload:            payload,
+	}
+}
+
+// SerializeCompact makes JWS Compact Serialization (https://tools.ietf.org/html/rfc7515#section-7.1)
+func (s *JSONWebSignature) SerializeCompact(signer Signer, detached bool) (string, error) {
+	err := s.sign(signer)
```
troyronda: It's odd that the object is mutated during serialization. Shouldn't this happen prior?

pkg/doc/jose/jws.go
```diff
@@ -0,0 +1,380 @@
...
+		b64Signature), nil
+}
+
+// Signature returns a copy of JWS signature.
+func (s *JSONWebSignature) Signature() []byte {
+	if s.signature == nil {
+		return nil
+	}
+
+	return append(s.signature[:0:0], s.signature...)
```
troyronda: please elaborate on this.

pkg/doc/jose/jws.go
```diff
@@ -0,0 +1,380 @@
...
+package jose
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/square/go-jose/v3"
+	"github.com/square/go-jose/v3/json"
```
troyronda: we should probably comment on why we need a fork of json encoders.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,271 @@
...
+
+package jwt
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"reflect"
+
+	"github.com/square/go-jose/v3/json"
```
troyronda: we should probably comment on why we need a fork of json encoders.

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func (j *JSONWebToken) serialize(signer jose.Signer, detached bool) (string, error) {
+	payloadBytes, err := j.marshalPayload()
+	if err != nil {
+		return "", fmt.Errorf("marshal JWT claims: %w", err)
+	}
+
+	jws := jose.NewJWS(j.Headers, nil, payloadBytes)
+	j.signature = jws.Signature()
```
troyronda: So this seems a bit odd currently. When I reviewed this code previously, I thought signature wasn't populated by the constructor ... and hence this would end up returning nil?

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,271 @@
...
+	return j.serialize(&unsecuredJWTSigner{extraHeaders}, detached)
+}
+
+func (j *JSONWebToken) serialize(signer jose.Signer, detached bool) (string, error) {
+	payloadBytes, err := j.marshalPayload()
+	if err != nil {
+		return "", fmt.Errorf("marshal JWT claims: %w", err)
+	}
+
+	jws := jose.NewJWS(j.Headers, nil, payloadBytes)
```
troyronda: unprotected headers are always nil?

pkg/doc/jose/jws.go
kdimak: Yeah, it makes sense for me. So we will have the following steps:

```
jws := NewJWS(...)

jws.Sign(signer)

jwsCompacted := jws.SerializeCompact(false) // no detach
```

This will also help to support multiple signings (e.g. for JWS JSON Serialization).

pkg/doc/jwt/jwt.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	typ, ok := headers[jose.HeaderType]
+	if ok && typ != TypeJWT {
+		return errors.New("typ is not JWT")
+	}
+
+	cty, ok := headers[jose.HeaderContentType]
+	if ok && cty == TypeJWT {
+		return errors.New("nested JWT is not supported")
```
troyronda: what is this case?

pkg/doc/jwt/verifier.go
```diff
@@ -0,0 +1,150 @@
...
+	"golang.org/x/crypto/ed25519"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
+)
+
+const (
+	// signatureEdDSA defines EdDSA alg
+	signatureEdDSA = "EdDSA"
+
+	// signatureRS256 defines RS256 alg
```
troyronda: it would be good to add a NIST P256 test case.

pkg/doc/jwt/jwt.go
kdimak: Yes, as JWT uses only JWS Compact serialization which builds JOSE headers and verification data from **protected** headers only.

pkg/doc/jose/jws.go
troyronda: This is better. I'm trying to better understand why we don't sign during construction.

pkg/doc/jwt/jwt.go
troyronda: ok. can you please add a comment and citation?

pkg/doc/jose/jws.go
kdimak: I changed it to the following:

```
jws := NewJWS(..., signer)

jwsCompacted := jws.SerializeCompact(false) // no detach
```

pkg/doc/jose/jws.go
kdimak: I agree, no sense to make it mutable. I removed `AddVerifier()` method.

pkg/doc/jose/jws.go
kdimak: Substituted with the usage of conventional `copy()`.

pkg/doc/jwt/jwt.go
kdimak: https://tools.ietf.org/html/rfc7519#section-5.2
I added a reference.

pkg/doc/jwt/jwt.go
kdimak: Sure! Added a reference to https://tools.ietf.org/html/rfc7515#section-3.1.

pkg/doc/jwt/verifier.go
kdimak: there is a TODO at `NewVerifier()` which references https://github.com/hyperledger/aries-framework-go/issues/1266.

pkg/doc/jwt/jwt.go
kdimak: not relevant anymore as `JSONWebToken` is made immutable.

Issue #35, opened 2019-07-25
Peer DIDs


llorllale:
_Peer DIDs_ are non-anchored DIDs shared directly between N parties. The Peer DID specification reserves `peer` as the method spec.

Peer DIDs are used to preserve privacy by making it harder for attackers to correlate your identity across interactions with different parties.

Agents need the ability to create and share Peer DIDs during the `did-exchange` protocol. They also need to synchronize state of these peer DIDs with each other:

* state is managed as a sequence of signed patches
* changes in state are communicated via the `sync-state` protocol

The Peer DIDs can be resolved in two ways:

* Resolution can directly query the owning Agent with a "resolve" message
* Resolution can query its local storage or cache

In all cases, the source against which the Peer DID will be resolved needs to maintain patches (as opposed to just keeping an "up-to-date" copy) because:

* It's required to maintain proof of authority over the evolution of the DID document
* Following from previous point: both the owning Agent and the Agent(s) to which it has been shared need to merge those patches and resolve any conflicts (there is a section on conflict resolution in the Peer DID spec).

(Interestingly, the peer DID spec defines several "levels of support" for peer DIDs)

**References**
* [Peer DIDs](https://openssi.github.io/peer-did-method-spec/index.html)
* [DID Resolution](https://w3c-ccg.github.io/did-resolution/
* [sync-state protocol](https://github.com/hyperledger/aries-rfcs/blob/master/features/0030-sync-connection/README.md#sync_state)

PR #1273, opened 2020-02-14
feat: mimetype for REST controller response

### LINKED ISSUE #1272: REST controllers response mimetype
mime-type in the response from REST controllers should be application/json.

sudeshrshetty:
- added `application/json` mimetype for REST controller response
- closes #1272

### DIFF
--- a/pkg/controller/rest/rest.go
+++ b/pkg/controller/rest/rest.go
@@ -31,6 +31,8 @@ func Execute(exec command.Exec, rw http.ResponseWriter, req io.Reader) {
 	if err != nil {
 		SendError(rw, err)
 	}
+
+	rw.Header().Set("Content-Type", "application/json")
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/rest/rest.go
```diff
@@ -59,7 +61,6 @@ func SendError(rw http.ResponseWriter, err command.Error) {
 // SendHTTPStatusError sends given http status code to response with error body
 func SendHTTPStatusError(rw http.ResponseWriter, httpStatus int, code command.Code, err error) {
 	rw.WriteHeader(httpStatus)
-	rw.Header().Set("Content-Type", "application/json")
```
troyronda: why is this removed?

sudeshrshetty: line #35 always gets called, so it gets set twice in case of errors

PR #1276, opened 2020-02-14
feat: #1268 aries-js-worker - parametrize assets path

llorllale:

### DIFF
--- a/cmd/aries-js-worker/src/aries.js
+++ b/cmd/aries-js-worker/src/aries.js
@@ -26,12 +26,12 @@ const __publicPath = _ => {
 
 __webpack_public_path__ = __publicPath()
 
-const { _getWorker } = require("worker_loader")
+const { loadWorker } = require("worker_loader")
 
 // TODO synchronize access on this map?
 const PENDING = new Map()
 const NOTIFICATIONS = new Map()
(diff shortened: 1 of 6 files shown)


cmd/aries-js-worker/src/aries.js
```diff
@@ -99,6 +99,24 @@ export const Aries = function(opts) {
...
+    if (!opts.assetsPath) {
+        throw new Error("aries: missing assets path")
+    }
+
+    WORKER = loadWorker(
+        PENDING,
+        NOTIFICATIONS,
+        {
+            dir: opts.assetsPath,
+            wasm: opts.assetsPath + "/aries-js-worker.wasm.gz",
```
troyronda: we should also allow configuration of whether or not the wasm was pre-compressed.
some servers prefer uncompressed and then they returned the compressed encoding as needed.

llorllale: new issue?

troyronda: sure.

(the other idea would be to allow customization of the filename including extension).

troyronda: #1277

PR #1283, opened 2020-02-19
feat: aries js worker start from constructor

sudeshrshetty:
- removed aries startup from start() to constructor
- aries stop() replaced by destroy function
- each instance will have localized worker, message and notification map
- fixed ununsed startup opts

TODO - wait for aries startup before returning or using
event/callback/flag to indicate if aries started successfully.

### DIFF
--- a/cmd/aries-js-worker/index.html
+++ b/cmd/aries-js-worker/index.html
@@ -13,25 +13,49 @@ This is a test page for developers to test WASM integration.
     <meta charset="utf-8"/>
     <script src="dist/web/aries.js"></script>
     <script>
-        // TODO fix the stutter in the API's exports ("Aries.Aries....")
-        const aries = Aries.Aries({assetsPath: "/dist/assets"})
+        var aries
         var stopNotifier
 
-        async function handleStart() {
+        async function startAries() {
+            if (aries) {
+                console.error("aries is already initialized")
+                document.querySelector("#output").value = "aries is already initialized"
+                return
+            }
+
             const request = document.querySelector("#opts").value
-            const response = await aries.start(request)
-            document.querySelector("#output").value = JSON.stringify(response)
+
+            // TODO fix the stutter in the API's exports ("Aries.Aries....")
+            aries = await Aries.Aries(JSON.parse(request))
+
+            //TODO wait for event/callback to confirm if aries is started
+            document.querySelector("#output").value = "aries started"
         }
 
-        async function handleStop() {
-            const response = await aries.stop()
+        function destroyAries() {
+            if (!aries) {
+                console.error("aries not initialized")
+                document.querySelector("#output").value = "aries not initialized"
+                return
+            }
+
+            const response = aries.destroy()
+            aries = null
+
             document.querySelector("#output").value = JSON.stringify(response)
         }
 
         async function handleOperationInput() {
+            if (!aries) {
+                console.error("aries not initialized")
+                document.querySelector("#output").value = "aries not initialized"
+                return
+            }
+
             const request = document.querySelector("#input").value
             const operation = document.querySelector("#operation").value
             const method = document.querySelector("#method").value
+
             const response = await aries[operation][method](request)
             document.querySelector("#output").value = JSON.stringify(response)
         }
@@ -45,6 +69,7 @@ This is a test page for developers to test WASM integration.
             }
 
             stopNotifier =  aries.startNotifier(pushToMsgPanel, topics)
+            document.querySelector("#output").value = "notifier started"
         }
 
         function pushToMsgPanel(val) {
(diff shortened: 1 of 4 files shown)


cmd/aries-js-worker/index.html
```diff
@@ -13,22 +13,39 @@
...
+        var aries
         var stopNotifier
 
-        async function handleStart() {
+        async function startAries() {
             const request = document.querySelector("#opts").value
-            const response = await aries.start(request)
-            document.querySelector("#output").value = JSON.stringify(response)
+
+            // TODO fix the stutter in the API's exports ("Aries.Aries....")
```
troyronda: the stutter looks even weirder now that we have a constructor function.

cmd/aries-js-worker/index.html
```diff
@@ -13,22 +13,39 @@
...
+            aries = Aries.Aries({assetsPath: "/dist/assets", startOpts: request})
```
troyronda: we don’t await like we did with start?

cmd/aries-js-worker/index.html
```diff
@@ -13,22 +13,39 @@
     <meta charset="utf-8"/>
     <script src="dist/web/aries.js"></script>
     <script>
-        // TODO fix the stutter in the API's exports ("Aries.Aries....")
-        const aries = Aries.Aries({assetsPath: "/dist/assets"})
+        var aries
         var stopNotifier
 
-        async function handleStart() {
+        async function startAries() {
```
troyronda: there isnt currently an await in this func

cmd/aries-js-worker/index.html
```diff
@@ -13,22 +13,39 @@
...
-        async function handleStop() {
-            const response = await aries.stop()
+        async function destroyAries() {
+            if (!aries) {
+                console.error("aries not initialized")
+                document.querySelector("#output").value = "aries not initialized"
+                return
+            }
+
+            const response = aries.destroy()
```
troyronda: we dont await destroy like we did for stop?

cmd/aries-js-worker/index.html
```diff
@@ -13,22 +13,39 @@
...
+            // TODO fix the stutter in the API's exports ("Aries.Aries....")
+            aries = Aries.Aries({assetsPath: "/dist/assets", startOpts: request})
+
+            //TODO wait for event/callback to confirm if aries is started
+            document.querySelector("#output").value = "aries started"
         }
 
-        async function handleStop() {
-            const response = await aries.stop()
+        async function destroyAries() {
```
troyronda: there is not currently an await in this func

cmd/aries-js-worker/index.html
llorllale: all parameters should be passed at the same level as `assetsPath`

cmd/aries-js-worker/src/aries.js
```diff
@@ -184,71 +180,71 @@ export const Aries = function(opts) {
...
             unregister: async function () {
-                return invoke(this.pkgname, "Unregister", "{}", "timeout while registering router")
+                return invoke(aw, pending,  this.pkgname, "Unregister", "{}", "timeout while registering router")
             }
         }
     }
 
-    return INSTANCE
-
+    // TODO wait till aries startup or send event once aries started successfully
```
llorllale: this constructor function should return a promise

cmd/aries-js-worker/index.html
sudeshrshetty: it was same before, I just moved it inside function. Will fix it in upcoming PR

cmd/aries-js-worker/index.html
sudeshrshetty: No, destroy function doesn't return promise, instead it will wait for promise(`aries.stop()`) to terminate worker.

sudeshrshetty: I am going to update upcoming patch where it constructor returns promise. Then will put await here.

sudeshrshetty: constructor to return promise, and added await.

cmd/aries-js-worker/index.html
sudeshrshetty: fixed

cmd/aries-js-worker/index.html
sudeshrshetty: No, destroy function doesn't return promise, instead it will wait for promise(aries.stop()) to terminate worker.

cmd/aries-js-worker/index.html
sudeshrshetty: fixed it

cmd/aries-js-worker/index.html
sudeshrshetty: fixed

cmd/aries-js-worker/src/aries.js
sudeshrshetty: fixed and removed TODO.

cmd/aries-js-worker/src/aries.js
```diff
@@ -133,17 +114,16 @@ export const Aries = function(opts) {
              * @private
              */
             _echo: async function (text) {
-                return invoke("test", "echo", text, "timeout while accepting invitation")
+                return invoke(aw, pending, "test", pending,"echo", text, "timeout while accepting invitation")
```
llorllale: the second `pending` seems an error

sudeshrshetty: will push fix in next PR

PR #1287, opened 2020-02-19
feat: Parse Verifiable Credential without checking proofs

### LINKED ISSUE #1286: Parse Verifiable Credential without checking proofs
`verifiable.NewCredential()` parses VC from input bytes (could be JSON, JWS, unsecured JWT) and checks the proofs (JWS or linked data proof).
We've got a use case when we need to load the VC from a wallet and make just parsing without a proof check (as we trust what is kept in the wallet).
We need to add an API function similar to the `NewCredential()` which makes unverified parsing of VC.

kdimak:
closes #1286

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -697,6 +697,34 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, []byte, e
 	return vc, vcDataDecoded, nil
 }
 
+// NewUnverifiedCredential decodes Verifiable Credential from bytes which could be marshalled JSON or serialized JWT.
+// It does not make a proof check though. Can be used for purposes of decoding of VC stored in a wallet.
+// Please use this function with caution.
+func NewUnverifiedCredential(vcBytes []byte) (*Credential, error) {
+	vcDataDecoded, err := decodeRaw(vcBytes, &credentialOpts{
+		disabledProofCheck: true,
+	})
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -697,6 +697,34 @@ func NewCredential(vcData []byte, opts ...CredentialOpt) (*Credential, []byte, e
 	return vc, vcDataDecoded, nil
 }
 
+// NewUnverifiedCredential decodes Verifiable Credential from bytes which could be marshalled JSON or serialized JWT.
+// It does not make a proof check though. Can be used for purposes of decoding of VC stored in a wallet.
+// Please use this function with caution.
+func NewUnverifiedCredential(vcBytes []byte) (*Credential, error) {
```
troyronda: Note: this is a `Parse` method that doesn't validate.
TODO: review method namings. And options strategy.

(I was also tempted to ask if it is an `Unsafe` method).

kdimak: Yeah, when I named it I recalled our today's call and the naming strategy in regards to `Parse()` ([]byte) and `Decode()` (struct).
I named it as `NewUnverifiedCredential()` in order to be consistent with already present `NewCredential()`. I agree that the naming should be revisited in `verifiable`. I remember we wanted to make the package refactoring after closing major dev issues.
Also, I named it "Unverifiable" to emphasize the contrast to the "verifiable" nature of Credentials.

PR #1328, opened 2020-02-24
chore: Hook the new KMS in the framework and Context

### LINKED ISSUE #1150: Hook the new KMS in framework and context
part of #985

Baha-sk:
closes #1150

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ We aim to provide Go implementations of:
 
 - Decentralized identity standards including [W3C decentralized identifiers](https://w3c.github.io/did-core/) (DIDs), [W3C DID resolution](https://w3c-ccg.github.io/did-resolution/), and [W3C verifiable credentials](https://w3c.github.io/vc-data-model/).
 - Decentralized data communication protocols anchored in DIDs: [DIDComm](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0005-didcomm).
-- A pluggable dependency framework, where implementors can customize primitives via Service Provider Interfaces (SPIs). We have a "batteries included" model where default primitives are included -- such as a key management system (KMS), crypto, data storage, digital hub integration, etc.
+- A pluggable dependency framework, where implementors can customize primitives via Service Provider Interfaces (SPIs). We have a "batteries included" model where default primitives are included -- such as a [key management system (KMS)](docs/kms_secretlock.md), crypto, data storage, digital hub integration, etc.
 
 We aim to enable usage of our protocol implementations in a wide variety of edge and cloud environments including servers, browsers, mobile, and devices.
 API bindings are supplied to enable these environments including:
--- a/docs/kms_secretlock.md
+++ b/docs/kms_secretlock.md
@@ -1,0 +1,181 @@
+# KMS, SecretLock and Crypto capabilities
+
+Keys used in the Aries framework must be secured from being compromised during their usage or storage. This goal can
+be achieved by avoiding the exchange of raw key bytes as an argument of functions requiring keys and instead accept a
+key reference. The functions in question are usually crypto operations that require these keys. One of the libraries 
+that provides a clear separation between keys and crypto operations is [Tink](https://github.com/google/tink). Keys in 
+Tink are referenced by Key Handles and are linked to primitives representing the crypto operation. In a nutshell, a 
+user will get a key handle from a source (or default key templates offered by Tink) and passes it to the crypto function
+that needs the underlying key. Tink is used in the default Crypto and KMS implementations of the framework to hide key 
+content from being accessed externally.
+
+The KMS service should handle how to fetch the key for the crypto operation needed. Therefore, it must never return 
+raw key contents to the user. The default implementation makes use of Tink to get corresponding key handles discussed
+above.
+
+In addition to the Crypto and KMS services, there is a need to secure and lock keys to protect the identity of the
+agent. The SecretLock service is offered in the framework for this purpose. It uses a master key to encrypt keys stored
+locally (or in the cloud) and it can be itself protected, for example, by a passphrase.
+
+To achieve these requirements, the following criteria must be met:
+ - The KMS service should support providing a key reference (or keyset handle) whenever a key is required. It must not
+   return or accept raw contents of keys within the framework and with the user.
+ - The SecretLock service is needed to protect keys by encrypting them when being stored.
+ - The SecretLock service must work with a master key that can be provided by the user.
+ - The master key should also be encrypted to prevent its compromise. This should be optional in the SecretLock. For
+   instance, a user may decide to use an unprotected master key for testing purposes but the recommended way is for the 
+   master key to be secured.
+ - An example of unlocking a master key by the KMS/SecretLock would be the use of a passphrase.
+ - Finally the crypto services must only accept key references provided by the KMS to execute operations. They must not
+   accept raw content of keys as function arguments.
+ 
+The default Crypto service accepts key handles and delegate crypto operations to Tink. These key handles are a  
+reference to real keys which are only available internally to Tink crypto operations. The default KMS service creates 
+and manages these key handles and serves as a provider of these key handles to the crypto service. Although the default
+Crypto and KMS services are implemented using Tink, the framework itself can accept any implementation being supplied 
+to `aries.WithKMS(customKMSImplementaiton)` and `aries.WithCrypto(customCryptoImplementation)`.
+ 
+There are several ways to prepare and use the new KMS and SecretLock services. The below sections describe each path:
+
+## No SecretLock/KMS options
+
+The framework default is no SecretLock or the use of a noop secret lock service. This means keys are
+not securely stored in the keystore. To secure keys storage, one must create a secure SecretLock option and add it to 
+the framework. See next sections for an example on how to pass one as a framework option.
+
+For agents tools that use the framework, eg [agent rest](../cmd/aries-agent-rest) and [js worker](../cmd/aries-js-worker),
+a custom SecretLock default implementation will be created to suit the target environment. More details will be added
+about these in the future.
+
+## Passing in a SecretLock option
+
+The previous way will create a default no SecretLock instance that does not protect keys. It should only be used in tests. 
+The recommended safe way is to create your own secure SecretLock instance by first creating your own encrypted master 
+key file. This key can be generated via password expansion as shown in the example below. Once the file is created with
+the protected master key, you can create your SecretLock service with a reader to the file and the masterLocker instance
+used to encrypt the master key. You can then pass in the created SecretLock as an option to the framework.
+Below is an example code snippet:
+
+```
+package mypackage
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock/local/masterlock/hkdf"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock/local"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
+ 
+)
+
+...
+
+// keySize to be used to create master key
+keySize := sha256.Size
+
+// create, an optional salt
+salt := make([]byte, keySize)
+_, err = rand.Read(salt)
+if err != nil {
+    return err
+}
+
+// create a master lock to protect the master key (salt is optional)
+masterLock, err := hkdf.NewMasterLock(passphrase, sha256.New, salt)
+if err != nil {
+    return err
+}
+
+// generate a random master key
+masterKeyContent := make([]byte, keySize)
+_, err = rand.Read(masterKeyContent)
+if err != nil {
+    return err
+}
+
+// encrypt it
+masterKeyEnc, err := masterLock.Encrypt("", &secretlock.EncryptRequest{
+    Plaintext: string(masterKeyContent)})
+
+// create a file to store it
+file, err := os.OpenFile("path/to/your/encrypted/masterKeyFile", os.O_WRONLY|os.O_CREATE, 0600)
+if err != nil {
+    return err
+}
+
+// write the encrypted master key in the file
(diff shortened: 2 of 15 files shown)


pkg/framework/aries/default.go
```diff
@@ -122,11 +150,102 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
...
+	_, err = file.Write([]byte(masterLockEnc.Ciphertext))
+	if err != nil {
+		return err
+	}
 
 	return nil
 }
 
+func getPassphraseAndSalt(opts *Aries, keySize int) (string, []byte, error) {
+	passphrase := "secretPwd"
```
llorllale: Shouldn't this input be required from the user?

pkg/framework/aries/default.go
```diff
@@ -122,11 +150,102 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
...
+	if opts.salt != "" {
+		if len(opts.salt) != keySize {
+			return "", nil, fmt.Errorf("invalid salt option provided")
+		}
+
+		copy(salt, opts.salt)
+		opts.salt = ""
+	} else {
+		// default -  no salt
+		salt = nil
```
llorllale: Why not generate a new random salt?

Baha-sk: salt is optional, if the user doesn't provide one, the framework shouldn't generate one.

salt and passphrase are not stored anywhere... they are the `key` to unlock the master key which can be stored either in a file or an env variable.. for the default behaviour, the framework will generate a master key and secure it in a file so that future New() calls can still read it (of course the user must pass in the same passphrase and salt options they provided the first time to unlock the master key)

Baha-sk: also notice that `passphrase` and `salt` options are cleared once used to avoid keep in them in memory
see lines 230 and 240 above

pkg/framework/aries/default.go
Baha-sk: to ease the creation of the Aries framework and not impact existing code (namely BDD tests) that don't use the new KMS (and SecretLock), the framework will use this default passphrase in case the passphrase option (and KMS/SecretLock options) are not passed

Baha-sk: to clarify, this is supposed to be temporary until we flip legacyKMS into KMS, then either passphrase or SecretLock options may become mandatory .. it's to be reviewed in the future

pkg/framework/aries/framework.go
```diff
@@ -208,7 +219,23 @@ func WithProtocols(protocolSvcCreator ...api.ProtocolSvcCreator) Option {
...
 func WithLegacyKMS(k api.KMSCreator) Option {
 	return func(opts *Aries) error {
-		opts.kmsCreator = k
+		opts.legacyKMSCreator = k
+		return nil
+	}
+}
+
+// WithPassphrase injects a passphrase used to secure the master key used in a SecretLock service
+func WithPassphrase(s string) Option {
```
troyronda: I don't think this shouldn't be top level - you should be providing your SecretLock. There is a particular SecretLock that is constructed with a passphrase and secret.

pkg/framework/aries/default.go
llorllale: Right, so what should be the default secretlock implementation?

pkg/framework/aries/framework.go
Baha-sk: function removed

pkg/framework/aries/default.go
```diff
@@ -7,8 +7,13 @@ SPDX-License-Identifier: Apache-2.0
 package aries
 
 import (
+	"crypto/sha256"
 	"fmt"
+	"io"
 	"net/http"
+	"os"
+
+	"github.com/google/tink/go/subtle/random"
```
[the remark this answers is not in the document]
Baha-sk: import removed

pkg/framework/aries/default.go
Baha-sk: salt and passphrase are removed from framework.go and default.go

pkg/framework/aries/default.go
Baha-sk: there is no more passphrase being passed as an option.

the default behaviour (if no SecretLock option is passed to `New()`) is to create a SecretLock with a master key file generated and stored `unprotected`  in a file. The default service will use the master key as is (without encryption/decryption of master key) to wrap KMS keys.

For a more secure approach, one must create a SecretLock option with secret expansion to protect the master key prior to storing in a path. Examples are in unit tests and a README document will be pushed soon to explain how to do this.

docs/kms_secretlock.md
```diff
@@ -0,0 +1,143 @@
+# Aries New Secure KMS, SecretLock and Crypto capabilities
```
troyronda: Remove “Aries New Secure” from the title.

docs/kms_secretlock.md
```diff
@@ -0,0 +1,143 @@
...
+
+The aries-framework-go project has a mandate to protect keys from being compromised during usage or storage of keys. 
+This goal can achieved by hiding raw key bytes from being exchanged between the user and the framework. Also, keys
+should be secured and locked by a master key that is knwon to the user and protected by a passphrase. In order
```
troyronda: Paraphrase is only an example protection.

docs/kms_secretlock.md
```diff
@@ -0,0 +1,143 @@
+# Aries New Secure KMS, SecretLock and Crypto capabilities
+
+The aries-framework-go project has a mandate to protect keys from being compromised during usage or storage of keys.
```
troyronda: We should just talk about the goal rather than Aries-framework-go mandates.

cmd/aries-js-worker/package-lock.json
```diff
@@ -446,6 +446,14 @@
       "integrity": "sha512-Wm6ukoaOGJi/73p/cl2GvLjTI5JM1k/O14isD73YML8StrH/7/lRFgmg8nICZgD3bZZvjwCGxtMOD3wWNAu8cg==",
       "dev": true
     },
+    "axios": {
```
troyronda: why was this file change as part of this PR?

README.md
```diff
@@ -15,7 +15,7 @@ We aim to provide Go implementations of:
 
 - Decentralized identity standards including [W3C decentralized identifiers](https://w3c.github.io/did-core/) (DIDs), [W3C DID resolution](https://w3c-ccg.github.io/did-resolution/), and [W3C verifiable credentials](https://w3c.github.io/vc-data-model/).
 - Decentralized data communication protocols anchored in DIDs: [DIDComm](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0005-didcomm).
-- A pluggable dependency framework, where implementors can customize primitives via Service Provider Interfaces (SPIs). We have a "batteries included" model where default primitives are included -- such as a key management system (KMS), crypto, data storage, digital hub integration, etc.
+- A pluggable dependency framework, where implementors can customize primitives via Service Provider Interfaces (SPIs). We have a "batteries included" model where default primitives are included -- such as a [key management system (KMS), secret lock](docs/kms_secretlock.md), crypto, data storage, digital hub integration, etc.
```
troyronda: let's just include secret lock as part of KMS instead of an explicit mention here. 

- it's a bit odd to have a link with two of the items.
- secret lock isn't really a normal term AFAIK.

docs/kms_secretlock.md
```diff
@@ -0,0 +1,143 @@
...
+The aries-framework-go project has a mandate to protect keys from being compromised during usage or storage of keys. 
+This goal can achieved by hiding raw key bytes from being exchanged between the user and the framework. Also, keys
+should be secured and locked by a master key that is knwon to the user and protected by a passphrase. In order 
+to achieve this, the following criteria must be met:
+ - The KMS service should support providing a key reference (or keyset handle) whenever a key is required. It must not
+   return or accept raw contents of keys within the framework and with the user.
+ - A SecretLock service is needed to protect keys by encrypting them when being stored.
+ - The SecretLock service must work with a master key that can be provided by the user.
+ - The master key should also be encrypted to protect it from being compromised.
+ - It should be unlocked by a passphrase known to the user.
```
troyronda: Again - passphrase is only one of the mechanisms.

docs/kms_secretlock.md
```diff
@@ -0,0 +1,143 @@
...
+ - Finally the crypto services must only accept key references provided by the KMS to execute operations.
+ 
+There are several ways to prepare and use the new KMS and SecreatLock services. The below sections describe each path:
```
troyronda: typo. SecreatLock.

pkg/framework/context/context.go
```diff
@@ -247,10 +254,18 @@ func WithProtocolServices(services ...dispatcher.ProtocolService) ProviderOption
 	}
 }
 
-// WithLegacyKMS injects a legacy kms service into the context.
+// WithLegacyKMS injects a legacy legacyKMS service into the context.
```
troyronda: a legacy legacyKMS?

pkg/framework/context/context.go
```diff
@@ -79,8 +81,13 @@ func (p *Provider) Service(id string) (interface{}, error) {
 	return nil, api.ErrSvcNotFound
 }
 
-// LegacyKMS returns a kms service.
+// LegacyKMS returns a legacyKMS service.
 func (p *Provider) LegacyKMS() legacykms.KeyManager {
+	return p.legacyKMS
+}
+
+// KMS returns a KMS service
```
troyronda: We could at least expand the acronym in these comments :).

pkg/framework/aries/default.go
```diff
@@ -122,7 +147,64 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
...
 	}
 
-	// TODO add SecretLock creation here.. #1150
+	return nil
+}
+
+func createDefSecretLock(opts *Aries) error {
+	keySize := sha256.Size
+
+	file, err := os.OpenFile(defMKPath, os.O_WRONLY|os.O_CREATE, 0600)
```
troyronda: we are hardcoding a file lookup as the default secret lock?
this will break WASM.

troyronda: Also, in combination with the other default - local DB storage, this default local file secret lock is arguably not adding much.

docs/kms_secretlock.md
```diff
@@ -0,0 +1,143 @@
...
+
+## No SecretLock/KMS options
+
+For a user who doesn't have a master key created, they can create an Aries instance without SecretLock/KMS options
+which will create a new unsecured master key behind the scenes and save it to local disk. This key will be used to
```
troyronda: Doesn't work for WASM.

cmd/aries-js-worker/package-lock.json
Baha-sk: not sure, but it resulted from `make all`

pkg/framework/aries/default.go
Baha-sk: another option would be to set an `env` variable, but it will be erased when the Go program shuts down. 

This means a new master key is generated on each aries.New() call without KMS/SecretLock options which makes keys stored in a local DB storage unlockable.

docs/kms_secretlock.md
Baha-sk: done

docs/kms_secretlock.md
Baha-sk: updated text to mention it as an example

docs/kms_secretlock.md
Baha-sk: updated text

README.md
Baha-sk: removed secret lock from the link

docs/kms_secretlock.md
Baha-sk: updated to reflect this as an example

docs/kms_secretlock.md
Baha-sk: fixed

pkg/framework/aries/default.go
Baha-sk: I will move this function to default_srv.go and use an the env reader master key in default_js_wasm.go

troyronda: @Baha-sk There are many options as SecretLock is an injectable interface. The problem is really that there might not be any acceptable default at the framework level (other than no-op).

There may be acceptable defaults for a particular target like cmd/agent-rest & the JS/WASM worker in cmd.

docs/kms_secretlock.md
troyronda: I think this section of text should be deferred to another PR once we figure out defaults for each target.

pkg/framework/aries/default_test.go
```diff
@@ -20,6 +20,10 @@ func TestDefaultFramework(t *testing.T) {
 		defer cleanup()
 		dbPath = path
 
+		defMKPath = path + "/mk"
```
troyronda: I see that this is only tested during non-wasm builds.

Should be deferred into another PR.

pkg/framework/aries/default.go
Baha-sk: sure, I'll add noop secretlock and make it as a framework default.

other specific targets may have their own secret lock instances (part of other PRs)..

I'll push another change in a bit...

pkg/framework/aries/default_test.go
Baha-sk: removed

pkg/framework/aries/default.go
Baha-sk: updated with noop SecretLock

pkg/framework/context/context.go
Baha-sk: done

pkg/framework/context/context.go
Baha-sk: updated

docs/kms_secretlock.md
Baha-sk: Updated with a more relevant comment.

Baha-sk: created #1358 to add a default js-worker secretLock
and #1359 to add a default agent-rest secretLock

PR #1333, opened 2020-02-24
fix: codecov reports

llorllale:


llorllale:
nevermind - coverage is uploaded from azure pipeline

Issue #1303, opened 2020-02-20
Controller APIs for VC Store


rolsonquadras:
Build REST/JS Controller APIs:
- VC Store 
  - Save with friendly name
  - Get based on vcID
  - Get all VC names

PR #1337, opened 2020-02-24
feat: Controller binding(rest/js) for VC Save operation

rolsonquadras:
part of #1303 

### DIFF
--- a/pkg/controller/rest/verifiable/operation.go
+++ b/pkg/controller/rest/verifiable/operation.go
@@ -7,17 +7,26 @@ SPDX-License-Identifier: Apache-2.0
 package verifiable
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/hyperledger/aries-framework-go/pkg/controller/command/verifiable"
 	"github.com/hyperledger/aries-framework-go/pkg/controller/internal/cmdutil"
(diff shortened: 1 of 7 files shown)


pkg/controller/rest/verifiable/operation.go
```diff
@@ -42,16 +56,28 @@ func (o *Operation) GetRESTHandlers() []rest.Handler {
 func (o *Operation) registerHandler() {
 	o.handlers = []rest.Handler{
 		cmdutil.NewHTTPHandler(validateCredentialPath, http.MethodPost, o.ValidateCredential),
+		cmdutil.NewHTTPHandler(saveCredentialPath, http.MethodPost, o.ValidateCredential),
 	}
 }
 
-// ValidateCredential swagger:route POST /verifiable/validateCredential verifiable validateCredentialReq
+// ValidateCredential swagger:route POST /verifiable/credential/validate verifiable validateCredentialReq
```
troyronda: Not related to this PR, but we should chat about the base path.

rolsonquadras: ok

troyronda: (how we best represent that these operations are the verifiable data registry).

Issue #1173, opened 2020-02-04
Service implementation for the Issue Credential Protocol


soluchok:
RFC https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential

Create a service according to RFC. 
Service should satisfy the following interface:
```
type ProtocolService interface {
	HandleInbound(msg DIDCommMsg, myDID, theirDID string) (string, error)
	HandleOutbound(msg DIDCommMsg, myDID, theirDID string) error
	Accept(msgType string) bool
	Name() string
}
```

PR #1343, opened 2020-02-25
feat: Issue Credential - protocol states skeleton

soluchok:
* added service common logic

Part of #1173 

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/service.go
+++ b/pkg/didcomm/protocol/issuecredential/service.go
@@ -7,6 +7,12 @@ SPDX-License-Identifier: Apache-2.0
 package issuecredential
 
 import (
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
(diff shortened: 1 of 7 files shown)


pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -0,0 +1,417 @@
...
+	md, err := s.doHandle(msg, true)
+	if err != nil {
+		return "", fmt.Errorf("doHandle: %w", err)
+	}
+
+	// sets outbound payload
+	md.MyDID = myDID
+	md.TheirDID = theirDID
+
+	return "", s.handle(md)
```
llorllale: what is the `string` return arg for?

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -0,0 +1,417 @@
...
+}
+
+func getPIID(msg service.DIDCommMsg) (string, error) {
+	if pthID := msg.ParentThreadID(); pthID != "" {
+		return pthID, nil
+	}
+
+	thID, err := msg.ThreadID()
+	if errors.Is(err, service.ErrThreadIDNotFound) {
+		// TODO: create a setter function e.g SetID
```
llorllale: shouldn't the messenger be setting the ID?

soluchok: When you send the first message you need to set this ID. I am going to add `SetID` function to the message interface.

pkg/didcomm/protocol/issuecredential/service.go
soluchok: It is because of the service interface. For did exchange we are returning `connectionID`. For the `introduce` and `issue-credential` protocols we do not need this. 

NOTE: I am going to refactor it later. In my previous PR, I have added this for HandleOutbound. But it seems to be redundant.

troyronda:
I assume tests are still coming in this PR?

soluchok:
> @soluchok the team just discussed this and agree that PRs must include tests.
> 
> That being said, please break this PR down into smaller chunks such that the final PR (with tests) does not exceed ~500 lines.

I am going to add tests. Ok, will separate it.

pkg/didcomm/protocol/issuecredential/service.go
llorllale: I still don't understand why the messenger can't set the ID.

It seems like we're setting the `@id` here and return it as the PIID for later use in the call to storage in `currentStateName`, where it will obviously not be found in this case.

So why not return an appropriate `ErrNotFound` from this method instead and earn two things:

* this logic won't have to be replicated in other protocols. Leave it up to the messenger to insert an `@id`
* save ourselves an unnecessary call to the storage

llorllale: Another option would be for `NewDIDCommMsgMap()` to set the `@id`.

soluchok: I forgot about an additional call to DB with not found result. Adding it to the construction function will not help, id and in that case, will not be empty. It makes sense to rely on the messenger. I am going to change this logic.

soluchok: @llorllale I have added `SetID` function. A call to DB is avoided now. I cannot rely on the messenger because I need to save data before action will be executed.

PR #1346, opened 2020-02-25
fix: multiple objectstores under same indexdb database

### LINKED ISSUE #1309: should not have a database per objectstore in indexDB

sudeshrshetty:
- due to complications in dynamically adding objectstores under already
opened indexdb database, jsindexdb provider will have predefined list of
objectstores. `onupgradeevent` during very first creation of database
will create those object stores in advance.
- attempt to create store not defined in object store will not return
any error, instead it will create new database for that object store.
- closes #1309

### DIFF
--- a/pkg/storage/jsindexeddb/jsindexeddb.go
+++ b/pkg/storage/jsindexeddb/jsindexeddb.go
@@ -11,22 +11,44 @@ package jsindexeddb
 import (
 	"errors"
 	"fmt"
+	"sync"
 	"syscall/js"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/messenger"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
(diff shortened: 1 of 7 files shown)


pkg/storage/jsindexeddb/jsindexeddb.go
```diff
@@ -217,3 +256,17 @@ func getResult(req js.Value) (*js.Value, error) {
 		return nil, errors.New("timeout waiting for eve")
 	}
 }
+
+// since jsindexdb doesn't support adding object stores on fly, using predefined object store names to
+//  create object store in advance instead of creating a database per store.
+func getStoreNames() []string {
```
troyronda: this array should be collected at a higher layer package and passed-in as an option.

sudeshrshetty:
> This code is used in https://github.com/trustbloc/edge-agent/blob/master/pkg/store/vc/store.go#L40
> we need to make sure it's working before release aries-framework-go

existing feature is still supported

pkg/storage/jsindexeddb/jsindexeddb.go
sudeshrshetty: it may require interface change, added a TODO and created issue #1347

sudeshrshetty:
@fqutishat Existing functionality is not broken, only pre initializing all aries object stores. Any other stores will still use older version of open db logic.
https://github.com/hyperledger/aries-framework-go/blob/f1d4e4f7552b7e50eccc31791026746787309567/pkg/storage/jsindexeddb/jsindexeddb.go#L73

Issue #1323, opened 2020-02-22
REST JS wrapper support for notifications


troyronda:
Depends on https://github.com/hyperledger/aries-framework-go/issues/471

sudeshrshetty:
- REST agent webhook to send topic information in messages, so that subscribers can be aware of the topic details.

PR #1353, opened 2020-02-26
feat: websocket support in webhooks

sudeshrshetty:
- websocket URLs are now supported in addition to existing HTTP webhook
URLs.
- Part of #1323


pkg/controller/webhook/webhook.go
```diff
@@ -89,6 +108,41 @@ func notify(destination string, message []byte) error {
 		destination, resp.Status)
 }
 
+func notifyWS(destination string, message []byte) error {
+	ctx, cancel := context.WithTimeout(context.Background(), notificationSendTimeout)
+	defer cancel()
+
+	c, rs, err := websocket.Dial(ctx, destination, nil)
```
troyronda: So this is useful variation where the REST server dials a WebSocket...

But we really need a version where the REST server exposes a WebSocket and the client dials it. This way the client does not need an exposed endpoint.

pkg/controller/webhook/webhook.go
```diff
@@ -89,6 +108,41 @@ func notify(destination string, message []byte) error {
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		err = c.Close(websocket.StatusNormalClosure, "closing connection")
```
troyronda: we should followup where we hold the connection open until an inactivity period.

pkg/controller/webhook/webhook.go
troyronda: > The implementation should work the same as it does in ACA-Py. The WebSocket support will allow for clients to register dynamically after the server has started. The full duplex functionality of WebSocket will not be used - it should be a one-way notification from server->client. The endpoint will be /ws.

pkg/controller/webhook/webhook.go
```diff
@@ -30,20 +33,31 @@ type Notifier interface {
...
+type WebNotifier struct {
+	httpURLs []string
+	wsURLS   []string
 }
 
-// NewHTTPNotifier returns a new instance of an HTTPNotifier.
-func NewHTTPNotifier(webhookURLs []string) HTTPNotifier {
-	return HTTPNotifier{WebhookURLs: webhookURLs}
+// NewWebNotifier returns a new instance of an WebNotifier.
+func NewWebNotifier(webhookURLs []string) WebNotifier {
```
troyronda: Should be returning a pointer.

pkg/controller/webhook/webhook.go
```diff
@@ -89,6 +108,41 @@ func notify(destination string, message []byte) error {
 		destination, resp.Status)
 }
 
+func notifyWS(destination string, message []byte) error {
+	ctx, cancel := context.WithTimeout(context.Background(), notificationSendTimeout)
```
troyronda: parent ctx should be an argument to Notify.

pkg/controller/webhook/webhook.go
sudeshrshetty: Will add that feature as part of separate PR

pkg/controller/webhook/webhook.go
sudeshrshetty: Added a follow up TODO

pkg/controller/webhook/webhook.go
sudeshrshetty: fixed

pkg/controller/webhook/webhook.go
sudeshrshetty: Added a follow up TODO and created an issue, applies to HTTP as well.

sudeshrshetty:
pushed this as part of other PRs

Issue #1238, opened 2020-02-11
Setup CI for npm package of aries-js-worker


llorllale:
- [x] Skeleton #1310
- [ ] Use feature files
   * [cucumber-js](https://github.com/cucumber/cucumber-js) is the only gherkin library for JS
   * very little support for karma+cucumber integration
   * 2 plugins for Karma:
      * [karma-cucumber-js](https://github.com/eugene-sea/karma-cucumber-js) - deprecated/archived
      * [karma-cucumber-js-latest](https://github.com/mitcoding/karma-cucumber-js-latest)
         * hasn't received updates in over 16 months
         * declares support up to v5.0.2 of cucumber-js although the latter is already at 6.0.5
         * the instructions in the README do not work
         * opened two issues: [#1](https://github.com/mitcoding/karma-cucumber-js-latest/issues/1), [#2](https://github.com/mitcoding/karma-cucumber-js-latest/issues/2)
- [ ] Real tests
   - [x] didexchange: edge agent - router #1330

llorllale:
blocked by #1268

llorllale:
blocked by #1277

PR #1357, opened 2020-02-26
chore: aries-js-worker - didexchange edgeagent/router

### LINKED ISSUE #1330: aries-js-worker: add bdd test: didexchange b/w edge agent  and router
part of #1238

llorllale:
closes #1330

* BDD test for did-exchange edgeagent <-> router
* fixes for properly propagating errors wasm -> JS when startup fails (eg. invalid startup options) and when `didexchange.acceptInvitation` fails
* `didexchange.acceptInvitation/queryConnectionByID` now accept a text argument and transform it to a properly built object that the wasm expects
* better test reporting to the terminal


cmd/aries-js-worker/src/aries.js
```diff
@@ -183,11 +181,16 @@ const Aries = function(opts) {
...
                 return invoke(aw, pending,  this.pkgname, "CreateInvitation", text, "timeout while creating invitation")
             },
-            receiveInvitation: async function (text) {
-                return invoke(aw, pending,  this.pkgname, "ReceiveInvitation", text, "timeout while receiving invitation")
+            receiveInvitation: async function (invitation) {
+                return invoke(aw, pending,  this.pkgname, "ReceiveInvitation", invitation, "timeout while receiving invitation")
             },
-            acceptInvitation: async function (text) {
-                return invoke(aw, pending,  this.pkgname, "AcceptInvitation", text, "timeout while accepting invitation")
+            acceptInvitation: async function (connectionID) {
```
rolsonquadras: The accept invitation API takes optional public DID as well (https://github.com/hyperledger/aries-framework-go/blob/master/pkg/controller/command/didexchange/models.go#L86) along with connection ID.

llorllale: renamed to "id"

llorllale: ok: renamed back to `connectionID` and added an optional `publicDID` arg

cmd/aries-js-worker/src/aries.js
```diff
@@ -183,11 +181,16 @@ const Aries = function(opts) {
...
                 return invoke(aw, pending,  this.pkgname, "CreateInvitation", text, "timeout while creating invitation")
             },
-            receiveInvitation: async function (text) {
-                return invoke(aw, pending,  this.pkgname, "ReceiveInvitation", text, "timeout while receiving invitation")
+            receiveInvitation: async function (invitation) {
+                return invoke(aw, pending,  this.pkgname, "ReceiveInvitation", invitation, "timeout while receiving invitation")
             },
-            acceptInvitation: async function (text) {
-                return invoke(aw, pending,  this.pkgname, "AcceptInvitation", text, "timeout while accepting invitation")
+            acceptInvitation: async function (connectionID, publicDID="") {
```
sudeshrshetty: As per my opinion, we should avoid using strict function args, because every time there is an update in command controller models we may have to come back here to make corresponding updates.

test/aries-js-worker/test/didexchange/edgeagent_router.js
```diff
@@ -0,0 +1,85 @@
...
+                aries.didexchange.acceptInvitation(connection.connection_id)
+            } catch (err) {
+                done(err)
+            }
+            done()
+        }, ["connections"])
+        aries.didexchange.receiveInvitation(invitation)
+    })
+
+    it("Edge Agent validates that the connection's state is 'completed'", function(done) {
```
rolsonquadras: Are we going to have the .feature file for this step/scenario ?

llorllale: See my findings so far under the second point here in #1238

llorllale: So short answer is - I couldn't make it work.

rolsonquadras: ok. so, for now we wont be able to reuse the steps.

PR #1362, opened 2020-02-26
feat: Issue Credential - protocol states skeleton (part 1)

soluchok:
* added state definitions

Part of #1173


pkg/didcomm/protocol/issuecredential/states_test.go
```diff
@@ -0,0 +1,287 @@
...
+	// states for Holder
+	require.True(t, st.CanTransitionTo(&proposalSent{}))
+	require.False(t, st.CanTransitionTo(&offerReceived{}))
+	require.False(t, st.CanTransitionTo(&requestSent{}))
+	require.False(t, st.CanTransitionTo(&credentialReceived{}))
+}
+
+func TestStart_ExecuteInbound(t *testing.T) {
+	followup, action, err := (&start{}).ExecuteInbound(&metaData{})
+	require.Error(t, err)
```
sandrask: Normally you should check for specific error message but since this is 'not implemented' it is ok to leave it as is.

pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+// state action for network call
+type stateAction func(messenger service.Messenger) error
+
+type metaData struct{}
+
+// nolint: deadcode,unused
+// represents zero action function
+func zeroAction(_ service.Messenger) error { return nil }
```
rolsonquadras: Is this no-op function for state transition ? Also, can we remove this if this is not used in the PR.

pkg/didcomm/protocol/issuecredential/states_test.go
soluchok: Done!

pkg/didcomm/protocol/issuecredential/states.go
soluchok: Removed!

PR #1372, opened 2020-02-26
feat: notifiers support in  aries rest client js worker

### LINKED ISSUE #1370: Aries Rest Client JS worker - Add websocket notifier

sudeshrshetty:
- new constructor opts `agent-rest-wshook` in rest client based aries js
worker. If provided then `aries.startNotifier` can be used to listen to
incoming messages

- closes #1370

### DIFF
--- a/cmd/aries-js-worker/index.html
+++ b/cmd/aries-js-worker/index.html
@@ -106,7 +106,7 @@ This is a test page for developers to test WASM integration.
             document.getElementById("mode-label").innerHTML = `Running in ${mode} mode`
 
             if (src == "rest"){
-                document.getElementById("opts").innerHTML = `{"assetsPath": "/dist/assets", "agent-rest-url": "http://localhost:8082"}`
+                document.getElementById("opts").innerHTML = `{"assetsPath": "/dist/assets", "agent-rest-url": "http://localhost:8082", "agent-rest-wshook":"ws://localhost:8881"}`
             } else {
                 document.getElementById("opts").innerHTML = `{"assetsPath": "/dist/assets", "agent-default-label":"dem-js-agent","http-resolver-url":[],"auto-accept":true,"outbound-transport":["ws","http"],"transport-return-route":"all","log-level":"debug"}`
             }
--- a/cmd/aries-js-worker/src/worker-impl-rest.js
+++ b/cmd/aries-js-worker/src/worker-impl-rest.js
@@ -9,7 +9,9 @@ self.importScripts("./agent-rest-client.js")
 postMessage({topic: "asset-ready"})
 
 // TODO synchronized access to controller
-let handler, controller
+let handler, controller, notifier
+
+const wsNormalClosureCode = 1000
 
 const ariesHandle = {
     aries: {
@@ -21,6 +23,13 @@ const ariesHandle = {
             if (!data.payload["agent-rest-url"]) {
                 return newResponse(data.id, null, "'agent-rest-url' is required");
             }
+
+            if (data.payload["agent-rest-wshook"]){
+                notifier = new wsnotifier(data.payload["agent-rest-wshook"], (msg) => {
+                    postMessage(msg)
+                });
+            }
+
 
             controller = new RESTAgent.Client(data.payload["agent-rest-url"], data.payload["agent-rest-token"]);
             return newResponse(data.id, "aries is started");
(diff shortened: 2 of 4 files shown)


cmd/aries-js-worker/src/worker-impl-rest.js
```diff
@@ -67,4 +74,18 @@ function newResponse(id, payload, errMsg, topic) {
...
     };
-}
\ No newline at end of file
+}
+
+const wsnotifier = class {
+    constructor(url, postMsg) {
+        this.socket = new WebSocket(url);
+        this.socket.addEventListener('message', function (event) {
+            // TODO REST agents are not currently revealing topic information on incoming messages,
```
troyronda: can you patch the REST agent?

cmd/aries-js-worker/src/worker-impl-rest.js
```diff
@@ -67,4 +74,18 @@ function newResponse(id, payload, errMsg, topic) {
...
+            //  Once REST supports this feature, topic value will be dynamic.
+            postMsg(newResponse("random-id",  event.data, "", "all"));
```
troyronda: random-id?

cmd/aries-js-worker/src/worker-impl-rest.js
```diff
@@ -67,4 +74,18 @@ function newResponse(id, payload, errMsg, topic) {
...
+        });
+    }
+    stop(){
+        this.socket.close(1011, "stopped notifier in aries-js-worker")
```
troyronda: 1011?

can we replace magic number with a named const?

cmd/aries-js-worker/src/worker-impl-rest.js
sudeshrshetty: as part of https://github.com/hyperledger/aries-framework-go/issues/1323

All BDD tests will fail, so will push it in next PR.

cmd/aries-js-worker/src/worker-impl-rest.js
troyronda: Also, you should patch REST agent to not throw a log on 1011 status code.

troyronda: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/controller/webnotifier/websocket.go#L94

troyronda: Never mind. Isn't 1011 an error code (https://github.com/nhooyr/websocket/blob/94f9b71576625ef4dbe5950253cbfa8b539b88b5/close.go#L42)

Shouldn't you use a non-error status?

cmd/aries-js-worker/src/worker-impl-rest.js
troyronda: ok

cmd/aries-js-worker/index.html
```diff
@@ -106,7 +106,7 @@
             document.getElementById("mode-label").innerHTML = `Running in ${mode} mode`
 
             if (src == "rest"){
-                document.getElementById("opts").innerHTML = `{"assetsPath": "/dist/assets", "agent-rest-url": "http://localhost:8082"}`
+                document.getElementById("opts").innerHTML = `{"assetsPath": "/dist/assets", "agent-rest-url": "http://localhost:8082", "agent-rest-wshook":"ws://localhost:8881"}`
```
troyronda: the rest code called it a notifier rather than a hook.

sudeshrshetty: in agent-rest start command we call it webhook, so reused that term here.
`notifier` term is use for "startNotifier/stopNotifier" where we watch for incoming messages.

cmd/aries-js-worker/src/worker-impl-rest.js
sudeshrshetty: using normal closure [code](https://tools.ietf.org/html/rfc6455#section-7.4.1)

cmd/aries-js-worker/src/worker-impl-rest.js
sudeshrshetty: fixed it

llorllale:
This new option requires a line in the docs

cmd/aries-js-worker/index.html
llorllale: Looking at these URLs, I see that we should document somewhere how to properly run this `index.html`

llorllale: ie. this goes way past just inflating the wasm file. It requires services up and running in addition to the inflated wasm.

llorllale: one other point: can we start using property names with no special characters such that the double quotes aren't needed? Like `assetsPath` for instance.

sudeshrshetty: updated docs

> This new option requires a line in the docs

PR #1376, opened 2020-02-26
feat: DIDKeyResolver: check key id in did authentication

### LINKED ISSUE #1351: DIDKeyResolver: check key id in did authentication
if you can't find the key id in did public key, try find it in did authentication

https://github.com/hyperledger/aries-framework-go/blob/5531f2c6db0351beec475715ca22162c2b20651c/pkg/doc/verifiable/common.go#L82

fqutishat:
closes #1351

### DIFF
--- a/pkg/doc/verifiable/common.go
+++ b/pkg/doc/verifiable/common.go
@@ -85,6 +85,13 @@ func (r *DIDKeyResolver) resolvePublicKey(issuerDID, keyID string) (interface{},
 		}
 	}
 
+	// if key not found in PublicKey try to find it in authentication
+	for _, auth := range doc.Authentication {
+		if auth.PublicKey.ID == keyID {
+			return auth.PublicKey.Value, nil
+		}
+	}
+
 	return nil, fmt.Errorf("public key with KID %s is not found for DID %s", keyID, issuerDID)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/doc/verifiable/common.go
```diff
@@ -85,6 +85,13 @@ func (r *DIDKeyResolver) resolvePublicKey(issuerDID, keyID string) (interface{},
 		}
 	}
 
+	// if key not found in PublicKey try to find it in authentication
```
troyronda: I'm a bit confused, how did you decide priority ordering?

fqutishat: priority not important here because we search by key id

troyronda: I think we will need to revisit.

llorllale:
My reading of the [did-core spec](https://www.w3.org/TR/did-core/#authentication) is that `authentication` is:

> a mechanism by which an entity can prove it is the DID subject.

It goes on to add:

> the verification method indicated by the authentication property of a DID document can only be used to authenticate the DID subject

Is there any reason why these v1 DIDs can't be updated with a publicKey similar to [this](https://w3c-ccg.github.io/did-method-v1/#example-11-did-document-update-request) example (except under `/publicKey`)?

One other point: we should conform to the [did resolution algorithm](https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm)

llorllale:
It seems like to resolving a key from a DID URL the way this PR wants to is dependent of the MIME type that was attached to the DID document: https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm-secondary

PR #1387, opened 2020-02-27
feat: Issue Credential - defines states transitions

soluchok:
This PR adds proper states transitions.

Part of #1173
Depends on https://github.com/hyperledger/aries-framework-go/pull/1343

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/service.go
+++ b/pkg/didcomm/protocol/issuecredential/service.go
@@ -346,12 +346,9 @@ func nextState(msg service.DIDCommMsg, outbound bool) (state, error) {
 
 // canTriggerActionEvents checks if the incoming message can trigger an action event
 func canTriggerActionEvents(msg service.DIDCommMsg) bool {
-	switch msg.Type() {
-	case ProposeCredentialMsgType, OfferCredentialMsgType, RequestCredentialMsgType, IssueCredentialMsgType:
-		return true
-	}
-
-	return false
+	return msg.Type() == ProposeCredentialMsgType ||
+		msg.Type() == OfferCredentialMsgType ||
+		msg.Type() == RequestCredentialMsgType
 }
 
 func (s *Service) processCallback(msg *metaData) {
--- a/pkg/didcomm/protocol/issuecredential/states.go
+++ b/pkg/didcomm/protocol/issuecredential/states.go
@@ -75,7 +75,16 @@ func (s *start) Name() string {
 }
 
 func (s *start) CanTransitionTo(st state) bool {
-	return st.Name() == stateNameProposalSent || st.Name() == stateNameProposalReceived
+	switch st.Name() {
+	// Issuer
+	case stateNameProposalReceived, stateNameOfferSent, stateNameRequestReceived:
+		return true
+	// Holder
+	case stateNameProposalSent, stateNameOfferReceived, stateNameRequestSent:
+		return true
+	}
+
+	return false
 }
 
 func (s *start) ExecuteInbound(_ *metaData) (state, stateAction, error) {
@@ -93,8 +102,8 @@ func (s *abandoning) Name() string {
 	return stateNameAbandoning
 }
 
-func (s *abandoning) CanTransitionTo(_ state) bool {
-	return false
+func (s *abandoning) CanTransitionTo(st state) bool {
+	return st.Name() == stateNameDone
 }
 
 func (s *abandoning) ExecuteInbound(_ *metaData) (state, stateAction, error) {
@@ -124,15 +133,15 @@ func (s *done) ExecuteOutbound(_ *metaData) (state, stateAction, error) {
 	return nil, nil, fmt.Errorf("%s: ExecuteOutbound is not implemented yet", s.Name())
 }
 
-// proposalReceived state
+// proposalReceived the Issuer's state
 type proposalReceived struct{}
 
 func (s *proposalReceived) Name() string {
 	return stateNameProposalReceived
 }
 
-func (s *proposalReceived) CanTransitionTo(_ state) bool {
-	return false
+func (s *proposalReceived) CanTransitionTo(st state) bool {
+	return st.Name() == stateNameOfferSent || st.Name() == stateNameAbandoning
 }
 
 func (s *proposalReceived) ExecuteInbound(_ *metaData) (state, stateAction, error) {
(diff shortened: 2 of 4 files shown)


test/aries-js-worker/test/didexchange/edgeagent_router.js
```diff
@@ -135,23 +135,15 @@ describe("DID-Exchange between two Edge Agents using the router", function () {
     var invitation, connectionID
 
     before(async () => {
-        return new Promise((resolve, reject) => {
-            newAries('alice').then(
-                a => {
-                    aliceAgent = a;
-                    resolve()
```
llorllale: how is the promise resolving then?

test/aries-js-worker/test/didexchange/edgeagent_router.js
```diff
@@ -135,23 +135,15 @@ describe("DID-Exchange between two Edge Agents using the router", function () {
     var invitation, connectionID
 
     before(async () => {
-        return new Promise((resolve, reject) => {
```
llorllale: what was the issue?

soluchok: https://github.com/hyperledger/aries-framework-go/pull/1387/checks?check_run_id=474809170#step:5:501

pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -75,7 +75,16 @@ func (s *start) Name() string {
 }
 
 func (s *start) CanTransitionTo(st state) bool {
-	return st.Name() == stateNameProposalSent || st.Name() == stateNameProposalReceived
+	switch st.Name() {
+	// Issuer
```
llorllale: we should consider specifying and persisting the role the participants in a protocol are fulfilling

test/aries-js-worker/test/didexchange/edgeagent_router.js
llorllale: Perhaps more investigation is needed (learning curve).

This looks like a simple refactor to my eyes.

Mocha also waits until the promise returned by an async function passed to `before()` to resolve.

The build passed on PR #1392

llorllale: Here is the merge build from PR #1392: https://github.com/hyperledger/aries-framework-go/actions/runs/46404517. Tests passed.

pkg/didcomm/protocol/issuecredential/states.go
soluchok: The role can be determined by message type and inbound/outbound value. Do you mean to include this check/condition?

llorllale: let's defer that for later

test/aries-js-worker/test/didexchange/edgeagent_router.js
llorllale: I notice that [your branch](https://github.com/soluchok/aries-framework-go/commits/issue-1173) was lagging behind a bit when it got merged today.

llorllale: I can't figure out what exactly was the issue.

soluchok: So, the problem here is that we have created the promise and then we are resolving it twice. It means after first resolve it is done. It is not waiting for the second one. It does not give us a guarantee that two agents were started successfully. I just not lucky enough :)

```
 return new Promise((resolve, reject) => {
      newAries('alice').then( ..  resolve() ..)
      newAries('alice').then( ..  resolve() ..)
}
```

llorllale: Ah! I got it. The promise was being resolved when the `alice` agent was fetch but *before* bob's agent was loaded.

soluchok: Yeah, but I am not :)

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -346,12 +346,7 @@ func nextState(msg service.DIDCommMsg, outbound bool) (state, error) {
 
 // canTriggerActionEvents checks if the incoming message can trigger an action event
 func canTriggerActionEvents(msg service.DIDCommMsg) bool {
-	switch msg.Type() {
-	case ProposeCredentialMsgType, OfferCredentialMsgType, RequestCredentialMsgType, IssueCredentialMsgType:
-		return true
-	}
-
-	return false
+	return msg.Type() == ProposeCredentialMsgType || msg.Type() == OfferCredentialMsgType
```
rolsonquadras: Don't we need user interaction/handling to pass the actual Credentials when `request credential` message type is received ?

llorllale: I understand the choreography diagram in the RFC has the Issuer automatically sending the credential when they receive `request-credential`, but really `propose-credential` and `offer-credential` are optional messages (see [here](https://github.com/hyperledger/aries-rfcs/blob/master/features/0036-issue-credential/README.md#messages)). This means `request-credential` is a valid entrypoint.

soluchok: It will be done by passing an option to the continue function or executing a specific method from the client. This PR is about `states transitions`. So, I have described the transitions between states.

test/aries-js-worker/test/didexchange/edgeagent_router.js
llorllale: @soluchok can you push a separate PR for this fix?

pkg/didcomm/protocol/issuecredential/service.go
rolsonquadras: Yes, but for that we need to trigger events right ? I see we are removing it in this PR.

soluchok: I am going to add it

test/aries-js-worker/test/didexchange/edgeagent_router.js
soluchok: Done!

pkg/didcomm/protocol/issuecredential/service.go
soluchok: Added `RequestCredentialMsgType`

soluchok: Done! I have added `RequestCredentialMsgType`

PR #1389, opened 2020-02-27
feat: 'db-path' start option for aries js worker WASM

### LINKED ISSUE #1388: Aries JS worker : IndexDB conflict in case of multiple aries instances

sudeshrshetty:
- added dbpath option in aries start opts to avoid collision between
multiple aries instance in same browser(context).
- closes #1388


sudeshrshetty:
> Sounds like it's more a prefix than a path

Using same option name as SDK, can rename it

PR #1393, opened 2020-02-27
feat: parse did v0.11

### LINKED ISSUE #1349: parse did v0.11
https://github.com/w3c-ccg/did-spec/releases/tag/v0.11

fqutishat:
closes #1349

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -23,25 +23,29 @@ import (
 const (
 	// Context of the DID document
 	Context             = "https://w3id.org/did/v1"
+	contextV011         = "https://w3id.org/did/v0.11"
 	jsonldType          = "type"
 	jsonldID            = "id"
+	jsonldPublicKey     = "publicKey"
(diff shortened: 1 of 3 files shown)


pkg/doc/did/doc.go
```diff
@@ -423,6 +599,11 @@ func decodePK(rawPK map[string]interface{}) ([]byte, error) {
 func validate(data []byte) error {
 	// Validate that the DID Document conforms to the serialization of the DID Document data model.
 	// Reference: https://w3c-ccg.github.io/did-spec/#did-documents)
+	schemaLoader := schemaLoaderV1
+	if strings.Contains(string(data), contextV011) {
```
rolsonquadras: May be better to check the context value `contextV011` against actual `context` key inside data rather than whole document/json ? we can avoid cases where the context is not `contextV011` but some other attribute in the json contains the text `contextV011`.

troyronda: Agree.

PR #1404, opened 2020-03-02
feat: verificationMethod for proof verification

### LINKED ISSUE #1156: verificationMethod for proof verification
Use `verificationMethod` element of signed linked data proofs to verify [Linked Data Signature](https://w3c-dvcg.github.io/ld-signatures).
Currently, at signature verification, we pass a value of `creator` field to resolve a public key. This is applicable to a specific use case of DID documents verification. We should make it more generic, configurable and flexible.

kdimak:
closes #1156

### DIFF
--- a/pkg/doc/signature/verifier/verifier.go
+++ b/pkg/doc/signature/verifier/verifier.go
@@ -71,7 +71,12 @@ func (dv *DocumentVerifier) verifyObject(jsonLdObject map[string]interface{}) er
 	}
 
 	for _, p := range proofs {
-		publicKey, err := dv.pkResolver.Resolve(p.Creator)
+		publicKeyID, err := p.PublicKeyID()
+		if err != nil {
+			return err
+		}
+
+		publicKey, err := dv.pkResolver.Resolve(publicKeyID)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


pkg/doc/signature/verifier/verifier.go
```diff
@@ -71,7 +71,7 @@ func (dv *DocumentVerifier) verifyObject(jsonLdObject map[string]interface{}) er
 	}
 
 	for _, p := range proofs {
-		publicKey, err := dv.pkResolver.Resolve(p.Creator)
+		publicKey, err := dv.pkResolver.Resolve(p.JSONLdObject())
```
kdimak: Here is a place of the main change - we pass a proof map to the `resolver` instead of concrete proof attribute. A resolver impl decides which proof attributes are needed for public key resolution (`creator`, `verificationMethod`, any other).

sandrask: I think you should first look if "verificationMethod" is present. If not then search for creator. If creator is not present, then fail. This will cover both cases (and older VC versions) and client will not have to worry about it.

kdimak: I changed it to the way you proposed @sandrask . Please take a look.

PR #1408, opened 2020-03-03
feat: Issue Credential - state implementation for offer sent/received

soluchok:
* `HandleOutbound` function can send an offer.
* `HandleInbound` function can receive an offer.

Part of #1173 

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/states.go
+++ b/pkg/didcomm/protocol/issuecredential/states.go
@@ -209,8 +209,13 @@ func (s *offerSent) ExecuteInbound(_ *metaData) (state, stateAction, error) {
 	return nil, nil, fmt.Errorf("%s: ExecuteInbound is not implemented yet", s.Name())
 }
 
-func (s *offerSent) ExecuteOutbound(_ *metaData) (state, stateAction, error) {
-	return nil, nil, fmt.Errorf("%s: ExecuteOutbound is not implemented yet", s.Name())
+func (s *offerSent) ExecuteOutbound(md *metaData) (state, stateAction, error) {
+	// creates the state's action
+	action := func(messenger service.Messenger) error {
+		return messenger.Send(md.Msg, md.MyDID, md.TheirDID)
+	}
+
+	return &noOp{}, action, nil
 }
 
 // requestReceived the Issuer's state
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -288,8 +300,19 @@ func (s *offerReceived) CanTransitionTo(st state) bool {
 		st.Name() == stateNameAbandoning
 }
 
-func (s *offerReceived) ExecuteInbound(_ *metaData) (state, stateAction, error) {
-	return nil, nil, fmt.Errorf("%s: ExecuteInbound is not implemented yet", s.Name())
+func (s *offerReceived) ExecuteInbound(md *metaData) (state, stateAction, error) {
+	if !isDataCorrect(md.Msg) {
```
llorllale: I know you left a TODO for `isDataCorrect`, but not returning an error if `!isDataCorrect` is really confusing. Where are we going with this?

pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -262,8 +267,15 @@ func (s *proposalSent) CanTransitionTo(st state) bool {
 	return st.Name() == stateNameOfferReceived || st.Name() == stateNameAbandoning
 }
 
-func (s *proposalSent) ExecuteInbound(_ *metaData) (state, stateAction, error) {
-	return nil, nil, fmt.Errorf("%s: ExecuteInbound is not implemented yet", s.Name())
+func (s *proposalSent) ExecuteInbound(md *metaData) (state, stateAction, error) {
+	// creates the state's action
+	action := func(messenger service.Messenger) error {
+		// sets message type
+		md.proposeCredential.Type = ProposeCredentialMsgType
```
llorllale: Are we replying to a proposal with another proposal?

soluchok: No, when we received an offer we have three ways:
1. Send a request credential (action is not triggered bc data is correct)
2. Send a propose credential  (e.g Continue())
3. Go to the abandoning state (e.g Stop())

This code is related to the `3` option. This means that we received an offer and data is not correct and we want to continue. This line shows that we are going back from `offerReceived` to `proposalSent`
https://github.com/hyperledger/aries-framework-go/pull/1408/files#diff-db914b8de9137376b5b65d627b223fdfR305

pkg/didcomm/protocol/issuecredential/states.go
soluchok: According to the diagram, if data is not correct, it is fine. In that case, we are going back to `proposalSent`  state and we are sending a proposal message.

soluchok: It looks like a negotiation, we may receive an offer and then go back and send a proposal and we can do it continuously until data is correct or someone of the participants stops the protocol. My question here what actually is the correct data, how to check it? For now, I have added this `0000-0000` id which allows me to mock this logic for the tests.

PR #1410, opened 2020-03-03
test: VC Store BDD tests for JS/REST API

### LINKED ISSUE #1400: VC Store - JS API BDD steps
BDD steps for VC Store features added as part of https://github.com/hyperledger/aries-framework-go/issues/1303

rolsonquadras:
closes #1400 


test/aries-js-worker/test/verifiable/verifiable.js
```diff
@@ -30,12 +33,20 @@ const vc = `
...
+    await verifiableStore(newAries())
+})
+
+async function verifiableStore(newAries, mode = wasmMode) {
     let aries
+    let modePrefix = '[' + mode + '] '
 
     before(async () => {
-        await newAries()
+        await newAries
```
llorllale: I find this pattern a bit confusing (`newAries` being a promise). Suggest minor refactor to make it clearer (eg. accept an async function as argument, or simply accept the aries object, etc.)

test/aries-js-worker/test/verifiable/verifiable.js
```diff
@@ -46,7 +57,7 @@ describe("Verifiable Store", function () {
         aries.destroy()
     })
 
-    it("Alice stores the verifiable credential received from the college", function (done) {
+    it(modePrefix + "Alice stores the verifiable credential received from the college", function (done) {
```
llorllale: nitpick: add an extra whitespace between the mode prefix and the rest of the description

test/aries-js-worker/fixtures/docker-compose.yml
```diff
@@ -24,3 +24,19 @@ services:
       - ${E2E_ROUTER_WS_INBOUND_PORT}:${E2E_ROUTER_WS_INBOUND_PORT}
       - ${E2E_ROUTER_API_PORT}:${E2E_ROUTER_API_PORT}
     command: start
+
+  user.agent.example.com:
```
llorllale: nitpick: rename to `user.agent.rest.example.com` or something that similarly communicates how this is the agent-rest we're interfacing with

test/aries-js-worker/test/verifiable/verifiable.js
rolsonquadras: It's there in line 46 - https://github.com/hyperledger/aries-framework-go/pull/1410/files#diff-307e2ee4ae6edb8abe0a65e7a61b69e6R46

test/aries-js-worker/fixtures/docker-compose.yml
rolsonquadras: All docker images are rest based. Only difference is while creating the Aries object in the js files (passing controller url(rest backend) v wasm pararm(wasm backend)).

test/aries-js-worker/test/verifiable/verifiable.js
rolsonquadras: we would need to pass the promise(line 40 and 41) in to some util function when we need to pass function arg. Since both create aries returns a promise, we can directly use it in func param and get the aries object. Also, this is used internally.

PR #1417, opened 2020-03-05
ci: use actions for codecov

troyronda:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,11 +29,9 @@ jobs:
 
     - name: Upload coverage to Codecov
       timeout-minutes: 10
-      # GitHub Actions currently only allow secret usage access during push events
-      if: github.event_name == 'push' && matrix.os == 'ubuntu-18.04' && github.repository == 'hyperledger/aries-framework-go'
-      uses: codecov/codecov-action@v1.0.2
+      if: matrix.os == 'ubuntu-18.04' && github.repository == 'hyperledger/aries-framework-go'
+      uses: codecov/codecov-action@v1.0.6
       with:
-        token: ${{secrets.CODECOV_TOKEN}}
         file: ./coverage.txt
 
   unitTestWasm:
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -3,13 +3,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 name: $(SourceBranchName)-$(Date:yyyyMMdd).$(Rev:r)
-#trigger: none
+trigger: none
 variables:
   GO_VERSION: 1.14
   GOPATH: $(Agent.BuildDirectory)/go
 
 jobs:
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -29,12 +29,12 @@ jobs:
...
     - name: Upload coverage to Codecov
       timeout-minutes: 10
-      # GitHub Actions currently only allow secret usage access during push events
-      if: github.event_name == 'push' && matrix.os == 'ubuntu-18.04' && github.repository == 'hyperledger/aries-framework-go'
-      uses: codecov/codecov-action@v1.0.2
+      if: matrix.os == 'ubuntu-18.04' && github.repository == 'hyperledger/aries-framework-go'
+      uses: codecov/codecov-action@v1.0.5
       with:
-        token: ${{secrets.CODECOV_TOKEN}}
+        token:
```
llorllale: empty token?

troyronda: https://github.com/codecov/codecov-action/issues/29#issuecomment-594958141
@llorllale

troyronda:
Bash uploader is now working. Waiting for proper Action. https://github.com/codecov/codecov-action/issues/29#issuecomment-594976661

.github/workflows/build.yml
```diff
@@ -29,12 +29,11 @@ jobs:
...
       timeout-minutes: 10
-      # GitHub Actions currently only allow secret usage access during push events
-      if: github.event_name == 'push' && matrix.os == 'ubuntu-18.04' && github.repository == 'hyperledger/aries-framework-go'
-      uses: codecov/codecov-action@v1.0.2
+      if: matrix.os == 'ubuntu-18.04' && github.repository == 'hyperledger/aries-framework-go'
+      uses: codecov/codecov-action@v1.0.6
       with:
-        token: ${{secrets.CODECOV_TOKEN}}
         file: ./coverage.txt
+        fail_ci_if_error: true
```
troyronda: not sure that we should actually fail the workflow if codecov fails.

troyronda:
Codecov Action is now working.

.github/workflows/build.yml
llorllale: We should definitely not be blocked for days while codecov integration is broken.

I believe there is the option to merge the PR even if the workflow fails. The onus is on the reviewer to inspect the reason why the workflow failed and ensure it's due to this bug.

troyronda: Or I remove this flag and you know that codecov failed because the comment isn't posted.

llorllale: A failed workflow raises a higher red flag than a missing codecov comment imo

llorllale: Actually, now that i think about it, i wouldn't want the build status badge to show "failed " just because this thing broke.

Maybe we should just remove this flag as you say

troyronda: Removed.

PR #1424, opened 2020-03-06
feat: messaging js worker test

### LINKED ISSUE #1412: Aries JS BDDTests - Messaging

sudeshrshetty:
- js test for messaging covering register/unregister/send/receive
features
- added test didexchange e2e client to serve presteps features involving
didexchange between 2 agents
- closes #1412

### DIFF
--- a/test/aries-js-worker/test/messaging/messaging.js
+++ b/test/aries-js-worker/test/messaging/messaging.js
@@ -1,0 +1,110 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+import {newDIDExchangeClient} from "../didexchange/didexchange_e2e.js";
+
+const basicMsgType = "https://didcomm.org/basicmessage/1.0/message"
+const basicMsgSvcName = "basic-msg-svc-demo"
+const sampleMsg = {
+    "@id": "1d9a1589-9d7b-4308-9fab-8ee9730720c2",
+    "@type":basicMsgType,
+    "~l10n":{"locale":"en"},
+    "sent_time":"2020-03-05T16:59:47.489789-05:00",
+    "content":"Your hovercraft is full of eels."
+}
+
+// scenarios
+describe("Basic Messaging", function () {
+    this.timeout(5000);
+
+    let destinationConnID
+    let didexClient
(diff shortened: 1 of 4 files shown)


pkg/framework/aries/default_js_wasm.go
```diff
@@ -23,7 +25,7 @@ func storeProvider() (storage.Provider, error) {
 }
 
 func transientStoreProvider() (storage.Provider, error) {
-	storeProv, err := jsindexeddb.NewProvider("temp")
+	storeProv, err := jsindexeddb.NewProvider("temp" + uuid.New().String())
```
llorllale: shouldn't this be under the same namespace as the persistent store?

troyronda: I don’t think we should create a new database every time.

llorllale: Yes - this will fill up storage needlessly.

Any reason why this can't be under the persistent store's namespace?

llorllale: Separate store, with a deterministic key under the same namespace.

sudeshrshetty: this is for transient db only, right now it is being used by multiple aries instances simultaneously, because of common db name `aries-temp`. btw, this is temp db should it be reused for new instances?

sudeshrshetty: reverted

test/aries-js-worker/test/messaging/messaging.js
```diff
@@ -0,0 +1,107 @@
...
+    "~l10n":{"locale":"en"},
+    "sent_time":"2020-03-05T16:59:47.489789-05:00",
+    "content":"Your hovercraft is full of eels."
+}
+
+// scenarios
+describe("Basic Messaging", function () {
+    this.timeout(5000);
+
+    let connections, didexClient, sender, receiver
```
llorllale: `connections` could be an object with distinct and descriptive properties instead of just a list. Eg.:

```js
{
    receiver: "",
    sender: ""
}
```

sudeshrshetty: here sender & receiver are aries instances
connections renamed to destinationConnID

PR #1427, opened 2020-03-06
feat: Issue Credential - adds an attachment to the message

soluchok:
* `requests attach` is populated from the inbound `OfferCredential` message
* `credentials attach` is populated from the inbound `RequestCredential` message

Part of #1173 

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/states.go
+++ b/pkg/didcomm/protocol/issuecredential/states.go
@@ -234,12 +234,15 @@ func (s *requestReceived) CanTransitionTo(st state) bool {
 }
 
 func (s *requestReceived) ExecuteInbound(md *metaData) (state, stateAction, error) {
+	if md.issueCredential == nil {
+		return nil, nil, errors.New("issue credential was not provided")
+	}
+
 	// creates the state's action
 	action := func(messenger service.Messenger) error {
 		// sets message type
-		return messenger.ReplyTo(md.Msg.ID(), service.NewDIDCommMsgMap(IssueCredential{
-			Type: IssueCredentialMsgType,
-		}))
+		md.issueCredential.Type = IssueCredentialMsgType
+		return messenger.ReplyTo(md.Msg.ID(), service.NewDIDCommMsgMap(md.issueCredential))
 	}
 
 	return &credentialIssued{}, action, nil
@@ -322,10 +325,16 @@ func (s *offerReceived) ExecuteInbound(md *metaData) (state, stateAction, error)
 		return &proposalSent{}, zeroAction, nil
 	}
 
+	var offer = OfferCredential{}
+	if err := md.Msg.Decode(&offer); err != nil {
+		return nil, nil, fmt.Errorf("decode: %w", err)
+	}
+
 	// creates the state's action
 	action := func(messenger service.Messenger) error {
 		return messenger.ReplyTo(md.Msg.ID(), service.NewDIDCommMsgMap(RequestCredential{
-			Type: RequestCredentialMsgType,
+			Type:           RequestCredentialMsgType,
+			RequestsAttach: offer.OffersAttach,
 		}))
 	}
 
--- a/pkg/didcomm/protocol/issuecredential/states_test.go
+++ b/pkg/didcomm/protocol/issuecredential/states_test.go
@@ -335,18 +335,27 @@ func TestRequestReceived_CanTransitionTo(t *testing.T) {
 }
 
 func TestRequestReceived_ExecuteInbound(t *testing.T) {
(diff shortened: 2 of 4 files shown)


pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -234,11 +234,17 @@ func (s *requestReceived) CanTransitionTo(st state) bool {
...
+		return nil, nil, fmt.Errorf("decode: %w", err)
+	}
+
 	// creates the state's action
 	action := func(messenger service.Messenger) error {
 		// sets message type
 		return messenger.ReplyTo(md.Msg.ID(), service.NewDIDCommMsgMap(IssueCredential{
-			Type: IssueCredentialMsgType,
+			Type:              IssueCredentialMsgType,
+			CredentialsAttach: req.RequestsAttach,
```
llorllale: The attachment should not be replayed back to the sender of the request. This new attachment should contain a credential.

soluchok: Done! I have added `WithIssueCredential` function.

pkg/didcomm/protocol/issuecredential/states_test.go
```diff
@@ -335,18 +335,27 @@ func TestRequestReceived_CanTransitionTo(t *testing.T) {
...
 func TestRequestReceived_ExecuteInbound(t *testing.T) {
-	followup, action, err := (&requestReceived{}).ExecuteInbound(&metaData{})
-	require.NoError(t, err)
-	require.Equal(t, &credentialIssued{}, followup)
-	require.NotNil(t, action)
+	t.Run("Successes", func(t *testing.T) {
+		followup, action, err := (&requestReceived{}).ExecuteInbound(&metaData{issueCredential: &IssueCredential{}})
+		require.NoError(t, err)
+		require.Equal(t, &credentialIssued{}, followup)
+		require.NotNil(t, action)
```
Baha-sk: NotNil or NotEmpty?

pkg/didcomm/protocol/issuecredential/states_test.go
```diff
@@ -335,18 +335,27 @@ func TestRequestReceived_CanTransitionTo(t *testing.T) {
...
+		messenger := serviceMocks.NewMockMessenger(ctrl)
+		messenger.EXPECT().ReplyTo(gomock.Any(), gomock.Any())
 
-	require.NoError(t, action(messenger))
+		require.NoError(t, action(messenger))
+	})
+
+	t.Run("IssueCredential is absent", func(t *testing.T) {
+		followup, action, err := (&requestReceived{}).ExecuteInbound(&metaData{})
+		require.Contains(t, fmt.Sprintf("%v", err), "issue credential was not provided")
```
Baha-sk: replace `fmt.Sprintf("%v", err)` with `err.Error()`

pkg/didcomm/protocol/issuecredential/states_test.go
```diff
@@ -496,6 +505,18 @@ func TestOfferReceived_ExecuteInbound(t *testing.T) {
...
 	})
+
+	t.Run("Decode error", func(t *testing.T) {
+		followup, action, err := (&offerReceived{}).ExecuteInbound(&metaData{
+			transitionalPayload: transitionalPayload{
+				Msg: service.DIDCommMsgMap{"@type": map[int]int{}},
+			},
+		})
+
+		require.Contains(t, fmt.Sprintf("%v", err), "got unconvertible type")
```
Baha-sk: same as above

pkg/didcomm/protocol/issuecredential/states_test.go
soluchok: This is not a good idea. `err.Error()` will cause a panic when `err` is `nil`. The best way to check it is to use `require.EqualError`. But `EqualError` requires full error string. Which causes a linter issue (lll e.g line is too long). 
So, I figured out that the best way is to use `Contains`. 
I can try to replace `Contains` with `EqualError`. But it will work only for errors with small length. Otherwise, I need to disable the linter.

pkg/didcomm/protocol/issuecredential/states_test.go
soluchok: Actually, it does not matter. `action` is a function. Which means that it can be only `nil` or a function.

pkg/didcomm/protocol/issuecredential/states_test.go
Baha-sk: this is an expected error test case, so err should be not nil, if it is, let the test panic, that's fine

PR #1429, opened 2020-03-06
fix: fix panic in case of parsing null did document

### LINKED ISSUE #1430: Got panic in case of parsing null as did document
**What I'm trying to do**

I'm parsing "null" string by did.ParseDocument

**Expected result**

I'm getting nil, error or empty struct.

**Actual result**

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x13e8130]

goroutine 6 [running]:
testing.tRunner.func1(0xc000154400)
	/usr/local/Cellar/go/1.13.4/libexec/src/testing/testing.go:874 +0x3a3
panic(0x1464900, 0x18609c0)
	/usr/local/Cellar/go/1.13.4/libexec/src/runtime/panic.go:679 +0x1b2
github.com/hyperledger/aries-framework-go/pkg/doc/did.ParseDocument(0xc0001ba818, 0x4, 0x4, 0xc000042760, 0x1083c26, 0x5e625ca9)
	/Users/belezeko/Projects/aries-framework-go/pkg/doc/did/doc.go:401 +0x1a0
github.com/hyperledger/aries-framework-go/pkg/doc/did.TestParseOfNull(0xc000154400)
	/Users/belezeko/Projects/aries-framework-go/pkg/doc/did/doc_test.go:144 +0x57
testing.tRunner(0xc000154400, 0x1503b48)
	/usr/local/Cellar/go/1.13.4/libexec/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/usr/local/Cellar/go/1.13.4/libexec/src/testing/testing.go:960 +0x350

Process finished with exit code 1
```

**Sample code / test case**

```
func TestParseOfNull(t *testing.T) {
	doc, err := ParseDocument([]byte("null"))
	require.NoError(t, err)
	require.Nil(t, doc)
}
```

AlexeyBelezeko:
Signed-off-by: Aleksei Belezeko <aleksei.belezeko@t-systems.com>

I found a bug that Parse Document get panic in case of null string. Now it returns nil in case of null string.

Closes #1430

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -386,16 +386,18 @@ type Proof struct {
 
 // ParseDocument creates an instance of DIDDocument by reading a JSON document from bytes
 func ParseDocument(data []byte) (*Doc, error) {
-	// validate did document
-	if err := validate(data); err != nil {
-		return nil, err
-	}
-
 	raw := &rawDoc{}
 
 	err := json.Unmarshal(data, &raw)
 	if err != nil {
 		return nil, fmt.Errorf("JSON marshalling of did doc bytes bytes failed: %w", err)
+	} else if raw == nil {
+		return nil, errors.New("document payload is not provided")
+	}
+	// validate did document
+	err = validate(data, raw.schemaLoader())
+	if err != nil {
+		return nil, err
 	}
 
 	publicKeys, err := populatePublicKeys(raw.Context[0], raw.PublicKey)
@@ -595,20 +597,18 @@ func decodePK(rawPK map[string]interface{}) ([]byte, error) {
 	return nil, errors.New("public key encoding not supported")
 }
 
-func validate(data []byte) error {
-	// Validate that the DID Document conforms to the serialization of the DID Document data model.
-	// Reference: https://w3c-ccg.github.io/did-spec/#did-documents)
+func (r *rawDoc) schemaLoader() gojsonschema.JSONLoader {
 	schemaLoader := schemaLoaderV1
-
-	var r rawDoc
-	if err := json.Unmarshal(data, &r); err != nil {
-		return fmt.Errorf("unmarshal failed: %w", err)
-	}
-
 	if len(r.Context) > 0 && r.Context[0] == contextV011 {
 		schemaLoader = schemaLoaderV011
 	}
 
+	return schemaLoader
+}
+
+func validate(data []byte, schemaLoader gojsonschema.JSONLoader) error {
+	// Validate that the DID Document conforms to the serialization of the DID Document data model.
+	// Reference: https://w3c-ccg.github.io/did-spec/#did-documents)
 	documentLoader := gojsonschema.NewStringLoader(string(data))
 
 	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
--- a/pkg/doc/did/doc_test.go
+++ b/pkg/doc/did/doc_test.go
@@ -140,6 +140,13 @@ const creator = did + "#key-1"
 const keyType = "Ed25519VerificationKey2018"
 const signatureType = "Ed25519Signature2018"
 
+func TestParseOfNull(t *testing.T) {
+	doc, err := ParseDocument([]byte("null"))
+	require.Error(t, err)
+	require.Nil(t, doc)
+	require.Contains(t, err.Error(), "document payload is not provided")
+}
+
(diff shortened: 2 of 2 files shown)


pkg/doc/did/doc.go
```diff
@@ -396,6 +396,8 @@ func ParseDocument(data []byte) (*Doc, error) {
 	err := json.Unmarshal(data, &raw)
 	if err != nil {
 		return nil, fmt.Errorf("JSON marshalling of did doc bytes bytes failed: %w", err)
+	} else if raw == nil {
+		return nil, nil
```
soluchok: We cannot return `nil, nil`. It can be  `&Doc{},nil` or `nil,err`.
We have a method `validate`. I guess this check should be filtered.
there @troyronda @llorllale What do you think?

Could you please attach the panic error you got?

AlexeyBelezeko: Thank you for quick response. I've created issue with the panic error.

pkg/doc/did/doc_test.go
```diff
@@ -140,6 +140,12 @@ const creator = did + "#key-1"
 const keyType = "Ed25519VerificationKey2018"
 const signatureType = "Ed25519Signature2018"
 
+func TestParseOfNull(t *testing.T) {
+	doc, err := ParseDocument([]byte("null"))
+	require.NoError(t, err)
```
llorllale: for sure there should be an error here.

pkg/doc/did/doc.go
soluchok: Here is my suggestion. You need to change the `validate` function as I did below:

```
	var r *rawDoc
	if err := json.Unmarshal(data, &r); err != nil {
		return fmt.Errorf("unmarshal failed: %w", err)
	}

	if r == nil {
		return fmt.Errorf("document payload is not provided")
	}
```

AlexeyBelezeko: Done.

soluchok: You can do even better. I see that `validate` and `ParseDocument` functions are doing the same job. They both do `json.Unmarshal`  into `rawDoc`. It would be nice to improve optimize that by pass `rawDoc` to the  `validate` function. If `rawDoc` is provided - ignore `Unmarshal`.

soluchok: e.g 
```
func validate(data []byte, r *rawDoc) error {
	if r == nil {
		if err := json.Unmarshal(data, &r); err != nil {
			return fmt.Errorf("unmarshal failed: %w", err)
		}
	}

	if r == nil {
		return errors.New("document payload is not provided")
	}
```
and

```
func ParseDocument(data []byte) (*Doc, error) {
	var raw *rawDoc

	err := json.Unmarshal(data, &raw)
	if err != nil {
		return nil, fmt.Errorf("JSON marshalling of did doc bytes bytes failed: %w", err)
	}

	// validate did document
	if err := validate(data, raw); err != nil {
		return nil, err
	}
```

pkg/doc/did/doc.go
```diff
@@ -386,16 +386,17 @@ type Proof struct {
...
 	raw := &rawDoc{}
 
 	err := json.Unmarshal(data, &raw)
 	if err != nil {
 		return nil, fmt.Errorf("JSON marshalling of did doc bytes bytes failed: %w", err)
+	} else if raw == nil {
+		return nil, errors.New("document payload is not provided")
+	}
+	// validate did document
+	if err := validate(data, raw.schemaLoader()); err != nil {
```
Baha-sk: replace the above line with:
```
err = validate(data, raw.schemaLoader())
if err != nil {
```
to remove the govet error

AlexeyBelezeko: Done

pkg/doc/did/doc_test.go
AlexeyBelezeko: Done

pkg/doc/did/doc.go
AlexeyBelezeko: Done

llorllale:
hi @AlexeyBelezeko 

Can you please squash your commits?

Thanks!

AlexeyBelezeko:
@llorllale squashed

Issue #1058, opened 2020-01-13
MessageService to use messenger api


sudeshrshetty:
- Remove `~purpose` from header and switch to map based lookup after (Ref: https://github.com/hyperledger/aries-framework-go/pull/1054)
- Use messenger instead of outbound dispatcher to send messages (Ref: https://github.com/hyperledger/aries-framework-go/issues/1039)

sudeshrshetty:
Send msg service endpoint in REST Binding
https://github.com/hyperledger/aries-framework-go/issues/973

PR #1436, opened 2020-03-06
feat: out of band messaging

sudeshrshetty:
- message service command controller to use messanger api
- send message to support out of band messages
- Part of #1058

### DIFF
--- a/pkg/controller/command/messaging/command.go
+++ b/pkg/controller/command/messaging/command.go
@@ -11,14 +11,13 @@ import (
 	"fmt"
 	"io"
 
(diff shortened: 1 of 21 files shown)


.golangci.yml
```diff
@@ -71,7 +71,7 @@ linters-settings:
       - unnamedResult
       - whyNoLint # TODO enable.
   funlen:
-    lines: 60
+    lines: 100
```
llorllale:
```suggestion
    lines: 60
```

sudeshrshetty: fixed it

PR #1441, opened 2020-03-10
feat: Issue Credential - client implementation

### LINKED ISSUE #1438: Issue Credential - client implementation
Need to implement a Client for the Issue Credential Protocol.

soluchok:
Implements  Issue Credential Client.
Includes only methods that start the protocol.

Part of #1438 


pkg/client/issuecredential/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Actions returns unfinished actions for the async usage
+func (c *Client) Actions() ([]issuecredential.Action, error) {
+	return c.service.Actions()
+}
+
+// SendOffer is used by the Issuer to send an offer.
+func (c *Client) SendOffer(offer *issuecredential.OfferCredential, myDID, theirDID string) error {
+	if offer == nil {
+		return errors.New("an empty offer received")
```
DRK3: I think `"received an empty offer"` sounds more correct to me.

pkg/client/issuecredential/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+	offer.Type = issuecredential.OfferCredentialMsgType
+
+	return c.service.HandleOutbound(service.NewDIDCommMsgMap(offer), myDID, theirDID)
+}
+
+// SendProposal is used by the Holder to send a proposal.
+func (c *Client) SendProposal(proposal *issuecredential.ProposeCredential, myDID, theirDID string) error {
+	if proposal == nil {
+		return errors.New("an empty proposal received")
```
DRK3: I think `"received an empty proposal"` sounds more correct to me.

pkg/client/issuecredential/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+	proposal.Type = issuecredential.ProposeCredentialMsgType
+
+	return c.service.HandleOutbound(service.NewDIDCommMsgMap(proposal), myDID, theirDID)
+}
+
+// SendRequest is used by the Holder to send a request.
+func (c *Client) SendRequest(request *issuecredential.RequestCredential, myDID, theirDID string) error {
+	if request == nil {
+		return errors.New("an empty request received")
```
DRK3: I think `"received an empty request"` sounds more correct to me.

DRK3: You can also define these error messages as unexported constants so that you don't have to repeat the text in the client_test.go file

pkg/client/issuecredential/client.go
```diff
@@ -0,0 +1,89 @@
...
+	}, nil
+}
+
+// Actions returns unfinished actions for the async usage
+func (c *Client) Actions() ([]issuecredential.Action, error) {
+	return c.service.Actions()
+}
+
+// SendOffer is used by the Issuer to send an offer.
+func (c *Client) SendOffer(offer *issuecredential.OfferCredential, myDID, theirDID string) error {
```
rolsonquadras: I feel having a target connection would make more sense than passing myDID and theirDID (internally service/client can derive this). 

ie, Issue credential to target connectionID - SendOffer(offer *issuecredential.OfferCredential, connectionID string)

soluchok: It is not a problem. It can be replaced at any time. Anyway, the connection should contain myDID and theirDID.

pkg/client/issuecredential/client.go
soluchok: Done!

PR #1443, opened 2020-03-11
feat: add ability to export and import public keys

### LINKED ISSUE #1442: Add support to extract a public key Handle as raw bytes and load it back into a key handle
in order to support exporting a public key (to be stored in DID documents or in DIDcomm envelopes) the localKMS must support exporting and importing Tink public keys to raw bytes and vice versa

Baha-sk:
Also updated kms_secretlock.md readme with more accurate description.

This patch adds `ExportPubKeyBytes()` and `PubKeyBytesToHandle()` functions to localKMS. The first is to export a Tink public key to bytes array and the second will create a Tink Keyset Handle from a public key as bytes array.

Only signing public keys are supported (EcDSA and EdDSA) in this change.

finally, updated Tink to v1.3.0

closes #1442


sandrask:
Functions names are bit long

Baha-sk:
> Functions names are bit long

They're long to be precise as of what kind of key type is exportable and in what format. Only asymmetric public keys can be exported in []byte format (and only for ECDSA and EDDSA curves). 

In the future, other asymmetric public keys may be exported if needed (like normal AES NIST approved or X25519 public keys used in Encryption)  and we can also support exporting in JWK format.

pkg/kms/localkms/pubkey_reader.go
```diff
@@ -0,0 +1,146 @@
...
+			return nil, "", err
+		}
+	default:
+		return nil, "", fmt.Errorf("invalid key type")
+	}
+
+	return keyValue, tURL, nil
+}
+
+func getMarshalledEcdsaKey(pubKey []byte, curveName string, c commonpb.EllipticCurveType,
```
troyronda: ECDSA should be caps.

Baha-sk: done

PR #1447, opened 2020-03-11
refactor: Public key resolver returning key type

### LINKED ISSUE #1446: Public key resolver returning key type
In `signature` and `verifiable` packages, we use a public key resolver abstraction that takes public key URI (e.g. DID) and resolves it into a public key ([]byte). 
In some cases, having public key bytes is not enough. For example, for KMS/Crypto based on https://github.com/google/tink, we need to convert the public key to key handle via exporting, the opposite operation is needed as well.

kdimak:
closes #1446

We will use the public key type in order to get KMS handle. A prototype function declaration from @Baha-sk :

```
// CreatePublicKeyHandle will create and return a key handle for pubKey of type kt
// it returns an error if it failed creating the key handle
// Note: The key handle created is not stored in the KMS
func (l *LocalKMS) CreatePublicKeyHandle(pubKey []byte, kt string)
```

This PR is a one of the preparation steps to use KMS/Crypto based signature verifiers in `verifiable`.


pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -591,7 +591,7 @@ func verifySignature(connSignature *ConnectionSignature, recipientKeys string) (
 	// TODO: Replace with signed attachments issue-626
 	signatureSuite := ed25519signature2018.New()
 
-	err = signatureSuite.Verify(pubKey, sigData, signature)
+	err = signatureSuite.Verify("ed25519", pubKey, sigData, signature)
```
sandrask: I would prefer to have PublicKey struct that contains Type and Value. Both type and value are properties of public key. Later on if we need additional properties it will be easy to add. @Baha-sk

Baha-sk: I'm also working on constraining the key types as constants in the KMS, we can reference these types instead of open string value (once my PR is merged you can take advantage of it).

kdimak: Agree, @sandrask .  I will refactor that.

pkg/doc/verifiable/support_test.go
```diff
@@ -202,3 +202,12 @@ func (s rs256TestSigner) Sign(data []byte) ([]byte, error) {
 
 	return rsa.SignPKCS1v15(rand.Reader, s.privKey, crypto.SHA256, hashed)
 }
+
+const (
+	rsaPublicKeyType     = "RSA"
+	ed25519PublicKeyType = "ed25519"
```
Baha-sk: probably expose these key types or better yet, use KeyType defined in pkg/kms/api.go as it should be the official key type in Aries-Framework-Go

and replace all references of "ed25519" in this patch.. there's at least 12, but in different packages than `verifiable`

pkg/kms/legacykms/kms_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"fmt"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
```
Baha-sk: move this import under line 20 below

kdimak: that's why the lint check failed... rebased

pkg/doc/verifiable/support_test.go
kdimak: Makes sense! I added `PublicKeyType` in a way similar to `kms.KeyType`.

pkg/doc/signature/verifier/verifier.go
```diff
@@ -23,17 +23,34 @@ type signatureSuite interface {
...
+// PublicKeyType is a type of a public key
+type PublicKeyType string
+
+const (
+	// RSAPublicKeyType public key type
+	RSAPublicKeyType = PublicKeyType("RSA")
+
+	// Ed25519PublicKeyType public key type
+	Ed25519PublicKeyType = PublicKeyType(kms.Ed25519Type)
+)
```
Baha-sk: Instead of creating new key type struct here, add `RSA` key type in kms/api.go and reference its KeyTypes directly.

We should have only one keyType struct across the framework to avoid duplication.

kdimak: OK, I wil change that!

kdimak: changed

pkg/doc/verifiable/linked_data_proof.go
```diff
@@ -38,7 +37,7 @@ type verifierSignatureSuite interface {
 	signatureSuite
 
 	// Verify will verify signature against public key
-	Verify(pubKey []byte, doc []byte, signature []byte) error
+	Verify(pubKey *verifier.PublicKey, doc []byte, signature []byte) error
```
Baha-sk: add a TODO to replace `verifier.PublicKey` with `keyset.Handle`.

`Verify()` implementations should call pkg/crypto's Verify() function with the instance of keyset.Handle passed here.

And the same thing for Sign() implementations, they should use pkg/crypto's default Sign() function that accepts a keyset.Handle as argument.

pkg/doc/verifiable/linked_data_proof.go
```diff
@@ -52,25 +51,20 @@ type keyResolverAdapter struct {
 	pubKeyFetcher PublicKeyFetcher
 }
 
-func (k *keyResolverAdapter) Resolve(id string) ([]byte, error) {
+func (k *keyResolverAdapter) Resolve(id string) (*verifier.PublicKey, error) {
```
Baha-sk: same here... add a TODO comment that says: replace `*verifier.PublickKey` with Tink's `*keyset.Handle`

pkg/doc/verifiable/linked_data_proof.go
kdimak: I think we wanted to don't make tight coupling with crypto/kms? Just make generic interfaces of signer / verifier without passing a private key as an argument to signer? @troyronda

Baha-sk: or if we don't want to limit our APIs to Tink, then make the pubKey argument to be `interface{}` type.

and default implementation will use `*keyset.Handle` Tink type

other implementations requiring []byte can use `*verifier.PublicKey`

Baha-sk: > I think we wanted to don't make tight coupling with crypto/kms? Just make generic interfaces of signer / verifier without passing a private key as an argument to signer? @troyronda

the concept of hiding keys entails using a key reference to execute a crypto primitive, which should be resolved by a kms implementation and primitive executed by a crypto implementation.

in order to not tie the code to any specific implementation, we should simply use `interface{}` type for keys. This way, kms/crypto Tink implementations can exchange `*keyset.Handle` with verifierSignatureSuite/signerSignatureSuite implemented with Tink key handles. Other KMS implementations will exchange their own key type with their respective signature suite implementations.

kdimak: I see your point. Thank you for the explanation! I created https://github.com/hyperledger/aries-framework-go/issues/1458.

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -90,7 +97,7 @@ func ExamplePresentation_JWTClaims() {
 `
 
 	// The Holder wants to send the presentation to the Verifier in JWS.
-	vp, err := verifiable.NewPresentation([]byte(vpStrFromWallet))
+	vp, err := verifiable.NewUnverifiedPresentation([]byte(vpStrFromWallet))
```
troyronda: Why this change?

pkg/doc/verifiable/presentation.go
```diff
@@ -306,6 +306,27 @@ func NewPresentation(vpData []byte, opts ...PresentationOpt) (*Presentation, err
 		return nil, err
 	}
 
+	return newPresentation(vpRaw, vpOpts)
+}
+
+// NewUnverifiedPresentation decodes Verifiable Presentation from bytes which could be marshalled JSON or
+// serialized JWT. It does not make a proof check though. Can be used for purposes of decoding of VP stored in a wallet.
+// Please use this function with caution.
+func NewUnverifiedPresentation(vpBytes []byte) (*Presentation, error) {
```
troyronda: We should have a TODO to re-do naming (e.g., with Parse prefix).

pkg/doc/verifiable/example_presentation_test.go
kdimak: `vpStrFromWallet` has no embedded proof. If we pass it to `NewPresentation()` as JSON bytes, it will raise an error saying that no proof was found.
This test did not fail before because of `//nolint:lll` before `//Output:` line which made the test function non-effective. I moved `//nolint:lll` to the line before the method declaration to enable the test. There were about 10 not working example-based tests in `verifiable`.

pkg/doc/verifiable/presentation.go
kdimak: Yeah, I will file the issue soon.

PR #1450, opened 2020-03-11
Fix: use v1 when context is empty

fqutishat:


pkg/doc/did/doc.go
```diff
@@ -670,14 +670,20 @@ func stringArray(entry interface{}) []string {
 
 // JSONBytes converts document to json bytes
 func (doc *Doc) JSONBytes() ([]byte, error) {
+	context := Context
+
+	if len(doc.Context) > 0 {
+		context = doc.Context[0]
```
troyronda: I don’t understand this change. Why are we removing contexts.

fqutishat: I didn't remove context just use v1 if context is empty

PR #1452, opened 2020-03-12
feat: Issue Credential - saves credential to the verifiable store

soluchok:
When the Holder receives a credential, the protocol saves it to the verifiable store.

Part of #1173 

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/service.go
+++ b/pkg/didcomm/protocol/issuecredential/service.go
@@ -15,7 +15,9 @@ import (
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
+	storeverifiable "github.com/hyperledger/aries-framework-go/pkg/store/verifiable"
 )
 
 const (
@@ -62,9 +64,11 @@ type transitionalPayload struct {
 // metaData type to store data for internal usage
 type metaData struct {
 	transitionalPayload
-	state    state
-	msgClone service.DIDCommMsg
-	inbound  bool
+	state           state
+	msgClone        service.DIDCommMsg
+	inbound         bool
+	credentials     []*verifiable.Credential
+	credentialNames []string
 	// keeps offer credential payload,
 	// allows filling the message by providing an option function
 	offerCredential   *OfferCredential
@@ -110,6 +114,14 @@ func WithIssueCredential(msg *IssueCredential) Opt {
 	}
 }
 
+// WithFriendlyNames allows providing names for the credentials.
+// USAGE: This function should be used when the Holder receives IssueCredential message
+func WithFriendlyNames(names ...string) Opt {
+	return func(md *metaData) {
+		md.credentialNames = names
+	}
+}
+
 // Provider contains dependencies for the protocol and is typically created by using aries.Context()
 type Provider interface {
 	Messenger() service.Messenger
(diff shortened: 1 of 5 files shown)


pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -380,7 +383,44 @@ func (s *credentialReceived) CanTransitionTo(st state) bool {
 	return st.Name() == stateNameDone || st.Name() == stateNameAbandoning
 }
 
+func toVerifiableCredentials(attachments []decorator.Attachment) ([]*verifiable.Credential, error) {
+	var credentials []*verifiable.Credential
+
+	// TODO: Currently, it supports only JSON payload. We need to add support for links and base64 as well.
```
troyronda: issue number?

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -295,6 +305,12 @@ func (s *Service) handle(md *metaData) error {
 		return fmt.Errorf("failed to persist state %s: %w", stateName, err)
 	}
 
+	for _, credential := range md.credentials {
+		if err := s.verifiable.SaveCredential(credential.ID, credential); err != nil {
```
rolsonquadras: Why does service need to store the credential directly? Shouldn't it be up-to the consumer to decide ?

soluchok: The consumer can decide whether to accept issued credentials or not. If he accepts it we simply store it into the verifiable store. 

Why does service need to store the credential directly? - it is according to the [diagram](https://github.com/hyperledger/aries-rfcs/blob/master/features/0036-issue-credential/credential-issuance.png).
NOTE: See when the Holder receives an issue-credential message.

pkg/didcomm/protocol/issuecredential/states.go
soluchok: Sure, I will add it!

soluchok: Done!

pkg/didcomm/protocol/issuecredential/service.go
rolsonquadras: Ideally we need to store it with a friendly name, so that the user can search easily. May be we need to take that input from the user while accepting.

soluchok: I skipped this part with a friendly name. I did not know what to put there so, I decided to use ID. Get a friendly name from the user sounds good. But as I understand we can receive several credentials.  Not sure whether it is ok to provide an option for that e.g `(credentialNames ...string)`. Looks like the user should count credentials a provide the same number of names. What if names did not match the count of credentials? Should it be an error or we can use ID? 

What do you think about the name? Is it ok to generate a name e.g based on the ID or any other field from the credential payload?
Or provide a function to e.g. `WithCustomName(credentialNames ...string) Opt`

llorllale: @soluchok fow now i would provide `WithFriendlyNames(names ...string)`

soluchok: Ok

soluchok: Done!

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -304,6 +327,21 @@ func (s *Service) handle(md *metaData) error {
...
 }
 
+func (s *Service) saveCredentials(md *metaData) error {
+	for i, credential := range md.credentials {
+		var name = credential.ID
+		if len(md.credentialNames) > i {
+			name = md.credentialNames[i]
+		}
+
+		if err := s.verifiable.SaveCredential(name, credential); err != nil {
```
troyronda: Any reason for inline err style here?

soluchok: No, I just prefer to use inline. Should not I?

Issue #1416, opened 2020-03-04
Use KMS / Crypto in JWS and LDP tests of VC


kdimak:
According to @Baha-sk , 

> It would be nice to use a KMS instance like the example code above to show that it works with VCs/VPs as opposed to creating signers using a private key .. if you don't have access to an aries instance, you can create the KMS and Crypto instances from scratch.

It's a follow-up issue of #339 .

Issue #1267, opened 2020-02-14
Pluggable JWS verifiers


kdimak:
`Verifier` interface is defined as 
```
type Verifier interface {
	Verify(j JSONWebToken) error
}
```
There is one basic implementation of it created by NewVerifier(). It now supports EdDSA and RS256. And currently, there is no way to plug in an extra concrete algorithm verifier.
We need to get a possibility to:
1) add a verifier for a new algorithm
2) override default verifier (e.g. for EdDSA signatures).

kdimak:
See https://github.com/hyperledger/aries-framework-go/pull/1265#discussion_r379847069

PR #1463, opened 2020-03-16
refactor: ed25519 signature verifier option

### LINKED ISSUE #1458: Take interface{} as public key for signature verifier
See https://github.com/hyperledger/aries-framework-go/pull/1447#discussion_r392448914

kdimak:
closes #1458

Follow-up issue to use Crypto for VC: https://github.com/hyperledger/aries-framework-go/issues/1416
Follow-up issue to use Crypto in JWS/JWT: https://github.com/hyperledger/aries-framework-go/issues/1267

### DIFF
--- a/pkg/didcomm/protocol/didexchange/states.go
+++ b/pkg/didcomm/protocol/didexchange/states.go
@@ -591,10 +591,11 @@ func verifySignature(connSignature *ConnectionSignature, recipientKeys string) (
 	pubKey := base58.Decode(recipientKeys)
 
 	// TODO: Replace with signed attachments issue-626
-	signatureSuite := ed25519signature2018.New()
+	suiteVerifier := &ed25519signature2018.PublicKeyVerifier{}
+	signatureSuite := ed25519signature2018.New(ed25519signature2018.WithVerifier(suiteVerifier))
 
 	err = signatureSuite.Verify(&verifier.PublicKey{
-		Type:  kms.Ed25519Type,
+		Type:  kms.ED25519,
 		Value: pubKey},
 		sigData, signature)
 	if err != nil {
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -20,7 +20,6 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
-	"github.com/hyperledger/aries-framework-go/pkg/kms"
 )
 
 const (
@@ -781,7 +780,7 @@ func (r *didKeyResolver) Resolve(id string) (*verifier.PublicKey, error) {
 	for _, key := range r.PubKeys {
 		if key.ID == id {
 			return &verifier.PublicKey{
-				Type:  kms.KeyType(key.Type),
+				Type:  key.Type,
 				Value: key.Value,
 			}, nil
 		}
--- a/pkg/doc/signature/ed25519signature2018/public_key_verifier.go
+++ b/pkg/doc/signature/ed25519signature2018/public_key_verifier.go
@@ -1,0 +1,35 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ed25519signature2018
+
+import (
+	"crypto/ed25519"
+	"errors"
+
+	sigverifier "github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
+)
+
+// PublicKeyVerifier verifies a Ed25519 signature taking Ed25519 public key bytes as input.
+// NOTE: this verifier is present for backward compatibility reasons and can be removed soon.
+// Please use CryptoVerifier or your own verifier.
+type PublicKeyVerifier struct {
+}
+
+// Verify will verify a signature.
+func (v *PublicKeyVerifier) Verify(pubKey *sigverifier.PublicKey, doc, signature []byte) error {
+	// ed25519 panics if key size is wrong
+	if len(pubKey.Value) != ed25519.PublicKeySize {
+		return errors.New("ed25519: invalid key")
+	}
+
+	verified := ed25519.Verify(pubKey.Value, doc, signature)
+	if !verified {
+		return errors.New("ed25519: invalid signature")
+	}
+
+	return nil
+}
--- a/pkg/doc/signature/ed25519signature2018/suite.go
+++ b/pkg/doc/signature/ed25519signature2018/suite.go
@@ -12,18 +12,18 @@ SPDX-License-Identifier: Apache-2.0
 package ed25519signature2018
 
 import (
-	"crypto/ed25519"
 	"crypto/sha256"
 	"errors"
 
 	"github.com/piprate/json-gold/ld"
 
-	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
+	sigverifier "github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
 )
 
 // SignatureSuite implements ed25519 signature suite
 type SignatureSuite struct {
-	signer signer
+	signer   signer
+	verifier verifier
 }
 
 const (
@@ -36,6 +36,11 @@ type signer interface {
 	Sign(data []byte) ([]byte, error)
 }
 
+type verifier interface {
+	// Verify will verify a signature.
+	Verify(pubKeyValue *sigverifier.PublicKey, doc, signature []byte) error
+}
+
 // SuiteOpt is the SignatureSuite option.
 type SuiteOpt func(opts *SignatureSuite)
 
@@ -43,6 +48,13 @@ type SuiteOpt func(opts *SignatureSuite)
 func WithSigner(s signer) SuiteOpt {
 	return func(opts *SignatureSuite) {
 		opts.signer = s
+	}
+}
+
+// WithVerifier defines a verifier for the Signature Suite.
+func WithVerifier(v verifier) SuiteOpt {
+	return func(opts *SignatureSuite) {
+		opts.verifier = v
 	}
 }
 
@@ -81,18 +93,12 @@ func (s *SignatureSuite) GetDigest(doc []byte) []byte {
 }
 
 // Verify will verify a signature.
-func (s *SignatureSuite) Verify(pubKey *verifier.PublicKey, doc, signature []byte) error {
-	// ed25519 panics if key size is wrong
-	if l := len(pubKey.Value); l != ed25519.PublicKeySize {
-		return errors.New("ed25519: bad public key length")
+func (s *SignatureSuite) Verify(pubKeyValue *sigverifier.PublicKey, doc, signature []byte) error {
+	if s.verifier == nil {
+		return ErrVerifierNotDefined
 	}
 
-	verified := ed25519.Verify(pubKey.Value, doc, signature)
-	if !verified {
-		return errors.New("signature doesn't match")
-	}
-
-	return nil
+	return s.verifier.Verify(pubKeyValue, doc, signature)
 }
 
 // Sign will sign input data.
@@ -111,3 +117,6 @@ func (s *SignatureSuite) Accept(t string) bool {
 
 // ErrSignerNotDefined is returned when Sign() is called but signer option is not defined.
 var ErrSignerNotDefined = errors.New("signer is not defined")
+
+// ErrVerifierNotDefined is returned when Verify() is called but verifier option is not defined.
+var ErrVerifierNotDefined = errors.New("verifier is not defined")
--- a/pkg/doc/signature/ed25519signature2018/suite_crypto.go
+++ b/pkg/doc/signature/ed25519signature2018/suite_crypto.go
@@ -1,0 +1,48 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ed25519signature2018
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/crypto"
+	sigverifier "github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
+)
+
+// CryptoSigner defines signer based on crypto.
+type CryptoSigner struct {
+	cr crypto.Crypto
+	kh interface{}
+}
+
+// NewCryptoSigner creates a new CryptoSigner.
+func NewCryptoSigner(cr crypto.Crypto, kh interface{}) *CryptoSigner {
+	return &CryptoSigner{
+		cr: cr,
+		kh: kh,
+	}
+}
+
+// Sign will sign document and return signature.
+func (s *CryptoSigner) Sign(msg []byte) ([]byte, error) {
+	return s.cr.Sign(msg, s.kh)
+}
+
+// CryptoVerifier defines signature verifier based on crypto.
+type CryptoVerifier struct {
+	cr crypto.Crypto
+}
+
+// NewCryptoVerifier creates a new CryptoVerifier.
+func NewCryptoVerifier(cr crypto.Crypto) *CryptoVerifier {
+	return &CryptoVerifier{
+		cr: cr,
+	}
+}
+
+// Verify will verify a signature.
+func (v *CryptoVerifier) Verify(kh *sigverifier.PublicKey, msg, signature []byte) error {
+	return v.cr.Verify(signature, msg, kh)
+}
--- a/pkg/doc/signature/ed25519signature2018/suite_crypto_test.go
+++ b/pkg/doc/signature/ed25519signature2018/suite_crypto_test.go
@@ -1,0 +1,119 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ed25519signature2018
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/google/tink/go/keyset"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto"
+	sigverifier "github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
+	kmsapi "github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/localkms"
+	mockkms "github.com/hyperledger/aries-framework-go/pkg/mock/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/mock/storage"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock/noop"
+)
+
+func TestNewCryptoSignerAndVerifier(t *testing.T) {
+	lKMS := createKMS()
+
+	kid, kh := createKeyHandle(lKMS, kmsapi.ED25519Type)
+
+	tinkCrypto, err := tinkcrypto.New()
+	if err != nil {
+		panic("failed to create tinkcrypto")
+	}
+
+	doc := []byte("test doc")
+
+	suiteSigner := NewCryptoSigner(tinkCrypto, kh)
+	suiteVerifier := NewCryptoVerifier(&Crypto{
+		Crypto:   tinkCrypto,
+		localKMS: lKMS,
+	})
+
+	ss := New(WithSigner(suiteSigner), WithVerifier(suiteVerifier))
+
+	docSig, err := ss.Sign(doc)
+	if err != nil {
+		panic("failed to create a signature")
+	}
+
+	pubKeyBytes, err := lKMS.ExportPubKeyBytes(kid)
+	if err != nil {
+		panic("failed to export public key bytes")
+	}
+
+	pubKey := &sigverifier.PublicKey{
+		Type:  kmsapi.ED25519,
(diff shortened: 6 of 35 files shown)


pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -81,18 +93,28 @@ func (s *SignatureSuite) GetDigest(doc []byte) []byte {
...
+			return errors.New("ed25519: bad public key length")
+		}
+
+		verified := ed25519.Verify(pubKey.Value, doc, signature)
+		if !verified {
+			return errors.New("signature doesn't match")
+		}
+
+		return nil
+
```
Baha-sk: I assume `case *keyset.Handle:` is coming up in another PR?

kdimak: I am basing on a more generic Crypto API interface which takes `kh interface{}`. In this way, not only **tinkrypto** is supported but any other crypto implementation as well. That's my idea.

Baha-sk: I'm ok with generic SignatureSuite API.

Similarly to KMS and Crypto APIs, they offer generic functions. The default implementation is using Tink (`*keyset.Handle`) as argument types.

same here, there should be a case for Tink's *keyset.Handle to call aries-framwork-go's default Crypto.Verify() instead of directly calling de25519 with []byte key

kdimak: It is what is happening in a `default` of the `switch`: we use Crypto instance passed as a functional argument to verify a signature. I think there is no need to introduce a direct dependency on `tinkcrypto` here.

kdimak: OK, we agreed with @Baha that can use `localkms.PubKeyBytesToHandle()` to convert public key bytes to key handle and use it with crypto. This will require some time to change, I will push the solution later on today.

kdimak: I think we should also check a type of the crypto impl. It it's a `tinkcrypto`, which can convert public key bytes to `*keyset.Handle` and call crypto.Verify(). In contrast, if other crypto impl is used then it could not accept `*keyset.Handle`, in this case, we will get into trouble. What do you think? @Baha-sk  @troyronda

pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -81,18 +93,28 @@ func (s *SignatureSuite) GetDigest(doc []byte) []byte {
...
 // Verify will verify a signature.
-func (s *SignatureSuite) Verify(pubKey *verifier.PublicKey, doc, signature []byte) error {
-	// ed25519 panics if key size is wrong
-	if l := len(pubKey.Value); l != ed25519.PublicKeySize {
-		return errors.New("ed25519: bad public key length")
+func (s *SignatureSuite) Verify(pubKeyValue interface{}, doc, signature []byte) error {
+	switch pubKey := pubKeyValue.(type) {
+	case *verifier.PublicKey:
+		// ed25519 panics if key size is wrong
+		if l := len(pubKey.Value); l != ed25519.PublicKeySize {
```
soluchok: `l := len(pubKey.Value); l != ed25519.PublicKeySize` -> `len(pubKey.Value) != ed25519.PublicKeySize`

pkg/doc/signature/ed25519signature2018/suite.go
Baha-sk: this is why `case *keyset.Handle:` will manage Tink wrapped keys, other implementations can support `case verify.PublicKey`

The API will be flexible with these two case conditions.

pkg/doc/signature/ed25519signature2018/suite_test.go
```diff
@@ -42,37 +49,102 @@ func TestSignatureSuite_Sign(t *testing.T) {
...
+	pubKeyBytes, err := lKMS.ExportPubKeyBytes(kid)
+	require.NoError(t, err)
+
+	pubKey := &sverifier.PublicKey{
+		Type:  kmsapi.Ed25519Type,
+		Value: pubKeyBytes,
+	}
+
+	err = ss.Verify(pubKey, doc, docSig)
+	require.NoError(t, err)
```
kdimak: @Baha-sk the test is failing here saying that signature is invalid:
```
verify msg: verifier_factory: invalid signature
```
Do you have any idea why it's happening?

kdimak: Filed an issue regarding this: https://github.com/hyperledger/aries-framework-go/issues/1483
I disabled accepting a public key bytes input when verifying a signature via `LocalCryptoVerifier`.

kdimak:
I added `verifier` concept for the Signature Suite. It is defined by functional option in the same way as `signer` before. There are two default implementations of the `verifier`:
1.  `LocalCryptoVerifier` uses crypto and Local KMS for signature verification. The latter should be used to convert public key bytes to key handle and pass it to tinkcrypto (i removed support of this because of https://github.com/hyperledger/aries-framework-go/issues/1483).
2. `PublicKeyVerifier` is an old verification way using public key bytes. Other packages that use signature verification (like `did`  or `verifiable`) now use this type of verifier (mainly in the tests).

pkg/doc/did/doc.go
```diff
@@ -790,6 +795,16 @@ func (r *didKeyResolver) Resolve(id string) (*verifier.PublicKey, error) {
 	return nil, ErrKeyNotFound
 }
 
+// MapKeyTypeToKMS maps key type of DID doc to the KMS one.
+func MapKeyTypeToKMS(t string) (kms.KeyType, error) {
+	switch t {
+	case "Ed25519VerificationKey2018":
+		return kms.Ed25519Type, nil
```
troyronda: I'm a bit confused. What's the purpose of using interface{} above when we are mapping to definitions in kms anyways?

pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -36,6 +42,89 @@ type signer interface {
...
+		localKMS: localKMS,
+	}
+}
+
+// Verify will verify a signature.
+func (v *LocalCryptoVerifier) Verify(pubKeyValue interface{}, doc, signature []byte) error {
+	var keyHandle *keyset.Handle
+
+	switch pubKey := pubKeyValue.(type) {
+	case *keyset.Handle:
```
troyronda: I don't really understand this change. 
It seems like we are using an interface{} and then requiring deep knowledge of the types (coupling the kms types).

ed25519 package should not know about tink.

pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -36,6 +42,89 @@ type signer interface {
...
+	return v.cr.Verify(signature, doc, keyHandle)
+}
+
+// PublicKeyVerifier verifies a Ed25519 signature taking Ed25519 public key bytes as input.
+type PublicKeyVerifier struct {
+}
+
+// Verify will verify a signature.
+func (v *PublicKeyVerifier) Verify(pubKeyValue interface{}, doc, signature []byte) error {
+	pubKey, ok := pubKeyValue.(*sigverifier.PublicKey)
```
troyronda: I'm a bit confused. It seems like we require a particular interface in this line, but not in the func signature?

pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -36,6 +42,89 @@ type signer interface {
...
+	if pubKey.Type != kms.Ed25519Type {
+		return fmt.Errorf("invalid public key type: %s", pubKey.Type)
+	}
+
+	// ed25519 panics if key size is wrong
+	if len(pubKey.Value) != ed25519.PublicKeySize {
+		return errors.New("ed25519: bad public key length")
+	}
+
+	verified := ed25519.Verify(pubKey.Value, doc, signature)
```
troyronda: this should be part of the kms/crypto implementation @Baha-sk

pkg/doc/signature/ed25519signature2018/suite.go
```diff
@@ -36,6 +42,89 @@ type signer interface {
...
+}
+
+type verifier interface {
+	// Verify will verify a signature.
+	Verify(pubKeyValue interface{}, doc, signature []byte) error
+}
+
+// LocalCryptoVerifier defines a verifier which is based on Local KMS and Crypto
+// which uses keyset.Handle as input for verification.
+type LocalCryptoVerifier struct {
```
troyronda: This doesn't belong in this package @Baha-sk @kdimak

pkg/doc/did/doc.go
```diff
@@ -747,7 +747,7 @@ type signatureSuite interface {
 	GetDigest(doc []byte) []byte
 
 	// Verify will verify signature against public key
-	Verify(pubKey *verifier.PublicKey, doc []byte, signature []byte) error
+	Verify(pubKey interface{}, doc []byte, signature []byte) error
```
troyronda: this change *only* makes sense if pubKey is an opaque object to every package except for a kms implementation.

pkg/doc/did/doc.go
```diff
@@ -777,11 +777,16 @@ type didKeyResolver struct {
 	PubKeys []PublicKey
 }
 
-func (r *didKeyResolver) Resolve(id string) (*verifier.PublicKey, error) {
+func (r *didKeyResolver) Resolve(id string) (interface{}, error) {
```
troyronda: this change __only__ makes sense if pubKey is an opaque object to every package except for a kms implementation.

but I see this package constructing a generic return value. this is confusing.

pkg/doc/did/doc.go
troyronda: and I also see resolve actually constructing a generic value... (so pubkey doesn't seem to be opaque at all).

pkg/doc/signature/ed25519signature2018/suite.go
Baha-sk: `PublicKeyVerifier` is for keys in []byte format

pkg/doc/signature/ed25519signature2018/suite.go
Baha-sk: this is because `publicKeyValue` can be either a key reference (Tink's *keyset.Handle) or the raw []byte publicKey, similar to Crypto API that takes a `kh interface` argument for the key reference and the default implementation uses Tink's *keyset.Handle as kh type

pkg/doc/signature/ed25519signature2018/suite.go
Baha-sk: Agree, this struct can do any signature verification, not just ed25519.

ed25519signaturesuite2018 can use this struct to do verification, as well as any NIST approved ecdsa signature suite.

pkg/doc/signature/ed25519signature2018/suite.go
Baha-sk: I think @kdimak wants to separate signature verification using raw key []byte from Tink keys and this verify uses []byte keys with crypto.ed25519 package.

So `PublicKeyVerifier` is for keys with raw bytes and `LocalKeyVerifier` is for Tink keys (with public key loaded from a DID doc as []byte and transformed by localkKMS into *keyset.Handle which requires a fix I'm currently working on)

pkg/doc/signature/ed25519signature2018/suite.go
troyronda: More specifically, this package should not know what kms is in use nor anything about tink.

pkg/doc/signature/ed25519signature2018/suite_test.go
Baha-sk: yeah, the signature has a special Tink prefix based on Tink key types, we can either fix it in Tink or manually remove the prefix from docSig and try `ss.Verify()` again... 

I will post an issue in Tink to see if we can get this added to Tink

pkg/doc/signature/ed25519signature2018/suite.go
troyronda: This package should not know such details (that would be the reason for an empty interface).

pkg/doc/did/doc.go
```diff
@@ -790,6 +795,16 @@ func (r *didKeyResolver) Resolve(id string) (*verifier.PublicKey, error) {
 	return nil, ErrKeyNotFound
 }
 
+// MapKeyTypeToKMS maps key type of DID doc to the KMS one.
+func MapKeyTypeToKMS(t string) (kms.KeyType, error) {
```
troyronda: This function doesn't belong in the did package.

kdimak:
Now there is no dependency on KMS/Crypto in `ed25519signature2018`.  `keyResolver` produces generic `interface{}` and passes it to the `verifier`. I left `PublicKeyVerifier` in the production code as it is still used in many places in aries. Also, crypto operations like `ed25519.Verify()` are still present as I don't see a way now of how to use Crypto directly for this purpose. E.g. I would like to have the possibility to call a Crypto function with the following definition: 
```
Verify(signature, msg []byte, kms.KeyType type, privKey []byte)
```
If present, I would call it `Verify(signature, msg, kms.Ed25519Type, privKey)` and that's it.
I also added naive implementation of `signer/verifier` based on `crypto.Crypto` in a separate file. In the tests, one can see how it can be implemented using tinkcrypto/localkms.

pkg/doc/did/doc.go
kdimak: Why? The key types are defined at https://www.w3.org/TR/did-core/#public-keys spec, and we need to map them to our internal key types.

pkg/doc/signature/ed25519signature2018/public_key_verifier.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+)
+
+// PublicKeyVerifier verifies a Ed25519 signature taking Ed25519 public key bytes as input.
+type PublicKeyVerifier struct {
+}
+
+// Verify will verify a signature.
+func (v *PublicKeyVerifier) Verify(pubKeyValue interface{}, doc, signature []byte) error {
+	pubKey, ok := pubKeyValue.(*sigverifier.PublicKey)
```
troyronda: This function should be calling a KMS implementation, not converting interfaces here... Is that the plan?

pkg/doc/signature/ed25519signature2018/suite_crypto.go
```diff
@@ -0,0 +1,45 @@
...
+	return s.cr.Sign(msg, s.kh)
+}
+
+// CryptoVerifier defines signature verifier based on crypto.
+type CryptoVerifier struct {
+	cr crypto.Crypto
+}
+
+// NewCryptoVerifier creates a new CryptoVerifier.
+func NewCryptoVerifier(cr crypto.Crypto) *CryptoVerifier {
```
troyronda: Why is this only called by tests?

pkg/doc/signature/ed25519signature2018/public_key_verifier.go
kdimak: `NewCryptoVerifier()` is what we would like users to use. `PublicKeyVerifier` is a sort-of backward compatibility struct which is used in other packages (`verifiable`, `did`, `didexchange`,  `legacykms`) and should be replaced in the follow-up PRs. I can leave a comment regarding this in the code.

pkg/doc/signature/ed25519signature2018/suite_crypto.go
kdimak: This will be changed in the follow-up PRs in order to not make the current PR that huge.

pkg/doc/did/doc.go
sandrask: I agree that we should not return interface here.

pkg/doc/did/doc.go
```diff
@@ -777,11 +777,16 @@ type didKeyResolver struct {
 	PubKeys []PublicKey
 }
 
-func (r *didKeyResolver) Resolve(id string) (*verifier.PublicKey, error) {
+func (r *didKeyResolver) Resolve(id string) (interface{}, error) {
 	for _, key := range r.PubKeys {
 		if key.ID == id {
+			keyType, err := MapKeyTypeToKMS(key.Type)
```
sandrask: We could defer key type  mapping when it is needed (sign, verify) and not tie it to resolution. 
Did package should not know about KMS types. This should be postponed until sign or verify. You can have helper function on crypto before PubKeyBytesToHandle(pubKey.Value, pubKey.Type) is called.

pkg/doc/signature/ed25519signature2018/suite_crypto_test.go
```diff
@@ -0,0 +1,110 @@
...
+type Crypto struct {
+	*tinkcrypto.Crypto
+	localKMS *localkms.LocalKMS
+}
+
+func (t *Crypto) Verify(sig, msg []byte, kh interface{}) error {
+	var keyHandle *keyset.Handle
+
+	switch pubKey := kh.(type) {
+	case *keyset.Handle:
```
sandrask: I am not sure in what case handle would be passed in. Normally use case:
1. retrieve key from did (pub key bytes, type) 
2. use pub key for verification.

Baha-sk: usually pub key bytes will be transformed into public keyset.Handle to use kms/crypto's default implementation services

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -591,10 +590,11 @@ func verifySignature(connSignature *ConnectionSignature, recipientKeys string) (
 	pubKey := base58.Decode(recipientKeys)
 
 	// TODO: Replace with signed attachments issue-626
-	signatureSuite := ed25519signature2018.New()
+	suiteVerifier := &ed25519signature2018.PublicKeyVerifier{}
+	signatureSuite := ed25519signature2018.New(ed25519signature2018.WithVerifier(suiteVerifier))
 
 	err = signatureSuite.Verify(&verifier.PublicKey{
-		Type:  kms.Ed25519Type,
+		Type:  "ED25519",
```
troyronda: why this change?

pkg/doc/did/doc.go
kdimak: I agree with you and @troyronda . I removed this conversion. I also changed the `Type` of the `PublicKey` to be just a string as we discussed on the call with @sandrask .

pkg/didcomm/protocol/didexchange/states.go
sandrask: I commented that we tie kms to our signature suite. Now that I see this it doesn't look good either. Dima please revert back. I expected this type to be passed in from resolver - not hard-coded.

pkg/doc/signature/ed25519signature2018/suite_crypto_test.go
kdimak: I changed `publicKeyResolver.Resolve()` and `verifier.Verify()` to take `PublicKey` instead of genetic `interface{}`. The reason is that we don't see any use case to take e.g.  `*keyset.Handle` for now.
It looks like that when we would like to use `tinkcypto.Verifier`, we can intercept the call and convert `PublicKey` into the `*keyset.Handle` using `PubKeyBytesToHandle()`. This approach is implemented in **suite_crypto_test.go**.

pkg/didcomm/protocol/didexchange/states.go
kdimak: The type is now a `string`, not `kms.KeyType`. I decided it will look simpler in the tests to use just constant instead of `string(kms.Ed25519Type)` (plus extra import).

kdimak: I updated constants usage. As @troyronda suggested, I added string key type constants to the KMS and re-used it for KeyType constants.

sandrask: sounds good

pkg/doc/signature/ed25519signature2018/public_key_verifier.go
```diff
@@ -0,0 +1,35 @@
...
+// NOTE: this verifier is present for backward compatibility reasons and can be removed soon.
+// Please use CryptoVerifier or your own verifier.
+type PublicKeyVerifier struct {
+}
+
+// Verify will verify a signature.
+func (v *PublicKeyVerifier) Verify(pubKey *sigverifier.PublicKey, doc, signature []byte) error {
+	// ed25519 panics if key size is wrong
+	if len(pubKey.Value) != ed25519.PublicKeySize {
+		return errors.New("ed25519: bad public key length")
```
Baha-sk: change "ed25519: bad public key length" to "ed25519: invalid key"

pkg/doc/signature/ed25519signature2018/public_key_verifier.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	verified := ed25519.Verify(pubKey.Value, doc, signature)
+	if !verified {
+		return errors.New("signature doesn't match")
```
Baha-sk: change "signature doesn't match" with "ed25519: invalid signature"

pkg/doc/signature/ed25519signature2018/public_key_verifier.go
kdimak: changed

pkg/doc/signature/ed25519signature2018/public_key_verifier.go
kdimak: changed

Issue #20, opened 2019-07-18
DID resolver - Validate input DID


fqutishat:
Validate input DID conforms to the did rule of the Generic DID Syntax.

//////////////////////////////////////////////////
did = "did:" method-name ":" method-specific-id
method-name = 1*method-char
method-char = %x61-7A / DIGIT
method-specific-id = *idchar *( ":" *idchar )
idchar = ALPHA / DIGIT / "." / "-" / ""
did-url = did ( ";" param ) path-abempty [ "?" query ]
[ "#" fragment ]
param = param-name [ "=" param-value ]
param-name = 1param-char
param-value = *param-char
param-char = ALPHA / DIGIT / "." / "-" / "" / ":" /
pct-encoded

////////////////////////////////////////////////////

There is implementation in java

https://github.com/decentralized-identity/did-common-java/tree/master/src/main/java/did/parser

llorllale:
This was done in #18

PR #1466, opened 2020-03-16
feat: client pkg for out-of-band protocol

### LINKED ISSUE #1456: out-of-band protocol: client package - support for 'request' message
part of #1454 

This is the first degenerate "protocol" we're implementing. It defines two messages.
An instance of this "protocol" has at most 1 message.
This "protocol" has no states.

We will add support for the `request` message in this issue.

llorllale:
closes #1456

* client package for the [Out-Of-Band protocol](https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md) 
* `did.Parse(string)` helper function (see #20)

TODO: add support for the creation of Out-Of-Band [`invitation`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md#message-type-httpsdidcommorgoob-invitationverinvitation)

### DIFF
--- a/pkg/client/outofband/client.go
+++ b/pkg/client/outofband/client.go
@@ -1,0 +1,145 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package outofband
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+const (
+	protocolURI    = "https://didcomm.org/oob-request/1.0"
+	requestMsgType = protocolURI + "/request"
+)
+
+// RequestOptions allow you to customize the way request messages are built.
+type RequestOptions func(*Request) error
+
+// ConnectionRecorder records connection records produced as byproducts of creation of messages.
+type ConnectionRecorder interface {
+	SaveInvitation(id string, i interface{}) error
+}
+
+// Provider provides the dependencies for the client.
+type Provider interface {
+	// DidDocServiceFunc returns a function that returns a DID doc `service` entry.
+	// Used when no service entries are specified when creating messages.
+	DidDocServiceFunc() func() (*did.Service, error)
+	ConnRecorder() ConnectionRecorder
+}
+
+// Client for the Out-Of-Band protocol:
+// https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md
+type Client struct {
+	didDocSvcFunc func() (*did.Service, error)
+	connRecorder  ConnectionRecorder
+}
+
+// New returns a new Client for the Out-Of-Band protocol.
+func New(p Provider) *Client {
+	return &Client{
+		didDocSvcFunc: p.DidDocServiceFunc(),
+		connRecorder:  p.ConnRecorder(),
+	}
+}
+
+// CreateRequest creates and saves an Out-Of-Band request message.
+// At least one attachment must be provided.
+// Service entries can be optionally provided. If none are provided then a new one will be automatically created for
+// you.
+func (c *Client) CreateRequest(opts ...RequestOptions) (*Request, error) {
+	req := &Request{}
+
+	for _, opt := range opts {
+		if err := opt(req); err != nil {
+			return nil, fmt.Errorf("failed to create request: %w", err)
+		}
+	}
+
+	if len(req.Requests) == 0 {
+		return nil, errors.New("must provide at least one attachment to create an out-of-band request")
+	}
+
+	if len(req.Service) == 0 {
+		svc, err := c.didDocSvcFunc()
+		if err != nil {
+			return nil, fmt.Errorf("failed to create a new inlined did doc service block : %w", err)
+		}
+
+		req.Service = []interface{}{svc}
+	}
+
+	req.ID = uuid.New().String()
+	req.Type = requestMsgType
+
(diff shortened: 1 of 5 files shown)


pkg/client/outofband/client.go
```diff
@@ -0,0 +1,145 @@
...
+// RequestOptions allow you to customize the way request messages are built.
+type RequestOptions func(*Request) error
+
+// ConnectionRecorder records connection records produced as byproducts of creation of messages.
+type ConnectionRecorder interface {
+	SaveInvitation(id string, i interface{}) error
+}
+
+// Provider provides the dependencies for the client.
+type Provider interface {
```
rolsonquadras: Does context fulfills this provider ? Usually [framework.Context()](https://github.com/hyperledger/aries-framework-go/blob/23fd9909a22991088bf646e981e479add5e1bc31/pkg/client/didexchange/client.go#L34) is used to create the client.

llorllale: Does it need to? Is there a reason why it can't be adapted in eg. `pkg/framework/aries/default.go`?

rolsonquadras: Yeah, usually we add the provider dependency in the context and the consumer can use the context object directly to initialize the client.

rolsonquadras: tagging - @troyronda @sudeshrshetty @fqutishat

pkg/client/outofband/client.go
```diff
@@ -0,0 +1,145 @@
...
+type Client struct {
+	didDocSvcFunc func() (*did.Service, error)
+	connRecorder  ConnectionRecorder
+}
+
+// New returns a new Client for the Out-Of-Band protocol.
+func New(p Provider) *Client {
+	return &Client{
+		didDocSvcFunc: p.DidDocServiceFunc(),
+		connRecorder:  p.ConnRecorder(),
```
sudeshrshetty: Why can't you use, `connection.NewRecorder(ctx)` here? And use provider matching aries context.

pkg/client/outofband/client.go
sudeshrshetty: What if someone wants to customize it?  only entry in defaults won't be enough.
You don't need to connection recorder in provider, instead you should use store providers here.
You can refer didexchange service to use `connection.NewRecorder(ctx)`

llorllale: I opened #1467

Can this PR be merged?

pkg/client/outofband/models.go
```diff
@@ -0,0 +1,20 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package outofband
+
+import "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+
+// Request embeds the sender's request.
+type Request struct {
```
sudeshrshetty: Aren's we going to use `"protocols"`?

pkg/client/outofband/client.go
sudeshrshetty: connectionRecorder never intended to be a interface

llorllale: It is an interface that this package declares as a dependency.

pkg/client/outofband/client.go
llorllale: the context currently doesn't provide a connection recorder. Maybe can be solved in #1467

pkg/client/outofband/models.go
sudeshrshetty: https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md#message-type-httpsdidcommorgoob-invitationverinvitation

llorllale: `invitation` will be added in a separate PR. This PR right now has 600 lines.

sudeshrshetty: looks good

pkg/client/outofband/client.go
llorllale: It also doesn't provide a function that creates a did doc `service` object (requires new keys and router info (if applicable)). This capability is also needed in the didexchange protocol service.

llorllale:
@fqutishat @rolsonquadras @sudeshrshetty @troyronda @soluchok  I'd like to move on with this ASAP.

Can this PR be merged and address followups in #1467 ?

pkg/client/outofband/client.go
troyronda: @llorllale do you have a BDD test of this package?

I'm not quite understanding the usage plan when the framework generated context cannot be used?

troyronda: Also - I would think the constructor would take care of details of consuming the framework context, rather than requiring an adapter utility function.

llorllale: @troyronda BDD tests TODO

> I'm not quite understanding the usage plan when the framework generated context cannot be used?

I plan on addressing the "context" problem in #1467. Does that answer your question?

pkg/doc/did/doc.go
```diff
@@ -328,6 +330,45 @@ const (
...
+}
+
+// String returns a string representation of this DID
+func (d *DID) String() string {
+	return fmt.Sprintf("%s:%s:%s", d.Scheme, d.Method, d.MethodSpecificID)
+}
+
+// Parse parses the string according to the generic DID syntax.
+// See https://w3c.github.io/did-core/#generic-did-syntax.
+func Parse(did string) (*DID, error) {
```
troyronda: Perhaps this should be something like `ParseDID` since it is in a package with multiple parsing-like functions?

llorllale: @troyronda there's one other parse function - `did.ParseDocument()`. 

To me, `did.Parse()` makes perfect sense as a function that parses a DID.

Let's see what the others think. @sudeshrshetty @rolsonquadras @soluchok @fqutishat

PR #1471, opened 2020-03-17
feat: support new hybriddh crypto primitive

### LINKED ISSUE #1469: Create new Key types to support ECDH-ES primitives (using NIST approved keys)
part of #987 

Baha-sk:
feat: support new hybriddh crypto primitive

inspired by Tink's Hybrid crypto, this change introduces hybriddh package 
for ECDH-ES crypto primitives needed for JWE Anoncrypt operations.

This change is a first crypto patch introducing new key types and new primitives.

A second change is coming up to update HKDF key derivation with ConcatKDF one that
follows JWA's key agreement using ECDH-ES: 
https://tools.ietf.org/html/rfc7518#section-4.6.

A third change will then follow to update the primitives logic to suit fields needed
in a JWE envelope (ciphertext, tag, iv, derivedKEK for recipients, etc).

closes #1469


pkg/crypto/tinkcrypto/primitive/hybriddh/api/ecdhes_decrypt.go
```diff
@@ -0,0 +1,62 @@
+/*
```
troyronda: Is this file an exact copy from another project?
If yes, why can't it be imported?
If it's been changed, there needs to be a comment summarizing the change (and additional copyright headers).

I also feel like we should separate copied files into an internal structure like we did in fabric-sdk-go.

pkg/crypto/tinkcrypto/primitive/hybriddh/api/ecdhes_decrypt.go
```diff
@@ -0,0 +1,62 @@
...
+Copyright 2017 Google Inc.
```
troyronda: There also seems to be a substantial number of these files in this PR.

Perhaps we should have an internal folder structure similar to what we have in fabric-sdk-go for copied files.
(does anything actually need to be exposed outside the repo - we might also need to think about third_party).

Baha-sk: I'll push a change that includes only the Hybrid crypto files from Tink unchanged, then I'll push new PR on top of it for ECDH-ES crypto

troyronda:
I assume a replacement PR is coming.

PR #1474, opened 2020-03-17
chore: out-of-band client: provide dependencies from framework Context

### LINKED ISSUE #1467: out-of-band client: provide dependencies from framework Context
Gathering requirements from comments in https://github.com/hyperledger/aries-framework-go/pull/1466#discussion_r393275487

The common theme seems to be that users should only have to call `outofband.New(ctx)` ("ctx" is the framework's context) and it should just *work*.

It would be nice to do this, but I do think it's contradictory to say that all packages are decoupled because they declare their own dependencies yet these dependencies have the *exact* same signature as the implementations.

I also disagree that the `outofband.Client` should care about how to properly build the customizable `did.Service` struct, or how connection records are saved.

The OOB client needs two capabilities injected:

1. capability that creates a `did.Service` block
   * new keys (KMS)
   * set routing info on request (if applicable)
   * registering the new keys with the router (if applicable)
   * sets correct `@type` ("did-communication")
   * notes:
     * `didexchange.Client` is already doing something *very* close to this, but `CreateInvitation()` will be removed at some point.
     * `getDIDDocAndConnection()` from `protocol/didexchange/states.go` is doing something very similar as well, but it's creating a full DID document. It relies on the VDRI implementation to handle all the details. This sets the precedent of having an injected dependency handle all the details of creating these structures
2. capability that saves a connection record from the request message created
   * this connection record is referenced later by other protocols
   * `outofband.Client` does not need to query connection records - see #943

llorllale:
closes #1467

### DIFF
--- a/pkg/client/outofband/client.go
+++ b/pkg/client/outofband/client.go
@@ -10,6 +10,11 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
+
 	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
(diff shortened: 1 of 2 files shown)


pkg/client/outofband/client.go
```diff
@@ -34,7 +34,7 @@ type Provider interface {
 	// DidDocServiceFunc returns a function that returns a DID doc `service` entry.
 	// Used when no service entries are specified when creating messages.
 	DidDocServiceFunc() func() (*did.Service, error)
-	ConnRecorder() ConnectionRecorder
+	ConnectionRecorder() (ConnectionRecorder, error)
```
sudeshrshetty: Should be storage provider here.

rolsonquadras: Here is a reference - https://github.com/hyperledger/aries-framework-go/blob/0fb64f2818eefefbddc4ce231642488c14362b63/pkg/client/didexchange/client.go#L91

llorllale: @rolsonquadras @sudeshrshetty @troyronda perhaps I did not explain myself correctly. I opened #1477 to explain my views.

Issue #1473, opened 2020-03-17
Issue Credential - add more scenarios for the BDD tests


soluchok:
Currently, BDD tests for the Issue Credential protocol cover only happy paths. We need to add more tests for negative paths.

PR #1475, opened 2020-03-17
chore: Issue Credential - adds BDD tests

soluchok:
* The Holder begins with a proposal (negotiation)
* The Holder begins with a request and the Issuer declines it
* The Holder begins with a proposal and the Issuer declines it
* The Issuer begins with an offer and the Holder declines it
* The Holder begins with a request and the Holder declines the credential

Closes https://github.com/hyperledger/aries-framework-go/issues/1473


test/bdd/pkg/issuecredential/issuecredential_sdk_steps.go
```diff
@@ -194,7 +245,7 @@ func (a *SDKSteps) createClient(agentID string) error {
 		return nil
 	}
 
-	const stateMsgChanSize = 10
+	const stateMsgChanSize = 12
```
troyronda: how is this being chosen?

PR #1479, opened 2020-03-17
refactor: JWE implementation in jose 

### LINKED ISSUE #1409: JWE implementation that is separate from DIDComm and crypto-implementation agnostic
Make it possible for another project to import a JWE package independent of DIDComm. The other project would then be able to decide what package it will use to build the envelope.

DRK3:
The existing JWE and legacy JWE will be removed in a future commit.

closes #1409

### DIFF
--- a/pkg/didcomm/packer/jwe/authcrypt/authcrypt.go
+++ b/pkg/didcomm/packer/jwe/authcrypt/authcrypt.go
@@ -14,6 +14,7 @@ import (
 	chacha "golang.org/x/crypto/chacha20poly1305"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/packer"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
 	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
 )
 
@@ -48,12 +49,12 @@ type Packer struct {
 
 // Envelope represents a JWE envelope as per the Aries Encryption envelope specs
 type Envelope struct {
-	Protected  string      `json:"protected,omitempty"`
-	Recipients []Recipient `json:"recipients,omitempty"`
-	AAD        string      `json:"aad,omitempty"`
-	IV         string      `json:"iv,omitempty"`
-	Tag        string      `json:"tag,omitempty"`
-	CipherText string      `json:"ciphertext,omitempty"`
+	Protected  string           `json:"protected,omitempty"`
+	Recipients []jose.Recipient `json:"recipients,omitempty"`
+	AAD        string           `json:"aad,omitempty"`
+	IV         string           `json:"iv,omitempty"`
+	Tag        string           `json:"tag,omitempty"`
+	CipherText string           `json:"ciphertext,omitempty"`
 }
 
 // jweHeaders are the Protected JWE headers in a map format
@@ -61,21 +62,6 @@ type jweHeaders struct {
 	Typ string `json:"typ,omitempty"`
 	Alg string `json:"alg,omitempty"`
 	Enc string `json:"enc,omitempty"`
-}
-
-// Recipient is a recipient of an envelope including the shared encryption key
-type Recipient struct {
-	EncryptedKey string           `json:"encrypted_key,omitempty"`
-	Header       RecipientHeaders `json:"header,omitempty"`
-}
-
-// RecipientHeaders are the recipient headers
-type RecipientHeaders struct {
-	APU string `json:"apu,omitempty"`
-	IV  string `json:"iv,omitempty"`
-	Tag string `json:"tag,omitempty"`
-	KID string `json:"kid,omitempty"`
-	SPK string `json:"spk,omitempty"`
 }
 
 // recipientSPKJWEHeaders are the Protected JWE headers of a recipient's SPK field (which is a JWE with a JWK payload)
--- a/pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
+++ b/pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/btcsuite/btcutil/base58"
 	"github.com/stretchr/testify/require"
 
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
 	"github.com/hyperledger/aries-framework-go/pkg/internal/cryptoutil"
 	mockkms "github.com/hyperledger/aries-framework-go/pkg/mock/kms/legacykms"
 )
@@ -569,9 +570,9 @@ func TestEncrypt(t *testing.T) {
 
 func deepCopy(envelope, envelope2 *Envelope) {
 	for _, r := range envelope2.Recipients {
-		newRe := Recipient{
+		newRe := jose.Recipient{
 			EncryptedKey: r.EncryptedKey,
-			Header: RecipientHeaders{
+			Header: jose.RecipientHeaders{
 				APU: r.Header.APU,
 				KID: r.Header.KID,
 				IV:  r.Header.IV,
--- a/pkg/doc/jose/common.go
+++ b/pkg/doc/jose/common.go
@@ -8,46 +8,61 @@ package jose
 
 // IANA registered JOSE headers (https://tools.ietf.org/html/rfc7515#section-4.1)
 const (
-	// HeaderAlgorithm identifies the cryptographic algorithm used to secure the JWS.
+	// HeaderAlgorithm identifies:
+	// For JWS: the cryptographic algorithm used to secure the JWS.
+	// For JWE: the cryptographic algorithm used to encrypt or determine the value of the CEK.
 	HeaderAlgorithm = "alg" // string
 
-	// HeaderJWKSetURL is a URI that refers to a resource for a set of JSON-encoded public keys, one of
-	// which corresponds to the key used to digitally sign the JWS.
+	// HeaderJWKSetURL is a URI that refers to a resource for a set of JSON-encoded public keys, one of which:
+	// For JWS: corresponds to the key used to digitally sign the JWS.
+	// For JWE: corresponds to the public key to which the JWE was encrypted.
 	HeaderJWKSetURL = "jku" // string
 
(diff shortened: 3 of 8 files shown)


pkg/doc/jose/jwe.go
```diff
@@ -0,0 +1,86 @@
...
+// LegacyRecipientHeader holds the header data for a recipient
+type LegacyRecipientHeader struct {
+	KID    string `json:"kid,omitempty"`
+	Sender string `json:"sender,omitempty"`
+	IV     string `json:"iv,omitempty"`
+}
+
+// ToDo/ Code review question... can this just be called a "standard JWE" struct?
+// ToDo (continued) Rather than an "Aries Encryption envelope"?
+// ToDo (continued) It looks like a standard JWE to me, unless I'm mistaken...
```
DRK3: @Baha-sk

pkg/doc/jose/jwk.go
```diff
@@ -0,0 +1,38 @@
...
+
+	err := convertMapToValue(jwkRaw, &jwk)
+	if err != nil {
+		return nil, false
+	}
+
+	return &jwk, true
+}
+
+// ToDo/Code Review question: can this somehow be merged with the jose.JSONWebKey type?
```
DRK3: @Baha-sk

pkg/doc/jose/jwe.go
Baha-sk: I wouldn't move old formats to keep all legacy structs and code in one place. It will make it easier to decommission later on.

pkg/doc/jose/jwk.go
Baha-sk: if this represents the old JWK used in legacy and current JWE implementation, then no need to move it here... we'll use the Jose.JSONWebKey for the future crypto implementation.

kdenhartog:
I like the direction this is going. Looks like it will make it simpler to transition to the new JWM based DIDComm work in DIF.

pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
```diff
@@ -15,6 +15,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
```
troyronda: the aries-framework-go imports should be in the third group. (also in other files in this PR.)

pkg/doc/jose/jwk.go
```diff
@@ -0,0 +1,36 @@
...
+	err := convertMapToValue(jwkRaw, &jwk)
+	if err != nil {
+		return nil, false
+	}
+
+	return &jwk, true
+}
+
+// AriesJWK formatted key
+type AriesJWK struct {
```
troyronda: I'm a bit surprised to see an `Aries` prefix in this package. What makes this struct Aries-specific?

pkg/doc/verifiable/credential.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
-	"github.com/hyperledger/aries-framework-go/pkg/doc/jwt"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose/jwt"
```
troyronda: I'm not sure why this folder was moved as part of the JWE PR?
@kdimak @DRK3 @Baha-sk

kdimak: The initial idea was to have `jose` package and put inside it the implementation of JWS and JWE. JWT is a bit different thing (it uses JWS and JWE to build claims) and it was placed in a separate package. Likewise, we are gonna put JWM implementation in a separate (non `jose`) package.

pkg/doc/jose/jwk.go
DRK3: That was just a temporary name I gave it in order to avoid a type name collision - it will be removed :)

pkg/doc/verifiable/credential.go
DRK3: I was thinking that JWT should go here since it's part of the JOSE framework of standards - my thinking is that if we have a JOSE folder, then all the JOSE stuff should go under there. Otherwise, if JWT is separate, then I think JWE and JWS should be separate too for consistency. What do you think? @kdimak @Baha-sk @troyronda

troyronda: @DRK3 Jose is both signing and encryption. Keep those together as `jose`.
jwt uses jose to produce JWS and JWE versions. 
Regardless moving unrelated folders shouldn't be in this PR.

Note, we will soon have `jwm` as a sibling to `jwt`. I'm not sure they are logically child folders of jose.

kdimak: I agree with @troyronda . Maybe if we want to be consistent regarding a separation, we could also have `jose`, `jws` and `jwe` packages as @DRK3 suggested above. What do you think @troyronda ?

DRK3: @kdimak What would go under the `jose` package?

Baha-sk: keep jws and jwe under jose package, jwt and jwm can have their own package

pkg/doc/jose/jwk.go
Baha-sk: yeah.. remove this struct.. it's not needed, we should only use JWK type at line 12 above

pkg/didcomm/packer/jwe/authcrypt/authcrypt.go
```diff
@@ -46,56 +46,6 @@ type Packer struct {
 	randReader io.Reader
 }
 
-// Envelope represents a JWE envelope as per the Aries Encryption envelope specs
```
Baha-sk: as discussed @DRK3 , let's not modify existing code in packer/jwe... 

simply use the new JSONWebEncryption you created in this patch for the new JWE documents we're going to use in EDV and the (new) didComm (which will be different than the existing code in packer/jwe/authcrypt)

pkg/doc/jose/common.go
```diff
@@ -57,3 +72,27 @@ const (
 	// Input as ASCII(BASE64URL(JWS Payload)) or as the JWS Payload value itself with no encoding performed.
 	HeaderB64Payload = "b64" // bool
 )
+
+// Headers represents JOSE headers.
```
kdimak: Good to see it moved to the common part! 👍

pkg/doc/jose/jwe.go
```diff
@@ -0,0 +1,131 @@
...
+	ProtectedHeaders   string          `json:"protected,omitempty"`
+	UnprotectedHeaders json.RawMessage `json:"unprotected,omitempty"`
+	Recipients         json.RawMessage `json:"recipients,omitempty"`
+	AAD                string          `json:"aad,omitempty"`
+	IV                 string          `json:"iv,omitempty"`
+	Ciphertext         string          `json:"ciphertext,omitempty"`
+	Tag                string          `json:"tag,omitempty"`
+}
+
+var errEmptyCiphertext = fmt.Errorf("ciphertext cannot be empty")
```
kdimak: We can use just `errors.New()` here.

pkg/doc/jose/jwe.go
```diff
@@ -0,0 +1,131 @@
...
+		Ciphertext:         b64Ciphertext,
+		Tag:                b64Tag,
+	}
+
+	serializedJWE, err := marshal(preparedJWE)
+	if err != nil {
+		return "", err
+	}
+
+	return string(serializedJWE), err
```
kdimak: We can `return string(serializedJWE), nil` for simplicity.

pkg/doc/jose/common.go
```diff
@@ -8,46 +8,61 @@ package jose
 
 // IANA registered JOSE headers (https://tools.ietf.org/html/rfc7515#section-4.1)
 const (
-	// HeaderAlgorithm identifies the cryptographic algorithm used to secure the JWS.
+	// HeaderAlgorithm identifies...
```
troyronda: nit: why `...` instead of `:`

pkg/doc/jose/jwe.go
troyronda: (since you are not formatting.)

pkg/doc/jose/jwe.go
DRK3: (done!... or "not done", technically)

pkg/doc/jose/jwk.go
DRK3: done

pkg/didcomm/packer/jwe/authcrypt/authcrypt_test.go
DRK3: They should all be good now

pkg/doc/jose/jwk.go
DRK3: Done

pkg/doc/verifiable/credential.go
DRK3: Done

pkg/didcomm/packer/jwe/authcrypt/authcrypt.go
DRK3: Done

pkg/doc/jose/jwe.go
DRK3: I debated between the two. errors.New adds an extra import, so that's why I went with fmt.errorf. Not sure if that's actually a real benefit, though. But I agree that errors.New looks clearer, especially since I'm not formatting.

I've changed it to errors.New :)

pkg/doc/jose/jwe.go
DRK3: Oops - that was unintentional... good catch!

Done!

pkg/doc/jose/common.go
DRK3: I actually went back and forth between `...` and `:`. It was a toss-up for me, but if the `...` looks odd to you then I'm sure it'll look odd to others as well.

I've gone ahead and changed the `...` to `:` :)

PR #525, opened 2019-10-18
bdd: Create Invitation with Public DID

sandrask:
BDD test for Create Invitation with public DID

Closes #519

Signed-off-by: Sandra Vrtikapa <sandra.vrtikapa@securekey.com>

PR #1480, opened 2020-03-17
feat: out-of-band protocol service skeleton

### LINKED ISSUE #1478: implement: out-of-band protocol service skeleton
part of #1454

llorllale:
closes #1478 

Skeleton for the Out-Of-Band protocol service.

* A single out-of-band `request` contains one or more (didcomm?) messages embedded in attachments
* the service needs to first run didexchange (which *may* reuse an existing connection)
* the service then needs to dispatch the embedded requests (assuming didcomm for now...) to their respective handlers

### DIFF
--- a/pkg/client/outofband/client.go
+++ b/pkg/client/outofband/client.go
@@ -10,20 +10,20 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
-	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
-	"github.com/hyperledger/aries-framework-go/pkg/storage"
-	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
-
 	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/outofband"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
 )
 
 const (
-	protocolURI    = "https://didcomm.org/oob-request/1.0"
-	requestMsgType = protocolURI + "/request"
+	// RequestMsgType is the request message's '@type'.
+	RequestMsgType = outofband.RequestMsgType
 )
 
 // RequestOptions allow you to customize the way request messages are built.
@@ -68,7 +68,7 @@ func New(p Provider) (*Client, error) {
 // Service entries can be optionally provided. If none are provided then a new one will be automatically created for
 // you.
 func (c *Client) CreateRequest(opts ...RequestOptions) (*Request, error) {
-	req := &Request{}
+	req := &Request{&outofband.Request{}}
 
 	for _, opt := range opts {
 		if err := opt(req); err != nil {
@@ -90,7 +90,7 @@ func (c *Client) CreateRequest(opts ...RequestOptions) (*Request, error) {
 	}
 
 	req.ID = uuid.New().String()
-	req.Type = requestMsgType
+	req.Type = RequestMsgType
 
 	err := c.connRecorder.SaveInvitation(req.ID, req)
 	if err != nil {
--- a/pkg/didcomm/protocol/outofband/service.go
+++ b/pkg/didcomm/protocol/outofband/service.go
@@ -1,0 +1,337 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package outofband
+
+import (
+	"encoding/json"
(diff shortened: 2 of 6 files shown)


pkg/client/outofband/client.go
```diff
@@ -67,8 +63,8 @@ func New(p Provider) (*Client, error) {
 // At least one attachment must be provided.
 // Service entries can be optionally provided. If none are provided then a new one will be automatically created for
 // you.
-func (c *Client) CreateRequest(opts ...RequestOptions) (*Request, error) {
-	req := &Request{}
+func (c *Client) CreateRequest(opts ...RequestOptions) (*outofband.Request, error) {
```
rolsonquadras: We shouldn't ideally expose service pkg to the clients. We can wrap these structs in client pkg (similar to [didex](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/client/didexchange/models.go#L51)).

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -0,0 +1,341 @@
...
+	bytes, err := s.store.Get(id)
+	if err != nil && !errors.Is(err, storage.ErrDataNotFound) {
+		return nil, fmt.Errorf("failed to fetch state data with id=%s : %w", id, err)
+	}
+
+	if errors.Is(err, storage.ErrDataNotFound) {
+		// this didexchange thread did not spawn from an out-of-band thread that we know of
+		return nil, fmt.Errorf("cannot find a state record for pthid=%s : %w", id, err)
+	}
+
```
soluchok: I don't see any reason to have different error messages. Why do we need this?

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -0,0 +1,341 @@
...
+func listener(
+	callbacks chan *callback,
+	didEvents chan service.StateMsg,
+	handleReqFunc func(*callback) error,
+	handleDidEventFunc func(msg service.StateMsg) error) func() {
+	return func() {
+		for {
+			select {
+			case c := <-callbacks:
+				// TODO add support for handling the 'invitation' message
```
rolsonquadras: issue number ?

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -0,0 +1,341 @@
...
+		Request:      req,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to save my state : %w", err)
+	}
+
+	return nil
+}
+
+func (s *Service) handleDidEvent(e service.StateMsg) error {
```
rolsonquadras: handleDidEvent should be handleDIDEvent.

pkg/didcomm/protocol/outofband/service.go
llorllale: I cannot give you a definitive answer. 

I CAN explain however that those two error-checks we initially inside `handleDidEvent()`, where their intention would be perhaps more clear. (I moved them in here to please the cyclo linter).

If you imagine they're still in `handleDidEvent()` then see that:

1. the store may throw a generic error (normal)
2. no state may be found for the given pthid

I believe all did-exchange threads should spawn from this new out-of-band protocol. Therefore the incoming didexchange msg SHOULD have a pthid known to this service. Hence why I differentiated the errors between (1) and (2).

Make sense?

pkg/didcomm/protocol/outofband/service.go

pkg/didcomm/protocol/outofband/service.go

(this should be a linter check!)

pkg/client/outofband/client.go
soluchok: @rolsonquadras  How does it help? It makes work even complicated.
For example, the usage below shows that you need to import two packages anyway.
```
didexchange1.Invitation{
    Invitation: &didexchange2.Invitation{ID:"1234"},
}
```

Sure, we can do something like that
```
inv := didexchange1.Invitation{}
inv.ID = "1234"
```

If we do not want to expose service pkg than we need to create package e.g `model` which can keep both client and service `types/models`. What we did in did-exchange can work only when we do read operation. But if we want to populate the structure we need to import the service package.

pkg/didcomm/protocol/outofband/service.go
rolsonquadras: https://github.com/hyperledger/aries-framework-go/issues/87

pkg/didcomm/protocol/outofband/service.go
soluchok: I would not do that. For me, it looks complicated.

pkg/client/outofband/client.go
llorllale: @rolsonquadras I don't understand the benefits this actually brings in practice. Can you elaborate?

pkg/didcomm/protocol/outofband/service.go
llorllale: Ok - removed (one less test case!)

pkg/client/outofband/client.go
rolsonquadras: Whole idea was to have the consumer use/look in client/framework pkg only (may be some util from framework). There was a discussion on this in one of the issues/PR. I'll try to find that.

rolsonquadras: Here it is - https://github.com/hyperledger/aries-framework-go/pull/525#discussion_r337071685.

llorllale: There isn't any dicussion there.

Nevertheless I refactored - please take a look.

We're wasting too much time on nitpicks.

Issue #1489, opened 2020-03-19
Update Tink signature key templates to `...WithoutPrefixTemplate()` when available 


Baha-sk:
related to https://github.com/hyperledger/aries-framework-go/issues/1483

signing keys with public key exported to raw []byte to be stored in DID documents or didComm must use RAW prefix type in Tink.

As a workaround to fix #1483, the key templates are manually modified in Aries.

This issue is created to remove manual changes in Aries and use Tink provided key templates (when they are available in Tink [#326](https://github.com/google/tink/pull/326/))

PR #1490, opened 2020-03-19
fix: loading raw pub key to Tink keyset.Handle fails verification

### LINKED ISSUE #1483: invalid public key by localkms.ExportPubKeyBytes()
Public key export doing by `localkms.ExportPubKeyBytes()` doesn’t seem to do a full extract of the public key.

How to check: extend `TestLocalKMS_Success`, test signing/verification using `tinkcrypto`:
```
if strings.Contains(string(v), "ECDSA") || v == kms.Ed25519Type {
	pubKeyBytes, e := kmsService.ExportPubKeyBytes(keyID)
	require.Errorf(t, e, "KeyID has been rotated. An error must be returned")
	require.Empty(t, pubKeyBytes)

	pubKeyBytes, e = kmsService.ExportPubKeyBytes(newKeyID)
	require.NoError(t, e)
	require.NotEmpty(t, pubKeyBytes)

	kh, e := kmsService.PubKeyBytesToHandle(pubKeyBytes, v)
	require.NoError(t, e)
	require.NotEmpty(t, kh)

	// NEW: verify the restored public key
	c, err := tinkcrypto.New()
	require.NoError(t, err)

	msg := []byte("test message")

	msgSig, err := c.Sign(msg, newKeyHandle)
	require.NoError(t, err)

	err = c.Verify(msgSig, msg, kh)
	require.NoError(t, err)
}

```

Baha-sk:
Since default signing key templates have Tink specific prefix,
need to use RAW type to avoid adding a prefix. This will fix
verification with a keyset.Handle created from raw public key []byte

To unblock dev, the temporary fix is to manually change the key templates
in aries-framework-go code. A followup issue is created to use Tink
created key templates without a prefix when the templates are
available in Tink (#1489).

closes #1483

### DIFF
--- a/pkg/kms/localkms/pubkey_export_import_test.go
+++ b/pkg/kms/localkms/pubkey_export_import_test.go
@@ -21,6 +21,16 @@ import (
 )
 
 func TestPubKeyExportAndRead(t *testing.T) {
+	// TODO remove below variables when corresponding key templates `WithoutPrefix` are available in Tink - issue #1489
+	ecdsaP256KeyTemplate := signature.ECDSAP256KeyTemplate()
+	ecdsaP256KeyTemplate.OutputPrefixType = tinkpb.OutputPrefixType_RAW
+	ecdsaP384KeyTemplate := signature.ECDSAP384KeyTemplate()
+	ecdsaP384KeyTemplate.OutputPrefixType = tinkpb.OutputPrefixType_RAW
+	ecdsaP521KeyTemplate := signature.ECDSAP521KeyTemplate()
+	ecdsaP521KeyTemplate.OutputPrefixType = tinkpb.OutputPrefixType_RAW
+	ed25519KeyTemplate := signature.ED25519KeyTemplate()
+	ed25519KeyTemplate.OutputPrefixType = tinkpb.OutputPrefixType_RAW
+
 	var flagTests = []struct {
 		tcName      string
 		keyType     kms.KeyType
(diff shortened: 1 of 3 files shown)


kdimak:
@Baha-sk  can we add a test case I mentioned at https://github.com/hyperledger/aries-framework-go/issues/1483#issue-583803837 to double-check your temporary fix?

Baha-sk:
> @Baha-sk can we add a test case I mentioned at [#1483 (comment)](https://github.com/hyperledger/aries-framework-go/issues/1483#issue-583803837) to double-check your temporary fix?

@kdimak the test already exists: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/kms/localkms/pubkey_export_import_test.go#L76

kdimak:
> > @Baha-sk can we add a test case I mentioned at [#1483 (comment)](https://github.com/hyperledger/aries-framework-go/issues/1483#issue-583803837) to double-check your temporary fix?
> 
> @kdimak the test already exists: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/kms/localkms/pubkey_export_import_test.go#L76

OK!

pkg/kms/localkms/pubkey_export_import_test.go
```diff
@@ -78,7 +96,7 @@ func TestPubKeyExportAndRead(t *testing.T) {
 				require.NotEmpty(t, verifier)
 
 				err = verifier.Verify(s, msg)
-				require.NotEmpty(t, err)
+				require.NoError(t, err)
```
Baha-sk: this fixes the current verification test... it was wrongly checking if err is not empty, instead of checking if there's no error

PR #1492, opened 2020-03-19
feat: hook out-of-band svc into didexchange svc

### LINKED ISSUE #1491: outofband protocol: hook into did-exchange (temporary)
part of #1454 

Hook the OOB protocol service into the didexchange service by having the former pass in a `didexchange.Invitation` object to the latter.

This is a temporary solution to minimize refactors. In the longer term, the didexchange protocol service should be refactored to accept an invitation object with a different shape than `didexchange.Invitation`'s.

llorllale:
closes #1491 

### DIFF
--- a/pkg/didcomm/protocol/outofband/service.go
+++ b/pkg/didcomm/protocol/outofband/service.go
@@ -11,6 +11,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/google/uuid"
+
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
 	"github.com/hyperledger/aries-framework-go/pkg/internal/logutil"
@@ -22,8 +24,6 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
 )
 
-var logger = log.New("aries-framework/did-exchange/service")
-
 const (
 	// Name of this protocol service.
 	Name = "out-of-band"
@@ -33,6 +33,8 @@ const (
 	// TODO channel size - https://github.com/hyperledger/aries-framework-go/issues/246
 	callbackChannelSize = 10
 )
+
+var logger = log.New(fmt.Sprintf("aries-framework/%s/service", Name))
 
 var errIgnoredDidEvent = errors.New("ignored")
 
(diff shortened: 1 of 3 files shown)


pkg/didcomm/protocol/outofband/service_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
```
troyronda: should be in the third import group.

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -209,23 +211,23 @@ func listener(
 }
 
 func (s *Service) handleRequestCallback(c *callback) error {
-	// TODO either transform to a didexchange.invitation object or refactor didexchange.Service to accept this object
-	connID, err := s.didSvc.HandleInbound(c.msg, c.myDID, c.theirDID)
+	// TODO refactor didexchange.Service to accept an object other than didexchange.Invitation
```
troyronda: issue number?

llorllale: #1501

pkg/didcomm/protocol/outofband/service_test.go
llorllale: done

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -335,3 +338,33 @@ func extractDIDCommMsgBytes(_ *decorator.Attachment) ([]byte, error) {
...
+			invitation.DID = did
+			break
+		}
+	}
+
+	if invitation.DID == "" {
+		return nil, nil, errors.New("ouf-of-band protocol only supports invitations with public DIDs for now")
+	}
+
+	// TODO support explicit invitations
```
troyronda: issue number?

pkg/didcomm/protocol/outofband/service_test.go
```diff
@@ -585,3 +664,39 @@ func newAck(pthid ...string) *model.Ack {
...
 	return a
 }
+
+type testDIDCommMsg struct {
+	errDecode error
+}
+
+func (t *testDIDCommMsg) ID() string {
+	panic("implement me")
+}
```
soluchok: I would recommend you to replace it with `gomock`

llorllale: noted - maybe in the future

PR #1494, opened 2020-03-20
feat: Support for RemoveConnection in DIDExchange clients

### LINKED ISSUE #553: Support for RemoveConnection in DIDExchange clients
Based on [TODO](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/client/didexchange/client.go#L185).

rossus:
**Title:**
Implementation of removing the connection records from the stores

**Description:**
Closes #553 
Removes connection and it's mapping records across store and transient store.

Remove connection sequence in _pkg/store/connection/connection_recorder.go_ mirrors the save connection one. It removes all records for any state of the connection from the transient store, as well as did mappings records and namespace, threadID and connectionID mappings if there are some.

### DIFF
--- a/pkg/client/didexchange/client.go
+++ b/pkg/client/didexchange/client.go
@@ -272,7 +272,11 @@ func (c *Client) GetConnectionAtState(connectionID, stateID string) (*Connection
 }
 
 // RemoveConnection removes connection record for given id
-func (c *Client) RemoveConnection(id string) error {
-	// TODO https://github.com/hyperledger/aries-framework-go/issues/553 RemoveConnection from did exchange service
+func (c *Client) RemoveConnection(connectionID string) error {
+	err := c.connectionStore.RemoveConnection(connectionID)
+	if err != nil {
+		return fmt.Errorf("cannot remove connection from the store: err=%w", err)
+	}
+
 	return nil
 }
--- a/pkg/store/connection/connection_recorder.go
+++ b/pkg/store/connection/connection_recorder.go
@@ -131,6 +131,44 @@ func (c *Recorder) SaveNamespaceThreadID(threadID, namespace, connectionID strin
 	return c.transientStore.Put(getNamespaceKeyPrefix(prefix)(key), []byte(connectionID))
 }
 
+// RemoveConnection removes connection record from the store for given id
+func (c *Recorder) RemoveConnection(connectionID string) error {
+	record, err := c.GetConnectionRecord(connectionID)
+	if err != nil {
+		return fmt.Errorf("unable to get connection record: connectionid=%s err=%w", connectionID, err)
+	}
+
+	if err = c.transientStore.Delete(getConnectionKeyPrefix()(connectionID)); err != nil {
+		return fmt.Errorf("unable to delete connection record from the transient store: connectionid=%s err=%w",
+			connectionID, err)
+	}
+
+	// remove connection records for different states from transient store
+	err = removeConnectionsForStates(c, connectionID)
+	if err != nil {
+		return fmt.Errorf("remove records for different connections states error: %w", err)
+	}
+
+	err = c.store.Delete(getConnectionKeyPrefix()(connectionID))
+	if err != nil {
+		return fmt.Errorf("unable to delete connection record from the store: connectionid=%s err=%w", connectionID, err)
+	}
+
+	err = c.store.Delete(getDIDConnMapKeyPrefix()(record.MyDID, record.TheirDID))
+	if err != nil {
+		return fmt.Errorf("unable to delete did mapping connection record from the store: connectionid=%s err=%w",
+			connectionID, err)
+	}
+
+	// remove namespace, threadID and connection ID mapping from transient store
+	err = removeMappings(c, record)
+	if err != nil {
+		return fmt.Errorf("unable to delete connection record with namespace mappings: %w", err)
+	}
+
+	return nil
+}
+
 func marshalAndSave(k string, v interface{}, store storage.Store) error {
 	bytes, err := json.Marshal(v)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


rolsonquadras:
Hi @rossus - Thanks for the contribution to aries-framework-go. The code coverage check is failing as the unit-test coverage for the patch is less than 85% ([55% currently](https://codecov.io/gh/hyperledger/aries-framework-go/compare/8560666461f495a4eb0724d5741b56b035f70d4a...75ae126f3a5e6cd547f3347a951753a4987bd56c/diff)). Would you please add more unit tests and improve the coverage ?

pkg/client/didexchange/client.go
```diff
@@ -272,7 +272,15 @@ func (c *Client) GetConnectionAtState(connectionID, stateID string) (*Connection
...
+func (c *Client) RemoveConnection(connectionID string) error {
+	err := c.connectionStore.RemoveConnection(connectionID)
+	if errors.Is(err, storage.ErrDataNotFound) {
+		return err
+	}
+
+	if err != nil {
+		return fmt.Errorf("cannot remove connection from the store: connectionid=%s err=%s", connectionID, err)
+	}
+
```
soluchok: No need to return two different error messages. 
Use `%w` to wrap an error. It allows you to get an initial error by calling errors.Unwrap().

pkg/store/connection/connection_recorder.go
```diff
@@ -131,6 +131,44 @@ func (c *Recorder) SaveNamespaceThreadID(threadID, namespace, connectionID strin
 	return c.transientStore.Put(getNamespaceKeyPrefix(prefix)(key), []byte(connectionID))
 }
 
+// RemoveConnection removes connection record from the store for given id
+func (c *Recorder) RemoveConnection(connectionID string) error {
+	record, err := c.GetConnectionRecord(connectionID)
+	if err != nil {
+		return err
```
soluchok: Better to wrap an error with some helpful message here.

pkg/store/connection/connection_recorder.go
```diff
@@ -131,6 +131,44 @@ func (c *Recorder) SaveNamespaceThreadID(threadID, namespace, connectionID strin
...
+
+	if err = c.transientStore.Delete(getConnectionKeyPrefix()(connectionID)); err != nil {
+		return fmt.Errorf("unable to delete connection record from the transient store: connectionid=%s err=%w",
+			connectionID, err)
+	}
+
+	// remove connection records for different states from transient store
+	err = removeConnectionsForStates(c, connectionID)
+	if err != nil {
+		return err
```
soluchok: same

pkg/store/connection/connection_recorder.go
```diff
@@ -131,6 +131,44 @@ func (c *Recorder) SaveNamespaceThreadID(threadID, namespace, connectionID strin
...
+
+	// remove namespace, threadID and connection ID mapping from transient store
+	err = removeMappings(c, record)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
soluchok: can be simplified 

```
return removeMappings(c, record)
```

pkg/store/connection/connection_recorder.go
```diff
@@ -161,3 +199,38 @@ func computeHash(bytes []byte) (string, error) {
...
+		getConnectionStateKeyPrefix()(connectionID)+"~",
+	)
+	defer itr.Release()
+
+	for itr.Next() {
+		key := string(itr.Key())
+		err := c.transientStore.Delete(key)
+
+		if err != nil {
+			return fmt.Errorf(
```
soluchok: Should not we check an error here? e.g
```
for itr.Next() {
 	if itr.Error() != nil {
 	 	return nil, itr.Error()
 	}
```

pkg/store/connection/connection_recorder.go
```diff
@@ -161,3 +199,38 @@ func computeHash(bytes []byte) (string, error) {
...
+		return err
+	}
+
+	err = c.store.Delete(getNamespaceKeyPrefix(record.Namespace)(key))
+	if err != nil {
+		return fmt.Errorf("unable to delete connection record with namespace mappings: %w", err)
+	}
+
+	return nil
+}
```
soluchok: It makes sense to change the order of wrapping errors. Looks much better.

```
func removeMappings(c *Recorder, record *Record) error {
	key, err := computeHash([]byte(record.ThreadID))
	if err != nil {
		return fmt.Errorf("compute hash: %w", err)
	}

	return c.store.Delete(getNamespaceKeyPrefix(record.Namespace)(key))
}
```

pkg/store/connection/connection_recorder.go
rossus: Probably it is better to wrap this error if it will be already simplified error return inside _removeMappings_ function.

soluchok: I am ok with a `wrap` or `simplification`

pkg/client/didexchange/client.go
rossus: Got it and fixed now.

pkg/store/connection/connection_recorder.go
rossus: Did it.

rossus:
Hi @rolsonquadras, I fixed it up to 93% now.

pkg/store/connection/connection_recorder.go
rossus: True, but it should be placed before itr.Next() sequence, and so I did.

soluchok: `Release()` should be placed after the loop. An error might occur during the iteration.
e.g 
```
for itr.Next() {
	...
}

return iter.Error()
```

rossus: Yes, you are right, I will place `itr.Error()` check after the loop. But `Release()` already has `defer`, so is it needed to be put after the loop?

soluchok: I am not sure what kind of error it might return. But the right place for it will be after the loop.

rossus: But it is one `return` inside the loop, so if `Release()` will be put after the loop we can it will not be called in case of we get error in the loop. But we need to call it after we done with iterator. That's how it is now:
```
func removeConnectionsForStates(c *Recorder, connectionID string) error {
	itr := c.transientStore.Iterator(getConnectionStateKeyPrefix()(
		connectionID),
		getConnectionStateKeyPrefix()(connectionID)+"~",
	)
	defer itr.Release() // Release() is placed before any return so in case of any outcome it will be called

	for itr.Next() {
		key := string(itr.Key())
		err := c.transientStore.Delete(key)

		if err != nil {
			return fmt.Errorf( // this return can be missed if Release() would be placed after the loop
				"unable to delete connection state record from the transient store: key=%s connectionid=%s err=%w",
				key, connectionID, err)
		}
	}

	return itr.Error()
}
```

rossus: As described here https://gobyexample.com/defer, deferred method will be executed at the end of enclosing function. So it is much safer to put it at the beginning of the function before any of the possible ends of this function (`return`s), than to put it after such point located inside the loop. In first case `itr.Release()` will be executed in any possibile end of the function, otherwise it will be executed only if we pass the loop.

soluchok: sorry for that typo)  `Release()` ->`Error()`
I meant `Error()` should be placed after the loop. An error might occur during the iteration.`

troyronda:
@rossus could you please squash your commits? thanks!

pkg/store/connection/connection_recorder.go
rossus: Ahh, got it :D

rossus:
@troyronda sure, just done it!

PR #1495, opened 2020-03-20
feat: Present Proof - implements a skeleton for the service

### LINKED ISSUE #1485: Present Proof - need to implement a skeleton for the service
We need to implement a service logic similar to other protocols (e.g introduce,issuecredential).

soluchok:
Implements common methods for the service such as:
* HandleInbound
* HandleOutbound
* Accept

Closes #1485 


pkg/didcomm/protocol/presentproof/service.go
```diff
@@ -24,3 +35,372 @@ const (
...
+		}
+
+		current = next
+	}
+
+	if err := s.saveStateName(md.PIID, stateName); err != nil {
+		return fmt.Errorf("failed to persist state %s: %w", stateName, err)
+	}
+
+	for _, action := range actions {
```
llorllale: q: why are we transitioning states before completing their actions?

soluchok: This is a known issue. It was done to avoid an issue related to the network.
e.g we have sent a request and response was received before the state was saved. 
NOTE: in didexchange,introduce etc. we have the same.

llorllale: I agree on the known issue where we're splitting a "transaction" in two (without rollbacks when the latter half fails), but this PR is doing something else in addition to this: it is transitioning through all states in a loop and collecting their actions for execution in a separate loop. The didexchange service isn't doing that - it's executing each state's action in the same loop.

soluchok: It collects it to make code more generic.
Let's imagine you have executed a network call (action) in state A and your loop goes to the next state and executes another network call (action) in state B. And before executing action (B) we received a response related to state A. This behavior causes a mess.

Rollback will work. I am going to add it.

llorllale: I don't like this because it is separating further the two halves of these transactions.

And I'd to see how this situation presents itself in `present-proof` protocol:

> Let's imagine you have executed a network call (action) in state A and your loop goes to the next state and executes another network call (action) in state B. And before executing action (B) we received a response related to state A.

Because as far as I am aware this protocol is a sequenced exchange of messages between two parties. This PR might be over-engineering things.

soluchok: I see that each protocol has some similarities. I would like to make this `service` logic common (somewhere in the future). For that, I am trying to reuse code from my previous PRs.

Added rollback logic. Actually, it is pretty simple. If it was the `start state` - delete the record, otherwise, we save to DB the old name.

llorllale: My concerns remained unanswered. I ask because I am implementing a protocol service that so far looks quite different from this one. And before I implement any of this I'd like to know why.

soluchok: @llorllale I agree with you that this logic is not necessarily needed for the `present-proof` protocol. I can do it similar to the did-exchange (execute the action in the loop). Just let me know.
But I did it more generic and it solves the problem when we transit between states in the loop.
e.g we received a `message1` this message executes state `A` and state A returns the next state `B` and `B` returns the next state `C`. All these states are executed consequently. Our chain looks like this: `message1 -> A -> B -> C`. 
The problem here is that the state machine was not finished yet when we already received a response related to action A. We received a response in the middle of the state machine execution. Now we do not know what we should execute next. Start with state B (according to the last record we saved)?  But `B` is probably executed by the previous state machine. Execute it twice? etc.
According to that possible issue, it makes sense to collect all actions together and then execute them. This is why I did this.

llorllale: @soluchok so in essence this PR is furthering this "split transaction" pattern which is already troublesome. My recommendation? Stick to the pattern in didexchange for now. "Split transaction" is already a topic in of itself that should be revisited at some point.

soluchok: ok, I'll do that!

soluchok: @llorllale  Done!

PR #1497, opened 2020-03-20
feat: HMAC-SHA256 support

### LINKED ISSUE #1468: Add HMAC-SHA256 support to the crypto/KMS library

DRK3:

closes #1468

### DIFF
--- a/pkg/crypto/tinkcrypto/crypto.go
+++ b/pkg/crypto/tinkcrypto/crypto.go
@@ -17,10 +17,13 @@ import (
 	"github.com/google/tink/go/aead"
 	"github.com/google/tink/go/core/primitiveset"
 	"github.com/google/tink/go/keyset"
+	"github.com/google/tink/go/mac"
 	"github.com/google/tink/go/signature"
 	aeadsubtle "github.com/google/tink/go/subtle/aead"
 	"golang.org/x/crypto/chacha20poly1305"
 )
+
+var errBadKeyHandleFormat = errors.New("bad key handle format")
 
 // Package provider/tinkcrypto includes implementation of spi/crypto. SPI implementation will be built
 // as a Framework option and fed into pkg/common/crypto implementation that includes a combined crypto
@@ -39,7 +42,7 @@ func New() (*Crypto, error) {
 func (t *Crypto) Encrypt(msg, aad []byte, kh interface{}) ([]byte, []byte, error) {
 	keyHandle, ok := kh.(*keyset.Handle)
 	if !ok {
-		return nil, nil, errors.New("bad key handle format")
+		return nil, nil, errBadKeyHandleFormat
 	}
 
 	a, err := aead.New(keyHandle)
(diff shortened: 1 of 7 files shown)


pkg/crypto/tinkcrypto/crypto.go
```diff
@@ -151,3 +154,35 @@ func (t *Crypto) Verify(sig, msg []byte, kh interface{}) error {
...
+func (t *Crypto) ComputeMAC(data []byte, kh interface{}) ([]byte, error) {
+	keyHandle, ok := kh.(*keyset.Handle)
+	if !ok {
+		return nil, errBadKeyHandleFormat
+	}
+
+	macPrimitive, err := mac.New(keyHandle)
+	if err != nil {
+		return nil, err
+	}
```
DRK3: I don't know to create the error condition here. I attempted passing an inappropriate key handle (for example, `keyset.NewHandle(aead.AES256GCMKeyTemplate()`), but unfortunately that causes the Tink code to panic instead of returning an err.

The only other thing I can think of is to mac that mac.New function a parameter so that I can pass in a mock for the unit test, but then I'll have to add that to the interface, but I think that would be a bit ugly and perhaps confusing...

pkg/crypto/tinkcrypto/crypto.go
```diff
@@ -151,3 +154,35 @@ func (t *Crypto) Verify(sig, msg []byte, kh interface{}) error {
...
+func (t *Crypto) VerifyMAC(macBytes, data []byte, kh interface{}) error {
+	keyHandle, ok := kh.(*keyset.Handle)
+	if !ok {
+		return errBadKeyHandleFormat
+	}
+
+	macPrimitive, err := mac.New(keyHandle)
+	if err != nil {
+		return err
+	}
```
DRK3: I don't know to create the error condition here. I attempted passing an inappropriate key handle (for example, `keyset.NewHandle(aead.AES256GCMKeyTemplate()`), but unfortunately that causes the Tink code to panic instead of returning an err.

The only other thing I can think of is to mac that mac.New function a parameter (either for the function or in the Crypto struct) so that I can pass in a mock for the unit test, but I think that would be a bit ugly and perhaps confusing... let me know what you think.

DRK3:
@troyronda @Baha-sk Note that the code coverage is a bit low. Please see the comments I made on the files explaining which parts are holding it back. I feel like the only solution I could come up with right now is worse than just accepting the slightly low code coverage. Please let me know what you think :)

pkg/crypto/tinkcrypto/crypto.go
Baha-sk: yes, I noticed that Tink doesn't validate key handles when doing for instance mac.New() or signature.New() and then panics when calling CompuateMac() or Sign() because it tries to convert the keyset to the primitive specific type. I will file an issue in Tink to fix this.

DRK3: Nice, thanks Baha :)

PR #1506, opened 2020-03-24
feat: Present Proof - state implementation for request sent/received

### LINKED ISSUE #1505: Present Proof - need to implement request-sent and request-received states
Need to implement the following states according to [RFC](https://github.com/hyperledger/aries-rfcs/tree/master/features/0037-present-proof):
* request-sent
* request-received

soluchok:
The following states were implemented:
* abandoning (inbound)
* request-received  (inbound)
* request-sent (outbound)
* presentation-sent  (inbound)
* proposal-sent (inbound)

Closes #1505 

### DIFF
--- a/pkg/didcomm/protocol/presentproof/states.go
+++ b/pkg/didcomm/protocol/presentproof/states.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/presentproof/states.go
```diff
@@ -31,7 +32,10 @@ const (
 	stateNameProposalSent     = "proposal-sent"
 )
 
-const codeInternalError = "internal"
+const (
```
Baha-sk: any reason for declaring const twice?

soluchok: what do you mean?

DRK3: I think he's suggesting that you can combine the two const (...) blocks into one, unless there's a specific reason you separated the two

soluchok: Oh, I got it. Actually, there is no reason. Code just looks more readable.

PR #1510, opened 2020-03-25
feat: DID schema to support publicKeyJwk & secp256k1

### LINKED ISSUE #1509: DID Schema: Support publicKeyJwk
- support public key encoding type publicKeyJwk (also support SECP256K1)
- As per spec only first context in array has to be ‘https://www.w3.org/ns/did/v1’ (Refer : https://www.w3.org/TR/did-core/#contexts)
- “- publicKey.0: Must have at most 4 properties” : As per spec ‘The public key objects MAY include additional properties.’ (Refer: https://www.w3.org/TR/did-core/#public-keys)

sudeshrshetty:
- PublicKeys to support `publicKeyJwk`
- Support for JWK public key decode including `secp256k1`
- Context URI updated to validate `https://www.w3.org/ns/did/v1` as well
as the older URI `https://w3id.org/did/v1`
- Removed public key max number of  properties validation, as per spec
'The public key objects MAY include additional properties'
- closes #1509


go.sum
```diff
@@ -5,22 +5,30 @@ cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSR
...
 github.com/VictoriaMetrics/fastcache v1.5.7 h1:4y6y0G8PRzszQUYIQHHssv/jgPHAb5qQuuDNdCbyAgw=
 github.com/VictoriaMetrics/fastcache v1.5.7/go.mod h1:ptDBkNMQI4RtmVo8VS/XwRY6RoTu1dAWCbrk+6WsEM8=
+github.com/aead/siphash v1.0.1 h1:FwHfE/T45KPKYuuSAKyyvE+oPWcaQ+CUmFW0bPlM+kg=
 github.com/aead/siphash v1.0.1/go.mod h1:Nywa3cDsYNNK3gaciGTWPwHt0wlpNV15vwmswBAUSII=
 github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412 h1:w1UutsfOrms1J05zt7ISrnJIXKzwaspym5BTKGx93EI=
 github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412/go.mod h1:WPjqKcmVOxf0XSf3YxCJs6N6AOSrOx3obionmG7T0y0=
 github.com/allegro/bigcache v1.2.1-0.20190218064605-e24eb225f156 h1:eMwmnE/GDgah4HI848JfFxHt+iPb26b4zyfspmqY0/8=
 github.com/allegro/bigcache v1.2.1-0.20190218064605-e24eb225f156/go.mod h1:Cb/ax3seSYIx7SuZdm2G2xzfwmv3TPSk2ucNfQESPXM=
 github.com/aws/aws-sdk-go v1.25.39/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
+github.com/btcsuite/btcd v0.20.1-beta h1:Ik4hyJqN8Jfyv3S4AGBOmyouMsYE3EdYODkMbQjwPGw=
```
troyronda: Where is this import coming from?

troyronda: it's from btcutil. that will be another thing to look into.

pkg/doc/jose/jwk.go
```diff
@@ -0,0 +1,163 @@
...
+// DecodePublicKey reads a public key from its JSON Web Key representation.
+func DecodePublicKey(jwkBytes []byte) ([]byte, error) {
+	key := &jsonWebKey{}
+
+	err := json.Unmarshal(jwkBytes, key)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read JWK, %w", err)
+	}
+
+	if strings.EqualFold(key.Kty, secp256k1Kty) && strings.EqualFold(key.Crv, secp256k1Crv) {
```
troyronda: some comments on the various conditions in this function would be helpful (linked to the DID document data model).

troyronda: Are there other cases we aren't covering yet?

Baha-sk: secp256k1 is not the only EC key type that should be supported by this JWK implementation. We need to support NIST curves as well (secp256r1 or P-2516, secp384r1 or P-384 and secp521r1 or P-521), these are EC curves as well.

pkg/doc/jose/jwk.go
```diff
@@ -0,0 +1,163 @@
...
+			return nil, fmt.Errorf("failed to read public key: %w", err)
+		}
+
+		return pubKBytes, nil
+	default:
+		return nil, fmt.Errorf("unsupported public key type in kid '%s'", key.Kid)
+	}
+}
+
+func ecPublicKey(xBuffer, yBuffer *byteBuffer) ([]byte, error) {
```
Baha-sk: rename this function to `btcPublickKey`

pkg/doc/jose/jwk_test.go
```diff
@@ -0,0 +1,238 @@
...
+			{
+				name: "get public key bytes EC SECP256K1 JWK",
+				jwkJSON: `{
+							"kty": "EC",
+    						"use": "enc",
+    						"crv": "secp256k1",
+    						"kid": "sample@sample.id",
+    						"x": "wQehEGTVCu32yp8IwTaBCqPUIYslyd-WoFRsfDKE9II",
+    						"y": "rIJO8RmkExUecJ5i15L9OC7rl7pwmYFR8QQgdM1ERWI",
+   						 	"alg": "EdDSA"
```
Baha-sk: This doesn't look right: `"alg": "EdDSA"`
with
`"crv": "secp256k1",`

secp256k1 is an ECDSA curve to my knowledge. We only support Ed25519 as EdDSA curve.

pkg/doc/jose/jwk_test.go
```diff
@@ -0,0 +1,238 @@
...
+			{
+				name: "attempt public key bytes from invalid JSON bytes",
+				jwkJSON: `{
+							"kty": "EC",
+    						"use": "enc",
+    						"crv": "secp256k1",
+    						"kid": "sample@sample.id",
+    						"x": "wQehEGTVCu32yp8IwTaBCqPUIYslyd-WoFRsfDKE9II",
+    						"y": "",
+   						 	"alg": "EdDSA"
```
Baha-sk: same here

pkg/doc/jose/jwk_test.go
```diff
@@ -0,0 +1,238 @@
...
+			{
+				name: "attempt public key bytes from invalid JSON bytes",
+				jwkJSON: `{
+							"kty": "EC",
+    						"use": "enc",
+    						"crv": "secp256k1",
+    						"kid": "sample@sample.id",
+    						"x": "",
+    						"y": "",
+   						 	"alg": "EdDSA"
```
Baha-sk: same here

pkg/doc/jose/jwk_test.go
```diff
@@ -0,0 +1,238 @@
...
+			{
+				name: "attempt public key bytes from invalid JSON bytes",
+				jwkJSON: `{
+							"kty": "EC",
+    						"use": "enc",
+    						"crv": "secp256k1",
+    						"kid": "sample@sample.id",
+    						"x": "x",
+    						"y": "y",
+   						 	"alg": "EdDSA"
```
Baha-sk: same here

pkg/doc/jose/jwk_test.go
```diff
@@ -0,0 +1,238 @@
...
+			{
+				name: "attempt public key bytes from invalid JSON bytes",
+				jwkJSON: `{
+							"kty": "EC",
+    						"use": "enc",
+    						"crv": "secp256k1",
+    						"kid": "sample@sample.id",
+    						"x": "{",
+    						"y": "y",
+   						 	"alg": "EdDSA"
```
Baha-sk: same here.. I think we need a mapping validation between `alg` and `crv` as they're related

pkg/doc/jose/jwk.go
kdimak: To make it more generic, I would override `UnmarshalJSON()` of `JWK` to support `secp256k1` elliptic curve (`MarshallJSON()` for symmetry). And I would create e.g. `Bytes()` method returning `[]byte` for `JWK` as well where `x509.MarshalPKIXPublicKey()` could be used.

pkg/doc/jose/jwk_test.go
Baha-sk: I would like to see a validation check between `crv` and `alg` that we currently support as follows:
ECDSA crv is valid only for P-256, P-384, P-521 and secp256k1 algs
EdDSA crv is valid only for Ed25519 alg

If we wanna support other key algs, we should add them to the list.

pkg/doc/jose/jwk.go
kdimak: @sudeshrshetty the above is just my personal opinion, I'm OK with your solution in general.

sudeshrshetty: @troyronda almost all cases are covered by go-jose, my implementation only covers 'EC/secp256k1', rest of the types will be handled by go-jose.

pkg/doc/jose/jwk_test.go
sudeshrshetty: that's only sample input I generated for tests, you are right EdDSA also works.

pkg/doc/jose/jwk_test.go
sudeshrshetty: I see, alg is ignored by JWK parser since it uses 'kty=EC' and logic based on curve, I fixed tests by regenerating JWK using ES256(ECDSA using P-256)

pkg/doc/jose/jwk_test.go
sudeshrshetty: I agree, go-jose library is doing that, I just branched out my logic when I see 'kty=EC,crv=secp256k1'. 'alg' is not mandatory field in jwk if crv/ec are explanatory enough.

pkg/doc/jose/jwk_test.go
sudeshrshetty: same as above

pkg/doc/jose/jwk_test.go
sudeshrshetty: same as above

pkg/doc/jose/jwk.go
sudeshrshetty: done

pkg/doc/jose/jwk.go
sudeshrshetty: @kdimak sounds good actually, I agree. I will add that as part of upcoming PR. Created an issue #1513  for this.

PR #1515, opened 2020-03-25
feat: Controller for generate public key API

### LINKED ISSUE #1508: Controller for generate public key API

fqutishat:
closes #1508

### DIFF
--- a/cmd/aries-js-worker/src/agent-rest-client.js
+++ b/cmd/aries-js-worker/src/agent-rest-client.js
@@ -128,6 +128,12 @@ const pkgs = {
             pathParam:"id"
         },
     },
+    kms: {
+        CreateKeySet: {
+            path: "/kms/keyset",
+            method: "POST",
+        },
+    },
 }
 
 /**
--- a/pkg/controller/command/kms/command.go
+++ b/pkg/controller/command/kms/command.go
@@ -1,0 +1,75 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package kms
+
+import (
+	"io"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/controller/command"
+	"github.com/hyperledger/aries-framework-go/pkg/controller/internal/cmdutil"
+	"github.com/hyperledger/aries-framework-go/pkg/internal/logutil"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
+)
+
+var logger = log.New("aries-framework/command/kms")
+
+// Error codes
+const (
+	// CreateKeySetError is for failures while creating key set
+	CreateKeySetError = command.Code(iota + command.KMS)
+)
+
+const (
+	// command name
+	commandName = "kms"
+
+	// command methods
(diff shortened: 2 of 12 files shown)


pkg/controller/command/kms/command.go
```diff
@@ -0,0 +1,75 @@
...
+
+// CreateKeySet create a new public/private encryption and signature key pairs set.
+func (o *Command) CreateKeySet(rw io.Writer, req io.Reader) command.Error {
+	encryptionPublicKey, signaturePublicKey, err := o.ctx.LegacyKMS().CreateKeySet()
+	if err != nil {
+		logutil.LogError(logger, commandName, createKeySetCommandMethod, err.Error())
+		return command.NewExecuteError(CreateKeySetError, err)
+	}
+
+	command.WriteNillableResponse(rw, &CreateKeySetResponse{
```
sudeshrshetty: just curious, are we encoding key set in any format here?

fqutishat: @sudeshrshetty  https://github.com/hyperledger/aries-framework-go/pull/1515/files#diff-87eaebcb26229a7feb61ec0d2c34e66fR11

cmd/aries-js-worker/src/agent-rest-client.js
```diff
@@ -128,6 +128,12 @@ const pkgs = {
             pathParam:"id"
         },
     },
+    kms: {
+        CreateKeySet: {
+            path: "/kms/createkeyset",
```
troyronda: shouldn't need `create` prefix in a REST path.

POST to keysets should imply creation.

fqutishat: I will change it

troyronda: or if there are multiple steps in creation (like in DID Exchange), then we would see something like:

https://github.com/hyperledger/aries-framework-go/blob/c18b4aed0de3022064f90d7bf2b89f40963b244b/pkg/controller/rest/didexchange/operation.go#L28

troyronda: (note the `-`separator in the existing APIs).

pkg/controller/command/kms/command.go
```diff
@@ -0,0 +1,75 @@
...
+	// command name
+	commandName = "kms"
+
+	// command methods
+	createKeySetCommandMethod = "CreateKeySet"
+)
+
+// provider contains dependencies for the kms command and is typically created by using aries.Context().
+type provider interface {
+	LegacyKMS() legacykms.KeyManager
```
Baha-sk: Instead of using legacykms why not use `KMS() kms.KeyManager`?

it's currently available in the framework , legacykms will go away in the future..

cmd/aries-js-worker/src/agent-rest-client.js
rolsonquadras: I would prefer HTTP GET with /generatekeyset path as we are not POSTing anything here.

pkg/controller/command/kms/command.go
fqutishat: we will use legacykms until you remove it

cmd/aries-js-worker/src/agent-rest-client.js
fqutishat: done

PR #1516, opened 2020-03-25
feat: Present Proof - state implementation for presentation/proposal-received

### LINKED ISSUE #1511: Present Proof - need to implement proposal-received and presentation-received states
Need to implement the following states according to [RFC](https://github.com/hyperledger/aries-rfcs/tree/master/features/0037-present-proof):
* proposal-received
* presentation-received

soluchok:
The following states were implemented:
* request-sent  (inbound )
* presentation-received  (inbound)
* proposal-sent (outbound)
* proposal-received (inbound )

Closes #1511 

### DIFF
--- a/pkg/didcomm/protocol/presentproof/service_test.go
+++ b/pkg/didcomm/protocol/presentproof/service_test.go
@@ -62,6 +62,7 @@ func TestNew(t *testing.T) {
 	})
 }
 
+// nolint: gocyclo
 func TestService_HandleInbound(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/presentproof/service_test.go
```diff
@@ -278,6 +279,205 @@ func TestService_HandleInbound(t *testing.T) {
 			t.Error("timeout")
 		}
 	})
+
+	t.Run("Receive Propose Presentation (continue)", func(t *testing.T) {
+		var done = make(chan struct{})
+
+		messenger.EXPECT().ReplyTo(gomock.Any(), gomock.Any()).
```
llorllale: all this setup is verbose and distracts from the test case. Any way we can reduce all these lines of setup code in the test case?

pkg/didcomm/protocol/presentproof/states.go
```diff
@@ -202,8 +202,19 @@ func (s *requestSent) CanTransitionTo(st state) bool {
 		st.Name() == stateNameAbandoning
 }
 
-func (s *requestSent) ExecuteInbound(_ *metaData) (state, stateAction, error) {
-	return nil, nil, fmt.Errorf("%s: ExecuteInbound is not implemented yet", s.Name())
+func (s *requestSent) ExecuteInbound(md *metaData) (state, stateAction, error) {
```
llorllale: Why is `requestSent` sending an outbound request message received via `ExecuteInbound()`? Shouldn't this activity be done in its own `ExecuteOutbound()`?

soluchok: `ExecuteInbound` -  is executed in response to incoming messages. In that case, `ExecuteInbound` is triggered by the incoming `proposal` message. In our state machine, the flag `Inbound` is equal to `true`.
The flow for the incoming message can be:
`StateA.ExecuteInbound->StateB.ExecuteInbound->StateC.ExecuteInbound`.

soluchok: `ExecuteOutbound` is implemented as well. See  https://github.com/hyperledger/aries-framework-go/blob/f456261c07c30037b38fc3547436313e10e9f06e/pkg/didcomm/protocol/presentproof/states.go#L220

llorllale: Right, so why is this state replaying the inbound request back to the other agent?

soluchok: Ok, our `state` has two functions. For instance, state `requestSent` has `ExecuteInbound` and `ExecuteOutbound` functions. I thought that if we handle the inbound message our state machine goes through all states and executes the function related to inbound processing (ExecuteInbound). If we are handling outbound messages (which is usually the message which starts the protocol) our state machine goes through all states and executes the `ExecuteOutbound` function. 
Otherwise, I do not see the purpose of having two functions. They can be merged into one Execute function (it makes sense to me). 
@llorllale @troyronda  So my question is why do we need two functions ExecuteOutbound/Inbound?
If messages should be sent through the outbound function the flow will be following:
`e.g (loop) StateA.ExecuteInbound->StateB.ExecuteOutbound->...?`. Is that ok?

llorllale: Initially we had a single `Handle` function. IIRC, @rolsonquadras split those into two. Maybe he recalls why.

But regarding this:

>  I thought that if we handle the inbound message our state machine goes through all states and executes the function related to inbound processing (ExecuteInbound). If we are handling outbound messages (which is usually the message which starts the protocol) our state machine goes through all states and executes the ExecuteOutbound function.

For sure this cannot be generalized to all protocols. Not sure, but *maybe* it applies to the introduce protocol somehow? But coming back to `present-proof`, this **definitely** does not apply. It's literally request-response all the way between two agents. Outbound -> Inbound -> Outbound -> .... etc.

llorllale: By the way, if you think about it, our current design attempts to identify the roles of each party in a protocol by the direction of messages. I feel confident that we can ditch directionality altogether (save for when receiving the first message) and simply assign "myRole", "theirRole" and save those in our state.

soluchok: ok, I am waiting for @rolsonquadras comment and probably will merge these functions.

rolsonquadras: > Initially we had a single `Handle` function. IIRC, @rolsonquadras split those into two. Maybe he recalls why.
> 

@llorllale I had [refactored](https://github.com/hyperledger/aries-framework-go/commit/8d7509f8f86ca51a300f727cfc31af86528c3f3d#diff-86ab318ebdf86972bc415c0a51a2f1ceR159-R179) the code to remove the outbound indicator added by you in [this](https://github.com/hyperledger/aries-framework-go/commit/4e8dfe5838ab8ca68cb546cdf10e090474191420#diff-86ab318ebdf86972bc415c0a51a2f1ceR99-R110) commit.

rolsonquadras: @soluchok @llorllale btw, the state functions are not exposed or expects to implement any interfaces. Each service can have it's own implementation.

llorllale: Right, makes sense.

@soluchok I am in favor of merging these functions in the long term.

Back to this PR: are you going to update it so that `requestSent.ExecuteInbound()` does not replay the message back to the other agent?

soluchok: `are you going to update it so that requestSent.ExecuteInbound()` -  It is not easy. The loop depends on it. I am gonna merge them.

soluchok: To be honest, I would like to merge it. I am afraid that refactoring might be huge. It would be easier to review it after that merge.

llorllale: @soluchok OK - please create a followup task and we'll proceed

soluchok: @llorllale Done! https://github.com/hyperledger/aries-framework-go/issues/1528

PR #1520, opened 2020-03-25
feat: refactor didexchange service to accept out-of-band invitation

### LINKED ISSUE #1501: outofband protocol: refactor didexchange service to accept out-of-band invitation
part of #1454 

The did-exchange protocol's `invitation` is being extracted into its own "out-of-band" protocol. The framework's didexchange protocol service has to be refactored to accept this new invitation object.

llorllale:
closes #1501

* defines a new type to start the did-exchange protocol with (suggestions on the name are welcome)
* out-of-band service refactored to use this new type

The did-exchange protocol service still accepts didexchange.Invitations - that would be removed in a followup PR.

### DIFF
--- a/pkg/didcomm/protocol/didexchange/models.go
+++ b/pkg/didcomm/protocol/didexchange/models.go
@@ -11,11 +11,32 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 )
 
+// OOBInvitation to connect with did-exchange.
+type OOBInvitation struct {
+	// ID of this invitation (for record-keeping purposes).
+	// TODO can we remove this?
+	ID string
+	// TODO remove this
+	Type string `json:"@type"`
+	// ID of the thread from which this invitation originated.
+	// This will become the parent thread ID of the didexchange protocol instance.
+	ThreadID string
+	// Label of the connection invitation.
+	Label string
+	// Target destination.
+	// This can be any on of:
+	// - a string with a valid DID
+	// - a valid `did.Service`
+	Target interface{}
+}
+
 // Invitation model
 //
 // Invitation defines DID exchange invitation message
 // https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange#0-invitation-to-exchange
 //
+// TODO all uses of this invitation struct should be replaced with the new OOB one. The new one should be renamed
+//  to 'Invitation'.
 type Invitation struct {
 	// the Image URL of the connection invitation
 	ImageURL string `json:"imageUrl,omitempty"`
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	vdriapi "github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
 	"github.com/hyperledger/aries-framework-go/pkg/internal/logutil"
 	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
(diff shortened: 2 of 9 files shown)


pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -11,11 +11,32 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 )
 
+// OOBInvitation to connect with did-exchange.
+type OOBInvitation struct {
```
llorllale: naming suggestions are welcome

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -415,6 +418,90 @@ func (s *Service) AcceptExchangeRequest(connectionID, publicDID, label string) e
...
+		NextStateName: stateNameInvited,
+		ConnRecord:    record,
+	}
+
+	err = s.storeEventTransientData(msg)
+	if err != nil {
+		return "", fmt.Errorf("failed to store transient event data : %w", err)
+	}
+
+	err = s.AcceptInvitation(record.ConnectionID, record.InvitationDID, record.TheirLabel)
```
rolsonquadras: I believe we had 2 APIs before for Invitation (cc: @sudeshrshetty ). 
1. Handle Invitation -> get the invitation and save
2. AcceptInvitation -> User approves the invitation and sends exchange request.

Are we going to merge these 2 actions in new OutOfBand Invitation ? (just trying to understand)

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -339,32 +342,40 @@ func extractDIDCommMsgBytes(_ *decorator.Attachment) ([]byte, error) {
...
 }
+
+func chooseTarget(svcs []interface{}) (interface{}, error) {
+	for i := range svcs {
+		switch svc := svcs[i].(type) {
+		case string:
+			return svc, nil
+		case *did.Service:
+			return svc, nil
+		}
```
rolsonquadras: Can these be combined into [one case](https://github.com/golang/go/wiki/Switch#multiple-cases) statement as these are doing same things ?

pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -11,11 +11,32 @@ import (
...
+	// ID of this invitation (for record-keeping purposes).
+	// TODO can we remove this?
+	ID string
+	// TODO remove
+	Type string `json:"@type"`
```
rolsonquadras: just curious, why only one filed has json tag and not others ?

pkg/didcomm/protocol/didexchange/models.go
rolsonquadras: Looks like this is going to be renamed to `Invitation` as per [this](https://github.com/hyperledger/aries-framework-go/pull/1520/files#diff-81d07647d9d693299e624e7e5d843199R38) TODO.

llorllale: Something like that, yes.

pkg/didcomm/protocol/didexchange/models.go
llorllale: To reuse the existing machinery in the didexchange protocol service (`service.DIDCommMsg.Type()`). The existing code only expects didcomm messages. This new struct field is *not* a didcomm message.

pkg/didcomm/protocol/outofband/service.go

pkg/didcomm/protocol/didexchange/service.go
llorllale: There's a `HandleInvitation()` in the didexchange client package: it is processed as an inbound didcomm message (creates connection record, sends didexchange request, etc)

There are two `AcceptInvitation()` - on the didexchange client and service - and they expect the connection record and all that transient data to exist prior to invocation. 

The out-of-band service does not store any data when it receives an inbound didcomm oob invitation.

rolsonquadras: Yes, I understand. But my question was specifically about `RespondTo` function - why can't it just create connection record and store metadata (basically handleInvitation logic) without `Accepting` the Invitation ? Now it looks like we are combining these 2 actions.

Current BDD test -
https://github.com/hyperledger/aries-framework-go/blob/b0374299d8ac005db9c583334bd5c73f0e4a73b3/test/bdd/features/didexchange_e2e_sdk.feature#L22-L23

llorllale: It is currently reusing existing code. Plus "accepting the invitation" seemed like a step in the right direction, semantically speaking.

This PR is just a step in a transition. Completing the refactor on this service is a big task.

How do you see this happening?

rolsonquadras: Oh ok. I guess, we will have clear picture of overall flow once we have OutOfBand-DIDExchange BDD tests.

llorllale: That's my next task

Issue #1521, opened 2020-03-25
JS BDD Tests failing frequently


llorllale:
Example build logs:

* https://github.com/hyperledger/aries-framework-go/pull/1520/checks?check_run_id=534579568
* https://github.com/hyperledger/aries-framework-go/runs/533770667

Error:
<details><summary>expand</summary>
<p>

```
...

  DID-Exchange between two Edge Agents using the router
    ✗ "before all" hook for "Router is running on "http://0.0.0.0:10091,ws://0.0.0.0:10092" with controller "http://0.0.0.0:10093""
	Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

    ✗ "after all" hook for "Alice Edge Agent validates that the connection's state is 'completed'"
	TypeError: Cannot read property 'destroy' of undefined
	    at Context.<anonymous> (test/didexchange/edgeagent_router.js:153:18)

.....

1) "before all" hook for "Router is running on "http://0.0.0.0:10091,ws://0.0.0.0:10092" with controller "http://0.0.0.0:10093""
     DID-Exchange between two Edge Agents using the router
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

2) "after all" hook for "Alice Edge Agent validates that the connection's state is 'completed'"
     DID-Exchange between two Edge Agents using the router
     TypeError: Cannot read property 'destroy' of undefined
    at Context.<anonymous> (test/didexchange/edgeagent_router.js:153:18)
```
</p>
</details>

rolsonquadras:
Fixed as part of #1523

PR #1522, opened 2020-03-25
refactor: Disable DIDExchange JS tests

rolsonquadras:


rolsonquadras:
Disabled the JS DIDExchange test as to unblock other PRs. This will be fixed as part of #1521 .

PR #1526, opened 2020-03-26
fix: VC Evidence field type update

rolsonquadras:
- The Evidence type is of interface{}. The field in VC is set as pointer to Evidence type. ie, pointer to interface{}. Removed the pointer.


troyronda:
Yes - this gets confusing :).

PR #1530, opened 2020-03-27
feat: Present Proof - functions ExecuteInbound/Outbound were merged

### LINKED ISSUE #1528: Present Proof - state interface needs to be changed
State interface needs to be changed to: 

```
type state interface {
	Name() string
	CanTransitionTo(next state) bool
	Execute(msg *metaData) (state, stateAction, error)
}
```

soluchok:
Functions ExecuteInbound/Outbound were merged.

Closes #1528 

### DIFF
--- a/pkg/didcomm/protocol/presentproof/states.go
+++ b/pkg/didcomm/protocol/presentproof/states.go
@@ -33,8 +33,11 @@ const (
 )
 
 const (
+	// error codes
 	codeInternalError = "internal"
 	codeRejectedError = "rejected"
+
+	jsonThread = "~thread"
 )
 
 // state action for network call
@@ -48,8 +51,7 @@ type state interface {
 	CanTransitionTo(next state) bool
 	// Executes this state, returning a followup state to be immediately executed as well.
 	// The 'noOp' state should be returned if the state has no followup.
-	ExecuteInbound(msg *metaData) (state, stateAction, error)
-	ExecuteOutbound(msg *metaData) (state, stateAction, error)
+	Execute(msg *metaData) (state, stateAction, error)
 }
 
 // represents zero state's action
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/presentproof/states.go
```diff
@@ -202,28 +181,25 @@ func (s *requestSent) CanTransitionTo(st state) bool {
...
 	}
+}
 
-	// creates the state's action
-	action := func(messenger service.Messenger) error {
-		// sets message type
-		md.request.Type = RequestPresentationMsgType
-		return messenger.ReplyTo(md.Msg.ID(), service.NewDIDCommMsgMap(md.request))
+func (s *requestSent) Execute(md *metaData) (state, stateAction, error) {
+	if !md.inbound {
```
llorllale: curious: I see we're still using directionality here. Did you consider using role names instead?

llorllale: related: is there a followup plan to also merge the two Handle* service functions?

soluchok: I did not think about it. `Handle*` functions have a common `service` interface. I can try to get rid of `inbound`.

soluchok: @llorllale I've merged Handle* functions and also removed `inbound` flag.

pkg/didcomm/protocol/presentproof/states.go
```diff
@@ -297,28 +265,24 @@ func (s *proposalSent) CanTransitionTo(st state) bool {
 		st.Name() == stateNameAbandoning
 }
 
-func (s *proposalSent) ExecuteInbound(md *metaData) (state, stateAction, error) {
+func canReplyTo(msg service.DIDCommMsgMap) bool {
+	_, ok := msg["~thread"]
```
troyronda: "~thread" should be a constant.

Issue #1513, opened 2020-03-25
Crypto: DecodeJWK custom logic to be part of JWK type.


sudeshrshetty:
https://github.com/hyperledger/aries-framework-go/pull/1510#discussion_r397854515

PR #1538, opened 2020-03-31
refactor: Support JwsVerificationKey2020 for DID public key

### LINKED ISSUE #1527: Support JwsVerificationKey2020 for DID public key
`JwsVerificationKey2020` is defined at https://github.com/transmute-industries/lds-jws2020.
Example of the public key definition with `type=JwsVerificationKey2020`:
```
{
  "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
  "type": "JwsVerificationKey2020",
  "controller": "did:example:123",
  "publicKeyJwk": {
    "crv": "Ed25519",
    "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ",
    "kty": "OKP",
    "kid": "_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A"
  }
}
```

Please note that currently `verifier.PublicKey` structure is passed to the signature verifiers. In this particular case, `Type=JwsVerificationKey2020z  and some public key bytes are filled. This could be not enough for Crypto based signature verifier as we don't pass `crv` and `kty` of the JSON Web Key.

kdimak:
For linked data signatures with `JwsVerificationKey2020` type, it's not enough information to make signature verification based on the public key bytes and a type (JwsVerificationKey2020). We need to know additionally what signature algorithm (and/or curve) was used - e.g. `ES256` or `EdDSA`.
In the case of DID documents, we can read this information from the public key if it's defined as `publicKeyJwk`.
As a result, we extend `PublicKey` struct with `Curve`, `Alg` and `KeyType` fields and use them in the signature verifier implementations (e.g. in `PublicKeyVerifierEC`).

closes #1527, #1513

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -112,7 +112,10 @@ type PublicKey struct {
 	ID         string
 	Type       string
 	Controller string
-	Value      []byte
+
+	Value []byte
+
+	jsonWebKey *jose.JWK
 }
 
 // Service DID doc service
@@ -351,61 +354,88 @@ func populatePublicKeys(context string, rawPKs []map[string]interface{}) ([]Publ
 	var publicKeys []PublicKey
 
 	for _, rawPK := range rawPKs {
-		decodeValue, err := decodePK(rawPK)
-		if err != nil {
-			return nil, err
-		}
-
 		controllerKey := jsonldController
 
 		if context == contextV011 {
 			controllerKey = jsonldOwner
 		}
 
-		publicKeys = append(publicKeys, PublicKey{ID: stringEntry(rawPK[jsonldID]), Type: stringEntry(rawPK[jsonldType]),
-			Controller: stringEntry(rawPK[controllerKey]), Value: decodeValue})
+		publicKey := PublicKey{ID: stringEntry(rawPK[jsonldID]), Type: stringEntry(rawPK[jsonldType]),
+			Controller: stringEntry(rawPK[controllerKey])}
+
+		err := decodePK(&publicKey, rawPK)
+		if err != nil {
+			return nil, err
+		}
+
+		publicKeys = append(publicKeys, publicKey)
 	}
 
 	return publicKeys, nil
 }
 
-func decodePK(rawPK map[string]interface{}) ([]byte, error) {
+func decodePK(publicKey *PublicKey, rawPK map[string]interface{}) error {
 	if stringEntry(rawPK[jsonldPublicKeyBase58]) != "" {
-		return base58.Decode(stringEntry(rawPK[jsonldPublicKeyBase58])), nil
+		publicKey.Value = base58.Decode(stringEntry(rawPK[jsonldPublicKeyBase58]))
+		return nil
 	}
 
 	if stringEntry(rawPK[jsonldPublicKeyHex]) != "" {
 		value, err := hex.DecodeString(stringEntry(rawPK[jsonldPublicKeyHex]))
 		if err != nil {
-			return nil, fmt.Errorf("decode public key hex failed: %w", err)
+			return fmt.Errorf("decode public key hex failed: %w", err)
 		}
 
-		return value, nil
+		publicKey.Value = value
+
+		return nil
 	}
 
 	if stringEntry(rawPK[jsonldPublicKeyPem]) != "" {
 		block, _ := pem.Decode([]byte(stringEntry(rawPK[jsonldPublicKeyPem])))
 		if block == nil {
-			return nil, errors.New("failed to decode PEM block containing public key")
+			return errors.New("failed to decode PEM block containing public key")
 		}
 
-		return block.Bytes, nil
+		publicKey.Value = block.Bytes
+
+		return nil
 	}
 
 	if jwkMap := mapEntry(rawPK[jsonldPublicKeyjwk]); jwkMap != nil {
-		jwkBytes, err := json.Marshal(jwkMap)
-		if err != nil {
-			return nil, fmt.Errorf("failed to marshal '%s', cause: %w ", jsonldPublicKeyjwk, err)
-		}
-
-		if string(jwkBytes) == "{}" {
-			return []byte(""), nil
-		}
-
-		return jose.DecodePublicKey(jwkBytes)
+		return decodePublicKeyJwk(jwkMap, publicKey)
 	}
 
-	return nil, errors.New("public key encoding not supported")
+	return errors.New("public key encoding not supported")
+}
+
+func decodePublicKeyJwk(jwkMap map[string]interface{}, publicKey *PublicKey) error {
+	jwkBytes, err := json.Marshal(jwkMap)
+	if err != nil {
+		return fmt.Errorf("failed to marshal '%s', cause: %w ", jsonldPublicKeyjwk, err)
+	}
+
+	if string(jwkBytes) == "{}" {
+		publicKey.Value = []byte("")
+		return nil
+	}
+
+	var jwk jose.JWK
+
+	err = json.Unmarshal(jwkBytes, &jwk)
+	if err != nil {
+		return fmt.Errorf("unmarshal JWK: %w", err)
+	}
+
+	pkBytes, err := jwk.PublicKeyBytes()
+	if err != nil {
+		return fmt.Errorf("failed to decode public key from JWK: %w", err)
+	}
+
+	publicKey.Value = pkBytes
+	publicKey.jsonWebKey = &jwk
+
+	return nil
 }
 
 func (r *rawDoc) ParseContext() []string {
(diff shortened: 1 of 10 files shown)


pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
```diff
@@ -11,24 +11,45 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"errors"
+	"fmt"
 	"math/big"
 
+	"github.com/decred/dcrd/dcrec/secp256k1/v2"
```
Baha-sk: have you tried  github.com/btcsuite ? I believe they have secp256k1 alg as well.

pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
```diff
@@ -11,24 +11,45 @@ import (
...
+		keySize = 48
+	case "P-521":
+		curve = elliptic.P521()
+		keySize = 66
+	case "secp256k1":
+		curve = secp256k1.S256()
+		keySize = 32
+	default:
+		return fmt.Errorf("ecdsa: unsupported elliptic curve '%s'", pubKey.Curve)
+	}
```
Baha-sk: this doesn't belong to Verify()

create a separate function

probably create a struct keyInfo that has a curve and keySize if needed in more than one place
probably expose it to be used for Sign as well ?

pkg/doc/jose/jwk.go
```diff
@@ -24,101 +24,211 @@ import (
...
-	if !ok {
-		return nil, false
-	}
-
-	var jwk JWK
+	x   *big.Int
+	y   *big.Int
+	Kty string
+	Crv string
+}
```
Baha-sk: there's no `kid` needed?

pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
kdimak: I didn't. `github.com/decred` was initially added by @sudeshrshetty  but I think we can switch to `github.com/btcsuite`.

pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
kdimak: OK! I will try to simplify it.

pkg/doc/jose/jwk.go
kdimak: It's already there, inside embedded `jose.JSONWebKey`. `x`, 'y', 'key type', 'curve' are not copied from `jose.rawJSONWebKey` to `jose.JSONWebKey` as a result of unmarshalling. That's why I had to add them to `JWK` struct to support `secp256k1` properly.

Baha-sk: there seem to be duplication of fields with JSONWebKey then.. it's better to remove duplicate fields

pkg/doc/did/doc.go
```diff
@@ -112,7 +112,12 @@ type PublicKey struct {
 	ID         string
 	Type       string
 	Controller string
-	Value      []byte
+
+	Value []byte
+
+	Algorithm string `json:"-"`
```
sudeshrshetty: Why we need to show Alg, Curve & KeyType in exported did.Doc.PublicKey? Why can't we just decode public key internally and stick to old structure?

pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
troyronda: @kdimak @Baha-sk Switching that import is not part of this PR. Consider followup if there is a reason to do so.

pkg/doc/did/doc.go
kdimak: I think we can at least make it hidden, why not?

kdimak: Those new fields are made private, thank you @sudeshrshetty

sudeshrshetty: looks good, but what is the use of those private fields?
 As I can see we are setting it from jwk instance but not being used anywhere afterwards.

pkg/doc/jose/jwk.go
```diff
@@ -18,107 +18,220 @@ import (
...
+func unmarshalSecp256k1(jwk *jsonWebKey) (*JWK, error) {
+	if jwk.X == nil {
+		return nil, fmt.Errorf("missed x value in EC key")
 	}
 
-	if strings.EqualFold(key.Kty, secp256k1Kty) && strings.EqualFold(key.Crv, secp256k1Crv) {
-		//  if kty="EC" and Crv="secp256k1", then handle differently
-		return btcPublicKey(key.X, key.Y)
+	if jwk.Y == nil {
+		return nil, fmt.Errorf("missed y value in EC key")
```
Baha-sk: I noticed there are many detailed error messages returned like the above line.

If these messages reveal key details, then it's a security vulnerability. Please review all the returned errors in this document and only return generic messages to the user (or probably log detailed error messages in `debug` mode only).

pkg/doc/did/doc.go
Baha-sk: usually `Curve` is exposed (like in ecdsa.PublicKey)

sudeshrshetty: Why clients should care about curve? For key type based logics - they should be able to figure out using public key type

kdimak: For `JsonWebSignature2020`, it's not enough to have only `type` value of the public key. Look at the example:
```
  "publicKey": [
    {
      "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
      "type": "JwsVerificationKey2020",
      "controller": "did:example:123",
      "publicKeyJwk": {
        "crv": "Ed25519",
        "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ",
        "kty": "OKP",
        "kid": "_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A"
      }
    },
    {
      "id": "did:example:123#4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A",
      "type": "JwsVerificationKey2020",
      "controller": "did:example:123",
      "publicKeyJwk": {
        "crv": "secp256k1",
        "x": "Z4Y3NNOxv0J6tCgqOBFnHnaZhJF6LdulT7z8A-2D5_8",
        "y": "i5a2NtJoUKXkLm6q8nOEu9WOkso1Ag6FTUT6k_LMnGk",
        "kty": "EC",
        "kid": "4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A"
      }
    }
  ]
```

The `type` is the same and is `JwsVerificationKey2020`. 
We can't deduce is it `Ed25519` or `secp256k1` etc. We need to extract extra information from `publicKeyJwk`.
That's the problem solved in this PR.

pkg/doc/jose/jwk.go
kdimak: I agree. I changed the detailed error messages to a single generic `invalid JWK`.

pkg/doc/did/doc.go
```diff
@@ -112,7 +112,12 @@ type PublicKey struct {
 	ID         string
 	Type       string
 	Controller string
-	Value      []byte
+
+	Value []byte
+
+	algorithm string
```
troyronda, line 118: this would seem like it could be a publicKeyJWK sub struct?

kdimak: yes, we can just add `publicKeyJwk` field of `*jose.JWK` type.

pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
```diff
@@ -65,3 +68,42 @@ func (v *PublicKeyVerifierP256) Verify(pubKey *sigverifier.PublicKey, doc, signa
...
+		return &ellipticCurve{
+			Curve:   elliptic.P384(),
+			KeySize: p384KeySize,
+		}
+	case "P-521":
+		return &ellipticCurve{
+			Curve:   elliptic.P521(),
+			KeySize: p521KeySize,
+		}
+	case "secp256k1":
```
troyronda, line 101: what about X25519 / Ed25519?

pkg/doc/did/doc.go
kdimak: I will also update `verifier.PublicKey` struct.

kdimak: Updated.

pkg/doc/signature/suite/jsonwebsignature2020/public_key_verifier.go
kdimak: Filed https://github.com/hyperledger/aries-framework-go/issues/1556.

PR #1539, opened 2020-03-31
fix: outofband + didexchange

llorllale:
Working on the BDD tests for outofband+didexchange made me realize some mistaken assumptions I had about how the didexchange state machine works, and also that there were missing pieces to implement.

This PR will:

* Add `outofband.Client.AcceptRequest()` to receive `request`s received via out-of-band channels. It returns the new connection record's ID (similar to `didexchange.Client.HandleInvitation()`)
* Add `outofband.Service.AcceptRequest()` to accept another agent's request. This method returns the new connection record's ID.
* Add `outofband.Service.SaveRequest()` to save a request created by yourself
* Add `didexchange.Service.Save()` to save OOBInvitations created by yourself
* Hook `outofband.Service` into the framwork context's defaults                                                                
* Refactor `didexchange.Service.RespondTo()` to match what its `Accept*` methods do. This came about from the realization that I had mistaken assumptions about how the didexchange state machine works. Additional changes in the `Service` and state implementations were done to integrate `OOBInvitation` as close as possible to how regular `didexchange.Invitation`s are done.
* Fix existing examples and tests that are incorrectly setting the signing key on the mock KMS
* Add a few debug logging statements to aid troubleshooting

In general, the storing and much of the validation logic was pushed to/kept in the protocol services to keep things DRY and to avoid circular dependencies between client and service packages.

### DIFF
--- a/pkg/client/outofband/client.go
+++ b/pkg/client/outofband/client.go
@@ -17,8 +17,6 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/route"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/kms/legacykms"
-	"github.com/hyperledger/aries-framework-go/pkg/storage"
-	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
 )
 
 const (
@@ -29,15 +27,13 @@ const (
 // RequestOptions allow you to customize the way request messages are built.
 type RequestOptions func(*Request) error
 
-// ConnectionRecorder records connection records produced as byproducts of creation of messages.
-type connectionRecorder interface {
-	SaveInvitation(id string, i interface{}) error
+type oobService interface {
+	AcceptRequest(request *outofband.Request) (string, error)
+	SaveRequest(request *outofband.Request) error
 }
 
 // Provider provides the dependencies for the client.
 type Provider interface {
-	StorageProvider() storage.Provider
-	TransientStorageProvider() storage.Provider
 	ServiceEndpoint() string
 	Service(id string) (interface{}, error)
 	LegacyKMS() legacykms.KeyManager
@@ -47,19 +43,24 @@ type Provider interface {
 // https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md
 type Client struct {
 	didDocSvcFunc func() (*did.Service, error)
-	connRecorder  connectionRecorder
+	oobService    oobService
 }
 
 // New returns a new Client for the Out-Of-Band protocol.
 func New(p Provider) (*Client, error) {
-	r, err := connection.NewRecorder(p)
+	s, err := p.Service(outofband.Name)
 	if err != nil {
-		return nil, fmt.Errorf("failed to initialize outofband client : %w", err)
+		return nil, fmt.Errorf("failed to look up service %s : %w", outofband.Name, err)
+	}
+
+	oobSvc, ok := s.(oobService)
+	if !ok {
+		return nil, fmt.Errorf("failed to cast service %s as a dependency", outofband.Name)
 	}
 
 	return &Client{
 		didDocSvcFunc: didServiceBlockFunc(p),
-		connRecorder:  r,
+		oobService:    oobSvc,
 	}, nil
 }
 
@@ -92,12 +93,30 @@ func (c *Client) CreateRequest(opts ...RequestOptions) (*Request, error) {
 	req.ID = uuid.New().String()
 	req.Type = RequestMsgType
 
-	err := c.connRecorder.SaveInvitation(req.ID, req)
+	err := c.oobService.SaveRequest(req.Request)
 	if err != nil {
-		return nil, fmt.Errorf("failed to save request : %w", err)
+		return nil, fmt.Errorf("outofband service failed to save request : %w", err)
 	}
 
 	return req, nil
+}
+
+// AcceptRequest from another agent and return the ID of a new connection record.
+func (c *Client) AcceptRequest(r *Request) (string, error) {
+	connID, err := c.oobService.AcceptRequest(&outofband.Request{
+		ID:       r.ID,
+		Type:     r.Type,
+		Label:    r.Label,
+		Goal:     r.Goal,
+		GoalCode: r.GoalCode,
+		Requests: r.Requests,
+		Service:  r.Service,
+	})
+	if err != nil {
+		return "", fmt.Errorf("out-of-band service failed to accept request : %w", err)
+	}
+
+	return connID, err
 }
 
 // WithLabel allows you to specify the label on the message.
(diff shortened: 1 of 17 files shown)


pkg/client/outofband/client_test.go
```diff
@@ -13,6 +13,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/outofband"
+
```
troyronda, line 17: extra space

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -169,6 +171,46 @@ func (s *Service) HandleOutbound(_ service.DIDCommMsg, _, _ string) error {
 	return errors.New("not implemented")
 }
 
+// AcceptRequest from another agent and return the connection ID.
+func (s *Service) AcceptRequest(r *Request) (string, error) {
+	connID, err := s.handleRequestCallback(&callback{
+		msg:      service.NewDIDCommMsgMap(r),
+		myDID:    "",
+		theirDID: "",
```
soluchok, line 179: remove these empty fields

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -169,6 +171,46 @@ func (s *Service) HandleOutbound(_ service.DIDCommMsg, _, _ string) error {
...
+		return "", fmt.Errorf("failed to accept request : %w", err)
+	}
+
+	return connID, err
+}
+
+// SaveRequest created by the outofband client.
+func (s *Service) SaveRequest(r *Request) error {
+	// TODO the ID when saving this request...
+	err := s.connections.SaveInvitation(r.ID+"-TODO", r)
```
soluchok, line 191: Please, add more description and create an issue for that. I don't understand the problem.

pkg/client/outofband/client.go
```diff
@@ -92,14 +93,32 @@ func (c *Client) CreateRequest(opts ...RequestOptions) (*Request, error) {
...
 	}
 
 	return req, nil
 }
 
+// AcceptRequest from another agent and return the ID of a new connection record.
+func (c *Client) AcceptRequest(r *Request) (string, error) {
+	connID, err := c.oobService.AcceptRequest(&outofband.Request{
+		ID:       r.ID,
+		Type:     r.Type,
```
soluchok, line 108: It makes sense to check that `r *Request` is not nil.

pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -15,7 +15,7 @@ import (
 type OOBInvitation struct {
 	// ID of this invitation (for record-keeping purposes).
 	// TODO can we remove this?
-	ID string
+	ID string `json:"@id"`
```
soluchok, line 18: TODO can we remove this? - sure, it will be set by the messenger. If you need this ID before the messenger sets it you need to do this manually. e.g

```
if msg.ID() == "" {
    msg.SetID(uuid.New().String())
}
```

llorllale: `OOBInvitation` is not a DIDComm message. I don't expect it to be sent nor received via the messenger.

pkg/client/outofband/client.go
llorllale: I generally stay away from that extreme level of defensive programming. I don't believe it's one of this project's guidelines.

pkg/didcomm/protocol/didexchange/models.go
llorllale: To be clear: that TODO is about removing `ID` entirely, not just the json tag.

pkg/client/outofband/client.go
troyronda: I also generally do not think we need to cover developer errors like passing a nil request. (we should be covering runtime issues.)

pkg/didcomm/protocol/outofband/service.go
llorllale: done: #1547

pkg/didcomm/protocol/outofband/service.go

pkg/client/outofband/client_test.go
llorllale: done

pkg/client/outofband/client.go
soluchok: got it

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -422,84 +426,19 @@ func (s *Service) AcceptExchangeRequest(connectionID, publicDID, label string) e
...
-		svc = target
-	default:
-		return nil, fmt.Errorf("unsupported target type: %+v", target)
-	}
-
-	record.RecipientKeys = svc.RecipientKeys
-	record.RoutingKeys = svc.RoutingKeys
-	record.ServiceEndPoint = svc.ServiceEndpoint
+// Save this invitation created by you.
+func (s *Service) Save(i *OOBInvitation) error {
```
rolsonquadras, line 433: I feel better to have a suffix in the name - SaveInvitation or SaveOOBInvitation or something similar.

pkg/client/outofband/client.go
```diff
@@ -29,15 +27,13 @@ const (
 // RequestOptions allow you to customize the way request messages are built.
 type RequestOptions func(*Request) error
 
-// ConnectionRecorder records connection records produced as byproducts of creation of messages.
-type connectionRecorder interface {
-	SaveInvitation(id string, i interface{}) error
+type oobService interface {
+	AcceptRequest(request *outofband.Request) (string, error)
+	SaveRequest(request *outofband.Request) error
```
troyronda, line 32: Just to double-check, have we been using the `Save` terminology elsewhere?

pkg/didcomm/protocol/didexchange/service.go
troyronda: Yeh - that would seem more clear.

pkg/client/outofband/client.go
llorllale: `connection.Recorder` has `Save*` methods.

There is `saveCredential` in the JS API and `SaveCredential` in the REST interface.

There are numerous other unexported `save*` methods elsewhere.

pkg/didcomm/protocol/didexchange/service.go
llorllale: I was led to believe we did that because golang doesn't feature method overloading.

rolsonquadras: Also, it would be inconsistent when someone else adds a new SaveXYZ API. We would end up with Save() and SaveXYZ() functions.

llorllale: Ok - I disagree, but I'm not dying on this hill. I renamed it to "SaveInvitation".

Here are my own thoughts on the matter:

* There is a TODO to rename `OOBInvitation` - I'm not even sure it'll end up being called "Invitation".
* People worrying about refactoring a method's name makes me worry about the fragility of the code and the process.
* There seems to be a strong desire to attempt to predict the future. Let's just embrace change (the agile way) :). I don't see this service saving anything else, but that is my own perception of things, and it's limited to *today*
* There is a tendency for overly verbose method and variable names that encumbers the code's legibility

PR #1540, opened 2020-03-31
feat: Add main crypto interfaces and proto buffs for a new ECDH-ES kw + AEAD crypto

### LINKED ISSUE #1469: Create new Key types to support ECDH-ES primitives (using NIST approved keys)

### LINKED ISSUE #1470: Update created Tink Key types to return JWE components from crypto primitives
As a follow to #1469 

The new created Tink Key type and primitive must be able to return JWE components needed by Anoncrypt's Pack/Unpack operations like:
IV, cipher text & tag, encrypted CEK and derived EPK(per recipient).

### LINKED ISSUE #1472: Replace HKDF with ConcatKDF in new ECDH-ES Tink primitive
For JWE key derivations, the common and standard ECDH-ES is ConcatKDF of z, APU, APV, APUInfo, APVInfo with a hash function.

This is different than the current patch's HKDF derivation used in ECIES.

replace HKDF with ConcatKDF.

This is a continuation of #1469

Baha-sk:
This is a first change for Anoncrypt crypto primitives mainly called Composite primitives.

Following changes will build on top of this change, namely the crypto primitives, Tink templates
and helpers to complete full ECDH-ES key wrapping + AEAD content encryption logic needed
for building and parsing JWE messages.

part of #1469


pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_private_key_manager.go
```diff
@@ -0,0 +1,200 @@
+/*
```
troyronda, line 1: I don't fully understand the jwe reference in the path. JWE is not a crypto primitive. Can you clarify the meaning?

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package ecdhes provides implementations of JWE encryption using ECDH-ES KW primitives.
```
troyronda, line 7: JWE is not, itself, a crypto primitive. JWE belongs at a higher layer. Again, please clarify the intended meaning.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
```diff
@@ -0,0 +1,74 @@
...
+//
+// The functionality of ecdhes Encryption is represented as a pair of
+// primitives (interfaces):
+//
+//  * ECDHESEncrypt for encryption of data as JWE
+//
+//  * ECDHESDecrypt for decryption of JWE data and returning decrypted plaintext
```
troyronda, line 14: There should not be explicit JWE data. JWE is a higher layer.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
```diff
@@ -0,0 +1,98 @@
...
+
+		err := proto.Unmarshal(k.Value, gcmKeyFormat)
+		if err != nil {
+			return nil, err
+		}
+
+		keyLen = gcmKeyFormat.KeySize
+		switch keyLen {
+		case 16: // nolint:gomnd
+			encAlg = jose.A128GCM
```
troyronda, line 58: I do not understand why KMS/Crypto is dealing with JOSE data structures. This should be refactored.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	rDem, err := newRegisterECDHESEncHelper(key.Params.DemParams.AeadDem)
+	if err != nil {
+		return nil, err
+	}
+
+	ptFormat := key.Params.EcPointFormat.String()
+
+	// TODO add support for multiple 'recipient' keys somehow.
```
troyronda, line 81: issue number?

proto/ecdhes_aead.proto
```diff
@@ -0,0 +1,92 @@
+// Copyright 2017 Google Inc.
```
troyronda, line 1: why do we have this directory at the top level? these are protos for tink.

proto/readme.md
```diff
@@ -0,0 +1,14 @@
+# How to generate ecdhes_aead protobuf
```
troyronda, line 1: usually README.md

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	// for now, assuming recipient kw alg is jose.ECDH_ES_A256KW using pubKey provided by Tink
+	rcpt := &jose.Recipient{Algorithm: jose.ECDH_ES_A256KW, Key: pubKey}
+
+	encOpts := &jose.EncrypterOptions{
+		Compression: jose.NONE,
+	}
+
+	// for now, assuming didcomm type, TODO: jwe type will need to be parametrized
```
troyronda, line 52: Can we fix this asap? This package should not be hardcoded to didcomm.

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	pubKey := &ecdsa.PublicKey{
+		Curve: recipientPubKey.Curve,
+		X:     recipientPubKey.Point.X,
+		Y:     recipientPubKey.Point.Y,
+	}
+
+	// for now, assuming recipient kw alg is jose.ECDH_ES_A256KW using pubKey provided by Tink
+	rcpt := &jose.Recipient{Algorithm: jose.ECDH_ES_A256KW, Key: pubKey}
```
troyronda, line 46: Why is this hardcoded to jose.ECDH_ES_A256KW? We will also need to support chacha key wrapping in aries.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_private_key_manager.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/google/tink/go/subtle/hybrid"
+	tinkpb "github.com/google/tink/proto/tink_go_proto"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes"
+	ecdhespb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/ecdhes_aead_go_proto"
+)
+
+const (
+	ecdhesPrivateKeyVersion = 0
+	ecdhesPrivateKeyTypeURL = "type.googleapis.com/google.crypto.tink.EcdhesAeadPrivateKey"
```
troyronda, line 24: Is this correct for a type defined outside of the tink project?

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
```diff
@@ -0,0 +1,115 @@
...
+	// 1 key template is for 1 recipient only. A possible solution would be to update Encrypt() interface to take
```
troyronda, line 82: Updating tink interfaces doesn't seem very generic.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
```diff
@@ -0,0 +1,98 @@
...
+	gcmpb "github.com/google/tink/proto/aes_gcm_go_proto"
+	tinkpb "github.com/google/tink/proto/tink_go_proto"
+	"github.com/square/go-jose/v3"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes"
+)
+
+const (
+	// nolint:gochecknoglobals
+	aesGCMTypeURL = "type.googleapis.com/google.crypto.tink.AesGcmKey"
```
troyronda, line 24: I assume this one is actually defined in upstream tink?

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
```diff
@@ -0,0 +1,98 @@
...
+		gcmKeyFormat := new(gcmpb.AesGcmKeyFormat)
+
+		err := proto.Unmarshal(k.Value, gcmKeyFormat)
+		if err != nil {
+			return nil, err
+		}
+
+		keyLen = gcmKeyFormat.KeySize
+		switch keyLen {
+		case 16: // nolint:gomnd
```
troyronda, line 57: instead of nolint, you could create constants.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes_key_template.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"github.com/golang/protobuf/proto"
+	"github.com/google/tink/go/aead"
+	commonpb "github.com/google/tink/proto/common_go_proto"
+	tinkpb "github.com/google/tink/proto/tink_go_proto"
+
+	ecdhespb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/ecdhes_aead_go_proto"
+)
+
+// AES256GCMECDHESKeyTemplate is a KeyTemplate that generates an ECDH-ES P-256 key wrapping and AES256-GCM CEK with
```
troyronda, line 18: we will need a todo for chacha key wrapping template.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
troyronda: @Baha-sk I think we should, at least, have a TODO to refactor JWE specifics out where possible.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+
+	"github.com/google/tink/go/core/registry"
+)
+
+// nolint: gochecknoinits
+func init() {
+	err := registry.RegisterKeyManager(newECDHESPrivateKeyManager())
```
troyronda, line 65: do we really need to use this registry singleton?

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_private_key_manager.go
Baha-sk: This new package represents a new `special` primitive not be used directly by Crypto service. The aim is to create a special keys/crypto primitive for JWE operations and be able to store and exchange keys in localKMS. The best way to do this is to create new fully Tink supported keys and primitives for this special operation. 

The aim is for components requiring JWE envelopes to directly call these new primitives  in as many steps as possible. These steps would be:
1. get a key handle that supports JWE building
2. call Encrypt()

Instead  of a multiple steps process where Encrypt() would be broken down to the separate parts of the JWE message, ie get recipient public key, create a nonce, create a CEK, encrypt plaintext, extract tag, derive recipient's wrapping key, etc.

That's why `jwe` is in the path of this new special Tink primitive.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
Baha-sk: see my previous comment

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
Baha-sk: Crypto package is not involved in this change. This is a special Tink key/primitive to do JWE operations. It's similar to Tink's hybrid primitive that does ECIES operation, only we use go-jose for building the final output while hybrid builds the final output internally in Tink.

proto/ecdhes_aead.proto
Baha-sk: I will move it.

proto/readme.md
Baha-sk: I'll rename

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
Baha-sk: sure.. I wanted to get something out as soon as possible. 

the same for encHelper and encrypter, they should be merged, probably encrypter should be a member of encHelper, I'll update this as well.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
Baha-sk: sure, I'll update

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
Baha-sk: Yes, it's not exported, so I had to copy it here.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
Baha-sk: I followed Tink's pattern with other primitives. They register key managers this way.

The registry is used to fetch primitives and their keys. We will need to change Tink to remove singleton.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_private_key_manager.go
Baha-sk: Should I change to `hyperledger.aries.com`? This identifies the new key types introduced in this change.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
Baha-sk: The interface is created in this change (see `pkg/crypto/tinkcrypto/primitive/jwe/ecdhes_decrypt.go` below), each primitive has its own interface.

There are two entries to Tink when getting a new primitive and executing it, one is when creating the key template as follows:
`kh, err := keyset.NewHandle(AES256GCMECDHESKeyTemplate())`

One option would be to add recipients argument to `AES256GCMECDHESKeyTemplate()`

The other option is when executing the primitive operation, the `Encrypt()` function interface as mentioned in the code comment.

But that would be for another issue, hence the TODO, I'll create an issue for this.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
Baha-sk: I'll add

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
Baha-sk: At the moment go-jose doesn't support chacha content encryption nor key wrapping (they only support chacha signatures),  this will be refactored when we introduce chacha algorithm.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
troyronda: Isn't the `Encrypt` method a defined interface? (I assume unlike the constructor `AES256GCMECDHESKeyTemplate`)

troyronda: BTW if you put recipients into the constructor, why would we still need JWE return values.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_private_key_manager.go
Baha-sk: JWE reference is now removed from this change. It will be introduced in a another PR specifically for JWE outside of crypt package.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
Baha-sk: JWE reference is now removed from this change. It will be introduced in a another PR specifically for JWE outside of crypt package.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
Baha-sk: JWE reference is now removed from this change. It will be introduced in a another PR specifically for JWE outside of crypt package.

proto/ecdhes_aead.proto
Baha-sk: moved under `protos/tink`

proto/readme.md
Baha-sk: updated

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
Baha-sk: outdated code and therefore removed

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
Baha-sk: chacha key wrapping is net new, go Jose doesn't support it. I haven't seen anyone implement it (except in my experimental JWE Pack() func)

We can add a key template for chacha AEAD content encryption though.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
Baha-sk: JWE reference is now removed from this change and now supports multiple recipients exchange in the crypto side.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes_key_template.go
Baha-sk: TODO added... chacha key wrapping is tbd

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/register_ecdhes_aead_helper.go
Baha-sk: there's no more hard coded key size in my latest change

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_private_key_manager.go
Baha-sk: updated to `type.hyperledger.org/hyperledger.aries.crypto.tink.EcdhesAeadPrivateKey`

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdh_aead_public_key_manager.go
Baha-sk: Multiple recipients keys are now supported in this change.

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes_key_template.go
troyronda: issue number?

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
troyronda: status?

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes_key_template.go
Baha-sk: created #1637 and added it as a comment in the key templates 

We will need to discuss how chacha key wrapping will be done if it must be supported. But for now we can add Chacha content encryption with ECDE-ES key wrapping using a key with P-256 curve.

pkg/crypto/tinkcrypto/primitive/jwe/subtle/ecdhes/ecdhes_kw_aead_jwe_encrypt.go
Baha-sk: created #1637 and added it as a comment in the key templates

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes_key_template.go
Baha-sk: created #1637 and added it as a comment in the key templates 

we will need to discuss how chacha key wrapping will be done if it must be supported. But for now we can add Chacha content encryption with ECDE-ES key wrapping using a key with P-256 curve

pkg/crypto/tinkcrypto/primitive/jwe/ecdhes/ecdhes.go
Baha-sk: JWE reference is now removed from this change. It will be introduced in a another PR specifically for JWE outside of crypt package.

Baha-sk: created #1638 to build and parse the JWE message (outside of Crypto package)

Baha-sk:
I reduced the scope of this PR and will push subsequent PRs for the rest of the Crypto primitives impelementations.

troyronda:
@Baha-sk the PR description is still the same?

protos/tink/README.md
```diff
@@ -0,0 +1,65 @@
+# How to generate ecdhes_aead protobuf
+
+To execute the proto generation of `ecdhes_aead.proto`, copy this file into
```
troyronda, line 3: Seems like we should also add a script and make target for these steps?

.golangci.yml
```diff
@@ -10,7 +10,7 @@ run:
   issues-exit-code: 1
   tests: true
   build-tags: [""]
-  skip-dirs:  [""]
+  skip-dirs:  ["proto"]
```
troyronda, line 13: this should be /protos

This is why your code coverage is broken.

Baha-sk: I'll fix

troyronda: Also are .pb.go files being ignored?

Baha-sk: I believe golangci ignores auto generated files..

as discussed, I will remove `skip-dirs:  ["proto"]` since golangci ignores non go files anyway.

protos/tink/README.md
Baha-sk: generating the protobuf at each built is probably overkill as it requires checking out Tink then add and run the new build target mentioned in this readme. 

The go proto is generated and included in this PR to avoid rebuilds. But we can probably automate this if needed later..

PR #1546, opened 2020-04-01
feat: Create DID store

### LINKED ISSUE #1543: Create DID store
Store DID using generic wasm in aries framework go. We need this later to sign the presentations.

talwinder50:
closes #1543

### DIFF
--- a/pkg/store/did/store.go
+++ b/pkg/store/did/store.go
@@ -1,0 +1,135 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package did
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+const (
+	// NameSpace for did store
+	NameSpace = "didstore"
+
+	didNameKey        = "didname_"
+	didNameKeyPattern = didNameKey + "%s"
+
+	// limitPattern for the iterator
+	limitPattern = "%s~"
+)
+
+// ErrDIDNotFound signals that the entry for the given DID and key is not present in the store.
+var ErrDIDNotFound = errors.New("did not found under given key")
+
+// Store stores did doc
+type Store struct {
+	store storage.Store
+}
+
+type provider interface {
+	StorageProvider() storage.Provider
+}
+
+// New returns a new did store
+func New(ctx provider) (*Store, error) {
+	store, err := ctx.StorageProvider().OpenStore(NameSpace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open did store: %w", err)
+	}
+
+	return &Store{store: store}, nil
+}
+
+// SaveDID saves a did doc.
+func (s *Store) SaveDID(name string, didDoc *did.Doc) error {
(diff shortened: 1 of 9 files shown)


pkg/store/did/store.go
```diff
@@ -0,0 +1,135 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+const (
+	// NameSpace for did store
+	NameSpace = "did"
```
sudeshrshetty, line 19: name space too generic

pkg/store/did/store.go
```diff
@@ -0,0 +1,135 @@
...
+	if id != "" {
+		return errors.New("did name already exists")
+	}
+
+	docBytes, err := didDoc.JSONBytes()
+	if err != nil {
+		return fmt.Errorf("failed to marshal didDoc: %w", err)
+	}
+
+	if err := s.store.Put(didDoc.ID, docBytes); err != nil {
```
sudeshrshetty, line 70: does it mean, someone can save DID by name and later also retrieve by ID?

pkg/store/did/store.go
talwinder50: how about didStore ?

pkg/store/did/store.go
```diff
@@ -0,0 +1,135 @@
...
+	store, err := ctx.StorageProvider().OpenStore(NameSpace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open did store: %w", err)
+	}
+
+	return &Store{store: store}, nil
+}
+
+// SaveDID saves a did doc.
+func (s *Store) SaveDID(name string, didDoc *did.Doc) error {
```
rolsonquadras, line 51: Do we need to store DID Document here? I understand the requirement to save Name and DID mapping. We can always fetch DID Document from VDRI.

talwinder50: @sandrask

pkg/store/did/store.go
talwinder50: yes you can retrieve by both name and id

pkg/store/did/store.go
sandrask: It is true that we can always resolve it. However it is more efficient to store it since the user is supposed to be managing their own did.

sandrask: @troy What is your take on this? Should we store did documents or just dids?

pkg/store/did/store.go
talwinder50: have changed that to didstore

pkg/store/did/store.go
```diff
@@ -0,0 +1,135 @@
...
+	if name == "" {
+		return errors.New("did name is mandatory")
+	}
+
+	id, err := s.GetDIDByName(name)
+	if err != nil && !errors.Is(err, storage.ErrDataNotFound) {
+		return fmt.Errorf("get did id using name : %w", err)
```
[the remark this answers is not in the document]
talwinder50: done

pkg/store/did/store.go
troyronda: I would say the user/holder should have a copy of their own DID documents somewhere locally, if that is the question.

Is this store only for the user’s own DIDs?

Beyond that, It is an interesting question on what functionality belongs to vdri vs here.

Issue #1548, opened 2020-04-01
Preparing JWS proof : include Proof Type in verification data?


sudeshrshetty:
https://github.com/hyperledger/aries-framework-go/blob/55fe9fad72dcc99eb06b6dc5cf1c55daae3fe4bb/pkg/doc/signature/proof/jws.go#L104

Unable to find it in RFC:
https://tools.ietf.org/html/rfc7797
https://w3id.org/security/v2
https://github.com/digitalbazaar/jsonld-signatures

sudeshrshetty:
Refer: https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/suites/LinkedDataSignature.js#L194-L196

Type still exists in proof value

PR #1550, opened 2020-04-01
feat: JWS signing/verification - include proof.Type in verification data

### LINKED ISSUE #1549: Preparing JWS proof : include Proof Type in verification data
Include proof type in verification data while signing and verification for now, since it is causing signature verification failures while verifying signatures from other systems.

sudeshrshetty:
According to Linked Data Signatures specification for JSON-LD
("https://github.com/digitalbazaar/jsonld-signatures") proof value type
shouldn't be exluded while calculating checksum of proof value.

This problem was found while running signature verification tests with
signatures created by other systems using 'Ed25519Signature2018'.

- above specification may not apply for 'JsonWebSignature2020'.
Created issue #1548 to address this issue.

- closes #1549 

### DIFF
--- a/pkg/doc/signature/proof/jws.go
+++ b/pkg/doc/signature/proof/jws.go
@@ -101,7 +101,9 @@ func prepareJWSProof(suite signatureSuite, proofOptions map[string]interface{}) 
 
 	delete(proofOptionsCopy, jsonldJWS)
 	delete(proofOptionsCopy, jsonldProofValue)
-	delete(proofOptionsCopy, jsonldType)
+	// TODO not any where mentioned in RFCs to exclude type, this is causing issue while validating signature from
+	// other systems [Issue#1548]
+	// delete(proofOptionsCopy, jsonldType)
 
 	return suite.GetCanonicalDocument(proofOptionsCopy)
 }
--- a/pkg/doc/verifiable/credential_ldp_test.go
+++ b/pkg/doc/verifiable/credential_ldp_test.go
@@ -6,29 +6,16 @@ SPDX-License-Identifier: Apache-2.0
 package verifiable
 
 import (
-	"crypto/ecdsa"
 	"crypto/ed25519"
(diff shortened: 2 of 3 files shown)


pkg/doc/signature/proof/jws.go
```diff
@@ -101,7 +101,9 @@ func prepareJWSProof(suite signatureSuite, proofOptions map[string]interface{})
 
 	delete(proofOptionsCopy, jsonldJWS)
 	delete(proofOptionsCopy, jsonldProofValue)
-	delete(proofOptionsCopy, jsonldType)
+	// TODO not any where mentioned in RFCs to exclude type, this is causing issue while validating signature from
+	// other systems []
+	// delete(proofOptionsCopy, jsonldType)
```
Baha-sk, line 106: should add a flag to execute `delete()` here instead of commenting out this line

sudeshrshetty: Refer: https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/suites/LinkedDataSignature.js#L194-L196

We are deleting it from copy to calculate hash, flag will change the hash. Refer above code from JSON-LD specifications.

sandrask: @sudeshrshetty they must have changed spec then since I remember clearly that you needed to remove proof section before calculating JSON-LD signature

sandrask: I see, they are not excluding "proof" section they are excluding parts of proof 
delete proof.jws;
delete proof.signatureValue;
delete proof.proofValue;

kdimak: @sandrask @sudeshrshetty yes, they recently changed this in the code (`LinkedDataSignature.js`) and in the spec as well (https://w3c-ccg.github.io/ld-proofs/).

Issue #1586, opened 2020-04-08
support "proofPurpose" of linked data proof


kdimak:
Default `assertionMethod` is assigned to `proofPurpose` when a new linked data proof is added. This should be made customizable.
Also, when validating the proof, we should be able to accept `assertionMethod` input and check it against found in the proof.

PR #1588, opened 2020-04-08
fix: proofPurpose not used for credentials and presentations

### LINKED ISSUE #1578: 'proofPurpose' not used for credentials and presentations
Looks like proofPurpose is not set while creating proof for VCs and VPs.(should be default "assertionMethod").

This is causing interop issue in thridparty systems where proof digest is included (default if missing) during signature verifications.

kdimak:
closes #1578

We put default `assertionMethod` into `proofPurpose`. At the follow-up issue https://github.com/hyperledger/aries-framework-go/issues/1586 we need to support it fully.

### DIFF
--- a/pkg/doc/signature/signer/signer.go
+++ b/pkg/doc/signature/signer/signer.go
@@ -15,6 +15,8 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/proof"
 )
 
+const defaultProofPurpose = "assertionMethod"
+
 // signatureSuite encapsulates signature suite methods required for signing documents
 type signatureSuite interface {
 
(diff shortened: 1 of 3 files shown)


pkg/doc/signature/signer/signer.go
```diff
@@ -102,6 +104,7 @@ func (signer *DocumentSigner) signObject(context *Context, jsonLdObject map[stri
 		Domain:                  context.Domain,
 		Nonce:                   context.Nonce,
 		VerificationMethod:      context.VerificationMethod,
+		ProofPurpose:            defaultProofPurpose,
```
troyronda, line 107: should have TODO.

sudeshrshetty, line 107: There is also one more place where we call "NewProof()", shouldn't we use default there too?

kdimak: It's the place where we roughly speaking unmarshal `Proof`. So we should leave existent assignment:
```
ProofPurpose:            stringEntry(emap[jsonldProofPurpose]),
```

kdimak: added

Issue #1592, opened 2020-04-09
Signinig & Verification: Incorrect canonicalization of credential status 


sudeshrshetty:
During interop tests noticed an issues with canonicalization of verifiable credentials with credentials status.

Sample Doc : 

```
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "credentialSchema": [],
  "credentialStatus": {
    "id": "http://issuer.vc.rest.example.com:8070/status/1",
    "type": "CredentialStatusList2017"
  },
  "credentialSubject": {
    "degree": {
      "degree": "MIT",
      "type": "BachelorDegree"
    },
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "name": "Jayden Doe",
    "spouse": "did:example:c276e12ec21ebfeb1f712ebc6f1"
  },
  "id": "https://example.com/credentials/932236e0-966c-44cf-9342-236c0a2c77a7",
  "issuanceDate": "2020-03-16T22:37:26.544Z",
  "issuer": {
    "id": "did:elem:EiBJJPdo-ONF0jxqt8mZYEj9Z7FbdC87m2xvN0_HAbcoEg",
    "name": "alice_f94db66c-be63-4f03-af10-4205d1f625e1"
  },
  "proof": {
    "created": "2020-04-08T04:00:22Z",
    "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..6ICruO8hE3mYEyWeM01aB9iPmsGqhZg8rpNXPTofUUulYXJUJ2DOLe0uv1KJ3bR4N71h_Hp_S3YmzIUpyeoNCw",
    "proofPurpose": "assertionMethod",
    "type": "Ed25519Signature2018",
    "verificationMethod": "did:elem:EiBJJPdo-ONF0jxqt8mZYEj9Z7FbdC87m2xvN0_HAbcoEg#xqc3gS1gz1vch7R3RvNebWMjLvBOY-n_14feCYRPsUo"
  },
  "type": [
    "VerifiableCredential",
    "UniversityDegreeCredential"
  ]
}
```

**After canonicalization using JSON LD LIB** (https://github.com/digitalbazaar/jsonld-signatures/tree/master/lib)

```
[...]

PR #1594, opened 2020-04-09
fix: temporary work around for json ld canonizing issue

### LINKED ISSUE #1593: interop tests : Aries VC with credential status fails signature verification
VC with credential status created/signed in aries fails signature verification in third-party systems.

#1592  created to fix doc canonizing issue.

Exclude credential status type from json ld doc to be canonized as a temporary workaround for this issue.

sudeshrshetty:
- Excluding "credentialStatus.type" from VC which is causing document
canonizing issue (#1592)
- closes #1593


troyronda:
We should fix as noted here:
https://github.com/hyperledger/aries-framework-go/issues/1592#issuecomment-611600581

Issue #1596, opened 2020-04-09
JsonWebSignature2020 and EcdsaSecp256k1Signature2019 signature suites in VC BDD test


kdimak:
For now only `Ed25519Signature2018` is tested.

PR #1597, opened 2020-04-09
fix: assign JWK from PublicKey of DID to resolved PublicKey

### LINKED ISSUE #1595: JsonWebSignature2020 verification fails
The JWK is not set in [verifiable.PublicKey](https://github.com/hyperledger/aries-framework-go/blob/7e4eafc4b66cf509b8010bb3fccf1265c1ccbf1f/pkg/doc/verifiable/common.go#L92) in common.go and this field is used in public key verifier to match the [verifiers](https://github.com/hyperledger/aries-framework-go/blob/7e4eafc4b66cf509b8010bb3fccf1265c1ccbf1f/pkg/doc/signature/verifier/public_key_verifier.go#L79).

kdimak:
closes #1595

It should be also validated for the signature suites which use `JWK` from `verifier.PublicKey` in the VC BDD tests - https://github.com/hyperledger/aries-framework-go/issues/1596

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -107,7 +107,9 @@ type Doc struct {
 	Proof          []Proof
 }
 
-// PublicKey DID doc public key
+// PublicKey DID doc public key.
+// The value of the public key is defined either as raw public key bytes (Value field) or as JSON Web Key.
+// In the first case the Type field can hold additional information to understand the nature of the raw public key.
 type PublicKey struct {
 	ID         string
 	Type       string
@@ -116,6 +118,37 @@ type PublicKey struct {
 	Value []byte
 
 	jsonWebKey *jose.JWK
+}
+
+// NewPublicKeyFromBytes creates a new PublicKey based on raw public key bytes.
+func NewPublicKeyFromBytes(id, kType, controller string, value []byte) *PublicKey {
+	return &PublicKey{
+		ID:         id,
+		Type:       kType,
+		Controller: controller,
+		Value:      value,
+	}
+}
+
+// NewPublicKeyFromJWK creates a new PublicKey based on JSON Web Key.
+func NewPublicKeyFromJWK(id, kType, controller string, jwk *jose.JWK) (*PublicKey, error) {
+	pkBytes, err := jwk.PublicKeyBytes()
+	if err != nil {
+		return nil, fmt.Errorf("convert JWK to public key bytes: %w", err)
+	}
+
+	return &PublicKey{
+		ID:         id,
+		Type:       kType,
+		Controller: controller,
+		Value:      pkBytes,
+		jsonWebKey: jwk,
+	}, nil
+}
+
+// JSONWebKey returns JSON Web key if defined
+func (pk *PublicKey) JSONWebKey() *jose.JWK {
+	return pk.jsonWebKey
 }
 
 // Service DID doc service
@@ -555,11 +588,21 @@ func (doc *Doc) JSONBytes() ([]byte, error) {
 		context = doc.Context[0]
 	}
 
+	publicKeys, err := populateRawPublicKeys(context, doc.PublicKey)
+	if err != nil {
+		return nil, fmt.Errorf("JSON unmarshalling of Public Key failed: %w", err)
+	}
+
+	auths, err := populateRawAuthentications(context, doc.Authentication)
+	if err != nil {
+		return nil, fmt.Errorf("JSON unmarshalling of Authentication failed: %w", err)
+	}
+
 	raw := &rawDoc{
 		Context:        doc.Context,
 		ID:             doc.ID,
-		PublicKey:      populateRawPublicKeys(context, doc.PublicKey),
-		Authentication: populateRawAuthentications(context, doc.Authentication),
+		PublicKey:      publicKeys,
+		Authentication: auths,
 		Service:        populateRawServices(doc.Service),
 		Created:        doc.Created,
 		Proof:          populateRawProofs(context, doc.Proof),
@@ -641,16 +684,22 @@ func populateRawServices(services []Service) []map[string]interface{} {
 	return rawServices
 }
 
-func populateRawPublicKeys(context string, pks []PublicKey) []map[string]interface{} {
+func populateRawPublicKeys(context string, pks []PublicKey) ([]map[string]interface{}, error) {
 	var rawPKs []map[string]interface{}
+
 	for i := range pks {
-		rawPKs = append(rawPKs, populateRawPublicKey(context, &pks[i]))
+		publicKey, err := populateRawPublicKey(context, &pks[i])
+		if err != nil {
+			return nil, err
+		}
+
+		rawPKs = append(rawPKs, publicKey)
 	}
 
-	return rawPKs
+	return rawPKs, nil
 }
 
-func populateRawPublicKey(context string, pk *PublicKey) map[string]interface{} {
+func populateRawPublicKey(context string, pk *PublicKey) (map[string]interface{}, error) {
 	rawPK := make(map[string]interface{})
 	rawPK[jsonldID] = pk.ID
 	rawPK[jsonldType] = pk.Type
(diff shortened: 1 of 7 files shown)


pkg/doc/did/doc.go
```diff
@@ -433,7 +433,7 @@ func decodePublicKeyJwk(jwkMap map[string]interface{}, publicKey *PublicKey) err
 	}
 
 	publicKey.Value = pkBytes
-	publicKey.jsonWebKey = &jwk
+	publicKey.JSONWebKey = &jwk
```
rolsonquadras, line 436: This will be set only when the rawDIDDoc has "publicKeyJwk". But when we create DIDDoc struct we set value and it internally converts to "publicKeyBase58" and we would be able to get back the "publicKeyJwk".

pkg/doc/did/doc.go
```diff
@@ -115,7 +115,7 @@ type PublicKey struct {
 
 	Value []byte
 
-	jsonWebKey *jose.JWK
+	JSONWebKey *jose.JWK `json:"-"`
```
sudeshrshetty, line 118: why we need it? why can't we simply figure out key type by "Type" field and parse value to JWK?

kdimak:
1. `jsonWebKey` field of the `did.PublicKey` is hidden, `JSONWebKey` getter is added.
2. Proper marshaling of `publicKeyJwk`, checked round-trip JSON conversion in `doc_test.go`.

pkg/doc/did/doc.go
```diff
@@ -677,21 +698,33 @@ func populateRawPublicKey(context string, pk *PublicKey) map[string]interface{}
 		rawPK[jsonldController] = pk.Controller
 	}
 
-	if pk.Value != nil {
+	if pk.jsonWebKey != nil {
```
rolsonquadras, line 701: The jsonWebKey is not exposed. How can the user set this ?

kdimak: Should we give a user ability to set it? It's got defined when being unmarshalled from bytes.

rolsonquadras: How would this work with current version ? How does `pk.jsonWebKey` get set ?

kdimak: Reverted back to exposed `JSONWebKey`.

kdimak:
> As discussed, are we gonna have a test to prove we don't lost any data while doing "bytes --> did.Doc --> bytes"?

We already had it at https://github.com/hyperledger/aries-framework-go/blob/397839f21ed4dd9e54e908a4a737702b42070d0b/pkg/doc/did/doc_test.go#L784, I added a test case to check conversion of `PublicKey` with `publicKeyJwk` defined.

rolsonquadras:
> > As discussed, are we gonna have a test to prove we don't lost any data while doing "bytes --> did.Doc --> bytes"?
> 
> We already had it at
> 
> https://github.com/hyperledger/aries-framework-go/blob/397839f21ed4dd9e54e908a4a737702b42070d0b/pkg/doc/did/doc_test.go#L784
> 
> , I added a test case to check conversion of `PublicKey` with `publicKeyJwk` defined.

Do we have any test to show how to initialize JWK inside the DIDDoc struct and verify the case mentioned by @sudeshrshetty ?

pkg/doc/did/doc.go
sudeshrshetty, line 118: As discussed expose it through function keep the variable private, to maintain standards among all format

kdimak: I really don't understand it. Why should we hide it? What's the standard? It's defined in https://www.w3.org/TR/did-core/#public-keys, `publicKeyJwk` field. Why should we expose `Value` field and don't export `JSONWebKey` field? What's the rule here?

kdimak:
> > > As discussed, are we gonna have a test to prove we don't lost any data while doing "bytes --> did.Doc --> bytes"?
> > 
> > 
> > We already had it at
> > https://github.com/hyperledger/aries-framework-go/blob/397839f21ed4dd9e54e908a4a737702b42070d0b/pkg/doc/did/doc_test.go#L784
> > 
> > , I added a test case to check conversion of `PublicKey` with `publicKeyJwk` defined.
> 
> Do we have any test to show how to initialize JWK inside the DIDDoc struct and verify the case mentioned by @sudeshrshetty ?

There is no direct example of filling `JSONWebKey` inside DIDDoc, I can add the one if it's missing in your opinion. The  "bytes --> did.Doc --> bytes" test case is present, see `TestJSONConversion()`.

pkg/doc/did/doc.go
rolsonquadras: + @troyronda @fqutishat

rolsonquadras:
> > > > As discussed, are we gonna have a test to prove we don't lost any data while doing "bytes --> did.Doc --> bytes"?
> > > 
> > > 
> > > We already had it at
> > > https://github.com/hyperledger/aries-framework-go/blob/397839f21ed4dd9e54e908a4a737702b42070d0b/pkg/doc/did/doc_test.go#L784
> > > 
> > > , I added a test case to check conversion of `PublicKey` with `publicKeyJwk` defined.
> > 
> > 
> > Do we have any test to show how to initialize JWK inside the DIDDoc struct and verify the case mentioned by @sudeshrshetty ?
> 
> There is no direct example of filling `JSONWebKey` inside DIDDoc, I can add the one if it's missing in your opinion. The "bytes --> did.Doc --> bytes" test case is present, see `TestJSONConversion()`.

yes, I was asking about JSONWebKey.

troyronda:
@kdimak we should add a JWK example.

pkg/doc/did/doc.go
troyronda: btw - why aren't we using publicKeyJWK format internally instead of base58?

pkg/doc/did/doc_test.go
```diff
@@ -804,6 +806,47 @@ func TestJSONConversion(t *testing.T) {
 	}
 }
 
+func TestJSONWebKey(t *testing.T) {
```
kdimak, line 809: Here is an example with DID Doc & JSON Web Key creation. @sudeshrshetty @rolsonquadras @troyronda

pkg/doc/did/doc.go
sudeshrshetty: Some one who iterates through public key will be confused to use which one to use.

troyronda: I agree it's confusing to have both JSONWebKey and value/type exposed - these fields have similar purpose.

kdimak: Added `NewPublicKeyFromBytes()` and `NewPublicKeyFromJWK()` constructor functions. `jsonWebKey` field of `PublicKey` is hidden, added `JSONWebKey()` getter.

pkg/doc/did/doc.go
```diff
@@ -118,6 +120,31 @@ type PublicKey struct {
...
+
+// NewPublicKeyFromJWK creates a new PublicKey based on JSON Web Key.
+func NewPublicKeyFromJWK(id, kType, controller string, jwk *jose.JWK) *PublicKey {
+	return &PublicKey{
+		ID:         id,
+		Type:       kType,
+		Controller: controller,
+		jsonWebKey: jwk,
+	}
+}
```
soluchok, line 141: nit: These functions do nothing. I would suggest to make `jsonWebKey` field public and remove these functions.

kdimak: Oh man, this is how it looked before, but this can confuse, please see starting comment https://github.com/hyperledger/aries-framework-go/pull/1597#discussion_r407567036 and scroll down :)

pkg/doc/did/doc.go
```diff
@@ -118,6 +120,31 @@ type PublicKey struct {
...
+
+// NewPublicKeyFromJWK creates a new PublicKey based on JSON Web Key.
+func NewPublicKeyFromJWK(id, kType, controller string, jwk *jose.JWK) *PublicKey {
+	return &PublicKey{
+		ID:         id,
+		Type:       kType,
+		Controller: controller,
+		jsonWebKey: jwk,
```
troyronda, line 139: Do we not expose a value based on the JWK?
@kdimak @rolsonquadras @sudeshrshetty

pkg/doc/did/doc.go
troyronda: I think that the specific problem is that we Value/Type has overlap with JWK contents. But I am confused about https://github.com/hyperledger/aries-framework-go/pull/1597#r408393621

pkg/doc/did/doc.go
rolsonquadras: It's exposed through a public key method - https://github.com/hyperledger/aries-framework-go/pull/1597/files#diff-ca45993820e1e4b4338acd4e485ae699R144

troyronda: But now there is an exposed Value field that is nil for this case?

kdimak: Yes.

rolsonquadras: From the code, it looks like Value is always set and jsonWebKey is available through publicKey method for some cases - https://github.com/hyperledger/aries-framework-go/blob/47364c99607be0ebeea2395d9e7e97be0f1ed94d/pkg/doc/did/doc.go#L430-L436

kdimak: But if `did.Doc` is created manually using `NewPublicKeyFromJWK()` then `Value` is `nil`. To be consistent, we can convert `JWK` to bytes and fill in the `Value`.

kdimak: Updated `NewPublicKeyFromJWK()`. So now `PublicKey.Value` is always not nil. @rolsonquadras @troyronda

PR #1607, opened 2020-04-13
refactor: support multiple signature suites for VC and VP verification

### LINKED ISSUE #1602: VC/VP verification supports only one signature suite
Current code, supports only one verifierSuite for verification.
1. The consumer need to retrieve the proof, check the type and then need to figure out the verifierSuite. This will add extra hop at the consumer layer.
2. Also, this will not work when VC/VP has multiple proofs with different types. 

https://github.com/hyperledger/aries-framework-go/blob/c043ad2f2e922bb2605ce13b253920cd28d0cc1d/pkg/doc/verifiable/credential.go#L575

Ideally, the API should be able to take multiple verifierSuites and internally need to find teh suite based on proof type.

kdimak:
closes #1602

### DIFF
--- a/pkg/doc/signature/verifier/verifier.go
+++ b/pkg/doc/signature/verifier/verifier.go
@@ -13,8 +13,8 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/proof"
 )
 
-// signatureSuite encapsulates signature suite methods required for signature verification
-type signatureSuite interface {
+// SignatureSuite encapsulates signature suite methods required for signature verification
+type SignatureSuite interface {
 
 	// GetCanonicalDocument will return normalized/canonical version of the document
 	GetCanonicalDocument(doc map[string]interface{}) ([]byte, error)
@@ -48,14 +48,14 @@ type keyResolver interface {
 
 // DocumentVerifier implements JSON LD document proof verification
 type DocumentVerifier struct {
-	signatureSuites []signatureSuite
+	signatureSuites []SignatureSuite
 	pkResolver      keyResolver
 }
 
 // New returns new instance of document verifier
-func New(resolver keyResolver, mainSuite signatureSuite, extraSuites ...signatureSuite) *DocumentVerifier {
+func New(resolver keyResolver, mainSuite SignatureSuite, extraSuites ...SignatureSuite) *DocumentVerifier {
 	return &DocumentVerifier{
-		signatureSuites: append([]signatureSuite{mainSuite}, extraSuites...),
+		signatureSuites: append([]SignatureSuite{mainSuite}, extraSuites...),
 		pkResolver:      resolver}
 }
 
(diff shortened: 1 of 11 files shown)


pkg/doc/signature/verifier/verifier.go
```diff
@@ -48,14 +48,14 @@ type keyResolver interface {
...
 // DocumentVerifier implements JSON LD document proof verification
 type DocumentVerifier struct {
-	signatureSuites []signatureSuite
+	signatureSuites []SignatureSuite
 	pkResolver      keyResolver
 }
 
 // New returns new instance of document verifier
-func New(resolver keyResolver, mainSuite signatureSuite, extraSuites ...signatureSuite) *DocumentVerifier {
+func New(resolver keyResolver, mainSuite SignatureSuite, extraSuites ...SignatureSuite) *DocumentVerifier {
```
rolsonquadras, line 56: Why do we have distinction between mainSuite and extraSuites? And how should the consumer decide between main and extra ?

pkg/doc/verifiable/linked_data_proof_test.go
```diff
@@ -78,21 +84,91 @@ func TestLinkedDataProofSignerAndVerifier(t *testing.T) {
...
+				return &verifier.PublicKey{
+					Type:  "EcdsaSecp256k1VerificationKey2019",
+					Value: pubKeyBytes,
+					JWK: &jose.JWK{
+						JSONWebKey: gojose.JSONWebKey{
+							Algorithm: "ES256K",
+						},
+						Crv: "secp256k1",
+						Kty: "EC",
+					},
```
rolsonquadras, line 142: The [spec](https://w3c-ccg.github.io/ld-cryptosuite-registry/#ecdsa-secp256k1) doesn't specify JWK for `EcdsaSecp256k1VerificationKey2019` key type. Is this a dummy data for unit test or does the code really expect the DIDDoc to have JWK value for this key type?

pkg/doc/signature/verifier/verifier.go
kdimak: I used this approach to force a definition at least one signature suite. There is no difference between the main and extra signature suites.

pkg/doc/verifiable/linked_data_proof_test.go
kdimak: At https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/, you can find an example with the JWK. `publicKeyHex` should be supported in the `EcdsaSecp256k1Signature2019` suite, I will create an issue for that.

kdimak: https://github.com/hyperledger/aries-framework-go/issues/1609

pkg/doc/signature/verifier/verifier.go
sandrask: I agree that we shouldn't have mainSuite and extra suites. You can provide a comment saying that at least one suite is required and check for suites length in New().

troyronda: the terminology "main" and "extra" is likely to be confusing.

kdimak: @rolsonquadras @sandrask @troyronda 
Created https://github.com/hyperledger/aries-framework-go/issues/1616

PR #1629, opened 2020-04-15
feat: update did.Doc with additional fields for verification relation…

aholovko:
Update did.Doc with the following fields: assertionMethod, capabilityDelegation, capabilityInvocation and keyAgreement. These fields are used when a did:key value is expanded to a DID document.

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -97,14 +97,18 @@ func Parse(did string) (*DID, error) {
 
 // Doc DID Document definition
 type Doc struct {
-	Context        []string
-	ID             string
-	PublicKey      []PublicKey
-	Service        []Service
-	Authentication []VerificationMethod
-	Created        *time.Time
-	Updated        *time.Time
-	Proof          []Proof
+	Context              []string
+	ID                   string
+	PublicKey            []PublicKey
+	Service              []Service
+	Authentication       []VerificationMethod
+	AssertionMethod      []VerificationMethod
+	CapabilityDelegation []VerificationMethod
+	CapabilityInvocation []VerificationMethod
+	KeyAgreement         []VerificationMethod
+	Created              *time.Time
+	Updated              *time.Time
+	Proof                []Proof
 }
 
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/minio/sha256-simd v0.1.1 // indirect
 	github.com/mitchellh/mapstructure v1.1.2
+	github.com/mr-tron/base58 v1.1.3
```
troyronda, line 21: we already have a base58 import.
"github.com/btcsuite/btcutil/base58"

aholovko: fixed

pkg/doc/did/doc.go
```diff
@@ -204,32 +212,77 @@ func ParseDocument(data []byte) (*Doc, error) {
...
+
+	return doc, nil
+}
+
+func populateVerificationRelationships(doc *Doc, raw *rawDoc, pks []PublicKey) error {
+	context := doc.Context[0]
+
+	authentications, err := populateVerificationMethods(context, raw.Authentication, pks)
+	if err != nil {
+		return fmt.Errorf("populate authentications failed: %w", err)
```
kdimak, line 259: I noticed that it's a common pattern to use the "failed" keyword at error wrapping here.
But according to [Uber Go Style Guide](https://github.com/uber-go/guide/blob/master/style.md#error-wrapping), it's better to 
> When adding context to returned errors, keep the context succinct by avoiding phrases like "failed to", which state the obvious and pile up as the error percolates up through the stack

pkg/doc/did/doc_test.go
```diff
@@ -1394,3 +1483,32 @@ const docV011WithInvalidUpdated = `{
 	"updated": "2019-9-23T14:16:59.261024-04:00",
 	"id": "did:method:abc"
 }`
+
+const docV011WithVerificationRelationships = `{
```
troyronda, line 1487: Should we also additionally have a test with the latest schema?

PR #1634, opened 2020-04-16
feat: Present Proof - adds command support

### LINKED ISSUE #1633: Present Proof - need to add command support
Need to add `command` support.

soluchok:
Adds `command` support for the Present Proof protocol.

Closes #1633 

### DIFF
--- a/pkg/controller/command/presentproof/command.go
+++ b/pkg/controller/command/presentproof/command.go
@@ -1,0 +1,417 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package presentproof
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/hyperledger/aries-framework-go/pkg/client/presentproof"
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/controller/command"
+	"github.com/hyperledger/aries-framework-go/pkg/controller/internal/cmdutil"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/internal/logutil"
+)
+
+const (
+	// InvalidRequestErrorCode is typically a code for validation errors
+	// for invalid present proof controller requests
+	InvalidRequestErrorCode = command.Code(iota + command.PresentProof)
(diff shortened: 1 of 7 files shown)


pkg/controller/command/presentproof/command.go
```diff
@@ -0,0 +1,417 @@
...
+
+	// creates action channel
+	actions := make(chan service.DIDCommAction)
+	// registers action channel to listen for events
+	if err := client.RegisterActionEvent(actions); err != nil {
+		return nil, fmt.Errorf("register action event: %w", err)
+	}
+
+	// this code listens for the action events and does nothing
+	// this trick is used to avoid error such as `no clients are registered to handle the message`
```
troyronda, line 99: @rolsonquadras @llorllale

llorllale: Is this an architecture design question? Should protocol services require registered listeners for `DIDCommAction` events when handling inbound messages?

soluchok: now, I am having trouble with that design in my next PR. I suggest removing that restriction `no clients are registered to handle the message`.  We do not need a client to be registered bc we can continue at any time by using rest. @troyronda @llorllale Do you agree?

soluchok: This trick doesn't work anyway.  I forgot that we can register only one channel.

soluchok: Or is it the idea? I mean that only one consumer can be registered.

pkg/controller/command/presentproof/models.go
```diff
@@ -0,0 +1,174 @@
...
+
+import "github.com/hyperledger/aries-framework-go/pkg/client/presentproof"
+
+// DeclinePresentationArgs model
+//
+// This is used when the presentation needs to be rejected
+//
+type DeclinePresentationArgs struct {
+	// PIID protocol state machine identifier
+	PIID string `json:"piid"`
```
Baha-sk, line 17: `omitempty` is not needed here and all json definitions below?

soluchok: No, we do not need `omitempty`. I need to Unmarshal incoming data/arguments, but I do not need to Marshal them.

pkg/controller/command/presentproof/command.go
rolsonquadras: Note - DIDExchange sends action event only when clients are registered. https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/service.go#L360

Issue #1568, opened 2020-04-06
outofband: refactor introduce service to respond with oob invitation


llorllale:
part of #1454

PR #1635, opened 2020-04-16
feat: outofband events

llorllale:
Part of #1568

### DIFF
--- a/pkg/didcomm/protocol/outofband/service.go
+++ b/pkg/didcomm/protocol/outofband/service.go
@@ -30,6 +30,9 @@ const (
 	// RequestMsgType is the '@type' for the request message.
 	RequestMsgType = "https://didcomm.org/oob-request/1.0/request"
 
+	// StateRequested is one of the possible states of this protocol.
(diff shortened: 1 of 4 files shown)


pkg/didcomm/protocol/outofband/service.go
```diff
@@ -160,11 +170,37 @@ func (s *Service) HandleInbound(msg service.DIDCommMsg, myDID, theirDID string)
...
 	return "", nil
 }
 
+func sendMsgEvent(t service.StateMsgType, listeners *service.Message, msg service.DIDCommMsg, p *eventProps) {
+	var stateName string
+
+	if msg.Type() == RequestMsgType {
+		stateName = StateRequested
+	}
+
```
soluchok, line 188: I am a bit confused about that piece of code. It seems like `StateID` can be empty. Is that expected behavior?

llorllale: @soluchok this should be refactored after the oob invitation msg type is added to the service

PR #1641, opened 2020-04-16
refactor: introduce protocol to use outofband

### LINKED ISSUE #1568: outofband: refactor introduce service to respond with oob invitation

llorllale:
closes #1568

This PR refactors the `introduce` implementations to use `out-of-band` instead of `did-exchange`.

It includes these additional fixes:

* `outofband.Client.Event` not assigned a value by constructor (added an extra assertion to the `TestNew` test case)
* `DIDCommAction` properties not set by outofband.Service (added additional assertions to the "fires off an action event" test)
* `outofband.Service`: inability to handle inbound outofband.Requests over DIDComm b/c the 'OOBMessage' property is unmarshalled as a `map[string]interface{} type` (added a `TestChooseTarget` set of tests)

And also:

* temporarily disabled generic DID syntax validation in `outofband.Client.CreateRequest()` because our sidetree implementation is outputting non-compliant DID IDs (TODO: #1642 )

### DIFF
--- a/test/bdd/pkg/introduce/introduce_sdk_steps.go
+++ b/test/bdd/pkg/introduce/introduce_sdk_steps.go
@@ -13,15 +13,19 @@ import (
 	"time"
 
 	"github.com/cucumber/godog"
+	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/client/didexchange"
 	"github.com/hyperledger/aries-framework-go/pkg/client/introduce"
+	"github.com/hyperledger/aries-framework-go/pkg/client/outofband"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	introduceService "github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/introduce"
 	"github.com/hyperledger/aries-framework-go/test/bdd/agent"
 	"github.com/hyperledger/aries-framework-go/test/bdd/pkg/context"
 	bddDIDExchange "github.com/hyperledger/aries-framework-go/test/bdd/pkg/didexchange"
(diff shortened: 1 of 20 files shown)


llorllale:
I need to fix the coverage.

test/bdd/pkg/introduce/introduce_sdk_steps.go
```diff
@@ -396,3 +407,43 @@ func (a *SDKSteps) sendRequest(introducee1, introducer, introducee2 string) erro
...
+			ID:          uuid.New().String(),
+			Description: "test",
+			Data: decorator.AttachmentData{
+				JSON: map[string]interface{}{},
+			},
+		},
+		))
+	if err != nil {
+		return nil, err
+	}
```
soluchok, line 429: nit: redundant condition

test/bdd/pkg/introduce/introduce_sdk_steps.go
```diff
@@ -396,3 +407,43 @@ func (a *SDKSteps) sendRequest(introducee1, introducer, introducee2 string) erro
...
+	if err := a.outofbandSDKS.CreateClients(participants); err != nil {
+		return err
+	}
+
+	if err := a.didExchangeSDKS.RegisterPostMsgEvent(participants, "completed"); err != nil {
+		return err
+	}
+
+	return nil
+}
```
soluchok, line 449: The last error check in the function usually is redundant. I prefer the following:
`return a.didExchangeSDKS.RegisterPostMsgEvent(participants, "completed");`

test/bdd/pkg/introduce/introduce_sdk_steps.go
llorllale: I get you, but I've had experience with people in the past wanting to ensure that the result value is `nil` if a function is returning an error.

test/bdd/pkg/introduce/introduce_sdk_steps.go
llorllale: See my previous comment.

PR #1653, opened 2020-04-17
feat: add Core ECDH-ES crypto logic

### LINKED ISSUE #1470: Update created Tink Key types to return JWE components from crypto primitives

### LINKED ISSUE #1472: Replace HKDF with ConcatKDF in new ECDH-ES Tink primitive

Baha-sk:
This is a second part of the ECDH-ES KW + AEAD crypto change.
It includes the core crypto logic for key wrapping and AEAD encryption
In a subtle package. This package is not meant to be used outside of ecdhes
package. Its use will be relavant in the next change.

Key wrapping is considered complete as part of this change.

part of ##1469
closes  #1472
closes  #1470

### DIFF
--- a/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
@@ -1,0 +1,30 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package subtle
+
+// EncryptedData represents the Encryption's output data as a result of ECDHESEncrypt.Encrypt(pt, aad) call
+// The user of the primitive must unmarshal the result and build their own ECDH-ES compliant message (ie JWE msg)
+type EncryptedData struct {
+	Ciphertext []byte                 `json:"Ciphertext,omitempty"`
+	IV         []byte                 `json:"IV,omitempty"`
+	Tag        []byte                 `json:"Tag,omitempty"`
+	Recipients []*RecipientWrappedKey `json:"Recipients,omitempty"`
+}
+
+// RecipientWrappedKey contains recipient key material required to unwrap CEK
+type RecipientWrappedKey struct {
+	EncryptedCEK []byte      `json:"EncryptedCEK,omitempty"`
+	EPK          ECPublicKey `json:"EPK,omitempty"`
+	Alg          string      `json:"Alg,omitempty"`
+}
+
+// ECPublicKey mainly to exchange EPK in RecipientWrappedKey
+type ECPublicKey struct {
+	X     []byte `json:"X,omitempty"`
+	Y     []byte `json:"Y,omitempty"`
+	Curve string `json:"curve,omitempty"`
+}
--- a/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_decrypt.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_decrypt.go
@@ -1,0 +1,80 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package subtle
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/google/tink/go/subtle/hybrid"
+)
+
+// ECDHESAEADCompositeDecrypt is an instance of ECDH-ES decryption with Concat KDF
+// and AEAD content decryption
+type ECDHESAEADCompositeDecrypt struct {
+	privateKey  *hybrid.ECPrivateKey
+	pointFormat string
+	encHelper   EncrypterHelper
+}
+
+// NewECDHESAEADCompositeDecrypt returns ECDH-ES composite decryption construct with Concat KDF/ECDH-ES key unwrapping
+// and AEAD payload decryption.
+func NewECDHESAEADCompositeDecrypt(pvt *hybrid.ECPrivateKey, ptFormat string,
+	encHelper EncrypterHelper) (*ECDHESAEADCompositeDecrypt, error) {
+	return &ECDHESAEADCompositeDecrypt{
+		privateKey:  pvt,
+		pointFormat: ptFormat,
+		encHelper:   encHelper,
+	}, nil
+}
+
+// Decrypt using composite ECDH-ES with a Concat KDF key unwrap and AEAD content decryption
+func (d *ECDHESAEADCompositeDecrypt) Decrypt(ciphertext, aad []byte) ([]byte, error) {
+	if d.privateKey == nil {
+		return nil, fmt.Errorf("ECDHESAEADCompositeDecrypt: missing recipient private key for key unwrapping")
+	}
+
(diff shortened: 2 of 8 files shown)


pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package subtle
```
troyronda, line 7: why is this package called subtle?

troyronda: (subtle seems like a container folder rather than a package)

Baha-sk: Just like Tink's `subtle` packages that contains the core crypto logic for each primitive, they're used by the outer layer of crypto primitives and key templates which are exposed to the API user. 

I'm following their pattern here.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_decrypt.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, fmt.Errorf("ECDHESAEADCompositeDecrypt: missing recipient private key for key unwrapping")
+	}
+
+	keySize := d.encHelper.GetSymmetricKeySize()
+
+	var cek []byte
+
+	encData := new(EncryptedData)
+
+	err := json.Unmarshal(ciphertext, encData)
```
troyronda, line 47: so its a cipher text container?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_encrypt.go
```diff
@@ -0,0 +1,103 @@
...
+
+	var recipientsWK []*RecipientWrappedKey
+
+	for _, rec := range e.recPublicKeys {
+		senderKW := &ECDHESConcatKDFSenderKW{
+			recipientPublicKey: rec,
+			cek:                cek,
+		}
+
+		kek, err := senderKW.wrapKey(A256KWAlg, keySize)
```
troyronda, line 73: when we get to chacha, where do we do the additional handling?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_decrypt.go
Baha-sk: yes.. it's just a struct that contains crypto data to build a structured (JWE) message

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+func buildRecipientsKeys(t *testing.T, nbOfRecipients int) ([]*hybrid.ECPrivateKey, []*hybrid.ECPublicKey) {
+	t.Helper()
+
+	var (
+		recipientsPrivKeys []*hybrid.ECPrivateKey
+		recipientsPubKeys  []*hybrid.ECPublicKey
+	)
+
+	curvProto := commonpb.EllipticCurveType_NIST_P256
```
troyronda, line 177: need an additional test for 25519

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_encrypt.go
Baha-sk: As far as the [JWE RFC](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope#content-encryption-algorithms) is concerned, chacha will be used for Content encryption only, not key wrapping.

Although, if we want to support chacha key wrapping, I believe the logic [here](https://github.com/hyperledger/aries-framework-go/pull/1653/files/c2362a0548b400a4f418d37f925b8242e2afd432#diff-ee10b84659202daf624ebaf5311c1c59R41-R51) will be the same for chacha, the alg name will be different though. To best tested if we want chacha key wrapping. But chacha key wrapping is a first.

troyronda: The DIDComm WG is also allowing for chacha key wrapping.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
troyronda: So the functions in this subtle package are not intended to be used except by its child packages?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_test.go
Baha-sk: That's for signing, not encryption. 

The new primitive only deals with encryption (to support JWE serialization). 

For JWS, we can use Tink's Ed25519 signer and verifier primitives directly which are currently supported by localkms and crypto. We have a test that exports and imports Ed25519 keys and do signing/verification [here](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/kms/localkms/pubkey_export_import_test.go#L50).

troyronda: X25519

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
Baha-sk: except by its parent package, `ecdhes`

But the jwe package will reference the common fields to unmarshal the ciphertext container data.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_encrypt.go
Baha-sk: I added a TODO for this.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_test.go
Baha-sk: I added a TODO for X25519 key wrapping.

PR #1657, opened 2020-04-18
feat: Add ECDH-ES+AEAD Tink factory, key managers and templates

### LINKED ISSUE #1469: Create new Key types to support ECDH-ES primitives (using NIST approved keys)

Baha-sk:
Part 4 of ECDH-ES + AEAD logic. This change concludes the outer layer of
the new Tink primitives required to execute ECDH-ES kw and AEAD content
encryption. This layer is exposed to the user of the primitives.

This change includes:
1. New key templates
2. Primitives factory + their Tink registration
3. Key managers
4. Public key exporter/importer

closes #1469

### DIFF
--- a/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
@@ -1,0 +1,107 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package ecdhes provides implementations of payload encryption using ECDH-ES KW key wrapping with AEAD primitives.
+//
+// The functionality of ecdhes Encryption is represented as a pair of
+// primitives (interfaces):
+//
+//  * ECDHESEncrypt for encryption of data and aad for a given list of recipients keys
+//
+//  * ECDHESDecrypt for decryption of data for a certain recipient key and returning decrypted plaintext
+//
+//
+// Example:
+//
+//  package main
+//
+//  import (
+//      "bytes"
+//
+//      "github.com/google/tink/go/keyset"
+//
+//	    ecdhessubtle "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle"
+//      "github.com/aries-framework-go/pkg/crypto/tinkcrypto/composite/ecdhes"
+//  )
+//
+//  func main() {
+//      // create recipient side keyset handle
+//      recKH, err := keyset.NewHandle(ecdhes.ECDHES256KWAES256GCMKeyTemplate())
+//      if err != nil {
+//          //handle error
+//      }
+//
+//      // extract recipient public keyset handle and key
+//      recPubKH, err := recKH.Public()
+//      if err != nil {
+//          //handle error
+//      }
+//
+//      buf := new(bytes.Buffer)
+//      pubKeyWriter := ecdhes.NewWriter(buf)
+//      err = recPubKH.WriteWithNoSecrets(pubKeyWriter)
+//      if err != nil {
+//          //handle error
+//      }
+//
+//      ecPubKey := new(ecdhessubtle.ECPublicKey)
+//      err := json.Unmarshal(buf.Bytes(), ecPubKey)
+//
+//      // now create sender keyset handle with recipient public key (ecPubKey)
+//      sKH, err := keyset.NewHandle(ECDHES256KWAES256GCMKeyTemplateWithRecipients(
+//     		[]ecdhessubtle.ECPublicKey{*ecPubKey}))
+//      if err != nil {
+//          // handle error
+//      }
+//
+//      // for more recipient keys pass in a list: []ecdhessubtle.ECPublicKey{*ecPubKey1, *ecPubKey2, *ecPubKey3, etc.})
+//      // at least 1 recipient is required.
+//
+//      // extract sender public keyset handle to encrypt
+//      senderPubKH, err := sKH.Public()
+//      if err != nil {
+//          //handle error
+//      }
+//
+//      e := ecdhes.NewECDHESEncrypt(senderPubKH)
+//
+//      ct, err = e.Encrypt([]byte("secret message"), []byte("some aad"))
+//      if err != nil {
+//          // handle error
+//      }
+//
+//      // get a handle on the decryption key material for a recipient
+//      // this is usually reloading the recipient's keyset handle (ie: `recKH` above) from a kms
+//      refRecKH , err := keyset.NewHandle( .....reference/rebuild `recKH` here...);
+//      d := ecdhes.NewECDHESDecrypt(refRecKH)
+//
+//      pt, err := d.Decrypt(ct)
+//      if err != nil {
+//          // handle error
+//      }
+//  }
+package ecdhes
+
+import (
+	"fmt"
+
+	"github.com/google/tink/go/core/registry"
+)
+
+// TODO - find a better way to setup tink than init.
+// nolint: gochecknoinits
+func init() {
+	// TODO - avoid the tink registry singleton.
+	err := registry.RegisterKeyManager(newECDHESPrivateKeyManager())
+	if err != nil {
+		panic(fmt.Sprintf("ecdhes.init() failed: %v", err))
+	}
+
+	err = registry.RegisterKeyManager(newECDHESPublicKeyManager())
+	if err != nil {
+		panic(fmt.Sprintf("ecdhes.init() failed: %v", err))
+	}
+}
--- a/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_aead_private_key_manager.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_aead_private_key_manager.go
@@ -1,0 +1,188 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ecdhes
+
+import (
+	"crypto/elliptic"
+	"fmt"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/google/tink/go/core/registry"
+	"github.com/google/tink/go/keyset"
+	tinkpb "github.com/google/tink/go/proto/tink_go_proto"
+	"github.com/google/tink/go/subtle/hybrid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle"
+	ecdhespb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/ecdhes_aead_go_proto"
+)
+
+const (
+	ecdhesPrivateKeyVersion = 0
+	ecdhesPrivateKeyTypeURL = "type.hyperledger.org/hyperledger.aries.crypto.tink.EcdhesAeadPrivateKey"
+)
+
+// common errors
+var errInvalidECDHESPrivateKey = fmt.Errorf("ecdhes_private_key_manager: invalid key")
+var errInvalidECDHESPrivateKeyFormat = fmt.Errorf("ecdhes_private_key_manager: invalid key format")
+
+// ecdhesPrivateKeyManager is an implementation of PrivateKeyManager interface.
+// It generates new ECDHESPrivateKey keys and produces new instances of ECDHESAEADCompositeDecrypt subtle.
+type ecdhesPrivateKeyManager struct{}
+
+// Assert that ecdhesPrivateKeyManager implements the PrivateKeyManager interface.
+var _ registry.PrivateKeyManager = (*ecdhesPrivateKeyManager)(nil)
+
+// newECDHESPrivateKeyManager creates a new ecdhesPrivateKeyManager.
+func newECDHESPrivateKeyManager() *ecdhesPrivateKeyManager {
+	return new(ecdhesPrivateKeyManager)
+}
+
+// Primitive creates an ECDHESPrivateKey subtle for the given serialized ECDHESPrivateKey proto.
+func (km *ecdhesPrivateKeyManager) Primitive(serializedKey []byte) (interface{}, error) {
+	if len(serializedKey) == 0 {
+		return nil, errInvalidECDHESPrivateKey
+	}
+
+	key := new(ecdhespb.EcdhesAeadPrivateKey)
+
+	err := proto.Unmarshal(serializedKey, key)
+	if err != nil {
+		return nil, errInvalidECDHESPrivateKey
+	}
+
+	curve, err := km.validateKey(key)
+	if err != nil {
+		return nil, errInvalidECDHESPrivateKey
+	}
+
+	pvt := hybrid.GetECPrivateKey(curve, key.KeyValue)
+
+	rEnc, err := newRegisterECDHESAEADEncHelper(key.PublicKey.Params.EncParams.AeadEnc)
+	if err != nil {
+		return nil, err
+	}
+
+	ptFormat := key.PublicKey.Params.EcPointFormat.String()
+
+	return subtle.NewECDHESAEADCompositeDecrypt(pvt, ptFormat, rEnc)
+}
+
+// NewKey creates a new key according to the specification of ECDHESPrivateKey format.
+func (km *ecdhesPrivateKeyManager) NewKey(serializedKeyFormat []byte) (proto.Message, error) {
+	if len(serializedKeyFormat) == 0 {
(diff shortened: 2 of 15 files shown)


pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_common.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"fmt"
+
+	commonpb "github.com/google/tink/go/proto/common_go_proto"
+)
+
+// GetCurveType is a utility function that converts a string EC curve name into an EC curve proto type
+func GetCurveType(curve string) (commonpb.EllipticCurveType, error) {
+	switch curve {
+	case "secp256r1", "NIST_P256", "P-256", "EllipticCurveType_NIST_P256":
```
troyronda, line 18: also need 25519

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
```diff
@@ -0,0 +1,94 @@
...
+	tinkpb "github.com/google/tink/go/proto/tink_go_proto"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle"
+	ecdhespb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/ecdhes_aead_go_proto"
+)
+
+// ECDHES256AES256GCMKeyTemplate is a KeyTemplate that generates an ECDH-ES P-256 key wrapping and AES256-GCM CEK with
+// the following parameters:
+//  - Key Wrapping: ECDH-ES over NIST P-256
+//  - Content Encryption: AES256-GCM
```
troyronda, line 22: also need 25519

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
```diff
@@ -0,0 +1,104 @@
...
+package ecdhes
+
+import (
+	"fmt"
+
+	"github.com/google/tink/go/core/registry"
+)
+
+// nolint: gochecknoinits
+func init() {
```
troyronda, line 94: TODO - find a better way to setup tink than init.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
```diff
@@ -0,0 +1,104 @@
...
+	err := registry.RegisterKeyManager(newECDHESPrivateKeyManager())
```
troyronda, line 95: TODO: avoid the tink registry singleton.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_aead_private_key_manager.go
```diff
@@ -0,0 +1,188 @@
...
+var errInvalidECDHESPrivateKeyFormat = fmt.Errorf("ecdhes_private_key_manager: invalid key format")
+
+// ecdhesPrivateKeyManager is an implementation of PrivateKeyManager interface.
+// It generates new ECDHESPrivateKey keys and produces new instances of ECDHESAEADCompositeDecrypt subtle.
+type ecdhesPrivateKeyManager struct{}
+
+// Assert that ecdhesPrivateKeyManager implements the PrivateKeyManager interface.
+var _ registry.PrivateKeyManager = (*ecdhesPrivateKeyManager)(nil)
+
+// newECDHESPrivateKeyManager creates a new aesGcmKeyManager.
```
troyronda, line 39: aesGcmKeyManager?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
```diff
@@ -0,0 +1,119 @@
...
+func getMarshalledECDHESKey(pubKey []byte) ([]byte, error) {
+	ecPubKey := new(ecdhessubtle.ECPublicKey)
+
+	err := json.Unmarshal(pubKey, ecPubKey)
+	if err != nil {
+		return nil, err
+	}
+
+	switch ecPubKey.Curve {
+	case "secp256r1", "NIST_P256", "P-256", "EllipticCurveType_NIST_P256":
```
troyronda, line 88: also need 25519

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
```diff
@@ -0,0 +1,119 @@
...
+	default:
+		return nil, fmt.Errorf("ECDHES supports AES256GCM content encryption only")
```
troyronda, line 90: whats AESGCM have to do with the case statement?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
```diff
@@ -0,0 +1,119 @@
...
+	if ecPubKey.X == nil || ecPubKey.Y == nil {
+		return nil, fmt.Errorf("invalid key")
+	}
+
+	curveType, err := GetCurveType(ecPubKey.Curve)
+	if err != nil {
+		return nil, fmt.Errorf("undefined curve type %w", err)
+	}
+
+	aeadEnc := aead.AES256GCMKeyTemplate()
```
troyronda, line 102: also need chacha

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/google/tink/go/aead"
+	commonpb "github.com/google/tink/go/proto/common_go_proto"
+	tinkpb "github.com/google/tink/go/proto/tink_go_proto"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle"
+	ecdhespb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/ecdhes_aead_go_proto"
+)
+
+// ECDHES256AES256GCMKeyTemplate is a KeyTemplate that generates an ECDH-ES P-256 key wrapping and AES256-GCM CEK with
```
troyronda, line 19: Why is 256 repeated twice?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
```diff
@@ -0,0 +1,94 @@
+/*
```
troyronda, line 1: what’s a comparable template in Tink that does key exchange / wrapping?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_common.go
Baha-sk: that's ed curve, not elliptic curve, will deal with this when we add chacha key support

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: will handle 25519 when we add chacha key support

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
Baha-sk: will deal with 25519 we add chacha key support

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: Because the key wrapping uses P-256 keys and content encryption uses AES256GCM primitive. 

We can add P-128 key wrapping and AES256GCM too where the template would be `ECDHES128AES256GCMKeyTemplate`

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: hybrid's [ECIESHKDFAES128CTRHMACSHA256KeyTemplate](https://github.com/google/tink/blob/master/go/hybrid/hybrid_key_templates.go#L45)

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: there's a todo at line 65 below for this

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
troyronda: Are you missing the "P"? "256" is ambiguous when you mean `P-256` / `secp256r1`.

What's `P-128` key wrapping mean?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
Baha-sk: This can only be done if Tink removes their singletons. The new keys templates created in this patch are 100% Tink supported and must follow their pattern. 

Another approach would be to call these registery functions in Aries.New() using a once.Do(), let me know if it makes sense and I'll change this code.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: I think ECDH is for elliptic curves only, the `P` is assumed here. 

> What's `P-128` key wrapping mean? 
I meant AES128GCM primitive.
Go Jose calls the ECDH-ES key wrapping algorithms this way:
```
ECDH_ES_A128KW     = KeyAlgorithm("ECDH-ES+A128KW")     // ECDH-ES + AES key wrap (128)
	ECDH_ES_A192KW     = KeyAlgorithm("ECDH-ES+A192KW")     // ECDH-ES + AES key wrap (192)
	ECDH_ES_A256KW     = KeyAlgorithm("ECDH-ES+A256KW")     // ECDH-ES + AES key wrap (256)
```

I think the AES key wrap is unnecessary, ECDH-ES is sufficient (ie key derivation only). In other words, this call in [wrap()](https://github.com/hyperledger/aries-framework-go/blob/1c70c234ef1f0b0edfe779394292c9208ccaddee/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_concatkdf_sender_kw.go#L48) function:
`wk, err := josecipher.KeyWrap(block, s.cek)`
and this call in [unwrap()](https://github.com/hyperledger/aries-framework-go/blob/1c70c234ef1f0b0edfe779394292c9208ccaddee/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_concatkdf_recipient_kuw.go#L59) function:
`return josecipher.KeyUnwrap(block, recWK.EncryptedCEK)` 
should be replaced by `block.Encrypt(s.cek, nil)` and `block.Decrypt(recK.EncryptedCEK, nil)` respectively.

This means I need to change the default key wrapping algorithm [ECDH-ES+A256KW](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_concatkdf_sender_kw.go#L19) to just `ECDH-ES`, but as far as Go Jose is concerned, there is no key wrapping with `ECDH-ES` as mentioned [here](https://github.com/square/go-jose/blob/master/asymmetric.go#L339). To follow Go Jose, I've used `ECDH-ES+A256KW` as the only key wrapping algorithm in our new Tink primitives. 

But we should probably do only key derivation and cek encryption/decryption with the derived kek without calling josesipher.KeyWrap()/KeyUnrwap(). What do you think @troyronda ?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_aead_private_key_manager.go
Baha-sk: Updated. Thanks for catching this!

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
Baha-sk: Updated error message.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
Baha-sk: I've been thinking of the other approach, I don't think we can move these calls to Aries.New() as all unit tests will fail unless we explicitly call Aries.New() in all the tests requiring Tink keys.

I think for now, we will have to keep this code as is until Tink changes their singleton pattern.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
troyronda: P is not assumed. there is also x25519, for example.

Baha-sk: added P in the template function name

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_common.go
Baha-sk: will be done as part of #1657

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: will be done as part of #1657

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
Baha-sk: will be done as part of #1657

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_import.go
Baha-sk: will be done as part of #1657

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes_key_template.go
Baha-sk: There's been some lint errors.. so for now I can't add a P... but 25519 key templates will be different.

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
Baha-sk: Added TODO

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
Baha-sk: Added TODO

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/ecdhes.go
```diff
@@ -0,0 +1,107 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package ecdhes provides implementations of payload encryption using ECDH-ES KW key wrapping with AEAD primitives.
+//
+// The functionality of ecdhes Encryption is represented as a pair of
+// primitives (interfaces):
+//
+//  * ECDHESEncrypt for encryption of data and aad for a given list of recipients keys
```
troyronda, line 12: To double check, will this solutions support recipients that each have a different key type (e.g.,  recipient 1 having X25519, recipient 2 having P256, recipient 3 having P384).

Baha-sk: They can be different types as long as they support the same primitive (CompositeEncrypt/Decrypt interfaces), theoretically. This is how Tink works, as long as the key supports the primitive, it can be used. 

But for key wrapping, we will have to make sure we can support different key types. It's to be discovered when we introduced more key templates to the primitive.

PR #1679, opened 2020-04-23
fix: ECDSA P256 signature verification using tinkcrypto and localkms

### LINKED ISSUE #1534: ECDSA P256 signature verification using tinkcrypto and localkms
At https://github.com/hyperledger/aries-framework-go/issues/1532, it was discovered that ECDSA P256 signature verification fails when using `local kms` to convert public key bytes to key handle and `tink crypto` to verify the signature based on that key handle. 
In contrast, when using `crypto/ecdsa` and `crypto/elliptic` packages directly to verify the signature, it works well (see `PublicKeyVerifierP256`).

To reproduce the error, go to `TestNewCredentialFromLinkedDataProof_JsonWebSignature2020_ecdsaP256` test and change `&jsonwebsignature2020.PublicKeyVerifierP256{}` to `suite.NewCryptoVerifier(createLocalCrypto())`.

kdimak:
closes #1534

### DIFF
--- a/pkg/doc/verifiable/credential_ldp_test.go
+++ b/pkg/doc/verifiable/credential_ldp_test.go
@@ -113,7 +113,7 @@ func TestNewCredentialFromLinkedDataProof_Ed25519Signature2018_Transmute(t *test
 
 	vcWithLdp, _, err := NewCredential([]byte(vcFromTransmute),
 		WithEmbeddedSignatureSuites(sigSuite),
-		WithPublicKeyFetcher(SingleKey(pubKeyBytes, kms.ED25519)))
+		WithPublicKeyFetcher(SingleKey(pubKeyBytes, "Ed25519Signature2018")))
 	r.NoError(err)
 	r.NotNil(t, vcWithLdp)
 }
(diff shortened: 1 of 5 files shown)


pkg/doc/verifiable/credential_ldp_test.go
```diff
@@ -113,7 +113,7 @@ func TestNewCredentialFromLinkedDataProof_Ed25519Signature2018_Transmute(t *test
 
 	vcWithLdp, _, err := NewCredential([]byte(vcFromTransmute),
 		WithEmbeddedSignatureSuites(sigSuite),
-		WithPublicKeyFetcher(SingleKey(pubKeyBytes, kms.ED25519)))
+		WithPublicKeyFetcher(SingleKey(pubKeyBytes, "Ed25519Signature2018")))
```
troyronda, line 116: I would make "Ed25519Signature2018" a const in the test file.

pkg/kms/localkms/localkms.go
```diff
@@ -158,6 +163,44 @@ func getKeyTemplate(keyType kms.KeyType) (*tinkpb.KeyTemplate, error) {
 	}
 }
 
+func ecdsaP256KeyWithoutPrefixTemplate() *tinkpb.KeyTemplate {
+	return createECDSAKeyTemplate(commonpb.HashType_SHA256,
+		commonpb.EllipticCurveType_NIST_P256,
+		ecdsapb.EcdsaSignatureEncoding_IEEE_P1363,
```
Baha-sk, line 169: for those interested,

new key templates with `ecdsapb.EcdsaSignatureEncoding_IEEE_P1363` encoding format is needed to support exporting marshalled elliptic curve keys since the default Tink templates use `ecdsapb.EcdsaSignatureEncoding_DER` format. This encoding change is needed to allow DID verification using public signing keys imported to Tink keyset.Handle instances.

kdimak: Thank you for the explanation, @Baha-sk !

pkg/doc/verifiable/credential_ldp_test.go
kdimak: Ah, I just found that @fqutishat already added the names of the signature suites as constants. I will use them.

Issue #1663, opened 2020-04-21
Marshal all verification methods of DID Doc


kdimak:
Now only `Authentication` is marshaled.

Issue #1646, opened 2020-04-17
inline public key not found in did


troyronda:
see https://github.com/trustbloc/edge-service/issues/286

PR #1685, opened 2020-04-23
feat: Support relative DID URLs in verification methods

### LINKED ISSUE #1662: Support relative DID URLs
https://www.w3.org/TR/did-core/#relative-did-urls

kdimak:
closes #1662, #1646, #1663

1. Relative DID URLs in Verification Method-s are supported.
2. Marshal all verification methods of DID Doc.

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -195,6 +195,7 @@ type VerificationMethod struct {
 	PublicKey    PublicKey
 	Relationship VerificationRelationship
 	Embedded     bool
+	RelativeURL  bool
 }
 
 type rawDoc struct {
@@ -255,7 +256,7 @@ func ParseDocument(data []byte) (*Doc, error) {
 
 	doc.PublicKey = publicKeys
 
-	err = populateVerificationRelationships(doc, raw, publicKeys)
+	err = populateVerificationRelationships(doc, raw)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/doc/did/doc.go
```diff
@@ -400,13 +399,15 @@ func populateVerificationMethods(context string, rawVerificationMethods []interf
...
 // getVerificationMethods gets verification methods from raw data
-// TODO support relative DID URLs (https://www.w3.org/TR/did-core/#relative-did-urls)
-//  (https://github.com/hyperledger/aries-framework-go/issues/1662)
-func getVerificationMethods(context string, rawVerificationMethod interface{},
-	pks []PublicKey, relationship VerificationRelationship) ([]VerificationMethod, error) {
+func getVerificationMethods(doc *Doc, rawVerificationMethod interface{},
+	relationship VerificationRelationship) ([]VerificationMethod, error) {
+	// context, docID string
+	pks := doc.PublicKey
+	context := doc.Context[0]
```
Baha-sk, line 406: is Context[0] always going to exist? if not, consider adding check if Context is empty

kdimak: If the `Doc` is unmarshaled from bytes (using `ParseDocument()`) then it must contain at least one context, so it's safe.
But in the case of manual construction of `Doc` structure it's possible... It relies on user's discipline. :)

PR #1690, opened 2020-04-24
refactor: KMS export key templates using DER format

### LINKED ISSUE #1689: Revert updated Key templates in Kms Import/Export and use DER format for signatures/verification
Currently, many instances in the code is using non DER format during signing/verification.

Revert the KMS signing key templates in Import/Export KMS to the default ones.

Also replace all manual signatures/verifications to use DER format.

Baha-sk:
Revert updated Key templates in Kms Import/Export and use DER format for
signatures/verification.

Also replace all manual signatures/verifications to use DER format.

closes #1689

### DIFF
--- a/pkg/doc/signature/verifier/public_key_verifier.go
+++ b/pkg/doc/signature/verifier/public_key_verifier.go
@@ -7,12 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package verifier
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/asn1"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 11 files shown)


pkg/doc/signature/verifier/public_key_verifier.go
```diff
@@ -228,6 +230,39 @@ const (
...
+type ECDSASignature struct {
+	R, S *big.Int
+}
+
+// NewECDSASignature creates a new ecdsaSignature object.
+func NewECDSASignature(r, s *big.Int) *ECDSASignature {
+	return &ECDSASignature{R: r, S: s}
+}
+
+func asn1decode(b []byte) (*ECDSASignature, error) {
```
troyronda, line 244: I don't understand this change. Where does JWS say to do asn1decode?

Was this tested for interop?

pkg/doc/signature/verifier/public_key_verifier_test.go
```diff
@@ -396,34 +402,28 @@ func getRSASignature(privKey *rsa.PrivateKey, payload []byte) []byte {
...
-	if curveBits%8 > 0 {
-		keyBytes++
-	}
-
-	copyPadded := func(source []byte, size int) []byte {
-		dest := make([]byte, size)
-		copy(dest[size-len(source):], source)
-
-		return dest
+	ret, err := asn1.Marshal(*ecdsaSig)
```
troyronda, line 423: https://github.com/hyperledger/aries-framework-go/pull/1690/files#r415352248

pkg/doc/signature/verifier/public_key_verifier.go
Baha-sk: JWS does not specify what signature encoding to use @troyronda

troyronda: What library did you test this change against?

Baha-sk: Tink uses DER keys by default which uses Asn1 encoding. This works 100% with Tink.

For the old signatures to work, we should use custom Tink key templates to specific IEE_1363 signing keys. 

Tink uses DER by default and this works perfectly with public key marshalled/parsed using x509 package.

PR #1695, opened 2020-04-24
fix: Fixes js BDD test

soluchok:
Fixes js BDD test.

```
fatal error: all goroutines are asleep - deadlock!
:9876/base/public/aries-framework-go/assets/wasm_exec.js:47 
:9876/base/public/aries-framework-go/assets/wasm_exec.js:47 goroutine 1 [select (no cases)]:
:9876/base/public/aries-framework-go/assets/wasm_exec.js:47 main.main()
:9876/base/public/aries-framework-go/assets/wasm_exec.js:47     ........./hyperledger/aries-framework-go/cmd/aries-js-worker/main.go:96 +0xe
:9876 ...../aries-framework-go/assets/wasm_exec.js:47
```

### DIFF
--- a/cmd/aries-js-worker/main.go
+++ b/cmd/aries-js-worker/main.go
@@ -17,9 +17,8 @@ import (
 	"syscall/js"
 	"time"
 
-	"github.com/mitchellh/mapstructure"
-
 	"github.com/google/uuid"
+	"github.com/mitchellh/mapstructure"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/controller"
@@ -38,6 +37,7 @@ const (
 	ariesCommandPkg  = "aries"
 	ariesStartFn     = "Start"
 	ariesStopFn      = "Stop"
+	workers          = 2
 )
 
 var logger = log.New("aries-js-worker")
@@ -104,6 +104,7 @@ func takeFrom(in chan *command) func(js.Value, []js.Value) interface{} {
 
 			return nil
 		}
+
 		in <- cmd
 
 		return nil
@@ -115,6 +116,12 @@ func pipe(input chan *command, output chan *result) {
 
 	addAriesHandlers(handlers)
 
+	for w := 0; w < workers; w++ {
+		go worker(input, output, handlers)
+	}
+}
+
+func worker(input chan *command, output chan *result, handlers map[string]map[string]func(*command) *result) {
 	for c := range input {
 		if c.ID == "" {
 			logger.Warnf("aries wasm: missing ID for input: %v", c)


cmd/aries-js-worker/main.go
```diff
@@ -93,7 +94,7 @@ func main() {
 		ready <- struct{}{}
 	}
 
-	select {}
+	time.Sleep(math.MaxInt64)
```
llorllale, line 97: if it works, it works. An explanation would be awesome though...

troyronda, line 97: this doesn't look right to me.

read from a channel instead.

troyronda: try a pattern similar to the done channel (`chan struct{}`)
`<- done`

soluchok: the channel creates deadlock if you have suggestion sent to me. I can try it. I tried with the channel it did not work.

soluchok: Hope it helps https://groups.google.com/forum/#!topic/golang-nuts/rtUIY5_m51I

soluchok: @troyronda it was the first thing I did`try a pattern similar to the done channel`

llorllale: Not really.

I'm ok with trying out this change but only if you put a comment with some explanation of the problem we're trying to solve and be clear that it is MAGIC and we don't really know how it works.

llorllale: Some quick links:

* https://stackoverflow.com/a/32148983/1623885
* https://github.com/golang/go/blob/master/src/runtime/time.go#L47
* https://golang.org/src/runtime/select.go#L118
* https://golang.org/src/runtime/proc.go#L287

cmd/aries-js-worker/main.go
```diff
@@ -104,7 +103,8 @@ func takeFrom(in chan *command) func(js.Value, []js.Value) interface{} {
 
 			return nil
 		}
-		in <- cmd
+
+		go func() { in <- cmd }()
```
soluchok, line 107: @troyronda @llorllale  I did another fix. It prevents deadlock.

llorllale: This seems to make more sense. See https://github.com/golang/go/issues/34478

llorllale: @soluchok how many times did you try it? What was the failure rate?

llorllale: Also - note the channel is unbuffered.

soluchok: @llorllale In progress. It seems to work fine. For now 7/7. I am gonna run it 100 times.

troyronda, line 107: normally it would be odd to spin a goroutine on every call, but at least this stabilizes the tests until there is a better idea.

soluchok: It can also be fixed with a buffered channel or worker pool. I guess that the problem is with the `pipe` function. I can try with the worker pool.

soluchok: @troyronda Done! I think this is the best we can do.

cmd/aries-js-worker/main.go
```diff
@@ -38,6 +37,7 @@ const (
 	ariesCommandPkg  = "aries"
 	ariesStartFn     = "Start"
 	ariesStopFn      = "Stop"
+	workers          = 9
```
troyronda, line 40: Why 9?

troyronda: eg., I think JS in worker only has one thread.

soluchok: `why 9?` just in case :) I can set it to `3`

fqutishat: why 3

troyronda: i assume 3 is the number of workers that rids us of the problem? if so, great.

soluchok: let me check it

troyronda: Somehow I thought it might be 2 :)

soluchok: @troyronda works with `2` as well

PR #1699, opened 2020-04-26
fix: Sign Verifiable Presentation bdd test bug

talwinder50:
Noticed a bug in 
`  await healthCheck(`${environment.HTTP_LOCAL_RESOLVER_URL}/` + did.id, 5000, "resolve did timeout!")`

This statement was added to check the availability of resolvable did.  But doesn't seem to work. 
There fore removed and added a loop to test sidetree did is available to resolve before signing presentation.


troyronda:
@talwinder50 this is a "fix:" rather than a "feat:". it would be helpful if you could update your commit description.

talwinder50:
> @talwinder50 this is a "fix:" rather than a "feat:". it would be helpful if you could update your commit description (similar to the updated PR title).

Done

test/aries-js-worker/test/verifiable/verifiable.js
```diff
@@ -133,6 +133,25 @@ async function verifiableStore(newAries, mode = wasmMode) {
         )
     })
 
+    it(modePrefix + "Alice makes sure that the DID is resolvable", async function () {
+        let success = false
+        for (var i = 0; i < 10; i++) {
```
soluchok, line 138: nit: there is a variable for  retries   `let retries = 10;`. Please use it in your test, and also fix it in the original test (from where you copied it).

test/aries-js-worker/test/verifiable/verifiable.js
```diff
@@ -133,6 +133,25 @@ async function verifiableStore(newAries, mode = wasmMode) {
...
+                })
+                .catch(function(error) {
+                    console.log('will try to resolve again : attempt=',i+1);
+                });
+
+            if (success) {
+                break
+            }
+
+            await new Promise(r => setTimeout(r, 2000));
```
llorllale, line 151: can we sleep for less time? Our build's duration has gotten longer and longer...

troyronda: we should be looping around smaller timeouts.

test/aries-js-worker/test/verifiable/verifiable.js
talwinder50: done

PR #1715, opened 2020-04-27
feat: JWE deserialize function

### LINKED ISSUE #1507: JWE deserialize function
The new JWE type in the jose package needs a deserialize function. It should be able to handle both compact and full JWE. It will deserialize into a rawJSONWebEncryption, and then manipulate the data into a JSONWebEncryption. Follow a similar pattern to the jose ParseEncrypted function.

DRK3:

closes #1507

### DIFF
--- a/pkg/doc/jose/jwe.go
+++ b/pkg/doc/jose/jwe.go
@@ -4,13 +4,21 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
+// General serialization/deserialization design based off of https://github.com/square/go-jose/blob/master/jwe.go.
+
 package jose
 
 import (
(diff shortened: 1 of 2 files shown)


DRK3:
> I'd suggest adding tests which serialize then deserialize and verify the data is the same, and tests which go vice-versa also (deserialize a JWE then serialize and verify it's the same).

I do have a test that deserializes a JWE and then serializes to verify that it's the same in https://github.com/hyperledger/aries-framework-go/pull/1715/files#diff-cae4e7ad4995f3dbf6d2cff8dcb81dac

Going the other way is a bit trickier since I'd have to manually check every field... or serialize it again and compare as a string... what do you think?

DRK3:
I'm going to add some jose interop tests

troyronda:
@DRK3 please add the interop tests.

PR #1723, opened 2020-04-28
feat: add JWE Decrypt support

### LINKED ISSUE #987: Add Default implementation for JWE Anoncrypt SPI
There should be a default implementation of JWE Anoncrypt envelope as described in [RFC-0334 JWE Envelope](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope)

required work:
- [x] Create new Key types to support ECDH-ES primitives (using NIST approved keys) #1469
- [x] Replace HKDF with ConcatKDF in new ECDH-ES Tink primitive #1472
- [x] Update created Tink Key types to return JWE components from crypto primitives #1470
- [ ] Add JWE build/parse support with new Tink key type created above #1638

### LINKED ISSUE #1638: Add JWE message build support with new Tink Crypto/Key templates API
When the new ECDHES key templates are available, create a JWE builder that will use them to build a JWE message and parse it.

Baha-sk:
This change completes adding JWE build/parse support with new Tink key type.
With this change, it can be considered that Default JWE Anoncrypt using
NIST curve is complete.

closes #1638
closes #987

### DIFF
--- a/pkg/doc/jose/decrypter.go
+++ b/pkg/doc/jose/decrypter.go
@@ -1,0 +1,133 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package jose
+
+import (
+	"crypto/ecdsa"
+	"encoding/json"
+	"fmt"
+
+	"github.com/google/tink/go/keyset"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/api"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/doc/jose/decrypter.go
```diff
@@ -0,0 +1,139 @@
...
+			return nil, err
+		}
+
+		rec.Alg = recJWE.Header.Alg
+		rec.EncryptedCEK = []byte(recJWE.EncryptedKey)
+
+		recipients = append(recipients, rec)
+	}
+
+	var ok bool
```
DRK3, line 100: Consider moving the var declaration closer to where it's used (right above)?

troyronda:
Needs followup with more extensive testing against go-jose.

pkg/doc/jose/decrypter.go
troyronda: also looking at the code, it could probably be implicit with `:=`

Baha-sk: `ok` variable has been removed

Baha-sk:
> Needs followup with more extensive testing against go-jose.

created #1726 for this task

PR #1727, opened 2020-04-28
test: JWE deserialize interop tests

DRK3:
Also includes some minor spec conformity fixes and an update to the JWE serialize method to output flattened syntax when there's only one recipient in order to match go-jose's JWE serialize behaviour.

### DIFF
--- a/pkg/doc/jose/jwe.go
+++ b/pkg/doc/jose/jwe.go
@@ -33,8 +33,8 @@ type JSONWebEncryption struct {
 
 // Recipient is a recipient of a JWE including the shared encryption key
 type Recipient struct {
-	EncryptedKey string           `json:"encrypted_key,omitempty"`
-	Header       RecipientHeaders `json:"header,omitempty"`
+	Header       *RecipientHeaders `json:"header,omitempty"`
+	EncryptedKey string            `json:"encrypted_key,omitempty"`
 }
 
 // RecipientHeaders are the recipient headers
@@ -52,9 +52,9 @@ type RecipientHeaders struct {
 type rawJSONWebEncryption struct {
 	B64ProtectedHeaders      string          `json:"protected,omitempty"`
 	UnprotectedHeaders       json.RawMessage `json:"unprotected,omitempty"`
-	Recipients               json.RawMessage `json:"recipients,omitempty"` //
+	Recipients               json.RawMessage `json:"recipients,omitempty"`
 	B64SingleRecipientEncKey string          `json:"encrypted_key,omitempty"`
-	SingleRecipientHeader    string          `json:"header,omitempty"`
+	SingleRecipientHeader    json.RawMessage `json:"header,omitempty"`
 	B64AAD                   string          `json:"aad,omitempty"`
 	B64IV                    string          `json:"iv,omitempty"`
 	B64Ciphertext            string          `json:"ciphertext,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/doc/jose/jwe_test.go
```diff
@@ -12,323 +12,454 @@ import (
...
+		`q7YN7e9jwQRb23nfa6c9d-StnImGyFDbSv04uVuxIp5Zms1gNxKKK2Da14B8S4rzVRltdYwam_lDp5XnZAYpQdb76FdIKLaVmqgfw` +
+		`X7XWRxv2322i-vDxRfqNzo_tETKzpVLzfiwQyeyPGLBIO56YJ7eObdv0je81860ppamavo35UgoRdbYaBcoh9QcfylQr66oc6vFWX` +
+		`RcZ_ZT2LawVCWTIy3brGPi6UklfCpIMfIjf7iGdXKHzg","iv":"48V1_ALb6US04U3b","ciphertext":"5eym8TW_c8SuK0ltJ` +
+		`3rpYIzOeDQz7TALvtu6UG9oMo4vpzs9tX_EFShS8iB7j6jiSdiwkIr3ajwQzaBtQD_A","tag":"XFBoMYUZodetZdvTiFvSkQ"}`
 )
 
 var errFailingMarshal = errors.New("i failed to marshal")
 
 func TestJSONWebEncryption_Serialize(t *testing.T) {
-	t.Run("Successfully serialize JWE, all fields filled", func(t *testing.T) {
```
DRK3, line 69: There's not actually as many lines changed as it might seen - I grouped some tests together under another t.Run, hence why it looks like I've changed so much

pkg/doc/jose/jwe.go
```diff
@@ -64,28 +64,16 @@ type rawJSONWebEncryption struct {
 type marshalFunc func(interface{}) ([]byte, error)
 
 // Serialize serializes the given JWE into JSON as defined in https://tools.ietf.org/html/rfc7516#section-7.2.
+// The full serialization syntax is used. If there is only one recipient, then the flattened syntax is used.
```
troyronda, line 67: why flattened instead of compact?

pkg/doc/jose/jwe.go
```diff
@@ -253,16 +286,14 @@ func deserializeAndDecodeHeaders(rawJWE *rawJSONWebEncryption) (*Headers, *Heade
 func deserializeRecipients(rawJWE *rawJSONWebEncryption) ([]*Recipient, error) {
 	var recipients []*Recipient
 
-	// If no recipients field, then we must be deserializing JWE with the flattened syntax as defined in
+	// If there is no recipients field, then we must be deserializing JWE with the flattened syntax as defined in
```
troyronda, line 289: https://github.com/hyperledger/aries-framework-go/pull/1727/files#r416945630

pkg/doc/jose/jwe.go
Baha-sk: compact is not a JSON format, while full and flattened are in JSON

DRK3: Two reasons...
1. Using compact instead of flattened isn't a lossless operation - it would discard the unprotected header and AAD values
2. To match what go-jose does. Go-jose's full serialization method automatically uses flattened syntax when there's only one recipient.

troyronda: we should have a followup to also support compact.

PR #1731, opened 2020-04-29
fix: JWE EPK and SPK deserialize error

DRK3:

### DIFF
--- a/pkg/doc/jose/jwe_test.go
+++ b/pkg/doc/jose/jwe_test.go
@@ -18,29 +18,32 @@ import (
 )
 
 const (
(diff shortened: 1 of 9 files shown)


pkg/doc/jose/jwe_test.go
```diff
@@ -105,7 +134,7 @@ func TestJSONWebEncryption_Serialize(t *testing.T) {
 					IV:  "TestIV",
 					Tag: "TestTag",
 					KID: "TestKID",
-					SPK: "TestSPK",
+					SPK: []byte(exampleSPK),
```
Baha-sk, line 137: Please use `EPK` instead of `SPK` which is not JWE/JWS compliant. `EPK` is compliant.

DRK3: Done

Issue #1705, opened 2020-04-26
VC/VP verification of context vocabulary


troyronda:
We need the ability to verify that only context vocabulary is present in a VC, VP, LD Signature. i.e., how do we validate that a JSON-LD document only contains defined vocabulary?

Example Scenarios:
- “newProp”: “bar” added to VC or VP (where newProp is not in the contexts).
- “newProp”: “bar” added to Proof (where newProp is not in the contexts).
- An existing property is set to a value not in context (see https://github.com/hyperledger/aries-framework-go/pull/1698).

note: we should retain the _option_ to ignore missing context vocabulary problems (in general JSON-LD document processing).

troyronda:
@kdimak i assume there is another PR coming for VPs?

PR #1743, opened 2020-04-30
feat: extra JSON-LD contexts and custom JSON-LD doc loaders

kdimak:

#1705

Added an ability to add supplement contexts on VC/VP parsing (in particular, JSON-LD context vocabulary extension) and custom JSON-LD document loaders (to allow in-memory contexts).


pkg/doc/verifiable/credential_ldp_test.go
```diff
@@ -208,6 +211,191 @@ func TestNewCredentialFromLinkedDataProof_JSONLD_Validation(t *testing.T) {
...
+	vcWithLdp, _, err = NewCredential(vcBytes,
+		WithEmbeddedSignatureSuites(sigSuite),
+		WithPublicKeyFetcher(SingleKey(pubKey, kms.ED25519)),
+		WithExtraJSONLDContexts("https://trustbloc.github.io/context/vc/examples-v1.jsonld"),
+		WithStrictValidation())
+	r.NoError(err)
+	r.NotNil(vcWithLdp)
+
+	// Use extra context.
+	vcWithLdp, _, err = NewCredential(vcBytes,
```
kdimak, line 359: An example of using extra remote context https://trustbloc.github.io/context/vc/examples-v1.jsonld to use custom fields like credential status type `CredentialStatusList2017`  to be used in Linked Data Signature verification data and also pass strict VC validation.

pkg/doc/verifiable/credential_ldp_test.go
```diff
@@ -208,6 +211,191 @@ func TestNewCredentialFromLinkedDataProof_JSONLD_Validation(t *testing.T) {
...
+		WithEmbeddedSignatureSuites(sigSuite),
+		WithPublicKeyFetcher(SingleKey(pubKey, kms.ED25519)),
+		WithExtraJSONLDContexts("https://trustbloc.github.io/context/vc/examples-v1.jsonld"),
+		WithStrictValidation())
+	r.NoError(err)
+	r.NotNil(vcWithLdp)
+
+	// Use extra in-memory context.
+	reader, err := ld.DocumentFromReader(strings.NewReader(`
```
kdimak, line 368: The most interesting part - use in-memory context.

pkg/doc/signature/jsonld/processor.go
```diff
@@ -26,23 +26,31 @@ var logger = log.New("aries-framework/json-ld-processor")
...
 // WithRemoveAllInvalidRDF option for removing all invalid RDF dataset from normalize document
-func WithRemoveAllInvalidRDF() CanonicalizationOpts {
+func WithRemoveAllInvalidRDF() ProcessorOpts {
 	return func(opts *normalizeOpts) {
 		opts.removeInvalidRDF = true
 	}
 }
 
-// WithRemoveInvalidRDF option for removing RDFs containing any one of the given matches
-func WithRemoveInvalidRDF(matches ...string) CanonicalizationOpts {
```
kdimak, line 43: `WithRemoveInvalidRDF()` is dropped. `WithRemoveAllInvalidRDF()` is used by default. If any element like `CredentialStatusList2017` should be included into canonical document, it should be explicitly supported by extra JSON-LD context (with the help of `WithExtraContext()` and optionally `WithDocumentLoader()`).

pkg/doc/verifiable/credential_ldp_test.go
```diff
@@ -208,6 +211,191 @@ func TestNewCredentialFromLinkedDataProof_JSONLD_Validation(t *testing.T) {
...
+`))
+	r.NoError(err)
+
+	loader := CachingJSONLDLoader()
+	loader.AddDocument("http://localhost:8652/dummy.jsonld", reader)
+
+	vcWithLdp, _, err = NewCredential(vcBytes,
+		WithEmbeddedSignatureSuites(sigSuite),
+		WithPublicKeyFetcher(SingleKey(pubKey, kms.ED25519)),
+		WithExtraJSONLDContexts("http://localhost:8652/dummy.jsonld"),
```
kdimak, line 392: I think we can add another option for in-memory context, e.g. `WithExtraInMemoryJSONLDContext()`, where we will pass just in-memory context (and won't pass this dummy context URL like `http://localhost:8652/dummy.jsonld`). Under the hood, it will generate this URL (with e.g. UUID identifier).

pkg/doc/signature/verifier/verifier.go
```diff
@@ -83,19 +67,19 @@ func New(resolver keyResolver, suites ...SignatureSuite) (*DocumentVerifier, err
...
+func (dv *DocumentVerifier) Verify(jsonLdDoc []byte, opts ...jsonld.ProcessorOpts) error {
 	var jsonLdObject map[string]interface{}
 
 	err := json.Unmarshal(jsonLdDoc, &jsonLdObject)
 	if err != nil {
 		return fmt.Errorf("failed to unmarshal json ld document: %w", err)
 	}
 
-	return dv.verifyObject(jsonLdObject)
+	return dv.verifyObject(jsonLdObject, append(opts, jsonld.WithRemoveAllInvalidRDF()))
```
sudeshrshetty, line 78: May be I am missing something here, shouldn't we keeping invalid RDFs as it is so that verifier should fail when it encounters one.

-> signer should remove invalid RDFs before signing
-> verifier should verify as it is.

pkg/doc/signature/jsonld/processor.go
```diff
@@ -26,23 +26,31 @@ var logger = log.New("aries-framework/json-ld-processor")
...
+// WithDocumentLoader option is for passing custom JSON-LD document loader.
+func WithDocumentLoader(loader ld.DocumentLoader) ProcessorOpts {
 	return func(opts *normalizeOpts) {
-		opts.removeRDFMatches = matches
+		opts.documentLoader = loader
+	}
+}
+
+// WithExtraContext option is for using extra context when doing JSON-LD operations.
+func WithExtraContext(context ...string) ProcessorOpts {
```
troyronda, line 51: I guess this is sometimes called external context: https://w3c.github.io/json-ld-syntax/#interpreting-json-as-json-ld

pkg/doc/signature/verifier/verifier.go
troyronda: Yes - that is what are trying to do.

pkg/doc/verifiable/credential_ldp_test.go
troyronda: right - this is similar to the embedded/inline context https://www.w3.org/TR/json-ld11/#example-7-in-line-context-definition

pkg/doc/verifiable/credential_ldp_test.go
```diff
@@ -208,6 +211,191 @@ func TestNewCredentialFromLinkedDataProof_JSONLD_Validation(t *testing.T) {
...
+		WithExtraJSONLDContexts("https://trustbloc.github.io/context/vc/examples-v1.jsonld"),
+		WithStrictValidation())
+	r.NoError(err)
+	r.NotNil(vcWithLdp)
+
+	// Use extra context.
+	vcWithLdp, _, err = NewCredential(vcBytes,
+		WithEmbeddedSignatureSuites(sigSuite),
+		WithPublicKeyFetcher(SingleKey(pubKey, kms.ED25519)),
+		WithExtraJSONLDContexts("https://trustbloc.github.io/context/vc/examples-v1.jsonld"),
```
troyronda, line 362: will the tests break if this file goes away?

kdimak: Yes, the same applies for `https://www.w3.org/2018/credentials/examples/v1` and `https://trustbloc.github.io/context/vc/examples-v1.jsonld` used in the tests.

troyronda: yup - we should probably have a todo/followup to cache contexts locally to remove this dependency.

pkg/doc/signature/jsonld/processor.go
kdimak: I think this is rather a hack solution: we artificially append extra contexts before doing any JSON-LD operation like compaction or normalization in order to use enlarged JSON-LD vocabularies and be capable of recognizing extra types like `name`, `description` (see https://github.com/w3c-ccg/vc-examples/issues/22), `CredentialStatusList2017`, ...
This approach should be used if we receive JSON-LD to parse and it contains out-of-vocabulary terms we still want to support (e.g. because they were used to build a signature verification data).

pkg/doc/signature/verifier/verifier.go
kdimak: I will re-check it in a few hours, come back with the answer. An important question!

pkg/doc/verifiable/credential_ldp_test.go
kdimak: https://github.com/hyperledger/aries-framework-go/issues/1753

pkg/doc/signature/jsonld/processor.go
kdimak: Agree! It would be more correct.

pkg/doc/signature/verifier/verifier.go
kdimak: I changed that, now `jsonld.WithRemoveAllInvalidRDF()` is applied on demand. To facilitate VC, I added `WithJSONLDOnlyValidRDF()` option.

pkg/doc/verifiable/credential_ldp_test.go
kdimak: Filed https://github.com/hyperledger/aries-framework-go/issues/1758

pkg/doc/signature/jsonld/processor.go
```diff
@@ -26,23 +26,31 @@ var logger = log.New("aries-framework/json-ld-processor")
...
 	return func(opts *normalizeOpts) {
-		opts.removeRDFMatches = matches
+		opts.documentLoader = loader
+	}
+}
+
+// WithExternalContext option is for definition of external context when doing JSON-LD operations.
+func WithExternalContext(context ...string) ProcessorOpts {
+	return func(opts *normalizeOpts) {
+		opts.extraContexts = context
```
troyronda, line 53: for consistency, would be good to replace-all on opts.extraContexts

kdimak:
Last update:
- "JSON-LD Only Valid RDF" on demand, added corresponding VC option.
- used `external context` term.
- added `ProofPurpose` field to DID Doc's `Proof` struct, in was missing and caused failing linked data proof checks.

One thing to note: in case of `JsonWebSignature2020`, when we verify the signature we should use `WithJSONLDOnlyValidRDF()` option. That's because this type of proof is not defined in security context (`https://w3id.org/security/v2`) that is used to build canonical documents for proofs. And we should drop it.

pkg/doc/signature/jsonld/processor.go
kdimak: ah, forgot to rename that.

pkg/doc/signature/jsonld/processor.go
```diff
@@ -97,16 +113,48 @@ func (p *Processor) GetCanonicalDocument(doc map[string]interface{}, opts ...Can
 	return []byte(result), nil
 }
 
+// AppendExtraContexts appends extra context(s) to the JSON_LD context which can have one or several contexts already.
+func AppendExtraContexts(context interface{}, extraContexts ...string) []interface{} {
```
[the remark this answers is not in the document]
kdimak: sure, thanks!

troyronda:
> And we should drop it.

No - include additional context.

kdimak:
> > And we should drop it.
> 
> No - include additional context.

The side (e.g. https://github.com/transmute-industries/lds-jws2020) which prepared VC with `JsonWebSignature2020` LDP dropped this field when building proof's canonical document. And we should do the same - otherwise we will get "invalid signature" error.

troyronda:
why do you get "invalid signature" error when you include the extra context?

troyronda:
I guess we should temporarily make an exception for `JsonWebSignature2020` when the context is missing. Do we still have a whitelist option? @kdimak

kdimak:
> I guess we should temporarily make an exception for `JsonWebSignature2020` when the context is missing during validation. Do we still have a whitelist option? @kdimak

i've dropped it :)

sudeshrshetty:
May need @rolsonquadras 's approval for this

kdimak:
To make forced cleanup of invalid RDF-s in proofs case only (for both signing and verification). This way we will clean JsonWebSignature2020 not declared in https://w3id.org/security/v2
For the VC content itself - we make this cleanup on demand only  (as it was before)
this way we don’t need to apply WithJSONLDOnlyValidRDF()  option for verification of VCs with JsonWebSignature2020 LDP.

PR #1746, opened 2020-04-30
feat: generate presentation to choose "authentication" from DID

### LINKED ISSUE #1745: Generate Presentation to use DID Authentication purpose

sudeshrshetty:
- Generate presentation to use did "authentication" method
- all opts will be validated based on did "authentication" method
- removed purpose override from opts

made other fixes like,
- Contexts from Credentials shouldn't be just copied over to
presentations
- Disabled 'Context' array to string compacting temporarily to address
interop issues
- fixed a bug in did store

- Closes #1745


pkg/controller/command/verifiable/command.go
```diff
@@ -763,11 +761,43 @@ func prepareOpts(opts *ProofOptions, didDoc *did.Doc) (*ProofOptions, error) {
...
+			break
+		}
+	}
+
+	if !vmMatched {
+		return nil, fmt.Errorf("unable to find matching 'authentication' key IDs for given verification method")
+	}
+
+	// this is the fallback logic kept for DIDs not having authentication method
+	// TODO to be removed [Issue #1693]
```
troyronda, line 795: we still need this?

pkg/controller/command/verifiable/models.go
```diff
@@ -55,8 +53,8 @@ type ProofOptions struct {
 	SignatureType string `json:"signatureType,omitempty"`
 	// PrivateKey is used to sign instead of DID
 	PrivateKey string `json:"privateKey,omitempty"`
-	// DIDKeyID id of the public key associated with the private key
-	DIDKeyID string `json:"didKeyID,omitempty"`
+	// proofPurpose is purpose of the proof.
+	proofPurpose string
```
troyronda, line 57: this doesn’t need json annotation or exporting?

pkg/controller/command/verifiable/command.go
sudeshrshetty: I didn't want to break any existing features, will be removed in next PR.

pkg/controller/command/verifiable/models.go
```diff
@@ -55,8 +53,8 @@ type ProofOptions struct {
 	SignatureType string `json:"signatureType,omitempty"`
 	// PrivateKey is used to sign instead of DID
 	PrivateKey string `json:"privateKey,omitempty"`
```
troyronda, line 55: there should be a followup to remove this.

pkg/controller/command/verifiable/models.go
sudeshrshetty: No, it is always "authentication", so didn't want to confuse command users with this option

pkg/controller/command/verifiable/models.go
sudeshrshetty: created : https://github.com/hyperledger/aries-framework-go/issues/1748

pkg/controller/command/verifiable/models.go
```diff
@@ -40,9 +40,7 @@ type IDArg struct {
 // ProofOptions is model to allow the dynamic proofing options by the user.
 type ProofOptions struct {
 	// VerificationMethod is the URI of the verificationMethod used for the proof.
-	VerificationMethod string `json:"verifiableMethod,omitempty"`
-	// ProofPurpose is purpose of the proof. If omitted "assertionMethod" will be used.
-	ProofPurpose string `json:"proofPurpose,omitempty"`
+	VerificationMethod string `json:"verificationMethod,omitempty"`
 	// Created date of the proof. If omitted current system time will be used.
 	Created *time.Time `json:"created,omitempty"`
 	// Domain is operational domain of a digital proof.
```
troyronda, line 46: do we have a followup to validate that the verificationmethod can be used for the purpose?

sudeshrshetty: Now we cross check if verification method belongs to "Authentication", if not we throw error.
Added tests for the same.

pkg/controller/command/verifiable/models.go
troyronda, line 57: we might need to revisit if all scenarios always have the same `authentication` purpose in the VP.

sudeshrshetty: I am already validating if privatekey & verification methods are matching DID authentication purpose.

troyronda: right - i meant we might need to revisit the assumption that `purpose` is always `authentication.`

troyronda: i.e., `authentication` seems like a good default but the question to be revisited: is it always `authentication`?

pkg/controller/command/verifiable/command.go
```diff
@@ -763,11 +761,43 @@ func prepareOpts(opts *ProofOptions, didDoc *did.Doc) (*ProofOptions, error) {
 		opts = &ProofOptions{}
 	}
 
+	opts.proofPurpose = "authentication"
+
+	if opts.PrivateKey != "" && opts.VerificationMethod == "" {
+		return nil, fmt.Errorf("verification method matching given private key is not provided")
+	}
+
+	authVMs := didDoc.VerificationMethods(did.Authentication)[did.Authentication]
```
troyronda, line 770: shouldn’t we look up by opts.proofPurpose?

pkg/controller/command/verifiable/models.go
sudeshrshetty: created https://github.com/hyperledger/aries-framework-go/issues/1750

pkg/controller/command/verifiable/command.go
sudeshrshetty: proofPurpose is no longer part of request opts. It is not exported. We default to authentication now.

troyronda: even if you hardcode opts.proofPurpose above, it is good to be generic :)

PR #1779, opened 2020-05-04
feat: support importing private key with a specific key ID

### LINKED ISSUE #1778: support importing private key with a specific key ID
As a follow up to #1764, the private key import function should accept a keyID to use to be used for the imported key.

The function should return an error if the keyID already exists.

Baha-sk:
closes #1778

### DIFF
--- a/pkg/kms/localkms/localkms_writer.go
+++ b/pkg/kms/localkms/localkms_writer.go
@@ -8,6 +8,8 @@ package localkms
 
 import (
 	"encoding/base64"
+	"errors"
+	"fmt"
 
 	"github.com/google/tink/go/subtle/random"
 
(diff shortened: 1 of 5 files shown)


pkg/kms/localkms/localkms_writer.go
```diff
@@ -17,21 +18,82 @@ import (
...
+
+	for _, opt := range opts {
+		opt(pOpts)
+	}
+
+	reqID := ""
+	if opts != nil && pOpts.ksID != "" {
+		reqID = pOpts.ksID
+	}
+
```
soluchok, line 32: This code is redundant. Better to do like below:
```
func newWriter(kmsStore storage.Store, opts ...PrivateKeyOpts) *storeWriter {
	pOpts := &privateKeyOpts{}

	for _, opt := range opts {
		opt(pOpts)
	}

	return &storeWriter{
		storage:           kmsStore,
		requestedKeysetID: pOpts.ksID,
	}
}
```

pkg/kms/localkms/localkms_writer.go
```diff
@@ -17,21 +18,82 @@ import (
...
+	if err != nil {
+		if err != storage.ErrDataNotFound {
+			return "", fmt.Errorf("got error while verifying requested ID: %w", err)
+		}
+	}
+
+	return l.requestedKeysetID, nil
+}
```
soluchok, line 94: better to avoid nested conditions
```
func (l *storeWriter) verifyRequestedID() (string, error) {
    if len(l.requestedKeysetID) > maxKeyIDLen {
        return "", fmt.Errorf("requested ID '%s' is longer than max allowed length of %d", l.requestedKeysetID,
            maxKeyIDLen)
    }

    _, err := l.storage.Get(l.requestedKeysetID)
        return l.requestedKeysetID, nil
    }

    if err != nil {
        return "", fmt.Errorf("got error while verifying requested ID: %w", err)
    }

    return "", fmt.Errorf("requested ID '%s' already exists, cannot write keyset", l.requestedKeysetID)
}
```

pkg/kms/localkms/localkms_writer.go
Baha-sk: updated

pkg/kms/localkms/localkms_writer.go
Baha-sk: updated

Issue #1133, opened 2020-01-29
Support to add business logic for Route Request Approval


rolsonquadras:
Currently, the framework approves all the route requests without any validation/logic. Framework should allow to inject business logic to make approval decision.

https://github.com/hyperledger/aries-framework-go/blob/6ee5e904c7a5de2d862f2abd57864baf5f0be148/pkg/didcomm/protocol/route/service.go#L187

PR #1780, opened 2020-05-04
feat: action events for routing service

llorllale:
closes: #1133

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -112,7 +112,7 @@ func GetRESTHandlers(ctx *context.Provider, opts ...Opt) ([]rest.Handler, error)
 	}
 
 	// route REST operation
-	routeOp, err := route.New(ctx)
+	routeOp, err := route.New(ctx, restAPIOpts.autoAccept)
 	if err != nil {
 		return nil, err
 	}
@@ -194,7 +194,7 @@ func GetCommandHandlers(ctx *context.Provider, opts ...Opt) ([]command.Handler, 
 	}
 
 	// route command operation
-	routecmd, err := routercmd.New(ctx)
+	routecmd, err := routercmd.New(ctx, cmdOpts.autoAccept)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/didcomm/protocol/route/service.go
+++ b/pkg/didcomm/protocol/route/service.go
@@ -93,7 +93,19 @@ type provider interface {
 	VDRIRegistry() vdri.Registry
 }
 
-// TODO add support for events: https://github.com/hyperledger/aries-framework-go/issues/1718
+// Options is a container for route protocol options.
+type Options struct {
+	ServiceEndpoint string
+	RoutingKeys     []string
+}
+
+type callback struct {
+	msg      service.DIDCommMsg
+	myDID    string
+	theirDID string
+	options  *Options
+	err      error
+}
 
 // Service for Route Coordination protocol.
 // https://github.com/hyperledger/aries-rfcs/tree/master/features/0211-route-coordination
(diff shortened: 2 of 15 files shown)


pkg/didcomm/protocol/route/service.go
```diff
@@ -133,18 +146,98 @@ func New(prov provider) (*Service, error) {
...
+
+	return nil
 }
 
 // HandleInbound handles inbound route coordination messages.
 func (s *Service) HandleInbound(msg service.DIDCommMsg, myDID, theirDID string) (string, error) { // nolint gocyclo (5 switch cases)
+	logger.Debugf("input: msg=%+v myDID=%s theirDID=%s", msg, myDID, theirDID)
+
+	if triggersActionEvent(msg.Type()) {
+		err := s.sendActionEvent(msg, myDID, theirDID)
```
soluchok, line 230: `return s.sendActionEvent(msg, myDID, theirDID)` or wrap the error

pkg/didcomm/protocol/route/service.go
```diff
@@ -205,33 +298,55 @@ func (s *Service) Name() string {
...
+		grant.Endpoint = defaultEndpoint
+	}
+
+	if len(grant.RoutingKeys) == 0 {
+		keys, err := defaultKeys()
+		if err != nil {
+			return nil, fmt.Errorf("failed to create keys : %w", err)
+		}
+
+		grant.RoutingKeys = []string{keys}
```
soluchok, line 346: I think it should be `defaultKey` instead of `defaultKeys` and `key` instead of `keys`.

test/bdd/pkg/route/route_sdk_steps.go
```diff
@@ -33,13 +34,48 @@ func (d *SDKSteps) CreateRouteClient(agentID string) error {
...
+}
+
+// ApproveRequest approves a routing protocol request with the given args.
+func (d *SDKSteps) ApproveRequest(agentID string, args interface{}) {
+	c, found := d.bddContext.RouteCallbacks[agentID]
+	if !found {
+		logger.Warnf("no callback channel found for %s", agentID)
+	}
+
+	c <- args
```
soluchok, line 72: It is a `deadlock`.  When `!found` condition is true, `c` will be a `nil`. Sending value to a nil channel blocks forever.

test/bdd/pkg/route/route_sdk_steps.go
```diff
@@ -33,13 +34,48 @@ func (d *SDKSteps) CreateRouteClient(agentID string) error {
...
+		logger.Debugf("received callback: %+v", c)
+		event.Continue(c)
+	}
+}
+
+// ApproveRequest approves a routing protocol request with the given args.
+func (d *SDKSteps) ApproveRequest(agentID string, args interface{}) {
+	c, found := d.bddContext.RouteCallbacks[agentID]
+	if !found {
+		logger.Warnf("no callback channel found for %s", agentID)
```
soluchok, line 69:
```suggestion
		logger.Warnf("no callback channel found for %s", agentID)
		return
```

llorllale: done

test/bdd/pkg/route/route_sdk_steps.go
llorllale: done

pkg/didcomm/protocol/route/service.go
llorllale: done

pkg/didcomm/protocol/route/service.go
llorllale: done

pkg/controller/controller.go
```diff
@@ -112,7 +112,7 @@ func GetRESTHandlers(ctx *context.Provider, opts ...Opt) ([]rest.Handler, error)
 	}
 
 	// route REST operation
-	routeOp, err := route.New(ctx)
+	routeOp, err := route.New(ctx, restAPIOpts.autoAccept)
```
rolsonquadras, line 115: I'm wondering whether "autoAccept" should be based on features (did-exchange, route, etc), rather than one for all. 

@troyronda @sudeshrshetty

PR #1788, opened 2020-05-06
doc: add export pubkey import privkey

### LINKED ISSUE #1794: KMS/Crypto
Add usage, api, godoc, etc documentation

https://github.com/hyperledger/aries-framework-go/issues/1785#issuecomment-624673537 - @Baha-sk

Baha-sk:
closes #1794

### DIFF
--- a/docs/kms_secretlock.md
+++ b/docs/kms_secretlock.md
@@ -12,7 +12,7 @@ To achieve these requirements, the following criteria must be met:
  - The SecretLock service must work with a master key that can be provided by the user.
  - The master key should also be encrypted to prevent its compromise. This should be optional in the SecretLock. For instance, a user may decide to use an unprotected master key for testing purposes but the recommended way is for the master key to be secured.
  - An example of unlocking a master key by the KMS/SecretLock would be the use of a passphrase.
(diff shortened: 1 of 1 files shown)


docs/kms_secretlock.md
```diff
@@ -232,4 +232,66 @@ a, err = aries.New(aries.WithKMS(func(ctx kms.Provider) (kms.KeyManager, error)
...
+    return err
+}
+
+pubKeyBytes, err := kmsInstance.ExportPubKeyBytes(ksID)
+
+```
+
+### Convert Public signing keys []byte to Tink's keyset.Handle
+
+The reverse of an export operation would be to get a keyset.Handle from a public key in raw []byte (DER or IEEE-P1363 signing format). The KMS service *does not* support importing *public keys*, but it can create a Tink `keyset.Handle` instance representing a public signing key to execute signature verification.
```
troyronda, line 257: > The KMS service *does not* support importing *public keys*

I assume this means only by themselves? i.e., there is private key import below, so I assume the public keys come along in that case?

Baha-sk: correct, importing private key includes its public key

Baha-sk: I've updated the readme to clarify this this detail.

PR #1793, opened 2020-05-06
feat: Credential and LDP dates to use custom object to avoid loss of data

### LINKED ISSUE #1772: Credential dates to use custom object to avoid loss of data
".000" millisecond precision in expirationDate and issuanceDate of credentials are getting removed/truncated while parsing.

Currently a temporary fix is in place to preserve raw string value string.

Long term fix: Should used custom Date objects which can be reused in other places too.

kdimak:
closes #1772

> ".000" millisecond precision in expirationDate and issuanceDate of credentials are getting removed/truncated while parsing.

For example, `2020-01-01T00:00:00.000Z` gets truncated to `2020-01-01T00:00:00Z` when formatted (marshalled) to string. It results in interoperability issues with e.g. JS libraries which provide VC with linked data proofs to validate.

In this PR, `TimeWithTrailingZeroMsec` is added that extends `time.Time` to handle this use case.

### DIFF
--- a/pkg/doc/signature/proof/proof.go
+++ b/pkg/doc/signature/proof/proof.go
@@ -8,7 +8,8 @@ package proof
 import (
 	"encoding/base64"
 	"errors"
-	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/util"
 )
 
 const (
@@ -37,7 +38,7 @@ const (
 // Proof is cryptographic proof of the integrity of the DID Document
 type Proof struct {
 	Type                    string
-	Created                 *time.Time
+	Created                 *util.TimeWithTrailingZeroMsec
 	Creator                 string
 	VerificationMethod      string
 	ProofValue              []byte
@@ -53,7 +54,7 @@ type Proof struct {
 func NewProof(emap map[string]interface{}) (*Proof, error) {
 	created := stringEntry(emap[jsonldCreated])
 
-	timeValue, err := time.Parse(time.RFC3339Nano, created)
+	timeValue, err := util.ParseTimeWithTrailingZeroMsec(created)
 	if err != nil {
 		return nil, err
 	}
@@ -87,7 +88,7 @@ func NewProof(emap map[string]interface{}) (*Proof, error) {
 
 	return &Proof{
 		Type:                    stringEntry(emap[jsonldType]),
-		Created:                 &timeValue,
+		Created:                 timeValue,
 		Creator:                 stringEntry(emap[jsonldCreator]),
 		VerificationMethod:      stringEntry(emap[jsonldVerificationMethod]),
 		ProofValue:              proofValue,
@@ -123,7 +124,7 @@ func (p *Proof) JSONLdObject() map[string]interface{} {
 	}
 
 	if p.Created != nil {
-		emap[jsonldCreated] = p.Created.Format(time.RFC3339Nano)
+		emap[jsonldCreated] = p.Created.Format(p.Created.GetFormat())
 	}
 
 	if len(p.ProofValue) > 0 {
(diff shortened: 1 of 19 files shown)


kdimak:
Naming improvements are welcome :) For me, `TimeWithTrailingZeroMsec` name is not a good one.

pkg/doc/signature/signer/signer.go
```diff
@@ -103,7 +103,7 @@ func (signer *DocumentSigner) signObject(context *Context, jsonLdObject map[stri
 		Type:                    context.SignatureType,
 		SignatureRepresentation: context.SignatureRepresentation,
 		Creator:                 context.Creator,
-		Created:                 created,
+		Created:                 &util.TimeWithTrailingZeroMsec{Time: *created},
```
rolsonquadras, line 106: Would be helpful to have a New() constructor rather than creating the struct.

kdimak: Good idea, I will take a look.

kdimak: Added `NewTime()` and `NewTimeWithTrailingZeroMsec()` factory methods.

sudeshrshetty:
> Naming improvements are welcome :) For me, `TimeWithTrailingZeroMsec` name is not a good one.

`CredentialTime`
`TimeExt`

pkg/doc/signature/proof/proof.go
```diff
@@ -53,7 +54,7 @@ type Proof struct {
 func NewProof(emap map[string]interface{}) (*Proof, error) {
 	created := stringEntry(emap[jsonldCreated])
 
-	timeValue, err := time.Parse(time.RFC3339Nano, created)
+	timeValue, err := util.ParseTimeWithTrailingZeroMsec(created)
```
sudeshrshetty, line 57: util.ParseTime() ??

pkg/doc/util/time.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"time"
+)
+
+// TimeWithTrailingZeroMsec overrides marshalling of time.Time. It keeps a format of initial unmarshalling
+// in case when date has zero a fractional second (e.g. ".000").
+// For example, time.Time marshals 2018-03-15T00:00:00.000Z to 2018-03-15T00:00:00Z
+// while TimeWithTrailingZeroMsec marshals to the initial 2018-03-15T00:00:00.000Z value.
+type TimeWithTrailingZeroMsec struct {
```
sudeshrshetty, line 11: Can also be called 'TimeExt', 'TimeX' etc, I am fine with current name too

pkg/doc/util/time.go
```diff
@@ -0,0 +1,140 @@
...
+// It functions as a normal time.Time object.
+func NewTime(t time.Time) *TimeWithTrailingZeroMsec {
+	return &TimeWithTrailingZeroMsec{Time: t}
+}
+
+// NewTimeWithTrailingZeroMsec creates TimeWithTrailingZeroMsec with certain zero sub-second precision.
+func NewTimeWithTrailingZeroMsec(t time.Time, trailingZerosMsecCount int) *TimeWithTrailingZeroMsec {
+	return &TimeWithTrailingZeroMsec{
+		Time:                   t,
+		trailingZerosMsecCount: trailingZerosMsecCount,
```
sudeshrshetty, line 27: just a thought:  preserving actual string value could have made logic more simpler

pkg/doc/util/time.go
kdimak: good idea!

pkg/doc/util/time.go
kdimak: I think the current solution is more generic - it keeps the format which can be obtained separately.

pkg/doc/signature/signer/signer.go
llorllale: ... but still pending to use it here @kdimak

PR #1796, opened 2020-05-07
feat: inverse verifiable credential dependency

AlexeyBelezeko:
Signed-off-by: Aleksei Belezeko <aleksei.belezeko@t-systems.com>

Summary:
To customize issue credential service and use issued credential we have to provide access for verifiable credential store.


pkg/storage/jsindexeddb/jsindexeddb.go
```diff
@@ -163,7 +163,7 @@ func (s *store) Get(k string) ([]byte, error) {
 
 // Iterator returns iterator for the latest snapshot of the underlying db.
 func (s *store) Iterator(start, limit string) storage.StoreIterator {
-	// TODO Change Store Iterator https://github.com/hyperledger/aries-framework-go/issues/852
+	// TODO Change storeImplementation Iterator https://github.com/hyperledger/aries-framework-go/issues/852
```
llorllale, line 166: this slipped through

pkg/framework/aries/default.go
```diff
@@ -8,6 +8,12 @@ package aries
 
 import (
 	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+
```
troyronda, line 13: extra blank line

pkg/framework/aries/default_test.go
```diff
@@ -9,12 +9,23 @@ SPDX-License-Identifier: Apache-2.0
...
 import (
+	"errors"
+
+	"github.com/golang/mock/gomock"
+
+	mocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/storage"
+
 	"testing"
 
+	verifiableStoreMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/store/verifiable"
```
troyronda, line 20: imports should be grouped.

pkg/framework/aries/framework.go
```diff
@@ -8,6 +8,9 @@ package aries
 
 import (
 	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/store/verifiable"
```
troyronda, line 12: imports should be grouped.

pkg/framework/aries/framework_test.go
```diff
@@ -14,6 +14,9 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+
+	verifiableStoreMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/store/verifiable"
```
troyronda, line 18: imports should be grouped.

pkg/framework/context/context_test.go
```diff
@@ -9,6 +9,9 @@ package context
 import (
 	"errors"
 	"fmt"
+
+	verifiableStoreMocks "github.com/hyperledger/aries-framework-go/pkg/internal/gomocks/store/verifiable"
```
troyronda, line 13: imports should be grouped.

pkg/store/verifiable/store.go
```diff
@@ -50,17 +62,17 @@ type provider interface {
 }
 
 // New returns a new vc store
-func New(ctx provider) (*Store, error) {
+func New(ctx provider) (Store, error) {
```
soluchok, line 65: We need to return an implementation here. Where `StoreImplementation` is public.
e.g. `func New(ctx provider) (StoreImplementation, error)`
To make sure that implementation satisfies the interface do the following:
`var _ Store = (*StoreImplementation)(nil)`

pkg/framework/aries/default.go
```diff
@@ -43,13 +49,21 @@ func defFrameworkOpts(frameworkOpts *Aries) error {
 		frameworkOpts.outboundTransports = append(frameworkOpts.outboundTransports, outbound)
 	}
 
-	if frameworkOpts.storeProvider == nil {
-		storeProv, err := storeProvider()
+	currentStoreProvider := frameworkOpts.storeProvider
+	if currentStoreProvider == nil {
+		var err error
+		currentStoreProvider, err = storeProvider()
+
```
soluchok, line 56: Please, revert this change. It does nothing.
The line below will look like:
`err := assignVerifiableStoreIfNeeded(frameworkOpts, frameworkOpts.storeProvider)`

pkg/framework/aries/default.go
```diff
@@ -167,6 +181,24 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
...
+	if aries.verifiableStore != nil {
+		return nil
+	}
+
+	provider, err := context.New(context.WithStorageProvider(storeProvider))
+	if err != nil {
+		return fmt.Errorf("verifiable store initialization failed : %w", err)
+	}
+
+	aries.verifiableStore, err = verifiable.New(provider)
```
soluchok, line 194: I don't like that idea. You are creating a context that will actually satisfy everything. But actually this context contains only the storage. It will be hard to debug if someone decides to use one more dependency in the verifiable store. The code will be compiled successfully instead of point out the error. It makes code more like dynamically typed instead of statically typed.

The simplest way to fix it. Is to receive `storage.Provider` in verifiable store.
`func New(ctx storage.Provider) (*StoreImpl, error)`

@troyronda @llorllale ^

AlexeyBelezeko: I agree that this is not the best idea but this approach already used in project. You can check `function func createVDRI(frameworkOpts *Aries) error` in framework.go. And if you add new dependency tests will fail.

soluchok: Tests might not fail. if you have BDD tests that cover that logic then yeah. Even in that case, you need to find out why they failed. It will take some time. This time can be saved.

pkg/storage/jsindexeddb/jsindexeddb.go
AlexeyBelezeko: Fixed

pkg/framework/context/context_test.go
AlexeyBelezeko: Fixed all import

pkg/store/verifiable/store.go
AlexeyBelezeko: Fixed

pkg/framework/aries/default.go
AlexeyBelezeko: Fixed

pkg/framework/aries/default.go
AlexeyBelezeko: This approach used a lot in this project. I can't fix this problem in this PR with 500 lines limit cause the provider for command which initialize verifiable Store has the same issue.

pkg/framework/aries/default.go
```diff
@@ -44,12 +47,17 @@ func defFrameworkOpts(frameworkOpts *Aries) error {
 	}
 
 	if frameworkOpts.storeProvider == nil {
-		storeProv, err := storeProvider()
+		newStoreProvider, err := storeProvider()
```
soluchok, line 50: nit: one more revert `storeProv` is needed. Everything else LGTM.

pkg/framework/aries/default.go
llorllale: @soluchok I think you're arguing for having a `context.Provider` -> `verifiable.provider` adapter.

I generally agree, but like @AlexeyBelezeko says, the approach taken in this PR is used by the framework everywhere, and I would advise to raise this point in a separate issue.

pkg/framework/aries/default.go
AlexeyBelezeko: fixed

AlexeyBelezeko:
Build fails with error: 
```
Checking committed files for traditional Apache License headers ...
The following files are missing traditional Apache 2.0 headers:
pkg/doc/verifiable/testdata/context/trustbloc_example.jsonld
pkg/doc/verifiable/testdata/context/vc_example.jsonld
Fatal Error - All files must have a license header
##[error]Makefile:39: recipe for target 'license' failed
make: *** [license] Error 1
##[error]Process completed with exit code 2.
```
But I don't have such files in my PR and in project locally. Any ideas? @soluchok @llorllale

llorllale:
@AlexeyBelezeko try rebasing on `master`

PR #1801, opened 2020-05-08
fix: remove check length for requested KeysetID

fqutishat:


troyronda:
what problem did you encounter @fqutishat

PR #1807, opened 2020-05-08
test: add BDD tests for Present Proof REST API

### LINKED ISSUE #1720: Present Proof REST - add BDD tests
Add BDD tests for Present Proof REST

aholovko:
Adds BDD tests for Present Proof REST API

Closes #1720

### DIFF
--- a/test/bdd/pkg/issuecredential/issuecredential_controller_steps.go
+++ b/test/bdd/pkg/issuecredential/issuecredential_controller_steps.go
@@ -36,7 +36,6 @@ const (
 	acceptOffer       = operationID + "/%s/accept-offer"
 	acceptRequest     = operationID + "/%s/accept-request"
 	acceptCredential  = operationID + "/%s/accept-credential"
-	declineRequest    = operationID + "/%s/decline-request"
 )
 
 var logger = log.New("aries-framework/issuecredential-tests")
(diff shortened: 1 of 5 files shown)


test/bdd/pkg/issuecredential/issuecredential_controller_steps.go
```diff
@@ -36,7 +36,6 @@ const (
 	acceptOffer       = operationID + "/%s/accept-offer"
 	acceptRequest     = operationID + "/%s/accept-request"
 	acceptCredential  = operationID + "/%s/accept-credential"
-	declineRequest    = operationID + "/%s/decline-request"
```
troyronda, line 39: what happened to decline?

aholovko: It was a cleanup from the previous commit. Currently, I am not checking decline cases for the reasons described below.

In a case when the credential was issued we can confirm that by asserting storage. In "decline" cases (i.e decline request/proposal/offer/credential) it can't be done and REST API, also, doesn't provide a way to check the state of protocol for them.

I think we have 2 options here:

1. Update REST API to expose the state of the protocol - perhaps worth considering when working on v2 of protocols.
2. Use a sort of "limited" or indirect assertion - just check if decline request doesn't fail.

Option 1 is definitely out of the scope of the current task. Troy, do you think it makes sense to work on the second option?

PR #1808, opened 2020-05-08
doc: update KMS/crypto/Context documentations

Baha-sk:
part of #1788

### DIFF
--- a/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
@@ -6,6 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 
 package subtle
 
+// package subtle provides the core crypto primitives to be used by composite primitives. It is intended for internal
+// use only.
+
 // EncryptedData represents the Encryption's output data as a result of ECDHESEncrypt.Encrypt(pt, aad) call
 // The user of the primitive must unmarshal the result and build their own ECDH-ES compliant message (ie JWE msg)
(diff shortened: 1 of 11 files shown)


pkg/secretlock/noop/noop_secret_lock.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
 )
 
+// package noop provides a noop secret lock service. This allows for quick testing of key storage using the KMS. Keys
+// stored with noop are unprotected. Therefore, this implementation is be used for testing purposes only.
+// Production code must always use pkg/secrelock/local implementation. In order to minimize the impact on existing
```
troyronda, line 14: typo `secrelock`

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aead_composite_common.go
```diff
@@ -6,6 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 
 package subtle
 
+// package subtle provides the core crypto primitives to be used by composite primitives. It is intended for internal
```
llorllale, line 9: @Baha-sk curious: why is it called `subtle`?

Baha-sk: Tink Primitives call their core package this way, it contains the core primitive logic intended to be used by the facing API package. It's not internal to allow interop between different Primitives (and, I think, expose calls to the core primitives by advanced users). Tink doesn't guarantee subtle packages won't change in the future and they're not part of the official API. They're not intended for external use. @llorllale

PR #1815, opened 2020-05-11
fix: do not fire didexchange event after accepting oob message

### LINKED ISSUE #1814: outofband: user should not have to approve did-exchange after accepting an out-of-band request or invitation

llorllale:
closes #1814

Note: `presentproof_sdk_steps.go#acceptRequestPresentation` was refactored because the JWS signature verification was failing for an as-yet unclear reason. Some of this method's assumptions[1] were strange to me, so I refactored and it started working. Please take a look.

[1]: it was assuming the first connection record retrieved from an agent's storage was the right one, and it was assuming consistency between these connection records and public DIDs registered in the BDD context by a different BDD package

### DIFF
--- a/pkg/client/outofband/client.go
+++ b/pkg/client/outofband/client.go
@@ -35,6 +35,18 @@ const (
 	InvitationMsgType = outofband.InvitationMsgType
 )
 
+// EventOptions are is a container of options that you can pass to an event's
+// Continue function to customize the reaction to incoming out-of-band messages.
+type EventOptions struct {
+	// Label will be shared with the other agent during the subsequent did-exchange.
+	Label string
+}
+
+// MyLabel will be shared with the other agent during the subsequent did-exchange.
+func (e *EventOptions) MyLabel() string {
+	return e.Label
+}
+
 // Event is a container of out-of-band protocol-specific properties for DIDCommActions and StateMsgs.
 type Event interface {
 	// ConnectionID of the connection record, once it's created.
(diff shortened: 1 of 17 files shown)


pkg/client/outofband/client.go
```diff
@@ -35,6 +35,18 @@ const (
 	InvitationMsgType = outofband.InvitationMsgType
 )
 
+// EventOptions are is a container of options that you can pass to an event's
+// Continue function to customize the reaction to incoming out-of-band messages.
+type EventOptions struct {
+	// Label will be shared with the other agent during the subsequent did-exchange.
```
troyronda, line 41: Do you have thoughts about this coupling? We don't know that the subsequent action will be a DID exchange right?

llorllale: It can be any protocol that forms a connection, really. Or it can be an ephemeral mode.

pkg/client/outofband/client.go
```diff
@@ -56,10 +68,10 @@ type message struct {
 
 type oobService interface {
 	service.Event
-	AcceptRequest(request *outofband.Request) (string, error)
-	AcceptInvitation(inv *outofband.Invitation) (string, error)
-	SaveRequest(request *outofband.Request) error
-	SaveInvitation(inv *outofband.Invitation) error
+	AcceptRequest(*outofband.Request, string) (string, error)
```
troyronda, line 71: Will this function signature be stable, or is there the possibility of more arguments? 

Since there is an `EventOptions` struct above, should there be some struct parameter here?

pkg/client/outofband/client.go
troyronda: So is Label intended to be generic or about did-exchange?

I would like to avoid this struct becoming a collection of all possible followup options. In that case, it seems better to pass options directly into the followup protocol implementation.

pkg/client/outofband/client.go
llorllale: I can't comment on its stability. I was following similar patterns in the didexchange client package.

troyronda: My concern stems more from this same theme: https://github.com/hyperledger/aries-framework-go/pull/1815#discussion_r423708740

(there could be other followup protocols).

PR #1830, opened 2020-05-14
feat: Credential and LDP dates to use custom object to avoid loss of data

### LINKED ISSUE #1772: Credential dates to use custom object to avoid loss of data

kdimak:
closes #1772

This is a re-opened PR from accidentally closed https://github.com/hyperledger/aries-framework-go/pull/1793

> ".000" millisecond precision in expirationDate and issuanceDate of credentials are getting removed/truncated while parsing.

For example, `2020-01-01T00:00:00.000Z` gets truncated to `2020-01-01T00:00:00Z` when formatted (marshalled) to string. It results in interoperability issues with e.g. JS libraries which provide VC with linked data proofs to validate.

In this PR, `TimeWithTrailingZeroMsec` is added that extends `time.Time` to handle this use case.


llorllale:
@kdimak you can reopen the other PR

PR #1834, opened 2020-05-14
feat: verifier/signer to fail on invalid RDF

### LINKED ISSUE #1818: Adding Proof: deny payload elements outside of the JSON-LD context
Currently during signing, we are removing invalid RDF and during verification we are not removing invalid RDFs before calculating digest.

**Change:** Invalid RDF should cause both signing and verification failure. 

Note: invalid RDF detection, filter logic and opts in json ld processor should be left as it is, so that it can be reused by other modules in future.

sudeshrshetty:
- Closes #1818

### DIFF
--- a/pkg/doc/signature/jsonld/processor.go
+++ b/pkg/doc/signature/jsonld/processor.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package jsonld
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -23,9 +24,13 @@ const (
 
 var logger = log.New("aries-framework/json-ld-processor")
 
+// ErrInvalidRDFFound is returned when normalized view contains invalid RDF
+var ErrInvalidRDFFound = errors.New("invalid JSON-LD context")
+
 // normalizeOpts holds options for canonicalization of JSON LD docs
 type normalizeOpts struct {
 	removeInvalidRDF bool
+	validateRDF      bool
 	documentLoader   ld.DocumentLoader
 	externalContexts []string
 }
@@ -51,6 +56,14 @@ func WithDocumentLoader(loader ld.DocumentLoader) ProcessorOpts {
 func WithExternalContext(context ...string) ProcessorOpts {
 	return func(opts *normalizeOpts) {
 		opts.externalContexts = context
+	}
+}
+
+// WithValidateRDF option validates result view and fails if any invalid RDF dataset found.
+// This option will take precedence when used in conjunction with 'WithRemoveAllInvalidRDF' option.
+func WithValidateRDF() ProcessorOpts {
+	return func(opts *normalizeOpts) {
+		opts.validateRDF = true
 	}
 }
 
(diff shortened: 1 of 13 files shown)


pkg/doc/signature/jsonld/processor.go
```diff
@@ -103,11 +116,9 @@ func (p *Processor) GetCanonicalDocument(doc map[string]interface{}, opts ...Pro
...
 	}
 
-	if procOptions.removeInvalidRDF {
-		result, err = p.removeMatchingInvalidRDFs(result)
-		if err != nil {
-			return nil, fmt.Errorf("failed to normalize due to invalid RDF dataset: %w", err)
-		}
+	result, err = p.removeMatchingInvalidRDFs(result, procOptions)
+	if err != nil {
+		return nil, err
```
troyronda, line 121: How do we allow the higher layer package to decide how to enable such cases? 

Are we just failing rather than enabling a warning (or otherwise enabling the higher layer package to decide how to handle)?

sudeshrshetty: Higher level packages will decide how to handle these cases.

Now signer & verifier will use default "WithValidateRDF" which validates and fails when invalid RDF found

sudeshrshetty: line number 179

pkg/doc/signature/signer/signer.go
```diff
@@ -121,7 +121,7 @@ func (signer *DocumentSigner) signObject(context *Context, jsonLdObject map[stri
 		p.JWS = proof.CreateDetachedJWTHeader(p) + ".."
 	}
 
-	message, err := proof.CreateVerifyData(suite, jsonLdObject, p, jsonld.WithRemoveAllInvalidRDF())
+	message, err := proof.CreateVerifyData(suite, jsonLdObject, p, jsonld.WithValidateRDF())
```
rolsonquadras, line 124: `WithValidateRDF` option is for jsonld consumers. Should this option be available to consumers of presentation/credential packages as well ?

sudeshrshetty: it is already available as remove all RDF, if provided then this option will not be used

pkg/doc/verifiable/support_test.go
```diff
@@ -41,6 +41,7 @@ const validCredential = `{
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
     "https://www.w3.org/2018/credentials/examples/v1",
+	"https://trustbloc.github.io/context/vc/credentials-v1.jsonld",
```
kdimak, line 44: Could you cache `https://trustbloc.github.io/context/vc/credentials-v1.jsonld` context for the VC tests? It's done at https://github.com/hyperledger/aries-framework-go/blob/c942c043873c23ccb7fe68bc177811efa89078d7/pkg/doc/verifiable/support_test.go#L366

sudeshrshetty: done

pkg/doc/signature/jsonld/processor.go
troyronda: Ok - can we provide two examples in the tests:

(a) "Client" who wants to print warnings about invalid RDF but allow signing/verifier to continue. The client should print the warnings not the library, of course.
(b) "Client" who wants to fail signing/verifier on invalid RDF but have an error object that indicates this is the failure (rather than invalid signature).

troyronda: @sudeshrshetty would you want to do these examples as the next PR?

sudeshrshetty: will add it in upcoming PR

PR #1841, opened 2020-05-18
feat: Present Proof saves the presentation to the verifiable store

soluchok:
Present Proof saves the presentation to the verifiable store.
This PR also improves BDD/BDD-js tests by adding verifiable presentation payload.

### DIFF
--- a/pkg/controller/rest/presentproof/models.go
+++ b/pkg/controller/rest/presentproof/models.go
@@ -140,6 +140,10 @@ type presentProofAcceptPresentationRequest struct { // nolint: unused,deadcode
 	// in: path
 	// required: true
 	PIID string `json:"piid"`
+
+	// in: body
+	// required: true
+	Names []string `json:""`
 }
 
 // presentProofAcceptPresentationResponse model
--- a/pkg/controller/rest/presentproof/operation.go
+++ b/pkg/controller/rest/presentproof/operation.go
@@ -180,9 +180,27 @@ func (c *Operation) AcceptProposePresentation(rw http.ResponseWriter, req *http.
 //    default: genericError
 //        200: presentProofAcceptPresentationResponse
 func (c *Operation) AcceptPresentation(rw http.ResponseWriter, req *http.Request) { // nolint: dupl
+	var buf bytes.Buffer
+
+	if req.Body != nil {
+		// nolint: errcheck
+		_, _ = io.Copy(&buf, req.Body)
+	}
+
+	if !isJSONArray(buf.Bytes()) {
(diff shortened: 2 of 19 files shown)


pkg/controller/rest/presentproof/models.go
```diff
@@ -140,6 +140,10 @@ type presentProofAcceptPresentationRequest struct { // nolint: unused,deadcode
 	// in: path
 	// required: true
 	PIID string `json:"piid"`
+
+	// in: body
+	// required: true
+	Names []string `json:""`
```
llorllale, line 146: missed the name on the json tag

pkg/controller/rest/presentproof/operation.go
```diff
@@ -180,9 +180,27 @@ func (c *Operation) AcceptProposePresentation(rw http.ResponseWriter, req *http.
 //    default: genericError
 //        200: presentProofAcceptPresentationResponse
 func (c *Operation) AcceptPresentation(rw http.ResponseWriter, req *http.Request) { // nolint: dupl
+	var buf bytes.Buffer
+
+	if req.Body != nil {
+		// nolint: errcheck
```
llorllale, line 186: why skip errcheck?

pkg/controller/rest/presentproof/operation.go
```diff
@@ -180,9 +180,27 @@ func (c *Operation) AcceptProposePresentation(rw http.ResponseWriter, req *http.
...
+		_, _ = io.Copy(&buf, req.Body)
+	}
+
+	if !isJSONArray(buf.Bytes()) {
+		rest.SendHTTPStatusError(rw,
```
llorllale, line 191: nit: I'm not sure we want to *force* the user to provide friendly names

pkg/controller/rest/presentproof/models.go
soluchok: It is not.  This is only for the `swagger`. The empty JSON tag means that the request body should be an array.

pkg/controller/rest/presentproof/operation.go
soluchok: Even if an error occurs the next line checks whether payload `the buf.Bytes()` is a JSON. It will return `InvalidRequestErrorCode` anyway. I skipped it because in my opinion it kind of redundant (I mean in that specific case).

pkg/controller/rest/presentproof/operation.go
soluchok: under the hood, it is not a required data. I added only for the rest API. This restriction can be removed easily.

PR #1842, opened 2020-05-19
feat: Issue Credential - Adds WithPublicKeyFetcher option

soluchok:
Adds WithPublicKeyFetcher option when we parse the credentials.

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/states_test.go
+++ b/pkg/didcomm/protocol/issuecredential/states_test.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package issuecredential
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"testing"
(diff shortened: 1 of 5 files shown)


pkg/didcomm/protocol/issuecredential/states_test.go
```diff
@@ -711,6 +714,94 @@ func TestCredentialReceived_ExecuteInbound(t *testing.T) {
...
+
+		require.NoError(t, err)
+		require.Equal(t, &done{}, followup)
+		require.NotNil(t, action)
+
+		messenger := serviceMocks.NewMockMessenger(ctrl)
+		messenger.EXPECT().ReplyTo(gomock.Any(), gomock.Any())
+
+		require.NoError(t, action(messenger))
+	})
```
kdimak, line 804: Maybe to add an error case here as well (i.e. corrupt `proof.jws` value and see that we've got an error with some clear description of the cause).

soluchok: I have a test when a credential is invalid  the https://github.com/hyperledger/aries-framework-go/pull/1842/files/4a36c0c1f382b9c8913b52f4e33ed3171921bd19#diff-f9dc1ab6bf3eb1acef404898701c3a3fR639

PR #1845, opened 2020-05-20
test: JsonWebSignature2020 and EcdsaSecp256k1Signature2019 signature suites in VC BDD tests

### LINKED ISSUE #1596: JsonWebSignature2020 and EcdsaSecp256k1Signature2019 signature suites in VC BDD test

kdimak:
closes #1596

This PR includes bugfix and extension of `jose.JWK`:
- added `JWKFromPublicKey()` function which takes a public key struct (e.g. `*ecdsa.PublicKey` and builds `JWK` from it
- fixed JSON marshalling for  secp256k1-based key.

For VC BDD tests, tinkcrypto/local KMS is used instead of legacy KMS.
Also, DID creation at sidetree is made less hardcoded, we can now create public keys with different types (previously `JwsVerificationKey2020` was hardcoded). This allows checking a different kind of VC Linked Data suites.

### DIFF
--- a/pkg/doc/did/contexts.go
+++ b/pkg/doc/did/contexts.go
@@ -1,0 +1,294 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package did
+
+const didV1Context = `
+{
+  "@context": {
+    "@version": 1.1,
+    "id": "@id",
+    "type": "@type",
+    "dc": "http://purl.org/dc/terms/",
+    "schema": "http://schema.org/",
+    "sec": "https://w3id.org/security#",
+    "didv": "https://w3id.org/did#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "EcdsaSecp256k1Signature2019": "sec:EcdsaSecp256k1Signature2019",
+    "EcdsaSecp256k1VerificationKey2019": "sec:EcdsaSecp256k1VerificationKey2019",
+    "Ed25519Signature2018": "sec:Ed25519Signature2018",
+    "Ed25519VerificationKey2018": "sec:Ed25519VerificationKey2018",
+    "RsaSignature2018": "sec:RsaSignature2018",
+    "RsaVerificationKey2018": "sec:RsaVerificationKey2018",
+    "SchnorrSecp256k1Signature2019": "sec:SchnorrSecp256k1Signature2019",
+    "SchnorrSecp256k1VerificationKey2019": "sec:SchnorrSecp256k1VerificationKey2019",
+    "ServiceEndpointProxyService": "didv:ServiceEndpointProxyService",
+    "allowedAction": "sec:allowedAction",
+    "assertionMethod": {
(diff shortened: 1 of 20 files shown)


pkg/doc/did/doc.go
```diff
@@ -53,6 +56,294 @@ const (
 	jsonldPublicKeyjwk    = "publicKeyJwk"
 )
 
+const didV1Context = `
```
troyronda, line 59: can we move contexts to a separate file?

kdimak: Sure, @troyronda . What's the best place for it? At the same level as the source files?

troyronda: sure for now.

pkg/doc/did/doc.go
```diff
@@ -1004,3 +1295,24 @@ func BuildDoc(opts ...DocOption) *Doc {
 
 	return doc
 }
+
+func createDocumentLoader() ld.DocumentLoader {
+	loader := ld.NewCachingDocumentLoader(ld.NewRFC7324CachingDocumentLoader(&http.Client{}))
```
llorllale, line 1300: Hardcoding the http client makes this harder to test and may impact future work that requires different implementations. I've already been bit by this sort of issue while working on unit tests and examples.

llorllale: Can't there be an option where you can provide your own document loader?

kdimak: Sure, I will add an option for this, at https://github.com/hyperledger/aries-framework-go/issues/1846.

pkg/doc/did/doc.go
kdimak: Added `context` folder.

pkg/doc/did/doc.go
```diff
@@ -1004,3 +1011,30 @@ func BuildDoc(opts ...DocOption) *Doc {
...
+
+	didV1Context, _ := ioutil.ReadFile(filepath.Clean(filepath.Join(
+		jsonldContextPrefix, "didv1.jsonld"))) //nolint:errcheck
```
llorllale, line 1019: These files are not bundled in the final binary. This won't work unless the files are distributed together with the binary. I believe all that @troyronda suggested was to put them in separate `.go` files.

kdimak: Ah, I see! My bad, one moment, I will fix it.

pkg/doc/did/doc.go
kdimak: Corrected - JSON-LD contexts definition is moved into separate `contexts.go`.

pkg/doc/did/doc.go
kdimak: Added `contexts.go`, thank you for the catch @llorllale !

pkg/doc/did/contexts.go
```diff
@@ -0,0 +1,289 @@
+package did
```
troyronda, line 1: missing license header.

kdimak: Added.

Issue #1846, opened 2020-05-20
Cache all contexts in uni-tests & bdd-tests


sudeshrshetty:
to avoid errors like these in CIs

```
--- FAIL: TestVerifyProof (0.49s)
panic: failed to normalize JSON-LD document: loading remote context failed: Dereferencing a URL did not result in a valid JSON-LD context: https://www.w3.org/ns/did/v1 [recovered]
	panic: failed to normalize JSON-LD document: loading remote context failed: Dereferencing a URL did not result in a valid JSON-LD context: https://www.w3.org/ns/did/v1

goroutine 59 [running]:
testing.tRunner.func1.1(0xbde4e0, 0xc00041e080)
	/opt/hostedtoolcache/go/1.14.3/x64/src/testing/testing.go:940 +0x421
testing.tRunner.func1(0xc000233c20)
	/opt/hostedtoolcache/go/1.14.3/x64/src/testing/testing.go:943 +0x600
panic(0xbde4e0, 0xc00041e080)
	/opt/hostedtoolcache/go/1.14.3/x64/src/runtime/panic.go:975 +0x3e3
github.com/hyperledger/aries-framework-go/pkg/doc/did.createSignedDidDocument(0xc00043ab80, 0x40, 0x40, 0xc0000fc680, 0x20, 0x20, 0x40, 0x40, 0x0)
	/home/runner/work/aries-framework-go/aries-framework-go/pkg/doc/did/doc_test.go:1434 +0x373
github.com/hyperledger/aries-framework-go/pkg/doc/did.TestVerifyProof(0xc000233c20)
	/home/runner/work/aries-framework-go/aries-framework-go/pkg/doc/did/doc_test.go:955 +0x15e
```

kdimak:
TODO: add an option for definition of custom JSONLD loader for DID doc verification.

PR #1851, opened 2020-05-22
test: cache all JSON-LD contexts in verifiable tests

kdimak:

#1846

On my local machine `verifiable` tests are got executed for about 7 seconds (more than 50 seconds before, and before the first caching stage it was about 3-4 minutes).

Besides context caching in the tests, the following is changed:
1) Added `jsonld.ProcessorOpts` parameter to `Credential.AddLinkedDataProof()` and `Presentation.AddLinkedDataProof()`. This allows passing JSON-LD options to be used when building linked data signature.
2) Fixed a bug - to pass `jsonld.ProcessorOpts` when building a canonized document with proof compaction.


pkg/doc/verifiable/example_credential_test.go
```diff
@@ -103,30 +73,31 @@ func ExampleCredential_embedding() {
...
 
 	// Parse JWS and make sure it's coincide with JSON.
 	vcParsed, err := verifiable.ParseCredential(
 		[]byte(jws),
-		verifiable.WithPublicKeyFetcher(verifiable.SingleKey(issuerPubKey, kms.ED25519)))
+		verifiable.WithPublicKeyFetcher(verifiable.SingleKey(issuerPubKey, kms.ED25519)),
+		verifiable.WithJSONLDDocumentLoader(getJSONLDDocumentLoader()))
 	if err != nil {
-		fmt.Println(fmt.Errorf("failed to encode VC from JWS: %w", err))
+		panic(fmt.Errorf("failed to encode VC from JWS: %w", err))
```
fqutishat, line 100: we you change it to panic

kdimak: Just because in many cases there is no more sense to continue the test execution. The same approach is used e.g. at https://github.com/hyperledger/aries-framework-go/blob/master/pkg/client/outofband/example_test.go

pkg/doc/verifiable/example_support_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return json.Marshal(&cp)
+}
+
+func getEd25519Signer(privKey []byte) *ed25519Signer {
+	return &ed25519Signer{privateKey: privKey}
+}
+
+type ed25519Signer struct {
```
Baha-sk, line 59: is it necessary to create new signer definition in every file requiring ed25519 signer? can't we call a single default definition?

pkg/doc/verifiable/example_support_test.go
```diff
@@ -0,0 +1,149 @@
...
+	privateKey []byte
+}
+
+func (s *ed25519Signer) Sign(doc []byte) ([]byte, error) {
+	if l := len(s.privateKey); l != ed25519.PrivateKeySize {
+		return nil, errors.New("ed25519: bad private key length")
+	}
+
+	return ed25519.Sign(s.privateKey, doc), nil
```
Baha-sk, line 68: code seems to be duplicated.

pkg/doc/verifiable/example_support_test.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func getECDSASecp256k1Signer(privKey *ecdsa.PrivateKey) *ecdsaSecp256k1Signer {
+	return &ecdsaSecp256k1Signer{
+		privKey: privKey,
+	}
+}
+
+type ecdsaSecp256k1Signer struct {
```
Baha-sk, line 77: why create new signer definition in tests?

pkg/doc/verifiable/example_support_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	copyPadded := func(source []byte, size int) []byte {
+		dest := make([]byte, size)
+		copy(dest[size-len(source):], source)
+
+		return dest
+	}
+
+	return append(copyPadded(r.Bytes(), keyBytes), copyPadded(s.Bytes(), keyBytes)...), nil
+}
```
Baha-sk, line 111: this code chunk seems to be duplicated in many places, consider consolidating it in 1 single place.

PR #1853, opened 2020-05-22
test: cache all JSON-LD contexts in DID tests

kdimak:

#1846

### DIFF
--- a/pkg/doc/did/contexts.go
+++ b/pkg/doc/did/contexts.go
@@ -292,3 +292,118 @@ const didV011Context = `
     }
   }
 }`
+
+const securityV1Context = `
+{
+  "@context": {
+    "id": "@id",
+    "type": "@type",
+
+    "dc": "http://purl.org/dc/terms/",
+    "sec": "https://w3id.org/security#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+
+    "EcdsaKoblitzSignature2016": "sec:EcdsaKoblitzSignature2016",
+    "Ed25519Signature2018": "sec:Ed25519Signature2018",
+    "EncryptedMessage": "sec:EncryptedMessage",
+    "GraphSignature2012": "sec:GraphSignature2012",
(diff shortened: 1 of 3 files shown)


pkg/doc/did/doc.go
```diff
@@ -762,7 +760,7 @@ func (doc *Doc) JSONBytes() ([]byte, error) {
 }
 
 // VerifyProof verifies document proofs
-func (doc *Doc) VerifyProof(suites ...verifier.SignatureSuite) error {
+func (doc *Doc) VerifyProof(suites []verifier.SignatureSuite, jsonldOpts ...jsonld.ProcessorOpts) error {
```
kdimak, line 763: Here we pass universal `jsonld.ProcessorOpts`, one of that is `WithDocumentLoader()`. The configured JSON-LD document loader with cached contexts is used in the tests, but one can pass arbitrary document loader in production code (with e.g. custom setup of HTTP client).

pkg/doc/did/contexts.go
```diff
@@ -1,294 +0,0 @@
-/*
```
llorllale, line 1: why was this deleted and then moved to .jsonld files? Don't we need these contexts in production?

kdimak: Right, it was moved to the `*.jsonld` files of test data directory. 
I have no strong opinion on whether we should have the cached contexts (by default) in the production.
What do you think @troyronda ?

llorllale: I think the default should be cached. The question now is where to maintain them, in files or in go code....

Go code is more straightforward imo.

fqutishat: I agree we need them in Go code

kdimak: Yes, this makes sense. I added back the caching to the production code. It's also exported now: `CachingJSONLDLoader()`. This allows the users to use the default caching or apply their own via providing custom loader as `jsonld.ProcessorOpts`.

Issue #1798, opened 2020-05-07
Remove LegacyKMS from context.Provider


Baha-sk:
Since the new KMS is now fully functional, LegacyKMS should not be presented as an Aries option. To maintain the existing Packer's compatibility with non go Aries agents, LegacyKMS can be initialized only for this packer.

Also update Packer to take a KMS as an option and allow the current Packer to accept LegacyKMS, in preparation for the new Packer implementation to be used with the new local KMS and JWE document.

PR #1854, opened 2020-05-22
refactor: switch LegacyKMS to new LocalKMS

Baha-sk:
closes: #1798

Signed-off-by: Baha Shaaban <baha.shaaban@securekey.com>

PR #1855, opened 2020-05-22
feature: create Anoncrypt Packer

Baha-sk:
This change introduces the new Anoncrypt Packer. It is yet not the default packer
since didComm requires Authcrypt.

The packer also introduces the use of the Compact JWE Serialization for single recipient envelopes.

closes: #1798

### DIFF
--- a/pkg/didcomm/packer/anoncrypt/pack_test.go
+++ b/pkg/didcomm/packer/anoncrypt/pack_test.go
@@ -1,0 +1,190 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package anoncrypt
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/google/tink/go/keyset"
+	"github.com/stretchr/testify/require"
+
+	ecdhessubtle "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/localkms"
+	mockkms "github.com/hyperledger/aries-framework-go/pkg/mock/kms"
+	mockprovider "github.com/hyperledger/aries-framework-go/pkg/mock/provider"
+	mockstorage "github.com/hyperledger/aries-framework-go/pkg/mock/storage"
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock/noop"
+)
+
+func TestAnoncryptPackerSuccess(t *testing.T) {
+	k := createKMS(t)
+	_, recipientsKeys, keyHandles := createRecipients(t, k, 10)
+
+	anonPacker := New(newMockProviderWithCustomKMS(k), jose.A256GCM)
+
+	origMsg := []byte("secret message")
+	ct, err := anonPacker.Pack(origMsg, nil, recipientsKeys)
+	require.NoError(t, err)
+
+	msg, err := anonPacker.Unpack(ct)
+	require.NoError(t, err)
+
+	recKey, err := exportPubKeyBytes(keyHandles[0])
+	require.NoError(t, err)
+
+	require.EqualValues(t, &transport.Envelope{Message: origMsg, ToVerKey: recKey}, msg)
+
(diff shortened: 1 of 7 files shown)


Baha-sk:
must be merged after #1854

pkg/didcomm/packer/anoncrypt/pack_test.go
```diff
@@ -0,0 +1,144 @@
...
+	msg, err := anonPacker.Unpack(ct)
+	require.NoError(t, err)
+
+	recKey, err := exportPubKeyBytes(keyHandles[0])
+	require.NoError(t, err)
+
+	require.EqualValues(t, &transport.Envelope{Message: origMsg, ToVerKey: recKey}, msg)
+
+	// try with only 1 recipient
+	// TODO uncomment code when CompactSerlize accepts recipientsKeys headers. It currently rejects recipient headers.
```
troyronda, line 47: when is this planned? issue number?

Baha-sk: From what I've seen in go-jose, compactSerialize() does not accept recipient headers. Not sure if they do key wrapping either. I'll discuss with Derek today and come up with a plan.

llorllale:
@Baha-sk #1798 is already closed?

Baha-sk:
> @Baha-sk #1798 is already closed?

@llorllale currently I'm trying to figure out the difference between compactSerialize and fullSerialize as per this commented code snippet: https://github.com/hyperledger/aries-framework-go/pull/1855/files#diff-bcfad0c31440fa7cb5d7a1b721b3dafeR47

I will be pushing an update when this is resolved and uncomment this snippet.

pkg/didcomm/packer/anoncrypt/pack_test.go
Baha-sk: fixed encrypter/decrypter to accept CompactSerialize, this test has been uncommented and is passing now.

Baha-sk:
interop with Go-JOSE using CompactSerialization (1 sender) is not included in this PR. It will be part of a future one to avoid increasing the size of this PR

pkg/doc/jose/decrypter.go
```diff
@@ -79,22 +74,52 @@ func (jd *JWEDecrypt) Decrypt(jwe *JSONWebEncryption) ([]byte, error) {
...
+
 	return decPrimitive.Decrypt(encryptedData, authData)
 }
 
 func buildEncryptedData(encAlg string, jwe *JSONWebEncryption) ([]byte, error) {
 	var recipients []*subtle.RecipientWrappedKey
 
-	for _, recJWE := range jwe.Recipients {
-		rec, err := convertMarshalledJWKToRecKey(recJWE.Header.EPK)
+	if len(jwe.Recipients) == 1 { // compact serialization: it has only 1 recipient with no headers
```
llorllale, line 92: nit: this comment and the one besides the `else` - while not technically wrong - are kind of misleading because there is no JWT serialization here

Baha-sk: This is not for JWT serialization, it's about JWE serialization. This function (`convertMarshalledJWKToRecKey`) is only a utility function to get a recipient instance, regardless of the serialization type used here.

pkg/doc/jose/decrypter.go
```diff
@@ -107,6 +132,52 @@ func buildEncryptedData(encAlg string, jwe *JSONWebEncryption) ([]byte, error) {
...
+	}
+
+	epk := json.RawMessage{}
+
+	err = epk.UnmarshalJSON(epkBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	alg := ""
```
llorllale, line 157: why are we allowing `alg` and `kid` to be empty?

pkg/doc/jose/encrypter.go
```diff
@@ -103,12 +112,9 @@ func (je *JWEEncrypt) Encrypt(plaintext, aad []byte) (*JSONWebEncryption, error)
 		HeaderEncryption: je.encAlg,
 	}
 
-	// TODO - Go jose adds CEK as part of protectedHeaders, see if this is valid. Also, for a single recipient,
```
llorllale, line 106: was this done in this PR?

pkg/doc/jose/decrypter.go
Baha-sk: we don't, but if the sender didn't specify them (the sender being not an aries agent), then the default should be empty string, not `"<nil>"`

llorllale: that is a strange decision - aries-framework-go is for agents, not a generic JOSE library.

Baha-sk: JWE is not just for aries agents, it's generic.. It's implied that aries agents will always set these headers for a recipient, these conditions are in case someone (being a non aries agent) is sending a JWE message without these headers.

PR #1859, opened 2020-05-26
feat: Issue Credential - adds middleware

soluchok:
Adds middleware support to the service. 
Add your custom logic to the protocol by using `Use(...Middleware)` function.
The middleware functions will be executed every time before state execution.

This PR also includes the helper function which saves the credentials to the store.
It was done as an example of how to use the middleware function.
This function does not depend on the format of the attachment and assumes that the payload is a verifiable credential.
It can be easily changed when we have document/RFC which describes the payload for the specific format.

Note: The `Use` function is not exposed to the context. But, there is a possibility to use that.
```
service, err := ctx.Service(issuecredential.name)
if err != nil {
	/// ....
}
service.(interface {
	Use(items ...issuecredential.Middleware)
}).Use(func(next issuecredential.Handler) issuecredential.Handler {
	return issuecredential.HandlerFunc(func(metadata issuecredential.MetaData) error {
		// some logic here
		return next.Handle(metadata)
	})
})
```


pkg/didcomm/protocol/middleware/issuecredential/middlewares.go
```diff
@@ -0,0 +1,96 @@
...
+
+			return next.Handle(metadata)
+		})
+	}
+}
+
+func toVerifiableCredentials(vReg vdri.Registry, attachments []decorator.Attachment) ([]*verifiable.Credential, error) {
+	var credentials []*verifiable.Credential
+
+	// TODO: Currently, it supports only JSON payload. We need to add support for links and base64 as well. [Issue 1455]
```
llorllale, line 78: there's already support for base64

llorllale: Use `AttachmentData.Fetch()`

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -168,6 +178,11 @@ func New(p Provider) (*Service, error) {
 	return svc, nil
 }
 
+// Use allows providing middlewares
+func (s *Service) Use(middlewares ...Middleware) {
+	s.middlewares = append(s.middlewares, middlewares...)
```
llorllale, line 183: suggest you use [`stack`](https://pkg.go.dev/github.com/golang-collections/collections/stack) here so that the ordering is clear and that for-loop where you build the handler chain makes more sense

soluchok: Changed to:
```
func (s *Service) Use(items ...Middleware) {
	s.stack = append(s.stack, items...)
}
```

llorllale: I meant to use a `stack` as a data structure (hence the link to https://pkg.go.dev/github.com/golang-collections/collections/stack)....

soluchok: Oh, I got it :) I don't think we really need that. This is one more dependency for the project.
But I did a small refactor. And now it undesirable.
```
func (s *Service) Use(items ...Middleware) {
	var handler Handler = initialHandler
	for i := len(items) - 1; i >= 0; i-- {
		handler = items[i](handler)
	}

	s.middleware = handler
}
```

Issue #1852, opened 2020-05-22
separate package for utility signature signers


kdimak:
There are signature signers like `ed25519` or `cdsaSecp256k1` which are declared and used in tests in `doc/signature` , `doc/verifiable`  and `doc/jose` packages.
Let's put the common code into e.g. `pkg/doc/util/signature`.
Motivated by https://github.com/hyperledger/aries-framework-go/pull/1851#discussion_r429273531

PR #1860, opened 2020-05-26
refactor: separate package for utility signature signers

kdimak:

#1852

`pkg/doc/util/signature` is created with signers based on `Ed25519`, `ECDSA P256`, `ECDSA Secp256k1` and `RS256` signature algorithms.
It's used by `verifiable` package in this PR, to be used by `doc/signature` and `doc/jose` in the follow-up PRs.


kdimak:
> these additions belong in test files, not production.

Yes, these signature verifiers are used in tests only. But they will be used by several packages. And potentially a user can use them in production code as well.
And I will add a signer based on crypto/kms in a separate PR. It already exists in another package, I just need to move it here.

pkg/doc/verifiable/example_support_test.go
```diff
@@ -52,64 +47,6 @@ func (udc *UniversityDegreeCredential) MarshalJSON() ([]byte, error) {
...
-	return &ecdsaSecp256k1Signer{
-		privKey: privKey,
-	}
-}
-
-type ecdsaSecp256k1Signer struct {
-	privKey *ecdsa.PrivateKey
-}
-
-func (es *ecdsaSecp256k1Signer) Sign(payload []byte) ([]byte, error) {
```
Baha-sk, line 81: I see this signing logic removed here and https://github.com/hyperledger/aries-framework-go/pull/1860/files#diff-bc435fee068227dfd806579fb2b37277L222

I believe there were more instances of this code snippet (3 or 4 at least), will they be removed in a later PR?

kdimak: Yes.

Baha-sk:
@kdimak @llorllale as discussed, since the signing function in this patch is only used by tests, it is better to move it to an internal folder.

kdimak:
> @kdimak @llorllale as discussed, since the signing function in this patch is only used by tests, it is better to move it to an internal folder.

Let me do this in a follow-up PR with the usage of the signers in `signature` (Linked Data Proofs) package, I will make it soon.

PR #1861, opened 2020-05-27
feat: Present Proof - adds middleware

soluchok:
Adds middleware support to the service.


pkg/didcomm/protocol/middleware/presentproof/middlewares.go
```diff
@@ -0,0 +1,93 @@
...
+type MetaData presentproof.MetaData
+
+// Provider contains dependencies for the SavePresentation middleware function
+type Provider interface {
+	VerifiableStore() storeverifiable.Store
+	VDRIRegistry() vdri.Registry
+}
+
+// SavePresentation the helper function for the present proof protocol which saves the presentations
+func SavePresentation(p Provider) presentproof.Middleware {
```
llorllale, line 32: this code looks a lot like `issuecredential.SaveCredentials()`. Did you think about extracting a common function?

soluchok: From my point of view, it will be more confused. It is not something reusable. I think that the final implementation can be a bit different. Probably the function will depend much more on the protocol.

pkg/didcomm/protocol/presentproof/middleware.go
```diff
@@ -0,0 +1,41 @@
...
+// HandlerFunc is a helper type which implements the middleware Handler interface
+type HandlerFunc func(metadata MetaData) error
+
+// Handle implements function to satisfy the Handler interface
+func (hf HandlerFunc) Handle(metadata MetaData) error {
+	return hf(metadata)
+}
+
+// MetaData provides helpful information for the processing
+type MetaData interface {
```
troyronda, line 28: Metadata is a single word.

pkg/didcomm/protocol/middleware/presentproof/middlewares.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/presentproof"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdri"
+	storeverifiable "github.com/hyperledger/aries-framework-go/pkg/store/verifiable"
+)
+
+const stateNamePresentationReceived = "presentation-received"
+
+// MetaData is an alias to the original MetaData
+type MetaData presentproof.MetaData
```
troyronda, line 23: Metadata is a single word.

PR #1863, opened 2020-05-27
fix: Prevents generating a new id if it already exists

soluchok:

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service.go
+++ b/pkg/didcomm/protocol/didexchange/service.go
@@ -143,7 +143,7 @@ func (s *Service) HandleInbound(msg service.DIDCommMsg, _, _ string) (string, er
 	logger.Debugf("receive inbound message : %s", msg)
 
 	// fetch the thread id
-	thID, err := threadID(msg)
+	thID, err := msg.ThreadID()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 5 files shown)


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -530,12 +530,14 @@ func isNoOp(s state) bool {
 	return ok
 }
 
+// nolint: unparam
 func threadID(didCommMsg service.DIDCommMsg) (string, error) {
-	if didCommMsg.Type() == InvitationMsgType || didCommMsg.Type() == oobMsgType {
+	tID, err := didCommMsg.ThreadID()
```
llorllale, line 535: when is `didCommMsg.ThreadID()` returning an error? Why can't we remove `threadID()`?

soluchok: we can remove it, it returns an error when id and thread id is absent.

llorllale: Let's please remove it.

llorllale:
@soluchok can you please remove https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/states.go#L163 as well?

PR #1867, opened 2020-05-29
refactor: use signature signers in linked data proof package

kdimak:

#1852

Signature signers are used in the tests of the linked data proof package.
Signature signers package is not made an internal because it's used by two packages `verifiable` and `signature`.

Follow-up PR-s:
- add signer based on crypto/kms
- use signature singer in JWS related code

### DIFF
--- a/pkg/doc/util/signature/ecdsa.go
+++ b/pkg/doc/util/signature/ecdsa.go
@@ -11,6 +11,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
+	"errors"
 
 	"github.com/btcsuite/btcd/btcec"
 )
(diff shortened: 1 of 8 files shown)


kdimak:
I have an idea of how to combine using the crypto signers and hide the signers internals (to use `internal` package). I will use this approach in the next follow-up PR.

pkg/doc/util/signature/ecdsa.go
```diff
@@ -45,6 +76,26 @@ func GetECDSASecp256k1Signer(privKey *ecdsa.PrivateKey) *ECDSASigner {
 	return &ECDSASigner{privateKey: privKey, PublicKey: &privKey.PublicKey, hash: crypto.SHA256}
 }
 
+// NewECDSASigner creates a new ECDSA signer based on the input elliptic curve.
+func NewECDSASigner(curve elliptic.Curve) (*ECDSASigner, error) {
```
troyronda, line 80: whats the plan for Ed25519?

kdimak: It's already present: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/doc/util/signature/ed25519.go

Issue #1134, opened 2020-01-29
Configure Timeout decorator in Route Messages


rolsonquadras:
Currently, the route messages have hardcoded timeout values for the message response. These needs to removed and configured through `timeout` decorator.

https://github.com/hyperledger/aries-framework-go/blob/6ee5e904c7a5de2d862f2abd57864baf5f0be148/pkg/didcomm/protocol/route/service.go#L443

PR #1884, opened 2020-06-02
proposed solution to https://github.com/hyperledger/aries-framework-g…

m00sey:
**Title:**
Setting timeout on router client/service

**Description:**
Resolves https://github.com/hyperledger/aries-framework-go/issues/1134

Summary:

Decorator on router client for setting a timeout. If set the timeout is then propagated down to the route/mediator service, the previous default remains in place.

I chose to modify the interface with an additional setTimeout method.

I preferred this to changing the Register() signature but options there could be:

explicitly with a timeout:
`Register(connectionID string, timeout time.Duration) error`

introduce some form of options for Register:
`Register(connectionID string, opts ..*.Option) error`

This seemed the most backward compatible with what is out there.

My impetus for needing to override the existing five second default was when testing internally with the route client.

Usage:

`myRouter, err := route.New(ctx, route.WithTimeout(30 * time.Second))`

### DIFF
--- a/pkg/client/mediator/client_test.go
+++ b/pkg/client/mediator/client_test.go
@@ -10,10 +10,12 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/mediator"
 	mockroute "github.com/hyperledger/aries-framework-go/pkg/mock/didcomm/protocol/mediator"
 	mockprovider "github.com/hyperledger/aries-framework-go/pkg/mock/provider"
 )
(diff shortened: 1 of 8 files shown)


pkg/client/mediator/client_test.go
```diff
@@ -154,3 +198,7 @@ func TestClient_GetConfig(t *testing.T) {
 		require.True(t, errors.Is(err, expected))
 	})
 }
+
+func TestWithTimeout(t *testing.T) {
+
+}
```
soluchok, line 204: empty test

pkg/mock/didcomm/protocol/mediator/mock_mediator.go
```diff
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package mediator
 
 import (
+	"time"
+
```
soluchok, line 11: remove extra line

m00sey: `make checks` fails if I do.

pkg/client/mediator/client_test.go
m00sey: sloppy, apologies. removed.

pkg/mock/didcomm/protocol/mediator/mock_mediator.go
soluchok: Oh sorry, my fault. That is correct.

llorllale:
@m00sey 

> introduce some form of options for Register:
> Register(connectionID string, opts ..*.Option) error

This option aligns better with the framework's design patterns.

Also, the [`~timing`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0032-message-timing/README.md) decorator should be added on these outgoing messages.

m00sey:
> @m00sey 
> 
> > introduce some form of options for Register:
> > Register(connectionID string, opts ..*.Option) error
> 
> This option aligns better with the framework's design patterns.
> 
> Also, the [`~timing`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0032-message-timing/README.md) decorator should be added on these outgoing messages.

I'll begin work on these changes, thanks for the feedback.

m00sey:
I wasn't really happy with where the naming ended up. Options existed in the package already, it felt odd to be defining ClientOptions in the protocol package.

I'll defer to you guys on preferred naming.

PR #1885, opened 2020-06-03
feat: support single recipient compact JWE format

### LINKED ISSUE #1872: Add injectable function to tinkcrypto's composite implementation to compute AAD with single recipient
Currently the new Tink composite primitive generates ephemeral keys to wrap the JWE cek then Encrypts the JWE envelope.

This works well for multi recipients encryption as their headers are not merged in the JWE's `protectedheaders` section which is fixed and is used to compute the AAD for the encryption.

The problem lies in Compact Serialization where the single recipient headers are merged in `protectedheaders` section. Since the ephemeral keys are not available in the packer that calls the tink primitive, there is no way to compute the AAD with the recipient headers merged in `protectedheaders` prior to calling Encrypt().

This causes a problem where composite encryption will not encrypt the message with the proper AAD value. This fails interop JWE testing with external libraries like `go-jose` using Compact Serialization.

We need to figure out a way to ensure AAD is correctly updated with the right `protectedheaders` value prior to calling Encrypt. 

Probably, adding an injectable function option to Encrypt() that is executed after key wrapping and right before encryption in the tinkcrypto composite primitive is a possible solution.

Baha-sk:
This change adds the single recipient headers to the AAD in the ECDHES
Tink primitive to AAD to support JWE encryption with a single recipient 
for Compact serialization.

closes #1872

### DIFF
--- a/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package subtle
 
 import (
+	"crypto/elliptic"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"testing"
@@ -73,7 +75,7 @@ func TestEncryptDecryptNegativeTCs(t *testing.T) {
 
 	// Encrypt should fail with empty recipients public keys
 	_, err := cEnc.Encrypt(pt, aad)
-	require.Error(t, err)
+	require.EqualError(t, err, "ECDHESAEADCompositeEncrypt: missing recipients public keys for key wrapping")
 
 	// test with large key size
 	mEncHelper.KeySizeValue = 100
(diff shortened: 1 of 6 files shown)


pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
```diff
@@ -87,6 +88,14 @@ func TestEncryptDecryptNegativeTCs(t *testing.T) {
 
 	mEncHelper.KeySizeValue = 32
 
+	// Encrypt should fail with bad key type
+	cEnc.keyType = ecdhespb.KeyType_UNKNOWN_KEY_TYPE
+
+	_, err = cEnc.Encrypt(pt, aad)
+	require.Error(t, err)
```
sandrask, line 95: Should you be checking for certain error type/message?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
```diff
@@ -176,6 +185,54 @@ func TestEncryptDecryptNegativeTCs(t *testing.T) {
...
+	pt := []byte("secret message")
+	aad := []byte("aad message")
+
+	// test with single recipient public key
+	cEnc := NewECDHESAEADCompositeEncrypt(recipientsPubKeys, commonpb.EcPointFormat_UNCOMPRESSED.String(),
+		mEncHelper, ecdhespb.KeyType_EC)
+
+	// Encrypt should fail with aad not base64URL encoded
+	_, err := cEnc.Encrypt(pt, aad)
+	require.Error(t, err)
```
sandrask, line 208: Should you be checking for certain error type/message?

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
```diff
@@ -176,6 +185,54 @@ func TestEncryptDecryptNegativeTCs(t *testing.T) {
...
+
+	newAAD := base64.RawURLEncoding.EncodeToString(aad)
+
+	// Encrypt should fail with aad not being a marshalled json
+	_, err = cEnc.Encrypt(pt, []byte(newAAD))
+	require.Error(t, err)
```
sandrask, line 214: same as above

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
Baha-sk: added

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_test.go
Baha-sk: added

Baha-sk:
> broken unit tests

unit tests are now fixed and code coverage increased as well @llorllale

PR #1886, opened 2020-06-03
feat: Outofband - adds command support

### LINKED ISSUE #1882: Out-of-band - add command support

soluchok:
Closes #1882 


pkg/controller/command/outofband/command.go
```diff
@@ -0,0 +1,223 @@
...
+	// registers action channel to listen for events
+	if err := client.RegisterActionEvent(actions); err != nil {
+		return nil, fmt.Errorf("register action event: %w", err)
+	}
+
+	// this code listens for the action events and does nothing
+	// this trick is used to avoid error such as `no clients are registered to handle the message`
+	go func() {
+		for range actions {
+		}
```
rolsonquadras, line 77: We need to notify the consumers(rest/wasm) when we receive the events. Is it planned for future PRs ?

soluchok: I think it is a common problem. We do not have that mechanism for now. We have something custom for didexchange(notifier) but I don't want to repeat that. There should be a common logic for all protocols. Since all protocols are using the same common structures for the `DIDCommAction` and `StateMsg`.  I have not planned it yet. But I know about it.

rolsonquadras: Ok. Just wanted to check as it's an essential components for the controller (wasm/rest controllers need to listen to events rather than polling for events).

llorllale: @soluchok please create a followup issue for this

soluchok: Already did) https://github.com/hyperledger/aries-framework-go/issues/1887

PR #1904, opened 2020-06-09
feat: Introduce - adds AcceptProposal method

soluchok:
* adds `AcceptProposal` method
* corrects swagger


pkg/controller/command/introduce/models.go
```diff
@@ -158,3 +158,18 @@ type DeclineRequestArgs struct {
...
 //
 type DeclineRequestResponse struct{}
+
+// AcceptProposalArgs model
+//
+// This is used for accepting a proposal.
+//
+type AcceptProposalArgs struct {
+	// PIID Protocol instance ID
+	PIID string `json:"piid,omitempty"`
```
llorllale, line 168: What's the plan for accepting a proposal by providing an oob message?

soluchok: We have a possibility to do that.
`AcceptProposal` can be used for accepting a proposal without providing an OOBRequest. e.g introducer sent two proposals to the introduces. One of the introduces uses AcceptProposalWithOOBRequest and another AcceptProposal.

PR #1908, opened 2020-06-11
refactor: Minor controller refactoring

soluchok:
* fixes swagger documentation
* changes `presentproof` and `issuecredential` request payload

### DIFF
--- a/cmd/aries-js-worker/src/agent-rest-client.js
+++ b/cmd/aries-js-worker/src/agent-rest-client.js
@@ -329,7 +329,7 @@ export const Client = class {
                 method: r.method,
                 url: url,
                 headers: headers,
-                data: (request.payload && 'body' in request.payload) ? request.payload.body : request.payload
+                data: request.payload,
             });
 
         return resp.data;
--- a/pkg/controller/command/introduce/command.go
+++ b/pkg/controller/command/introduce/command.go
@@ -8,6 +8,7 @@ package introduce
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 
@@ -148,7 +149,7 @@ func (c *Command) SendProposal(rw io.Writer, req io.Reader) command.Error {
 
 	if len(args.Recipients) != maxRecipients {
 		logutil.LogDebug(logger, commandName, sendProposal, errTwoRecipients)
-		return command.NewValidationError(InvalidRequestErrorCode, fmt.Errorf(errTwoRecipients))
+		return command.NewValidationError(InvalidRequestErrorCode, errors.New(errTwoRecipients))
 	}
 
 	if err := c.client.SendProposal(args.Recipients[0], args.Recipients[1]); err != nil {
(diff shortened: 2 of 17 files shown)


pkg/controller/command/issuecredential/models.go
```diff
@@ -14,7 +14,7 @@ import "github.com/hyperledger/aries-framework-go/pkg/client/issuecredential"
 //
 type AcceptProposalArgs struct {
 	// PIID Protocol instance ID
-	PIID string `json:"piid"`
+	PIID string `json:"piid,omitempty"`
```
llorllale, line 17: why declare this as optional when we still have this check? https://github.com/hyperledger/aries-framework-go/blob/4a9b7fa924cd8aa3f218fb910eff42060cb01266/pkg/controller/command/issuecredential/command.go#L275-L278

pkg/controller/command/issuecredential/models.go
```diff
@@ -32,7 +32,7 @@ type AcceptProposalResponse struct{}
 //
 type AcceptOfferArgs struct {
 	// PIID Protocol instance ID
-	PIID string `json:"piid"`
+	PIID string `json:"piid,omitempty"`
```
llorllale, line 35: same here - PIID is required

pkg/controller/command/issuecredential/models.go
soluchok: It allows reusing `AcceptProposalArgs ` structure. We are getting `piid` from the URL.

llorllale: That's not really RESTful, is it (then again, the whole API isn't...). If the ID is in the URI, then the more appropriate HTTP method would be `PUT` but it still remains `POST`. See [REST – PUT vs POST](https://restfulapi.net/rest-put-vs-post/).

Any reason why you can't set the PIID in the BDD tests?

llorllale: I see the piid on line 161 in `issuecredential_controller_steps.go`. This seems like an unnecessary change.

soluchok: Removed `omitempty`. 
`I see the piid on line 161 in issuecredential_controller_steps.go. This seems like an unnecessary change.` this change is necessary.

llorllale: @soluchok what I mean is the change to have the option of providing the piid in the URI is unnecessary. The PIID can be passed in the request body, and I don't see a reason why it isn't done so in the BDD tests.

soluchok: I am not sure about it. Currently, the URL looks like the following:
`/protocolname/{piid}/accept-proposal`
If we want to provide `piid` in the body we need to change the URL to
`/protocolname/accept-proposal`. 
Which actually affects everything. (unit/BDD/js tests, swagger etc.)
I cannot provide `piid` without those changes. 
If we really need it it can be done in a separate task.

llorllale: Oh I see :/

soluchok: How about `omitempty`? I would like to add it to every field. It will not affect anything.

Issue #811, opened 2019-11-14
Support for User Authentication


rolsonquadras:
DIDComm mediator stores and forwards the messages. There should be a way to authenticate the user, so that they can access their store/mailbox.

rolsonquadras:
This feature can be divided into following components:
- Service Layer
- Store the messages when registered agent is unavailable
- Integrate with framework
- Client Layer
- BDD tests

m00sey:
Can you elaborate on what is required for "Integrate with framework"

Internally we initialized using

```	
framework, err := aries.New(
		aries.WithTransportReturnRoute("all"),
		aries.WithOutboundTransports(ws.NewOutbound()),
		aries.WithStoreProvider(mem.NewProvider()),
		aries.WithProtocols(messagepickupSvc.ServiceCreator(), basicmessageSvc.ServiceCreator()),
	)
if err != nil {
    nil, err
}
```

llorllale:
@m00sey the question on my mind is: how will the `messagepickup` protocol be integrated with the `mediator` one?

m00sey:
We had our own router implementation but didn't get it submitted before you published `mediator`, I am going to be working on that integration.

rolsonquadras:
> Can you elaborate on what is required for "Integrate with framework"
> 
> Internally we initialized using
> 
> ```
> framework, err := aries.New(
> 		aries.WithTransportReturnRoute("all"),
> 		aries.WithOutboundTransports(ws.NewOutbound()),
> 		aries.WithStoreProvider(mem.NewProvider()),
> 		aries.WithProtocols(messagepickupSvc.ServiceCreator(), basicmessageSvc.ServiceCreator()),
> 	)
> if err != nil {
>     nil, err
> }
> ```

It was been added to the default framework in https://github.com/hyperledger/aries-framework-go/pull/2001/files#diff-da8e5280b0328cb616f0b42390a35b00R72.

PR #1926, opened 2020-06-15
feat:  messagebox

m00sey:
**Title:**
proposed implementation for a messagebox feature

**Description:**
implements https://github.com/hyperledger/aries-framework-go/issues/811 service layer

Summary:

Introduction of a messagepickup service for storing messages until and edge agent collects.

I'd like to credit @pfeairheller for the original implementation.

Since this violates the ~500 line PR rule I wanted to get it out early for discussion.
I am still evaluating what BDD tests might be needed and will mark this current PR as draft.

I am making a corresponding PR to aries-rfcs, to update the type values of messages.
https://github.com/hyperledger/aries-rfcs/pull/500

### DIFF
--- a/pkg/didcomm/protocol/messagepickup/service.go
+++ b/pkg/didcomm/protocol/messagepickup/service.go
@@ -1,0 +1,569 @@
+/*
+Copyright Scoir Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package messagepickup
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	commtransport "github.com/hyperledger/aries-framework-go/pkg/didcomm/common/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
+)
+
+const (
+	// MessagePickup defines the protocol name
+	MessagePickup = "messagepickup"
+	// Spec defines the protocol spec
+	Spec = "https://didcomm.org/messagepickup/1.0/"
+	// StatusMsgType defines the protocol propose-credential message type.
+	StatusMsgType = Spec + "status"
+	// StatusRequestMsgType defines the protocol propose-credential message type.
+	StatusRequestMsgType = Spec + "status-request"
+	// BatchPickupMsgType defines the protocol offer-credential message type.
+	BatchPickupMsgType = Spec + "batch-pickup"
+	// BatchMsgType defines the protocol offer-credential message type.
+	BatchMsgType = Spec + "batch"
+	// NoopMsgType defines the protocol request-credential message type.
+	NoopMsgType = Spec + "noop"
+)
+
+const (
+	updateTimeout = 50 * time.Second
+
+	// Namespace is namespace of messagepickup store name
+	Namespace = "mailbox"
+)
+
+// ErrConnectionNotFound connection not found error
+var ErrConnectionNotFound = errors.New("connection not found")
+var logger = log.New("aries-framework/messagepickup")
+
+type provider interface {
+	OutboundDispatcher() dispatcher.Outbound
+	StorageProvider() storage.Provider
+	TransientStorageProvider() storage.Provider
+}
+
+type connections interface {
+	GetConnectionRecord(string) (*connection.Record, error)
+}
+
+// Service for the messagepickup protocol
+type Service struct {
+	connectionLookup connections
+	outbound         dispatcher.Outbound
+	msgStore         storage.Store
+	packager         commtransport.Packager
+	msgHandler       transport.InboundMessageHandler
+	batchMap         map[string]chan Batch
+	batchMapLock     sync.RWMutex
+	statusMap        map[string]chan Status
+	statusMapLock    sync.RWMutex
+	inboxLock        *lockbox
+}
+
+// ServiceCreator for the messagepickup protocol
+func ServiceCreator() api.ProtocolSvcCreator {
+	return func(prv api.Provider) (dispatcher.ProtocolService, error) {
+		tp, ok := prv.(transport.Provider)
+		if !ok {
+			return nil, errors.New("failed to cast transport provider")
+		}
+
+		return New(prv, tp)
+	}
+}
+
+// New returns the messagepickup service
+func New(prov provider, tp transport.Provider) (*Service, error) {
+	store, err := prov.StorageProvider().OpenStore(Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("open mailbox store : %w", err)
+	}
+
+	connectionLookup, err := connection.NewLookup(prov)
+	if err != nil {
+		return nil, err
+	}
+
+	svc := &Service{
+		outbound:         prov.OutboundDispatcher(),
+		msgStore:         store,
+		connectionLookup: connectionLookup,
+		packager:         tp.Packager(),
+		msgHandler:       tp.InboundMessageHandler(),
+		batchMap:         make(map[string]chan Batch),
+		statusMap:        make(map[string]chan Status),
+		inboxLock:        newLockBox(),
+	}
+
+	return svc, nil
+}
+
+// HandleInbound handles inbound message pick up messages
+func (s *Service) HandleInbound(msg service.DIDCommMsg, myDID, theirDID string) (string, error) {
+	// perform action asynchronously
+	go func() {
+		var err error
+
+		switch msg.Type() {
+		case StatusMsgType:
+			err = s.handleStatus(msg)
+		case StatusRequestMsgType:
+			err = s.handleStatusRequest(msg, myDID, theirDID)
+		case BatchPickupMsgType:
+			err = s.handleBatchPickup(msg, myDID, theirDID)
+		case BatchMsgType:
+			err = s.handleBatch(msg)
+		case NoopMsgType:
+			err = s.handleNoop(msg)
+		}
+
+		if err != nil {
+			logger.Errorf("Error handling message: (%w)\n", err)
+		}
+	}()
+
+	return msg.ID(), nil
+}
+
+// HandleOutbound adherence to dispatcher.ProtocolService
+func (s *Service) HandleOutbound(_ service.DIDCommMsg, _, _ string) error {
+	return errors.New("not implemented")
+}
+
+// Accept checks whether the service can handle the message type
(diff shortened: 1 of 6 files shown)


llorllale:
@m00sey I think it's headed in the right direction. I am afraid of the coupled `getInbox/putInbox` pairs in `AddMessage` and `pullMessages`, especially the latter. Looks dangerous and prone to data loss. I think a lock on `theirDID` is needed.

m00sey:
> @m00sey I think it's headed in the right direction. I am afraid of the coupled `getInbox/putInbox` pairs in `AddMessage` and `pullMessages`, especially the latter. Looks dangerous and prone to data loss. I think a lock on `theirDID` is needed.

I added additional locks around each `theirDID` and a map to manage those mutex.

I am looking into why it thinks I have touched the additional six files, I don't see them in my `Files Changed` list and `make checks` passes locally.

llorllale:
@m00sey 

> > @m00sey I think it's headed in the right direction. I am afraid of the coupled `getInbox/putInbox` pairs in `AddMessage` and `pullMessages`, especially the latter. Looks dangerous and prone to data loss. I think a lock on `theirDID` is needed.
> 
> I added additional locks around each `theirDID` and a map to manage those mutex.
> 
> I am looking into why it thinks I have touched the additional six files, I don't see them in my `Files Changed` list and `make checks` passes locally.

Sorry for the late reply. I think the code looks correct, but now I regret having suggested to use locks. The code gets so complicated. 

But don't worry - we can merge this because it looks like it works without data loss. Just need to pass the checks and I think we can merge it.

m00sey:
Agreed locking has increased the complexity, open to other thoughts on how to block access to avoid data loss.

llorllale:
@m00sey please rebase on `master`

m00sey:
> @m00sey please rebase on `master`

Hm, so I rebased using `git rebase=true upstream master` where `upstream` is `git@github.com:hyperledger/aries-framework-go.git` resolved the conflicts and pushed, I wasn't expecting the additional changes to be tracked in my PR...

Is there a different/preferred way? Apologies.

llorllale:
> I'll work on the coverage today, I thought I had scraped past 85%, I am guessing it's a lines vs statements difference?
> 
> ```
> ┌─[±][issue-811 → origin ✓][messagepickup][]
> └─▪ go test -coverprofile=coverage.out && go tool cover -html=coverage.out
> PASS
> coverage: 87.0% of statements
> ok  	github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/messagepickup	0.004s
> ```

It's the coverage on the *patch* (delta) that's holding this back. See the report [here](https://codecov.io/gh/hyperledger/aries-framework-go/compare/b180f85613300a68bca65ed7eabf9cdeeea13435...b11f4b06e69b14257f68afc9b3ed412a7a610f78/diff).

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+}
+
+type connections interface {
+	GetConnectionRecord(string) (*connection.Record, error)
+}
+
+// Service for the messagepickup protocol
+type Service struct {
+	service.Action
+	service.Message
```
soluchok, line 72: I don't see the usage of `service.Action` or `service.Message`. Should be removed.

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+	batchMapLock     sync.RWMutex
+	statusMap        map[string]chan Status
+	statusMapLock    sync.RWMutex
+	inboxMap         map[string]*lockBox
+	inboxMapLock     sync.RWMutex
+}
+
+type lockBox struct {
+	mux sync.RWMutex
+}
```
soluchok, line 88: If you need lock by key please check  the following links:
https://github.com/kubernetes/utils/blob/master/keymutex/hashed.go
https://github.com/im7mortal/kmutex

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+	return MessagePickup
+}
+
+func (s *Service) handleStatus(msg service.DIDCommMsg) error {
+	// unmarshal the payload
+	statusMsg := &Status{}
+
+	err := msg.Decode(statusMsg)
+	if err != nil {
+		return fmt.Errorf("status message unmarshal: %v", err)
```
soluchok, line 181: Use `%w` instead of `%v`

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+func (r *inbox) DecodeMessages() ([]*Message, error) {
+	var out []*Message
+	err := json.Unmarshal(r.Messages, &out)
+
+	if err != nil {
+		out = []*Message{}
+	}
+
+	return out, nil
+}
```
soluchok, line 302: The error should be returned

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+
+// EncodeMessages Messages
+func (r *inbox) EncodeMessages(msg []*Message) error {
+	d, err := json.Marshal(msg)
+	if err != nil {
+		return fmt.Errorf("unable to marshal: %v", err)
```
soluchok, line 308: again  %w

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+	m := Message{
+		ID:        uuid.New().String(),
+		AddedTime: time.Now(),
+		Message:   message,
+	}
+
+	msgs = append(msgs, &m)
+
+	outbox.LastDeliveredTime = time.Now()
+	outbox.LastRemovedTime = time.Now()
```
soluchok, line 339: outbox.LastRemovedTime = outbox.LastDeliveredTime

pkg/didcomm/protocol/messagepickup/service_test.go
```diff
@@ -0,0 +1,906 @@
...
+		OutboundDispatcherValue:       nil,
+	}, &mockTransportProvider{
+		packagerValue: &mockPackager{},
+	})
+
+	return svc, err
+}
+
+// mockProvider mock provider
+type mockTransportProvider struct {
```
soluchok, line 849: Why not to use `gomock` ?

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+
+	trans := &decorator.Transport{}
+	err = json.Unmarshal(unpackMsg.Message, trans)
+
+	if err != nil {
+		return fmt.Errorf("unmarshal transport decorator : %v", err)
+	}
+
+	// TODO: what does this do?
+	// if trans != nil && trans.ReturnRoute != nil && trans.ReturnRoute.Value == decorator.TransportReturnRouteAll {
```
soluchok, line 600: @rolsonquadras ^

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+		if errors.Is(err, storage.ErrDataNotFound) {
+			return nil, ErrConnectionNotFound
+		}
+
+		return nil, fmt.Errorf("fetch connection record from store : %w", err)
+	}
+
+	return conn, nil
+}
+
```
soluchok, line 544: Can be simplified 
```
func (s *Service) getConnection(routerConnID string) (*connection.Record, error) {
	conn, err := s.connectionLookup.GetConnectionRecord(routerConnID)
		return nil, ErrConnectionNotFound
	}

	return conn, err
}
``

File: pkg/didcomm/protocol/messagepickup/service.go [hunk 10]
```diff
@@ -0,0 +1,612 @@
...
+	defer s.releaseInboxLock(theirDID)
+
+	err = s.msgStore.Put(theirDID, b)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
soluchok, line 438: `return s.msgStore.Put(theirDID, b)` or wrap an error

pkg/didcomm/protocol/messagepickup/service.go
```diff
@@ -0,0 +1,612 @@
...
+	}
+
+	msgs, err := outbox.DecodeMessages()
+	if err != nil {
+		return nil, fmt.Errorf("unable to pull messages: %v", err)
+	}
+
+	c := len(msgs)
+	end := int(math.Min(float64(size), float64(c)))
+
```
soluchok, line 367: Please remove it.

pkg/didcomm/protocol/messagepickup/service.go
rolsonquadras: This is used to store the the socket connection in pool(implemented only in websocket implementation). When an [edge agents](https://github.com/hyperledger/aries-framework-go/blob/master/docs/didcomm_mediator.md#edge-agents-without-inbound-capability) (without inbound capability opens a websocket connection with router), the router will keep the connection in the pool, so that it can send messages back to edge-agent. This logic is not required here and we can remove this commented code.

rolsonquadras:
> **Title:**
> proposed implementation for a messagebox feature
> 
> **Description:**
> closes #811
> 

@m00sey Thanks for the pull request. 

One minor note - This PR doesn't close #811 as this only implements service layer. We would need to add following to make this feature complete.
- Store the messages when registered agent is unavailable
- Integrate with framework 
- Client Layer
- BDD tests

pkg/didcomm/protocol/messagepickup/service_test.go
llorllale: @soluchok this mocking is fine. I'm not a fan of gomock. It's slowing down the builds and there is *baaaarely* any return on the investment.

pkg/didcomm/protocol/messagepickup/service.go
llorllale: I think it's a good idea to wrap the error that is not `storage.ErrDataNotFound`.

pkg/didcomm/protocol/messagepickup/service.go
m00sey: Are you saying `end := int(math.Min(float64(size), float64(len(msgs))))` is preferred, or do you not like the math dance in general?

m00sey:
@rolsonquadras understood, I was overzealous. I will remove that from the description in order to keep this PR moving. I plan to address the subsequent parts in future PRs.

pkg/didcomm/protocol/messagepickup/service.go
soluchok: Sorry I got confused. I see what you doing.
nit: it would be more readable 
```
	end := len(msgs)
	if size <end {
		end = size
	}
```

The lock/unlock work incorrectly. You're locking the database instead of your logic. The database is thread-safe. 
Locks in `getInbox` and `putInbox` don’t  make sense.

llorllale: I think they allow thread-safe operation within putInbox()  and getInbox() respectively, but should be refactored such that thread-safety is guaranteed *across* putInbox() and getInbox() (eg. in `AddMessage()`).

llorllale: I think the easiest way would be to have a queue of operations and 1 go routine processing them. Each protocol message will have its own operation.

pkg/didcomm/protocol/messagepickup/service.go
m00sey: @soluchok / @llorllale can one of you reslove this, I am easy as to which to adopt.

soluchok: it is ok, you can leave as it is

llorllale:
@m00sey thanks!

PR #1933, opened 2020-06-17
feat: Added MyDID and TheirDID to the action event

### LINKED ISSUE #1929: Unable to get connection of the DidComm message
**What I'm trying to do**

I get issue credential message in an action event. I want to get some info of the connection that send it to me.

**Expected result**

I have a field or method in DidCommMessage or DidCommAction to get connection info .

**Actual result**

I have no way to figure out the connection of the message.

soluchok:
The following protocols were modified:
* presentproof
* introduce
* issuecredential

Closes #1929 

### DIFF
--- a/pkg/didcomm/protocol/introduce/service.go
+++ b/pkg/didcomm/protocol/introduce/service.go
@@ -68,18 +68,16 @@ type Recipient struct {
 // Action contains helpful information about action
 type Action struct {
 	// Protocol instance ID
-	PIID string                `json:"piid"`
-	Msg  service.DIDCommMsgMap `json:"msg"`
+	PIID     string
+	Msg      service.DIDCommMsgMap
+	MyDID    string
+	TheirDID string
 }
 
 // transitionalPayload keeps payload needed for Continue function to proceed with the action
 type transitionalPayload struct {
-	// Protocol instance ID
-	PIID      string
+	Action
 	StateName string
-	Msg       service.DIDCommMsgMap
-	MyDID     string
-	TheirDID  string
 }
 
 // metaData type to store data for internal usage
@@ -231,8 +229,10 @@ func (s *Service) doHandle(msg service.DIDCommMsg, outbound bool) (*metaData, er
 	return &metaData{
 		transitionalPayload: transitionalPayload{
 			StateName: next.Name(),
-			Msg:       msg.(service.DIDCommMsgMap),
-			PIID:      piID,
+			Action: Action{
+				Msg:  msg.(service.DIDCommMsgMap),
+				PIID: piID,
+			},
 		},
 		state:    next,
 		msgClone: msg.Clone(),
@@ -349,6 +349,10 @@ func (s *Service) newDIDCommActionMsg(md *metaData) service.DIDCommAction {
 			s.processCallback(md)
 		},
 		Stop: func(err error) { actionStop(customError{error: err}) },
+		Properties: &eventProps{
+			myDID:    md.MyDID,
+			theirDID: md.TheirDID,
+		},
 	}
 }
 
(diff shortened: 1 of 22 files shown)


pkg/didcomm/protocol/introduce/service.go
```diff
@@ -349,6 +351,10 @@ func (s *Service) newDIDCommActionMsg(md *metaData) service.DIDCommAction {
 			s.processCallback(md)
 		},
 		Stop: func(err error) { actionStop(customError{error: err}) },
+		Properties: &eventProps{
+			myDID:    md.MyDID,
```
rolsonquadras, line 355: Why can't we just return connectionID instead of myDID and theirDID ? The consumers can find these details using the connectionID, if required. In most of the cases, the consumer would just be interested in knowing the sender. ie, `alice` or `bob` etc.

llorllale: myDID + theirDID are readily available in the protocol implementations (`HandleInbound(msg, myDID, theirDID)`)

rolsonquadras: Yeah, but I feel those lower layer details are not meaningful to the consumers. As a consumer, I just want to know who sent me the message without knowing how to derive it by myself. I guess, it's a discussion about who(consumers or service) should fetch the connectionID from myDID and theirDID value.

llorllale: `connectionID` a lower level concept than DIDs.

Besides that, let's say we return the connectionID. The consumer would use `connection.Lookup.GetConnectionByID()` and get a `connection.Record`... where they see DIDs.

This solution is a good, speedy one that can be uniformly implemented with minimum hassle across all protocol services.

rolsonquadras: > `connectionID` is a lower level concept than DIDs.
> 

I see this differently. DID is just a mean to connect to other agent. At higher level, the user would need the name/label of the sender. This label is tied to connectionID in the framework and connectionID in-turn has myDID and theirDID to help facilitate the communication.

> Besides that, let's say we return the connectionID. The consumer would use `connection.Lookup.GetConnectionByID()` and get a `connection.Record`... where they see DIDs.
> 
> This solution is a good, speedy one that can be uniformly implemented with minimum hassle across all protocol services.
I agree that myDID and theirDID is easy to implement.

llorllale: Both roads lead to `connection.Record`. It's easier with myDID + theirDID.

pkg/didcomm/protocol/issuecredential/service.go
```diff
@@ -58,11 +58,11 @@ type customError struct{ error }
 // transitionalPayload keeps payload needed for Continue function to proceed with the action
 type transitionalPayload struct {
 	// Protocol instance ID
-	PIID      string
-	StateName string
-	Msg       service.DIDCommMsgMap
-	MyDID     string
-	TheirDID  string
+	PIID      string                `json:"piid"`
```
llorllale, line 61: is there really a need for these struct tags?

llorllale:
Is there a need for all those struct tags?

pkg/didcomm/protocol/issuecredential/service.go
soluchok: I am reusing the payload. I am saving data as `transitionalPayload` structure. But when I read the data I am using a different `Action` structure. 
https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/issuecredential/service.go#L539

soluchok: Removed!

PR #1935, opened 2020-06-17
feat: add ECDH1PU key templates + P-384 and P-521 key templates

### LINKED ISSUE #1912: Create ECDH-1PU default key templates
for now P-256 keys, support for more keys will be added later.

### LINKED ISSUE #1934: Support P-384 and P-521 curves in Tink Composite key templates
Currently, only P-256 key template is available. Add support for P-384 and P-521 key templates.

Baha-sk:
- Introduce ECDH-1PU key templates.

- Separate Composite key writer into its own package (keyio) to reduce code duplication as much as possible.

- Add P-384 and P-521 key templates for both ECDH-ES and ECDH-1PU primitives.

closes #1912
closes #1934


pkg/crypto/tinkcrypto/primitive/composite/keyio/composite_key_pubkey_to_handle.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// ECDH1PU composite primitive type
+	ECDH1PU
+
+	// ECDHES composite primitive type
+	ECDHES
+)
+
+// publicKeyBytesToHandle for reading pubKey and get keyset.Handle of either ECDHES or ECDH1PU composite public key type
+func publicKeyBytesToHandle(pubKey []byte, compositeType PrimitiveCompositeType) (*keyset.Handle, error) {
```
llorllale, line 41: only used within tests

Baha-sk: Good catch, this functionally was never used by ECDH-ES originally. Similar function exists in localkms's public key reader to convert a Public key to a Tink key handle. It should be updated to support ECDH-ES/ECDH-1PU keys as well (if needed in the future).

I will remove this function.

Baha-sk: function is now removed

PR #1938, opened 2020-06-18
feat: Setup Mysql storage

### LINKED ISSUE #1166: Support SQL Storage Provider (sqlite)
Support usage of a SQL storage provider backend.

In particular, start by supporting sqlite.

talwinder50:
closes #1166


pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+		db:     db,
+		dbs:    map[string]*sqlDBStore{}}
+
+	return p, nil
+}
+
+// OpenStore opens and returns new db for given name space.
+func (p *Provider) OpenStore(name string) (storage.Store, error) {
+	// creating the database
+	_, err := p.db.Exec("CREATE DATABASE IF NOT EXISTS `name`")
```
llorllale, line 71: need to `fmt.Sprintf` the `name` arg

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create db: %w", err)
+	}
+
+	// Use is used to select the created database without this DDL operations are not permitted
+	_, err = p.db.Exec("USE `name`")
```
llorllale, line 77: same here with the `name` arg

llorllale: actually, I think this is not needed. You should open a new `*sql.DB` and assign it to `sqlDBStore.db` below

llorllale: The new `*sql.DB` should point to the newly created database, of course

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to use db: %w", err)
+	}
+
+	store := &sqlDBStore{
+		db:     p.db,
+		dbName: name}
+
+	p.dbs[name] = store
```
llorllale, line 86: forgot to guard this with `p.Lock()`

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	defer p.Unlock()
+
+	store, exists := p.dbs[name]
+	if !exists {
+		return nil
+	}
+
+	delete(p.dbs, name)
+
+	conn, err := store.db.Conn(context.Background())
```
llorllale, line 124: should just be closing `store.db` - remember what I mentioned here: https://github.com/hyperledger/aries-framework-go/pull/1938/files#r442128944

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	return conn.Close()
+}
+
+// Put stores the key and the record
+func (s *sqlDBStore) Put(k string, v []byte) error {
+	if k == "" || v == nil {
+		return errors.New("key and value are mandatory")
+	}
+
+	createTableStmt := "CREATE Table IF NOT EXISTS `tablePrefix s.dbName` (`key` varchar(50) NOT NULL, `value` BLOB, " +
```
llorllale, line 138: this should've been done when instantiating `sqlDBStore`, not on every `Put()`

llorllale: also, is `varchar(50)` enough width?

llorllale: MySQL will complain if you try to create indexes that are wider than its configured maximum.

Instead, control the width of the key inserted into the database by hashing that. Set the width of the varchar appropriately.

llorllale: ... which reminds me, I don't see the creation of an index on column `key`

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	if k == "" {
+		return errors.New("key is mandatory")
+	}
+
+	res, err := s.db.Exec("DELETE FROM `tablePrefix s.dbName` WHERE `key`=?", k)
+	if err == nil {
+		count, e := res.RowsAffected()
+		if e == nil {
+			if count == 0 {
+				return errors.New("failed to delete row")
```
llorllale, line 201: neither the leveldb implementation nor the mem implementation of `Store.Delete()` return an error if the key doesn't exist. I'm not sure if the couchDB one does.

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	resultRows *sql.Rows
+	err        error
+}
+
+func (s *sqlDBStore) Iterator(startKey, endKey string) storage.StoreIterator {
+	resultRows, err := s.db.Query("SELECT * FROM `tablePrefix s.dbName` WHERE `key` LIKE ?", startKey+"%")
+	if err != nil {
+		if err = resultRows.Err(); err != nil {
+			return &sqlDBResultsIterator{store: s, resultRows: &sql.Rows{},
+				err: fmt.Errorf("no rows found")}
```
llorllale, line 222: please confirm if this is behavour is expected

pkg/storage/mysql/mysqlstore_test.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/stretchr/testify/require"
+
+	_ "github.com/go-sql-driver/mysql"
+)
+
+const (
+	sqlStoreDBURL = "root:my-secret-pw@tcp(127.0.0.1:3306)/"
+)
+
+// For these unit tests to run, you must ensure you have a SQLDB instance running at the URL specified in sqlStoreDBURL.
```
llorllale, line 24: Please have MySQL spooled up by the tests in this package (`TestMain`) because they're the only unit tests that require it.

pkg/storage/mysql/mysqlstore.go
talwinder50: Confirmed couch db does provide this feature

pkg/storage/mysql/mysqlstore.go
soluchok: `endKey ` here is unused. Replace `endKey` with`_`. The query will not work. 
As I know you need to find data between `startKey and endKey`. `LIKE` doesn't do that.

pkg/storage/mysql/mysqlstore.go
talwinder50: As per the other store implementations we have two functions close provider which basically closes the store.db and then close store which closes the specific opened connection.

pkg/storage/mysql/mysqlstore.go
talwinder50: Some how this doesn't seem to work without the USE Statement. I always get NO Database selected even when i opened the new sql connection with newly created database. Even the official documentation suggest this https://dev.mysql.com/doc/refman/8.0/en/creating-database.html

llorllale: That doc assumes you're still on the same connection. I'm suggesting you open a separate connection for the new database. You'd need to provide a new URL to open that connection.

pkg/storage/mysql/mysqlstore.go
soluchok: @talwinder50  The simplest idea is to use `BETWEEN` operator. You need to create the following table:

|  ID   | KEY | VAL |
| ---- | ---- | ---- |
| 1      | start | v1    |
| 2      | end | v1    |
Where :
ID - should be an integer (auto-increment).
KEY - unique index 

```
SELECT * FROM table
WHERE key BETWEEN "start" AND "end"
ORDER BY ID;
```

Notice that we also need to find data by the key which is not fully presented. e.g `start` -> `sta`.
In that case, you can extend the query above with the subquery (using `LIKE` operator).

llorllale: @soluchok why should we order by `ID` instead of `key`?

soluchok: @llorllale As I know we need to keep consistency. My point is to add an ID(auto increment) which will guarantee the sequence.
But probably it will even work without an ID field. I did not check it.

llorllale: It will work without the `ID` field. The `k` arg is the key'.

pkg/storage/mysql/mysqlstore.go
talwinder50: done

pkg/storage/mysql/mysqlstore.go
talwinder50: done

pkg/storage/mysql/mysqlstore.go
talwinder50: Creating new connection as suggested

pkg/storage/mysql/mysqlstore.go
talwinder50: updated

pkg/storage/mysql/mysqlstore.go
talwinder50: This PR is getting bigger, I have increased the varchar(50) to varchar(255). Created TODO #1940. Which i will update right after this PR.

pkg/storage/mysql/mysqlstore.go
talwinder50: As discussed this is removed

pkg/storage/mysql/mysqlstore.go
talwinder50: Have updated this using sub query to simulate the range. Unit test is updated

pkg/storage/mysql/mysqlstore_test.go
talwinder50: Spooling inside the test file now

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return err
+	}
+
+	return conn.Close()
+}
+
+// Put stores the key and the record
+func (s *sqlDBStore) Put(k string, v []byte) error {
+	if k == "" || v == nil {
```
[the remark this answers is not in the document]
talwinder50: Updated

pkg/storage/mysql/mysqlstore.go
llorllale: Actually, I realized storing the hash would break the `Iterator()` function. So, don't use the hash. But we still need a way to implement indexes on keys of arbitrary lengths.

talwinder50: Can i address that in the separate PR. Will need to dig deeper how to get achieve that

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,275 @@
...
+	tablePrefix               = "t_"
+	sqlDBNotFound             = "no rows"
+	createDBQuery             = "CREATE DATABASE IF NOT EXISTS "
+	useDBQuery                = "USE "
+)
+
+var errKeyRequired = errors.New("key is mandatory")
+
+// NewProvider instantiates Provider
+func NewProvider(dbPath string) (*Provider, error) {
```
rolsonquadras, line 51: dbURL would be better term than dbPath. Path usually refers to directories/files (unique location in a file system).

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,275 @@
...
+		return fmt.Errorf("failed to insert key and value record into %s %w ", s.tableName, err)
+	}
+
+	return nil
+}
+
+// Get fetches the record based on key
+func (s *sqlDBStore) Get(k string) ([]byte, error) {
+	if k == "" {
+		return nil, errKeyRequired
```
rolsonquadras, line 168: I feel better to use consistent error messages across the implementations. May be by defining it in `storage` pkg.

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	var res result
+	// nolint: gosec select query to fetch the record by key
+	err := s.db.QueryRow("SELECT * FROM "+s.tableName+" "+
+		" WHERE `key` = ?", k).Scan(&res.key, &res.value)
```
soluchok, line 174: we are not using key. `.Scan(&res.key` is redundant

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,275 @@
...
+	queryStmt := "SELECT DISTINCT `key`, value FROM " +
+		"(SELECT * FROM " + s.tableName + " WHERE `key` LIKE ? " +
+		"UNION ALL SELECT * FROM " + s.tableName + " WHERE `key` LIKE ?) as d"
+
+	if startKey != "" && endKey != "" {
+		return s.prepareQueryResult(queryStmt, "%"+startKey+"%", "%"+endKey+"%")
+	}
+
+	return s.prepareQueryResult(queryStmt, startKey, endKey)
+}
```
soluchok, line 218: 1. replace `"%"+startKey+"%"` with `startKey+"%"` 
2.  replace `"%"+endKey+"%` with `"endKey+"%` 
3. endKey has suffix `!!`. Take a look of how we are using it `storage.EndKeySuffix`.

 The query will work. You are using LIKE again. 
The payload can be like this:
| key | val |
| ---- | ----|
| key_1 | val_1|
| key_2 | val_2|
| opps | val_3|
| key_4 | val_4|
| key_5 | val_5|
Make sure that `Iterator("key_2","key_4!!")` will return `key_2,opps,key_4`

pkg/storage/mysql/mysqlstore.go
talwinder50: Its actually not redundant , we get even when we are fetching only one column. sql: expected 2 destination arguments in Scan, not 1. Just how go mysql is working

pkg/storage/mysql/mysqlstore.go
talwinder50: @soluchok Something is wrong, as per the current store implementation does nt seem like it is picking range its rather like of matching the end key with the start key. 

https://github.com/hyperledger/aries-framework-go/blob/d258db55ea379933cde5bacfd7030a91d19c446c/pkg/storage/couchdb/couchdbstore_test.go#L259

talwinder50: @soluchok there is something missing, memstore is looking for prefix of the string to look through, where as couch db is also matching string . Are you sure that is the behavior we are looking in mysql.  

https://github.com/hyperledger/aries-framework-go/blob/d258db55ea379933cde5bacfd7030a91d19c446c/pkg/storage/mem/mem_store.go#L129

talwinder50: As per the other implementation this works slightly different than this.

talwinder50: In the above scenario response will be different as keys are stored or entered into the db in ascending order. So oops will be shifted to the end. There for the return value will be different  This works I have tested multiple scenarios. You can check the output here http://www.sqlfiddle.com/#!9/c13b02/10
SELECT * FROM test 
WHERE `key` >= 'abc_' AND `key` < 'mno~'

talwinder50: I have added this and ran the exact same test other implementation are running for the iterator. They passed.

pkg/storage/mysql/mysqlstore.go
```diff
@@ -0,0 +1,264 @@
...
+type sqlDBResultsIterator struct {
+	store      *sqlDBStore
+	resultRows *sql.Rows
+	err        error
+}
+
+func (s *sqlDBStore) Iterator(startKey, endKey string) storage.StoreIterator {
+	endKey = strings.ReplaceAll(endKey, storage.EndKeySuffix, "~")
+	//nolint:gosec sub query to fetch the all the keys that have start and end key reference, simulating range behavior.
+	queryStmt := "SELECT * FROM " + s.tableName + " WHERE `key` >= ?  AND `key` < ?"
```
soluchok, line 208: I have checked it. You need to add `ORDER BY key` as well. Also, I would suggest using `BETWEEN` operator. 

```
SELECT * FROM test 
WHERE `key` BETWEEN 'abc_' AND 'mno_123~' ORDER BY `key`
```

talwinder50: As discussed offline, will stick to operand query for now. Using Between always include the endkey value which we dont want

pkg/storage/mysql/mysqlstore.go
talwinder50: updated that

pkg/storage/mysql/mysqlstore.go
llorllale: @talltree sure - can leave a TODO

pkg/storage/mysql/mysqlstore.go
llorllale: note: #1946 was opened to address this inconsistency

pkg/storage/mysql/mysqlstore_test.go
```diff
@@ -0,0 +1,587 @@
...
+*/
+
+package mysql
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
```
llorllale, line 13: I am concerned about this usage.

Using this as a backend for the tests then places responsibility squarely on the maintainers to know how MySQL works. This carries the real risk of improper tests. Eg.: the current implementation will fail with keys of sufficient length.

Please - as a TODO - refactor to use a real MySQL instance.

llorllale:
@soluchok @talwinder50 I've opened #1948 with followup issues. Feel free to add or modify as needed.

llorllale:
@talwinder50 we can merge this now and continue in followup PRs - up to you. Some of the points in #1948 are blockers.

talwinder50:
> @talwinder50 we can merge this now and continue in followup PRs - up to you. Some of the points in #1948 are blockers.

Ok i will create the follow up pr for the issues you mentioned after this merge. Thanks. Will be working on them today

PR #1939, opened 2020-06-18
Initial mobile support setup

### LINKED ISSUE #1921: Add scripts to generate mobile bindings
**What I'm trying to do**

Create an executable script or make directive that will generate necessary files for iOS and Android SDKs.

**Expected result**

A one line command that runs a script which will produce **.framework, .jar** and **.aar** files.

**Checklist**
- [x] Include dependencies in script
- [x] Implement the SDK strategy from [Go mobile](https://github.com/golang/go/wiki/Mobile)
- [x] Support major targets
  - [x] iOS
  - [x] Android

ToJen:
**Title:**
Created `aries-agent-mobile` folder and supporting files

**Description:**
This should close #1921 

**Summary**:

Added a Makefile that runs necessary commands and an initial folder layout for packages that will be exposed in the mobile SDKs.


cmd/aries-mobile-sdk/.gitignore
```diff
@@ -0,0 +1,7 @@
+#
```
troyronda, line 1: i wonder if the “-sdk” suffix will cause confusion.

cmd/aries-mobile-sdk/.gitignore
```diff
@@ -0,0 +1,7 @@
...
+# Copyright SecureKey Technologies Inc. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+build/
```
soluchok, line 7: Is there a need for `.gitignore`? We have it the root directory.

cmd/aries-mobile-sdk/Makefile
```diff
@@ -0,0 +1,78 @@
+# Copyright SecureKey Technologies Inc.
```
rolsonquadras, line 1: It would be good if we can link this in the Makefile in root folder, so that we don't need to change the directory while building this.

cmd/aries-mobile-sdk/.gitignore
rolsonquadras: Current 2 bindings doesn't have `-framework` or `-sdk` suffix (`aries-agent-rest` and `aries-js-worker`). So, I guess, just `aries-mobile` or `aries-agent-mobile` would be ok.

cmd/aries-mobile-sdk/main.go
```diff
@@ -0,0 +1,11 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package main
+
+import (
+	//"github.com/hyperledger/aries-framework-go/cmd/aries-mobile-sdk/pkg/framework"
```
soluchok, line 10: You are adding go.sum and go.mod  files but the main file is empty. Please add at least one import. e.g
`_ "github.com/hyperledger/aries-framework-go/cmd/aries-mobile-sdk/pkg/framework"`

cmd/aries-mobile-sdk/.gitignore
ToJen: i see what you mean @troyronda.
@rolsonquadras  i was thinking of `aries-mobile-agent`

cmd/aries-mobile-sdk/.gitignore
ToJen: i thought about this and noticed the js-worker folder has a gitignore as well. my thinking was that this could grow to become it's own repo because of the builds and tests. should I  leave that for when the time comes and move remove this?

cmd/aries-mobile-sdk/Makefile
ToJen: agreed, i'll find a way to link the makefiles together. i expect that there will be many more mobile related targets and don't want to pollute the main makefile.
i'm also avoiding using shell scripts in the mobile section for now because of Windows.

cmd/aries-mobile-sdk/main.go
ToJen: i'll use this `_ "github.com/hyperledger/aries-framework-go"` thanks

cmd/aries-mobile-sdk/.gitignore
soluchok: Just comment on your section. We do not need the second file.

```
# mobile-related files 
cmd/aries-mobile-sdk/build/
```

PR #1947, opened 2020-06-23
refactor: VC extension of Credential Subject

### LINKED ISSUE #1943: VC extension of Credential Subject
At https://github.com/hyperledger/aries-framework-go/blob/afa883b058ee47cb78246f1effa318f85af45a16/pkg/doc/verifiable/credential_extension_test.go#L50, in order to build credential subject of a custom structure, we need to marshal VC and then unmarshal it back into extended VC with the custom credential subject field.
We need a better approach to avoid extra marshalling / unmarshalling of the whole VC.

kdimak:
closes #1943

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"reflect"
 	"time"
 
 	"github.com/piprate/json-gold/ld"
@@ -396,7 +397,7 @@ type Issuer struct {
 }
 
 // MarshalJSON marshals Issuer to JSON.
-func (i Issuer) MarshalJSON() ([]byte, error) {
+func (i *Issuer) MarshalJSON() ([]byte, error) {
 	if len(i.CustomFields) == 0 {
 		// as string
 		return json.Marshal(i.ID)
@@ -405,7 +406,7 @@ func (i Issuer) MarshalJSON() ([]byte, error) {
 	// as object
 	type Alias Issuer
 
-	alias := Alias(i)
+	alias := Alias(*i)
 
 	data, err := marshalWithCustomFields(alias, i.CustomFields)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -445,15 +446,63 @@ func (i *Issuer) UnmarshalJSON(bytes []byte) error {
...
 // Subject of the Verifiable Credential
-type Subject interface{}
+type Subject struct {
+	ID string `json:"id,omitempty"`
+
+	CustomFields CustomFields `json:"-"`
+}
+
+// MarshalJSON marshals Subject to JSON.
+func (s Subject) MarshalJSON() ([]byte, error) {
```
llorllale, line 456: receiver is not a pointer

pkg/doc/verifiable/credential.go
```diff
@@ -445,15 +446,63 @@ func (i *Issuer) UnmarshalJSON(bytes []byte) error {
...
-	Context        []string
-	CustomContext  []interface{}
-	ID             string
-	Types          []string
-	Subject        Subject
+	Context       []string
+	CustomContext []interface{}
+	ID            string
+	Types         []string
+	// Subject can be a string, map, slice of maps, struct (Subject or any custom), slice of structs.
```
llorllale, line 504: can you show an example of where it's just a string?

pkg/doc/verifiable/credential.go
```diff
@@ -665,6 +714,41 @@ func parseIssuer(issuerBytes json.RawMessage) (Issuer, error) {
 	return issuer, err
 }
 
+// parseSubject parses raw credential subject
+
```
llorllale, line 718: extra empty line

pkg/doc/verifiable/credential.go
```diff
@@ -445,15 +446,63 @@ func (i *Issuer) UnmarshalJSON(bytes []byte) error {
...
-	CustomContext  []interface{}
-	ID             string
-	Types          []string
-	Subject        Subject
+	Context       []string
+	CustomContext []interface{}
+	ID            string
+	Types         []string
+	// Subject can be a string, map, slice of maps, struct (Subject or any custom), slice of structs.
+	Subject        interface{}
```
llorllale, line 505: how many things will break? :D

pkg/doc/verifiable/credential.go
```diff
@@ -1005,6 +1094,86 @@ func issuerToRaw(issuer Issuer) (json.RawMessage, error) {
...
+// SubjectToBytes converts subject(s) to bytes.
+// A subject can be of a different kind:
+// - string (represents subject id)
+// - map
+// - slice of maps
+// - Subject
+// - slice of Subjects
+// - custom struct
+// - slice of custom structs
+func SubjectToBytes(subject interface{}) ([]byte, error) {
```
llorllale, line 1106: Does this really need to be exported? I believe `Subject.MarshalJSON()` covers this use case.

pkg/store/verifiable/store.go
```diff
@@ -238,6 +238,8 @@ func (s *StoreImplementation) getAllRecords(searchKey string, keyPrefix func(str
 	for itr.Next() {
 		var r record
 
+		fmt.Printf("itr.Value()=%s\n", string(itr.Value()))
```
llorllale, line 241: please remove this

llorllale:
@kdimak this looks like it solves #1943:

>  in order to build credential subject of a custom structure, we need to marshal VC and then unmarshal it back into extended VC with the custom credential subject field.
> We need a better approach to avoid extra marshalling / unmarshalling of the whole VC.


But as a user I have a request: can we make `verifiable.ParseCredential()` output a custom credential struct?

pkg/doc/verifiable/credential.go
kdimak: fixed

pkg/doc/verifiable/credential.go
kdimak: According to [VC data model documentation](https://www.w3.org/TR/vc-data-model/#credential-subject)
> Each object MAY contain an id
so according to JSON-LD spec, if the JSON object contains just `id` field, it can be simplified to just a string containing that id.

pkg/doc/verifiable/credential.go
kdimak: fixed

pkg/doc/verifiable/credential.go
kdimak: I think - none. `Subject` still has an `interface{}` type as it had before. And we just extended type variety support for the subject.

pkg/doc/verifiable/credential.go
kdimak: A subject here could be just of `[]Subject` type (as a default result of credential parsing), but can also be a string, custom structure etc. So this is a useful utility function. But after additional consideration, I decided to hide it.

pkg/store/verifiable/store.go
kdimak: removed! thank you for the catch!

kdimak:
> But as a user I have a request: can we make verifiable.ParseCredential() output a custom credential struct?
It's complicated. Just because we are using `rawCredential` structure for the initial VC parsing and then map it to the `Credential`. So we need to apply "injection points" where we could change the default mapping. Also, it will lead to the API break (of `ParseCredential()`) as we need to pass a custom credential struct as a parameter I believe. In contrast, we now return created `Credential` structure.

PR #1949, opened 2020-06-23
feat: Common notifier for the protocols

### LINKED ISSUE #1887: We need common notifier for the protocols
We do not have a mechanism that will notify the clients about `DIDCommAction` and `StateMsg` events when we use the Rest API. There should be a common logic for all protocols. Since all protocols are using the same common structures for the DIDCommAction and StateMsg events.

soluchok:
This PR adds notification support for the `StateMsg` and `Action`.
The following protocols are supported:
* introduce
* issuecredential
* outofband
* resentproof

The topics you might listen to are:
* present-proof_actions
* present-proof_states
* introduce_actions
* introduce_states
* issue-credential_actions
* issue-credential_states
* out-of-band_actions
* out-of-band_states

This PR doesn’t include `didexchange`. Since it will break all tests.

Closes #1887


llorllale:
@soluchok 

> This PR doesn’t include didexchange. Since it will break all tests.

Is there an issue for the followup?

soluchok:
> @soluchok
> 
> > This PR doesn’t include didexchange. Since it will break all tests.
> 
> Is there an issue for the followup?

I've created an issue for that https://github.com/hyperledger/aries-framework-go/issues/1950

pkg/controller/webnotifier/observer.go
```diff
@@ -0,0 +1,53 @@
...
+func (o *Observer) RegisterStateMsg(topic string, ch <-chan service.StateMsg) {
+	go func() {
+		for msg := range ch {
+			o.notify(topic, msg)
+		}
+	}()
+}
+
+func (o *Observer) notify(topic string, v interface{}) {
+	src, err := json.Marshal(v)
```
llorllale, line 43: is there a followup to properly serialize `service.DIDCommAction` and `service.StateMsg` ?

pkg/didcomm/common/service/event.go
```diff
@@ -49,24 +49,24 @@ type DIDCommAction struct {
 	//
 	// Supported protocols
 	//   - DID Exchange :  didexchange.DIDExchange
-	ProtocolName string
+	ProtocolName string `json:",omitempty"`
```
llorllale, line 52: why not do similar to service.StateMsg?

note:

* service.StateMsg.Type will be serialized as an int
* service.StateMsg.Msg will leak internal details from service.DIDCommMsgMap
* not sure if all service.StateMsg.EventProperties will serialize properly

pkg/didcomm/common/service/event.go
```diff
@@ -49,24 +49,24 @@ type DIDCommAction struct {
...
-	Stop func(err error)
+	Stop func(err error) `json:"-"`
 
 	// Properties contains value based on specific protocol. The consumers need to call the protocol client
 	// functions to get the data.
 	//
 	// Clients function to retrieve data based on protocol.
 	//   - DID Exchange :  didexchange.EventProperties
-	Properties EventProperties
+	Properties EventProperties `json:",omitempty"`
```
llorllale, line 69: gotta make sure implementations serialize properly

pkg/didcomm/common/service/event.go
soluchok: I added `omitempty` everywhere. We do not need empty values, right?

I would not say that is a leak of internal details. 
We actually provide the same functionality. We just replacing channels with WebSocket.
Currently, EventProperties are working fine. 
I can create different structures for that.
Just let me know.

soluchok: About properties, I am pretty sure they should be serialized and exposed as well. But I understand that because of interface it cannot be safe. What we can do here is to add a comment that properties should be serialized otherwise event can be skipped for the WebSocket.

llorllale: @soluchok 

> We actually provide the same functionality. We just replacing channels with WebSocket.

It's not the same. `service.DIDCommMsgMap` is actually exposed as the interface `service.DIDCommMsg` in the framework. From there users cannot access the metadata (unless they follow your example and start casting it to DIDCommMsgMap...).

> About properties, I am pretty sure they should be serialized and exposed as well. But I understand that because of interface it cannot be safe. What we can do here is to add a comment that properties should be serialized otherwise event can be skipped for the WebSocket.

I'm not sure what you mean regarding safety, but I'm sure, as an example, the following two properties won't be serialized: https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/protocol/didexchange/event.go#L19

soluchok: Done! no more casting

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -348,7 +354,13 @@ func (s *Service) newDIDCommActionMsg(md *metaData) service.DIDCommAction {
 
 			s.processCallback(md)
 		},
-		Stop: func(err error) { actionStop(customError{error: err}) },
+		Stop: func(err error) {
+			if err == nil {
+				err = errProtocolStopped
+			}
+
+			actionStop(customError{error: err})
```
Baha-sk, line 362: if err != nil, the Stop action doesn't include errProtocStopped, ie the stop error is hidden, is this the intended behaviour?

pkg/controller/webnotifier/observer.go
```diff
@@ -0,0 +1,118 @@
...
+		for msg := range ch {
+			o.notify(topic, toStateMsg(msg))
+		}
+	}()
+}
+
+func (o *Observer) notify(topic string, v interface{}) {
+	src, err := json.Marshal(v)
+	if err != nil {
+		logger.Errorf("notify marshal: %s", err)
```
Baha-sk, line 56: %w for error wrapping everywhere please..

pkg/didcomm/protocol/presentproof/service.go
```diff
@@ -524,6 +522,10 @@ func (s *Service) ActionStop(piID string, cErr error) error {
 		return fmt.Errorf("delete transitional payload: %w", err)
 	}
 
+	if cErr == nil {
+		cErr = errProtocolStopped
+	}
+
 	md.err = customError{error: cErr}
```
Baha-sk, line 529: same here, if cErr!= nil, shouldn't we wrap `errProtocolStopped` to it? .. I see this repeated elsewhere in this pr.

pkg/controller/webnotifier/observer.go
soluchok: it is logger, no need to wrap

pkg/didcomm/protocol/introduce/service.go
soluchok: when we stop the protocol we need to provide an error (reason), if the user doesn’t provide it we will use default message e.g `errProtocolStopped`. It is not about behavior, it is more as info.

pkg/didcomm/messenger/messenger.go
```diff
@@ -104,8 +104,6 @@ func (m *Messenger) saveMetadata(msg service.DIDCommMsgMap) error {
 		return fmt.Errorf("threadID: %w", err)
 	}
 
-	delete(msg, jsonMetadata)
```
llorllale, line 107: why was this removed?

soluchok: I have added it on the marshal level https://github.com/hyperledger/aries-framework-go/pull/1949/files#diff-fc8f058cfffcbd9722de36d08af83891R60

PR #1953, opened 2020-06-24
bug: remove nolint gosec and fix possible sql injection vulnerability

### LINKED ISSUE #1952: bug: remove nolint gosec and fix possible sql injection attack
current mysql storage implementation injects the storage as a free form string in the SQL queries.

golint sec has been wrongly disabled on these queries.

Remove the nolint gosec comments and fix the storage name in the statements to be added as part of the prepared statement.

Baha-sk:
closes #1952

### DIFF
--- a/pkg/storage/mysql/mysqlstore.go
+++ b/pkg/storage/mysql/mysqlstore.go
@@ -149,10 +149,10 @@ func (s *sqlDBStore) Put(k string, v []byte) error {
 		return storage.ErrKeyRequired
 	}
 
-	//nolint: gosec create upsert query to insert the record, checking whether the key is already mapped to a value in the store.
-	createStmt := "INSERT INTO " + s.tableName + " VALUES (?, ?) ON DUPLICATE KEY UPDATE value=?"
+	// create upsert query to insert the record, checking whether the key is already mapped to a value in the store.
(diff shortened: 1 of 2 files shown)


pkg/storage/mysql/mysqlstore.go
```diff
@@ -149,10 +149,10 @@ func (s *sqlDBStore) Put(k string, v []byte) error {
 		return storage.ErrKeyRequired
 	}
 
-	//nolint: gosec create upsert query to insert the record, checking whether the key is already mapped to a value in the store.
-	createStmt := "INSERT INTO " + s.tableName + " VALUES (?, ?) ON DUPLICATE KEY UPDATE value=?"
+	// create upsert query to insert the record, checking whether the key is already mapped to a value in the store.
+	createStmt := "INSERT INTO ? VALUES (?, ?) ON DUPLICATE KEY UPDATE value=?"
```
llorllale, line 153: @talwinder50

talwinder50: https://github.com/go-sql-driver/mysql/issues/848

I have already tried This query wont work see the spec they dont support placeholders on table name they only support on where and like. If we run this against the real mysql it will throw syntax error

llorllale: @talwinder50 thanks. Looks like we need to rollback this PR....

PR #1955, opened 2020-06-24
feat: add controller interface and protocol wrappers

### LINKED ISSUE #1923: Find solution for gomobile's type restrictions
**What I'm trying to do**

Since gomobile has [restrictions](https://godoc.org/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions) on what it can generate, there needs to be a way to make sure important functionality from the Aries library is not ignored.

Possible solutions include
- ~~refactoring exposed packages to be compliant with gomobile (high effort)~~
- write separate packages that will be compliant and not affect the library (medium effort)

**Expected result**

A `wrappers` package in [cmd/aries-agent-mobile/pkg](https://github.com/hyperledger/aries-framework-go/tree/master/cmd/aries-agent-mobile/pkg/) that has functions/methods with signatures that gomobile can bind.

**Sample code / test case**

```go
// an example of a possible wrapper for a method from pkg/framework/aries/framework.go

type CloseError struct {
  Message string
}

func CloseAries(a Aries) CloseError {
  if err := a.Close(); err != nil {
    return CloseError{Message: err.Error()};
  }
  return CloseError{};
}

```

**Checklist**
- [X] What's the best solution here?
  - use wrapper methods
- [x] Test it works
    - [x] err
    - [x] number of returned values

### LINKED ISSUE #1941: Create a controller interface for mobile
**What I'm trying to do**

Similar to the JS worker, the mobile agent should be able to execute calls internally or delegate them to an external agent. This means that users of the mobile agent can choose between an HTTP client implementation of the Aries library and the default "regular" implementation.

**Expected result**

An interface in [cmd/aries-agent-mobile](https://github.com/hyperledger/aries-framework-go/tree/master/cmd/aries-agent-mobile) that will have two implementations.

**Sample code / test case**

```go

type AriesController interface {
    Introduce(...)
}

// Original/internal implementation
type Aries struct {}

func (af *Aries) Introduce(...) {
    // ....
}

// HTTP Client implementation
type AriesHttp struct {}

func (ah *AriesHttp) Introduce(...) {
    // ....
}
```

**Checklist**
- [x] Add interface
- [x] Add HTTP client implementation
- [x] Add regular implementation (based on wrappers)

ToJen:

**Title:**
Created a controller interface and wrapper implementations. 

**Description:**
There are [restrictions](https://godoc.org/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions) on code that can be bound by gomobile. Hence, this change shows how we can create wrappers around each protocol's exported methods in Aries. A method from the Introduce protocol is included. The signatures of these wrapped methods are successfully parsed by gomobile. This can be tested by cloning this branch and running `make bindings` within the **/cmd/aries-agent-mobile**.
 
Closes #1923 
Closes #1941
Part of #1958

**Summary**:

- Relevant packages:
  - **/cmd/aries-agent-mobile/api** will contain interfaces that define agent operations.
  - **/cmd/aries-agent-mobile/wrappers/command** contains the implementation using the internal Aries framework.
  - **/cmd/aries-agent-mobile/wrappers/rest** contains the implementation that makes http requests to an external server.
- Relevant Structs:
  - Aries: wrapped version of the actual Aries framework and a map of handler functions
  - AriesREST: contains a map of endpoints to delegate requests to
  - Introduce: wrapped introduce command
  - IntroduceREST: http client that is modelled after the introduce operation in **/pkg/controller/rest/introduce**


cmd/aries-agent-mobile/pkg/wrappers/aries.go
```diff
@@ -0,0 +1,66 @@
...
+
+	command, err := commandintroduce.New(context)
+	if err != nil {
+		return fmt.Errorf("failed to create introduce command instance: %v", err)
+	}
+
+	a.Introduce.command = command
+
+	return nil
+}
```
soluchok, line 61: I think we do not need to load each client. Try to use `GetCommandHandlers` instead. it will provide all everything.

cmd/aries-agent-mobile/pkg/wrappers/ariesrest.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// AriesREST is an Aries instance with HTTP clients to execute operations
+type AriesREST struct {
+	Introduce *IntroduceREST
+	Aries     *Aries
```
soluchok, line 17: You don’t need `Aries`. I don’t see any reason for that.

cmd/aries-agent-mobile/pkg/wrappers/rest/rest.go
```diff
@@ -0,0 +1,41 @@
...
+
+package rest
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+func makeHTTPRequest(httpClient *http.Client, method, url, token string, _ interface{}) ([]byte, error) {
```
talwinder50, line 15: Why there is non variable interface  this one in the makeHTTPRequest signature  _ interface{}??

cmd/aries-agent-mobile/pkg/wrappers/rest/endpoints.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http"
+
+	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers"
+)
+
+// Endpoint describes the fields for making calls to external agents
+type Endpoint struct {
+	Path   string
+	Method string
+	PIID   string
```
talwinder50, line 19: variable stutter , PID should be enough ..

ToJen: hey Talwinder, I'm referring to the field representing a protocol instance id as these do: https://github.com/hyperledger/aries-framework-go/search?q=piid&unscoped_q=piid

cmd/aries-agent-mobile/pkg/wrappers/rest/rest.go
ToJen: that's supposed to be request body. i'll remove it since it's not needed/used yet

llorllale:
@ToJen 

> This can be tested by cloning this branch and running make bindings within the /cmd/aries-agent-mobile.

```
llorllale:~/dev/hyperledger/aries-framework-go/cmd/aries-agent-mobile$ make bindings
----------------------------------------
   gomobile: Creating Android bindings  
----------------------------------------
gomobile bind -v -target=android -javapkg=com.hyperledger.aries -o=/home/llorllale/dev/hyperledger/aries-framework-go/cmd/aries-agent-mobile/build/android/aries-agent.aar github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/ github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/command github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/rest
gomobile: no Android NDK found in $ANDROID_HOME/ndk-bundle nor in $ANDROID_NDK_HOME
Makefile:72: recipe for target 'bindings-android' failed
make: *** [bindings-android] Error 1
```

Can the README be updated to describe how/where exactly to install the dependencies?

cmd/aries-agent-mobile/Makefile
```diff
@@ -2,25 +2,29 @@
...
+MAKEFILE_DIR := ${CURDIR}
 BUILD_DIR = $(MAKEFILE_DIR)/build
 
 ANDROID_BUILD_DIR = $(BUILD_DIR)/android
-ANDROID_OUTPUT_FILE_NAME = aries-agent-mobile.aar
+ANDROID_OUTPUT_FILE_NAME = aries-agent.aar
 ifeq ($(ANDROID_TARGET),)
 	ANDROID_TARGET = android
 endif
+ANDROID_JAVA_PKG = com.hyperledger.aries
```
llorllale, line 24: org.hyperledger.aries

ToJen: thanks, missed that. i'll update it

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
```diff
@@ -0,0 +1,37 @@
+/*
```
llorllale, line 1: This is essentially a REST client.

It's a bad idea to add the cross-language barrier (java-go) on top of the REST network call. It's better to just implement a REST client in the Java/Android layer, similar to how aries-js-worker has a REST client version.

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
```diff
@@ -0,0 +1,80 @@
+/*
```
llorllale, line 1: Following the existing architecture, this package belongs under `pkg/controller`

ToJen: do you mean i should move `wrappers` into the main aries pkg or i should instead have cmd/aries-agent-mobile/pkg/wrappers/**_controller_**/command/aries.go

llorllale: @ToJen 

> do you mean i should move `wrappers` into the main aries pkg or i should instead have cmd/aries-agent-mobile/pkg/wrappers/**_controller_**/command/aries.go

Move them under the existing `pkg/controller` folder. Something like this: `cmd/aries-agent-mobile/pkg/wrappers/command` -> `pkg/controller/mobile`

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
ToJen: @llorllale this is related to what myself and @troyronda discussed. how many libraries to have per target and how much of the library we want written outside of go. what gomobile does is to convert golang code into a .so (shared object) which cgo uses to create a .framework or .aar.

if the http client is written outside of our bindings, we would have to import our generated code into a new objectivec/java library. this library will be the final product that exposes both implementations. However with iOS, we will be creating something called an Umbrella Framework which has been [discouraged by Apple](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/CreationGuidelines.html).

ToJen:
What we want to know are the following:

- should there be only one library containing implementations for the local and remote agent?
- how much latency would be experienced on a mobile app if the remote agent is written in go?

ToJen:
@llorllale 

here's the update for [the discussion](https://github.com/hyperledger/aries-framework-go/pull/1955#issuecomment-651242054). What I did was to test a simple GET request by
- calling `goFunction1` which is written in go and receives a URL and N, the number of times to call that URL
- calling `goFunction2` which is written in go and receives just a URL, i called this N times from objective-c
- calling `iOSFunction` which is written in objective-c and takes a url and calls it N times

> * how much latency would be experienced on a mobile app if the remote agent is written in go?

N=10
```
res: [goFunction1] url:http://www.google.com N:10 seconds:4.579783
res: [goFunction2] url:http://www.google.com N:10 seconds:4.766893
res: [iOSFunction] url:http://www.google.com N:10 seconds:4.703360
```

N=50
```
res: [goFunction1] url:http://www.google.com N:50 seconds:25.057063
res: [goFunction2] url:http://www.google.com N:50 seconds:24.392210
res: [iOSFunction] url:http://www.google.com N:50 seconds:23.741404
```
There's barely any noticeable difference

> * should there be only one library containing implementations for the local and remote agent?

Based on the results, we should write the api, local agent and remote agent in golang and produce a single `.framework` file

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// NewAries returns a new Aries instance that contains handlers and an Aries framework instance.
+func NewAries() (*Aries, error) {
+	// TODO receive options
+	storageProvider := mem.NewProvider()
+
+	framework, err := aries.New(aries.WithStoreProvider(storageProvider))
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize Aries framework: %v", err)
```
Baha-sk, line 34: use %w when wrapping errors (instead of %v)

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
```diff
@@ -0,0 +1,73 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package command
+
+import (
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage/mem"
+
```
Baha-sk, line 13: any reason for this blank line? usually imports from `aries-framework-go` are grouped together and are the last group of imports.

cmd/aries-agent-mobile/pkg/wrappers/rest/rest.go
```diff
@@ -0,0 +1,50 @@
...
+	response, err := httpClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("error while making request to [%s]: %v", url, err)
+	}
+
+	defer func() {
+		e := response.Body.Close()
+		if e != nil {
+			// TODO use `/pkg/common/log` package instead
+			fmt.Print(fmt.Errorf("failed to close response body: %v", e))
```
Baha-sk, line 40: Please fix this TODO, we should use proper log package. See other packages using log examples.

llorllale:
I got `make bindings-android` to work on Ubuntu 18.04. Cannot confirm with `bindings-ios`.

I noted that `make bindings-android` takes a LONG time: 
```
$ time make bindings-android
... lots and lots of stuff...
real    0m59.528s
user    2m49.010s
sys     0m20.962s
```

I also noted the ["Command line tools"](https://developer.android.com/studio#downloads) - you really do need to download the IDE package.

cmd/aries-agent-mobile/pkg/wrappers/rest/introduce_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"testing"
+
+	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers"
+
+	"github.com/stretchr/testify/require"
+)
```
soluchok, line 12: Wrong import. Should be:
```
import (
	"bytes"
	"io/ioutil"
	"net/http"
	"testing"

	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers"
)

```

cmd/aries-agent-mobile/Makefile
```diff
@@ -58,21 +73,31 @@ dependencies:
...
 bindings-ios:
 	@echo '----------------------------------------'
 	@echo '   gomobile: Creating iOS bindings      '
 	@echo '----------------------------------------'
+ifeq ($(IS_MACOS),true)
 	@mkdir -p $(IOS_BUILD_DIR)
-	$(GOMOBILE_CMD) bind -v -target=$(IOS_TARGET) -o=$(IOS_BUILD_DIR)/$(IOS_OUTPUT_FILE_NAME) $(MOBILE_BINDINGS_ROOT_PACKAGE)
+	$(GOMOBILE_CMD) bind -v -target=$(IOS_TARGET) -o=$(IOS_BUILD_DIR)/$(IOS_OUTPUT_FILE_NAME) $(MOBILE_BUILD_PACKAGES)
+else
+	@echo 'Skipping iOS bindings. Only available for Darwin operating systems'
```
ToJen, line 95: @llorllale iOS bindings require Darwin i.e macOS. Apple [doesn't allow](https://stackoverflow.com/a/2406839) Windows or Linux users to build iOS or Mac software. I will make it more explicit in the README that only macOS users will be able to create iOS bindings.

ToJen: this is in response to: https://github.com/hyperledger/aries-framework-go/pull/1955#issuecomment-653537494

ToJen: for the second part of that comment, i added in the requirements section of the README that Android Studio should be installed. I can make it more explicit. The android sdk can be installed without Android Studio but it's more of a hassle to get other things working like emulators and the rest.

the speed of the bindings is what's currently available. it pulls packages from the internet so a slow connection speed and low bandwidth could make the process seem slow. these were my results:

```bash
$ time make bindings-android
.....
real	1m7.049s
user	1m42.522s
sys	0m45.966s
```

llorllale: @ToJen I think my https://github.com/hyperledger/aries-framework-go/pull/1955#issuecomment-653537494 may be misinterpreted as a critique of this PR, or otherwise an account of "points against" this change. It is not. It's just my notes. :)

ToJen: thank you, i got that. i wanted to respond just for documentation purposes. i'll use them as suggestions for the README  file 👍

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
ToJen: i was having issues with the linter. i think this is an oversight. will clean up 👍

cmd/aries-agent-mobile/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ariesagent
+
+import (
+	"testing"
+
+	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/command"
+	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/rest"
```
Baha-sk, line 13: move these lints under testify/require import with and remove //nolint below

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
ToJen: thanks, i will use this instead. i didn't know about it actually

cmd/aries-agent-mobile/pkg/wrappers/rest/introduce_test.go
ToJen: thanks for pointing this out, i was trying to understand the linter's message about this

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
Baha-sk: "github.com/hyperledger/aries-framework-go/pkg/storage/mem"

should go right under (with no blank lines):
"github.com/hyperledger/aries-framework-go/pkg/framework/aries"

imports are sorted alphabetically, you don't need to worry about their order, `go import` does it for you, or if you use an IDE, there should be a shortcut to auto format your files. It's much easier to let the tool handle the imports sorting.

cmd/aries-agent-mobile/main_test.go
Baha-sk: don't resolve this comment, it should be fixed.. you should move Aries-framework-go imports as the last group of imports. This is a general rule, local imports as the last group of imports, preceded by (a blank line and) other libraries imports, preceded by (blank line and) core Go imports. Always follow this rule, otherwise you will get linter errors.

replace these imports as follows:
```
 	"testing"

 	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/command"
 	"github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/rest"
```

never use //nolint on imports

Issue #943, opened 2019-12-06
Decouple connection storage from DID Exchange protocol


troyronda:
We should not assume that the DID Exchange protocol is the only mechanism or protocol for establishing connection records.

llorllale:
Agreed.

An exchange of DIDs between a mobile agent and a server agent need NOT implement the did-exchange protocol as defined in RFC0023 to achieve the same goals. Considering that the security properties are imbued on the messages as opposed to the transports, a simple REST endpoint on the server side should suffice in this scenario.

Of course, a stateful protocol such as RFC0023 is still useful for simplex transports typically used between peers (eg. bluetooth in half-duplex mode. Bluetooth speeds vary between full-duplex and half-duplex).

fqutishat:
@sudeshrshetty

llorllale:
@troyronda was this task resolved as requested? The `didexchange.Client` is still the user-facing API for fetching connection records.

PR #1976, opened 2020-07-02
feat: Save Connection API

### LINKED ISSUE #1972: Save Connection API
Currently, the connection can only be created through DIDExchange protocol. There could be some cases when the users has the required data(DIDs, serviceEndpoints, keys etc) and need to create a connection record completed state. This API need to expose through sdk/rest/wasm bindings.

rolsonquadras:
closes #1972 

### DIFF
--- a/pkg/controller/rest/didexchange/models.go
+++ b/pkg/controller/rest/didexchange/models.go
@@ -200,3 +200,15 @@ type RemoveConnectionRequest struct { // nolint: unused,deadcode
 // swagger:response removeConnectionResponse
 type RemoveConnectionResponse struct { // nolint: unused,deadcode
 }
+
+// saveConnectionResp model
+//
+// This is used as the response model for save connection api.
+//
+// swagger:response saveConnectionResp
+type saveConnectionResp struct { // nolint: unused,deadcode
+	// The ID of the connection to get
+	//
+	// in: body
+	ID string `json:"id"`
+}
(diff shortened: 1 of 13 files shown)


rolsonquadras:
Issue for BDD tests https://github.com/hyperledger/aries-framework-go/issues/1977

pkg/controller/rest/didexchange/models.go
```diff
@@ -200,3 +200,15 @@ type RemoveConnectionRequest struct { // nolint: unused,deadcode
...
+// saveConnectionResp model
+//
+// This is used as the response model for save connection api.
+//
+// swagger:response saveConnectionResp
+type saveConnectionResp struct { // nolint: unused,deadcode
+	// The ID of the connection to get
+	//
+	// in: body
+	ID string `json:"id"`
```
soluchok, line 213: It will not works. Please check it and correct me if I wrong.
```
	// in: body
	Body struct {
		ID string `json:"id"`
	}
```

rolsonquadras:
> conflicts with #943

I don't understand how this conflicts with that. This API provides mechanism to Save connection records without DIDExchange. 

I added this in DIDExchange client because we have all Connection related APIs in DIDExchange client (query, remove).

pkg/controller/rest/didexchange/models.go
rolsonquadras: Thanks. I'll push changes with bdd test PR, as this is merged now.

PR #1985, opened 2020-07-06
fix: Mysql Iterator and adding db prefix

### LINKED ISSUE #1984: mysql iterator should not advance db cursor on calls to `Key()` or `Value()
https://github.com/hyperledger/aries-framework-go/blob/a92610a7ec561c76801ab18c79d0886808a0c721/pkg/storage/mysql/mysqlstore.go#L252-L262

With this implementation, two successive calls to Key() and Value() in any order will not correspond to the same db row.

Looks like sqlDBResultsIterator needs to hold on to the result until Next() is called

cc @talwinder50

talwinder50:
closes #1984

### DIFF
--- a/pkg/storage/mysql/mysqlstore.go
+++ b/pkg/storage/mysql/mysqlstore.go
@@ -18,11 +18,12 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
-// Provider represents an MySQL DB implementation of the storage.Provider interface
+// Provider represents a MySQL DB implementation of the storage.Provider interface
 type Provider struct {
-	dbURL string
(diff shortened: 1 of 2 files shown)


pkg/storage/mysql/mysqlstore.go
```diff
@@ -86,7 +104,7 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
 		return nil, fmt.Errorf("failed to create new connection %s: %w", p.dbURL, err)
 	}
 
-	// Use is used to select the created database without this DDL operations are not permitted
+	// Use query is used to select the created database without this DDL operations are not permitted
```
sandrask, line 107: comment is confusing

PR #1988, opened 2020-07-06
feat: add Authcrypt packer using ECDH1PU

### LINKED ISSUE #986: Add Default implementation for JWE Authcrypt Packer
There should be a default implementation of JWE Authcrypt envelope as described in [RFC-0334](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope) JWE Envelope

- [x] Create ECDH-1PU Tink primitive #1905

Baha-sk:
This change also introduces a thirdPartyKeysDB storage to store third party keys
It must be updated with the senderKey prior to Authcrypt.Decrypt() execution

closes #986

### DIFF
--- a/pkg/didcomm/packer/authcrypt/pack.go
+++ b/pkg/didcomm/packer/authcrypt/pack.go
@@ -1,0 +1,230 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package authcrypt
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/google/tink/go/keyset"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/keyio"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/packer"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// Package authcrypt includes a Packer implementation to build and parse JWE messages using Authcrypt. It allows sending
+// messages between parties with non-repudiation messages, ie the sender identity is revealed (and therefore
+// authenticated) to the recipient(s). The assumption of using this package is that public keys exchange has previously
+// occurred between the sender and the recipient(s).
+
+const (
+	encodingType = "didcomm-envelope-enc"
+	// ThirdPartyKeysDB is a store name containing keys of third party agents
+	ThirdPartyKeysDB = "thirdPartyKeysDB"
+)
+
+var logger = log.New("aries-framework/pkg/didcomm/packer/authcrypt")
+
+// Packer represents an Authcrypt Pack/Unpacker that outputs/reads Aries envelopes
+type Packer struct {
+	kms    kms.KeyManager
+	encAlg jose.EncAlg
+	store  storage.Store
+}
+
+// New will create an Packer instance to 'AuthCrypt' payloads for a given sender and list of recipients keys.
+// It will open a store (fetch cached one) that will contain third party keys. This store must be pre-populated with
+// the sender key required by a recipient to Unpack a JWE envelope. It is not needed by the sender (as the sender packs
+// the envelope with its own key).
+// The returned Packer contains all the information required to pack and unpack payloads.
+func New(ctx packer.Provider, encAlg jose.EncAlg) (*Packer, error) {
+	k := ctx.KMS()
+
+	store, err := ctx.StorageProvider().OpenStore(ThirdPartyKeysDB)
+	if err != nil {
+		return nil, fmt.Errorf("authcrypt: %w", err)
+	}
+
+	return &Packer{
+		kms:    k,
+		encAlg: encAlg,
+		store:  store,
+	}, nil
+}
+
+// Pack will encode the payload argument
+// Using the protocol defined by the Authcrypt message of Aries RFC 0334
+// with the following arguments:
+// payload: the payload message that will be protected
+// senderID: the key id of the sender (stored in the KMS)
+// recipientsPubKeys:
+func (p *Packer) Pack(payload, senderID []byte, recipientsPubKeys [][]byte) ([]byte, error) {
+	if len(recipientsPubKeys) == 0 {
+		return nil, fmt.Errorf("authcrypt Pack: empty recipientsPubKeys")
+	}
+
+	recECKeys, err := unmarshalRecipientKeys(recipientsPubKeys)
+	if err != nil {
+		return nil, fmt.Errorf("authcrypt Pack: failed to convert recipient keys: %w", err)
+	}
+
+	kh, err := p.kms.Get(string(senderID))
+	if err != nil {
+		return nil, fmt.Errorf("authcrypt Pack: failed to get sender key from KMS: %w", err)
+	}
+
+	jweEncrypter, err := jose.NewJWEEncrypt(p.encAlg, string(senderID), kh.(*keyset.Handle), recECKeys)
+	if err != nil {
+		return nil, fmt.Errorf("authcrypt Pack: failed to new JWEEncrypt instance: %w", err)
+	}
+
+	jwe, err := jweEncrypter.Encrypt(payload)
+	if err != nil {
+		return nil, fmt.Errorf("authcrypt Pack: failed to encrypt payload: %w", err)
+	}
+
+	var s string
+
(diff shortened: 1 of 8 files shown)


pkg/didcomm/packer/authcrypt/pack.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Packer represents an Authcrypt Pack/Unpacker that outputs/reads Aries envelopes
+type Packer struct {
+	kms    kms.KeyManager
+	encAlg jose.EncAlg
+	store  storage.Store
+}
+
+// New will create an Packer instance to 'AuthCrypt' payloads for a given sender and list of recipients keys.
+// It will open a store (fetch cached one) that will contain third party keys. This store must be pre-populated with
```
llorllale, line 48: @Baha-sk how will this store be pre-populated?

Foreign DIDs that have been exchanged with us are - I *think* - always stored in the vdri registry and the did connection store. The latter maps the keys to the DID IDs.

How does this store relate to the above?

pkg/didcomm/packer/authcrypt/pack.go
```diff
@@ -0,0 +1,231 @@
...
+		kid, err = getKID(i, jwe)
+		if err != nil {
+			return nil, fmt.Errorf("authcrypt Unpack: %w", err)
+		}
+
+		kh, err = p.kms.Get(kid)
+		if err != nil {
+			if strings.EqualFold(err.Error(), fmt.Sprintf("cannot read data for keysetID %s: %s", kid,
+				storage.ErrDataNotFound)) {
+				retriesMsg := ""
```
soluchok, line 153: It would be enough if you want to check `storage.ErrDataNotFound` error.
```
```

Baha-sk: the error is wrapped, can't use errors.Is(), unless the wrapping message is removed.

soluchok: this is exactly what errors.Is() does. From the doc: `Is reports whether any error in err's chain matches target.`

pkg/didcomm/packer/authcrypt/pack.go
Baha-sk: @llorllale regardless of DIDs, the packer (on the recipient side for now at least) needs to resolve/fetch the sender public it has (locally) for the `skid` that was (previously) received from the sender. I think the best place is to populate it during the invite process.

I've contacted the ECDH-1PU ietf author to clarify the `skid` resolution. His intention of using the `skid` for the sender is to ensure the static public key of the sender is not sent within the envelope to protect the sender's identity (as opposed to adding the sender public key as a jwk field, for example, in the envelope). The recipient should be able to retrieve the sender's key locally stored with that `skid` (which is nothing else than a `kid` value that was shared by the sender)

The aim in the Authcrypt packer is to have a local store for third party (encryption only) keys on an agent where it would store senders keys to be used for Authcrypt.Unpack() calls (as a recipient).  Future work can add maintenance of this store to update/remove senders keys from the storage.

This storage should not be viewed as part of the vdri registry as it only maintains `encryption` keys for the purpose of protecting didComm messages. This is why I want a clear separation between verification (signing) keys in DIDs and encryption keys (that can also be set in a DID document but used solely for Pack/Unpack). I believe the Aries community is in-line with this separation of key purposes (signature/verification vs encryption) in DID docs and didComm. I have not figured out how/where the encryption keys would go in the DID doc as of yet.

pkg/didcomm/packer/authcrypt/pack.go
Baha-sk: I have tried it and also thought the same thing, unfortunately `err.Is(storage.ErrDataNotFound)` returns false

Baha-sk: seems to work.. I must have tried something else.. will update..

pkg/didcomm/packer/authcrypt/pack.go
Baha-sk: @llorllale are you suggesting did connection store can potentially contain the sender key? This can replace the storage I introduced here if it makes sense and can be accessible from the packer.

pkg/didcomm/packer/authcrypt/pack.go
Baha-sk: updated

pkg/didcomm/packer/authcrypt/pack.go
llorllale: @Baha-sk 

> @llorllale are you suggesting did connection store can potentially contain the sender key? This can replace the storage I introduced here if it makes sense and can be accessible from the packer.

Are you asking if I, as a recipient, have a did connection store populated with another agent's keys who is sending me a message?

I'm suggesting it's a real possibility. I recently added the ability to manually create connections between `myDID` and `theirDID` (the latter being a did document) and one of the things it does it populate the did connection store with `theirDID`->`theirRecipientKeys` mappings.

llorllale: @Baha-sk 

> I've contacted the ECDH-1PU ietf author to clarify the `skid` resolution. His intention of using the `skid` for the sender is to ensure the static public key of the sender is not sent within the envelope to protect the sender's identity (as opposed to adding the sender public key as a jwk field in the envelope that poses a potential eavesdropping risk). The recipient should be able to retrieve the sender's key locally stored with that `skid` (which is nothing else than a `kid` value that was shared by the sender in a previous step, potentially during the invite process).

I think this sounds like a departure from the traditional didcomm model where the sender's key is inlined (encrypted) in the message. 

The description here says that this package assumes the keys have been exchanged previously. My question regarding that is - with which protocol? The did-exchange protocol is done over didcomm. Can you provide details of how it will be carried out with ecdh-1pu?

Another question is - how can Alice send Bob updates to her DID that rotate her `recipientKeys`?

> This storage should not be viewed as part of the vdri registry as it only maintains `encryption` keys for the purpose of protecting didComm messages. This is why I want a clear separation between verification (signing) keys in DIDs and encryption keys (that can also be set in a DID document but used solely for Pack/Unpack). I believe the Aries community is in-line with this separation of key purposes (signature/verification vs encryption) in DID docs and didComm.

Agreed on the clear separation of signing and verification keys. The did connection store - like I said above - already has the other agent's public keys used for didcomm key agreement.

>  I have not figured out how/where the encryption keys would go in the DID doc as of yet.

Aries RFCs place them in a `service` entry [as per convention](https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions).

Baha-sk: > 
> I'm suggesting it's a real possibility. I recently added the ability to manually create connections between `myDID` and `theirDID` (the latter being a did document) and one of the things it does it populate the did connection store with `theirDID`->`theirRecipientKeys` mappings.

this sounds like a good idea. I can update the third party store to be the did connection store (provided I can get the sender's key using `skid` where the intent for it to be a kid of a key, not sure how to map this from a DID document in the Packer as it knows nothing about DIDs).

>
> I think this sounds like a departure from the traditional didcomm model where the sender's key is inlined (encrypted) in the message.
> 
> 
> The description here says that this package assumes the keys have been exchanged previously. My question regarding that is - with which protocol? The did-exchange protocol is done over didcomm. Can you provide details of how it will be carried out with ecdh-1pu?

The ecdh-1pu(and es) primitive expects a list of recipients (we can assume they're jwk format) to execute the primitives.

The Pack() call (executed by the sender) will set the `skid` protected header in the JWE envelope where the ecdh-1pu Encrypt() primitive will use the sender key to wrap the JWE cek (ecdh-es does not use the sender key as it's anoncrypt).

The Unpack call (executed by the recipient) will use the `skid` header to fetch the sender static key from a local storage (this is where the sender key must be previously known to the recipient) in order for the ecdh-1pu Decrypt() primitive to unwrap the cek.

> Another question is - how can Alice send Bob updates to her DID that rotate her recipientKeys?

This is not yet determined, but I believe via a DID exchange that contains updated keys and kids.

>Agreed on the clear separation of signing and verification keys. The did connection store - like I said above - already has the other agent's public keys used for didcomm key agreement.

The issue with the current DID format is `recipientKeys` as used by the current (legacy) packer are `ed25519` and used for verification. They are converted to X25519 when needed by the packer for encryption. The new packer needs a new set of recipient keys for encryption (separate from verification keys), key conversion doesn't work for NIST approved keys.

Unless we create two types of resolution of the `recipientKeys` (since they're DID prefixed) one for encryption and another for verification, e.g. the same recipientKey can resolve to a verification key and an encryption key. 

> Aries RFCs place them in a service entry as per convention.

Service conventions use `recipientKeys` which are, by definition, verification keys. We need to be able to resolve two sets of keys, or as suggested above, the recipient keys can be resolved either for verification or encryption keys (we need two distinct sets of keys). This needs an RFC update (once we have the new Authcrypt packer and update how didComm should work with encryption keys).

Baha-sk:
> Adds Authcrypt packer using ECDH1PU. However, integration into the rest of the framework is pending - see discussion at https://github.com/hyperledger/aries-framework-go/pull/1988/files#r451608859

The Authcrypt packer is only concerned with JWE packing, it does not know DIDs. Therefore the sender public key (at the recipient side) should be pre populated (somewhere/somehow during invitation probably) in the recipient's thirdpartykeydb in order to decrypt messages. 

This is a pre requisite for Authcrypt packing (similar to the list of recipients needed to Pack() a message).

PR #2002, opened 2020-07-08
mediator - message pick up integration

m00sey:
**Title:**
mediator - message pick up integration

Summary:

We are attempting to integrate the message pick up and mediator services. So that when configured, the message pickup service can receive forward messages, from the mediator service and store them.

In our internal implementation of router/mediation, we used the `RegisterActionEvent` mechanism to register a channel and have the messages sent over to the message pickup service. This has been superseded by the use of `RegisterActionEvent` to allow businesses logic to be run during the RequestMessage flow during mediation.

We are looking for input on how best to proceed.

### DIFF
--- a/pkg/didcomm/protocol/mediator/service.go
+++ b/pkg/didcomm/protocol/mediator/service.go
@@ -20,6 +20,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/messagepickup"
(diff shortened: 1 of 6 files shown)


m00sey:
@llorllale You specifically asked about this integration before - https://github.com/hyperledger/aries-framework-go/issues/811#issuecomment-653557902 - would love to get your thoughts.

llorllale:
@m00sey 

> @llorllale You specifically asked about this integration before - [#811 (comment)](https://github.com/hyperledger/aries-framework-go/issues/811#issuecomment-653557902) - would love to get your thoughts.

If it were me, I'd invert the dependency relation such that the `mediator` service looks up a handle to the `messagepickup` service and adds messages to it if `s.outbound.Forward()` fails.

m00sey:
I see where you are coming from. I really like relying on `Forward` to fail and if configured, pushing them to the inbox. Thank you.

m00sey:
@llorllale sorry for the delay, here is an implementation based off your suggestion.

llorllale:
@m00sey thanks!

PR #2027, opened 2020-07-21
fix: Fixes storage behavior

### LINKED ISSUE #852: Change Store Iterator method to accept startKey and endKey

### LINKED ISSUE #1946: indexeddb and mem have different behavior
We have several implementations of the storage.
couchdb,leveldb and mysql have the same behavior. But indexeddb and mem  are not. 
We need to make sure that all implementations have the same behavior.

soluchok:
Closes #1946 
Closes #852 

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -83,6 +83,9 @@ linters-settings:
 linters:
   enable-all: true
   disable:
+    - godot
+    - nestif
+    - testpackage
     - maligned
     - prealloc
 
--- a/pkg/storage/couchdb/couchdbstore.go
+++ b/pkg/storage/couchdb/couchdbstore.go
@@ -157,7 +157,7 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
 
 	var valueToPut []byte
 	if isJSON(v) {
-		valueToPut = v
+		valueToPut = []byte(`{"payload":` + string(v) + `}`)
 	} else {
 		valueToPut = wrapTextAsCouchDBAttachment(v)
 	}
@@ -168,10 +168,7 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
 	}
 
 	if revID != "" {
-		valueToPut, err = c.addRevID(valueToPut, revID)
-		if err != nil {
-			return err
-		}
+		valueToPut = []byte(`{"_rev":"` + revID + `",` + string(valueToPut[1:]))
 	}
 
 	_, err = c.db.Put(context.Background(), k, valueToPut)
@@ -183,7 +180,7 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
 }
 
 func isJSON(textToCheck []byte) bool {
-	var js map[string]interface{}
+	var js struct{}
 	return json.Unmarshal(textToCheck, &js) == nil
 }
 
@@ -216,22 +213,6 @@ func (c *CouchDBStore) Get(k string) ([]byte, error) {
 	return c.getStoredValueFromRawDoc(rawDoc, k)
 }
 
-func (c *CouchDBStore) addRevID(valueToPut []byte, revID string) ([]byte, error) {
-	var m map[string]interface{}
-	if err := json.Unmarshal(valueToPut, &m); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal put value: %w", err)
-	}
-
-	m["_rev"] = revID
-
-	newValue, err := json.Marshal(m)
-	if err != nil {
-		return nil, fmt.Errorf("failed to marshal put value: %w", err)
-	}
-
-	return newValue, nil
-}
-
 // get rev ID
 func (c *CouchDBStore) getRevID(k string) (string, error) {
 	rawDoc := make(map[string]interface{})
@@ -259,6 +240,11 @@ func (c *CouchDBStore) Delete(k string) error {
 	revID, err := c.getRevID(k)
 	if err != nil {
 		return err
+	}
+
+	// no error if nothing to delete
+	if revID == "" {
+		return nil
 	}
 
 	_, err = c.db.Delete(context.TODO(), k, revID)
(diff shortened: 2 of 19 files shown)


.golangci.yml
```diff
@@ -83,6 +83,9 @@ linters-settings:
 linters:
   enable-all: true
   disable:
+    - godot
+    - nestif
+    - testpackage
```
llorllale, line 88: why are we disabling these?

llorllale:
@soluchok it seems like the test cases are duplicated among the stores. We should have all test cases in one place and subject all implementations to those.

.golangci.yml
soluchok: I have updated golangci to support t.Cleanup function (added in go 1.4). These linters were added in the new version.

soluchok:
> @soluchok it seems like the test cases are duplicated among the stores. We should have all test cases in one place and subject all implementations to those.

I moved common tests to `pkg/storage/store_test.go`. If I remove duplicates it will decrease test coverage. The idea is to keep common tests in `pkg/storage`. (responsible for behavior). And each storage implementation should have its own unit tests.

pkg/storage/couchdb/couchdbstore.go
```diff
@@ -155,12 +155,7 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
 		return errors.New("key and value are mandatory")
 	}
 
-	var valueToPut []byte
-	if isJSON(v) {
-		valueToPut = v
-	} else {
-		valueToPut = wrapTextAsCouchDBAttachment(v)
-	}
+	valueToPut := wrapTextAsCouchDBAttachment(v)
```
fqutishat, line 158: we shouldn't save data in Attachment it's bad for performance. we should check if data format is json will save in fields

fqutishat: you delete the correct behaviour

soluchok: in that case, implementation is incompatible since we cannot use keys such as "_key, _rev, etc."

soluchok: it looks like one implementation allows you to do that and another does not.

pkg/storage/mem/mem_store.go
```diff
@@ -127,19 +128,49 @@ func (s *memStore) Get(k string) ([]byte, error) {
 
 // Iterator returns iterator for the latest snapshot of the underlying db.
 func (s *memStore) Iterator(start, limit string) storage.StoreIterator {
-	// TODO Change Store Iterator https://github.com/hyperledger/aries-framework-go/issues/852
+	if start == "" {
```
ToJen, line 131: just wondering if `limit` may also need to be checked

pkg/storage/couchdb/couchdbstore.go
```diff
@@ -155,38 +154,30 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
...
-	} else {
-		valueToPut = wrapTextAsCouchDBAttachment(v)
-	}
-
 	revID, err := c.getRevID(k)
 	if err != nil {
 		return err
 	}
 
+	doc := map[string]interface{}{"payload": v}
```
fqutishat, line 162: you can put all data in one field

fqutishat: you need to check data type

pkg/storage/couchdb/couchdbstore.go
```diff
@@ -155,45 +154,30 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
 		return errors.New("key and value are mandatory")
 	}
 
-	var valueToPut []byte
```
fqutishat, line 158: we need this logic

pkg/storage/couchdb/couchdbstore.go
soluchok: @fqutishat  I have changed the logic, review it again

pkg/storage/couchdb/couchdbstore.go
soluchok: we do not need to, it is always `[]byte`

soluchok: @fqutishat  why I cannot put data to one field? it is always `[]byte` I put `[]byte` I get `[]byte`.
It solves the problem and it even improves performance (less Marshall/Unmarshall).

fqutishat: it's not good if data is big

fqutishat: in some cases you need to save big binary in attachment like images

fqutishat: @troyronda

soluchok: ok, I will combine it

llorllale: @soluchok please put a comment in there with the rationale, much appreciated :)

soluchok: I rollbacked changes. But I wrapped JSON payload with the `payload` field.
@llorllale the problem was that we couldn't use keys such as "_key, _rev, etc."

pkg/storage/mem/mem_store.go
soluchok: Fixed. It supposed to be a `limit` not a `start`. Added common tests for this logic.

.golangci.yml
troyronda: Can we add an issue and comment to followup on these news linters? I haven't looked into what they analyze.

soluchok: Sure, looks like we need `nestif` (for sure). We do not need `testpackage`(for sure). And probably we need `godot`.
godot is about to add `.` at the end of comments.

soluchok: @troyronda FYI: it is not the latest version of `golangci.` There are even more linters. I will create an issue to update the  `golangci` to the latest version.

troyronda: Sounds good. @soluchok

soluchok: I've created an issue https://github.com/hyperledger/aries-framework-go/issues/2032

PR #2028, opened 2020-07-22
remove Verifiable Credentials and Presentations from store

arttor:
**Add methods to remove VC and VP from the store**

Currently, there is no way to remove issued VC and VP from the store so I've added required methods

### DIFF
--- a/pkg/store/verifiable/store.go
+++ b/pkg/store/verifiable/store.go
@@ -43,6 +43,8 @@ type Store interface {
 	GetPresentationIDByName(name string) (string, error)
 	GetCredentials() ([]*Record, error)
 	GetPresentations() ([]*Record, error)
+	RemoveCredentialByName(name string) error
+	RemovePresentationByName(name string) error
 }
 
 type record struct {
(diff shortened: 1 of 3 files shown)


troyronda:
@arttor Thanks for your contribution. Please note that Hyperledger repositories require DCO sign-off. See [failing check](https://github.com/hyperledger/aries-framework-go/pull/2028/checks?check_run_id=899390945).

pkg/store/verifiable/store.go
```diff
@@ -43,6 +43,8 @@ type Store interface {
 	GetPresentationIDByName(name string) (string, error)
 	GetCredentials() ([]*Record, error)
 	GetPresentations() ([]*Record, error)
+	RemoveCredential(name string) error
+	RemovePresentation(name string) error
```
soluchok, line 47: function's names should be `RemoveCredentialByName` and `RemovePresentationByName`

pkg/store/verifiable/store.go
```diff
@@ -229,6 +231,62 @@ func (s *StoreImplementation) GetPresentations() ([]*Record, error) {
...
+// RemoveCredential removes the verifiable credential and its records containing given name.
+func (s *StoreImplementation) RemoveCredential(name string) error {
+	if name == "" {
+		return errors.New("credential name is mandatory")
+	}
+
+	id, err := s.GetCredentialIDByName(name)
+	if err != nil && !errors.Is(err, storage.ErrDataNotFound) {
+		return fmt.Errorf("get credential id using name : %w", err)
+	}
```
soluchok, line 243: why do we need this condition  `!errors.Is(err, storage.ErrDataNotFound)` ?

pkg/store/verifiable/store.go
```diff
@@ -229,6 +231,62 @@ func (s *StoreImplementation) GetPresentations() ([]*Record, error) {
...
+}
+
+// RemovePresentation removes the verifiable presentation and its records containing given name.
+func (s *StoreImplementation) RemovePresentation(name string) error {
+	if name == "" {
+		return errors.New("presentation name is mandatory")
+	}
+
+	id, err := s.GetPresentationIDByName(name)
+	if err != nil && !errors.Is(err, storage.ErrDataNotFound) {
```
soluchok, line 260: same here `!errors.Is(err, storage.ErrDataNotFound)` ?

pkg/store/verifiable/store.go
arttor: Done

pkg/store/verifiable/store.go
arttor: Correct. we don't need this check here and it is ok to rerun this error message for missing ID too.
Fixed.

pkg/store/verifiable/store.go
arttor: Fixed

pkg/store/verifiable/store.go
```diff
@@ -229,6 +231,62 @@ func (s *StoreImplementation) GetPresentations() ([]*Record, error) {
...
+	}
+
+	return nil
+}
+
+func (s *StoreImplementation) remove(id, recordKey string) error {
+	if id == "" {
+		return errors.New("id is required")
+	}
+
```
soluchok, line 276: nit: It will never happen, I would remove it.

arttor: Done.

arttor: Can you please merge PR if everything is ok?

soluchok: I can't :) @troyronda @fqutishat @llorllale ^

Issue #2034, opened 2020-07-23
update Verifiable controllers


ToJen:
**What I'm trying to do**

Include the two newly added methods (`RemoveCredentialByName` and `RemovePresentationByName`) from the recently merged https://github.com/hyperledger/aries-framework-go/pull/2028.

**Expected result**
Updates to both `pkg/controller/command/verifiable` and `pkg/controller/rest/verifiable` packages.

PR #2042, opened 2020-07-24
feat: updated verifiable controllers with the new methods

ToJen:

**Title:**
Update verifiable controllers

**Description:**
Closes https://github.com/hyperledger/aries-framework-go/issues/2034

Summary:
- Implement the two methods recently added to the verifiable store.
- updated the command controller
- updated the rest controller

### DIFF
--- a/pkg/controller/rest/verifiable/operation.go
+++ b/pkg/controller/rest/verifiable/operation.go
@@ -31,12 +31,13 @@ const (
 	verifiablePresentationPath = VerifiableOperationID + "/presentation"
 
 	// credential paths
-	ValidateCredentialPath  = verifiableCredentialPath + "/validate"
-	SaveCredentialPath      = verifiableCredentialPath
-	GetCredentialPath       = verifiableCredentialPath + "/{id}"
-	GetCredentialByNamePath = verifiableCredentialPath + "/name" + "/{name}"
-	GetCredentialsPath      = VerifiableOperationID + "/credentials"
-	SignCredentialsPath     = VerifiableOperationID + "/signcredential"
+	ValidateCredentialPath     = verifiableCredentialPath + "/validate"
+	SaveCredentialPath         = verifiableCredentialPath
+	GetCredentialPath          = verifiableCredentialPath + "/{id}"
+	GetCredentialByNamePath    = verifiableCredentialPath + "/name" + "/{name}"
+	GetCredentialsPath         = VerifiableOperationID + "/credentials"
+	SignCredentialsPath        = VerifiableOperationID + "/signcredential"
+	RemoveCredentialByNamePath = verifiableCredentialPath + "/remove/name" + "/{name}"
 
 	// presentation paths
 	GeneratePresentationPath     = verifiablePresentationPath + "/generate"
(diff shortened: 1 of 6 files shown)


fqutishat:
@ToJen where is the integration test

ToJen:
@fqutishat 
Should I create a new feature file here: https://github.com/hyperledger/aries-framework-go/tree/master/test/bdd/features?

It appears verifiable has no existing bdd tests. Or should I add a new scenario in `issue_credential.feature`

fqutishat:
> @fqutishat
> Should I create a new feature file here: https://github.com/hyperledger/aries-framework-go/tree/master/test/bdd/features?
> 
> It appears verifiable has no existing bdd tests. Or should I add a new scenario in `issue_credential.feature`

@ToJen https://github.com/hyperledger/aries-framework-go/tree/master/test/aries-js-worker

pkg/controller/rest/verifiable/operation.go
```diff
@@ -240,3 +244,33 @@ func (o *Operation) GeneratePresentation(rw http.ResponseWriter, req *http.Reque
 func (o *Operation) GeneratePresentationByID(rw http.ResponseWriter, req *http.Request) {
 	rest.Execute(o.command.GeneratePresentationByID, rw, req.Body)
 }
+
+// RemoveCredentialByName swagger:route POST /verifiable/credential/remove/name/{name}
```
soluchok, line 248: this comment should be like this:
`RemoveCredentialByName swagger:route POST /verifiable/credential/remove/name/{name} verifiable RemoveCredentialByNameReg `
where `verifiable` groups your APIs by name `verifiable`
and `RemoveCredentialByNameReq` describes the request 

the swagger spec will be incorrect, you need to check it by `make run-openapi-demo` https://github.com/hyperledger/aries-framework-go/blob/master/docs/rest/openapi_demo.md

NOTE: it is related to every operation (models are missing)

pkg/controller/rest/verifiable/operation.go
```diff
@@ -240,3 +244,33 @@ func (o *Operation) GeneratePresentation(rw http.ResponseWriter, req *http.Reque
...
+//
+// Removes a verifiable credential by name.
+//
+// Responses:
+//    default: genericError
+//        200: emptyRes
```
soluchok, line 254: also, I see that `emptyRes` is not working properly.
the structure should be with the body inside.
e.g 
```
// emptyRes model
//
// swagger:response emptyRes
type emptyRes struct {
Body struct {}
}
```
I also noticed that we are using that structure which means we need to duplicate it.
One structure should be for swagger another structure should be for `Go` usage.

soluchok: the working solution can be like this:
```
type emptyRes struct {}

// emptyResponse model
//
// swagger:response emptyResponse
type emptyResponse struct { // nolint:unused,deadcode
	// in: body
	Body emptyRes
}
```
In that case, you need to replace `200: emptyRes` with `200: emptyResponse`

pkg/controller/rest/verifiable/operation.go
ToJen: i'll add these

pkg/controller/rest/verifiable/operation.go
ToJen: updated to reflect this

PR #2046, opened 2020-07-27
feat: Present Proof - added WillConfirm to the RequestPresentation

soluchok:
Added `WillConfirm` field to the `RequestPresentatation` model.


pkg/didcomm/protocol/presentproof/models.go
```diff
@@ -38,6 +41,7 @@ type RequestPresentation struct {
 }
 
 // Presentation is a response to a RequestPresentation message and contains signed presentations.
+// TODO: Add ~please_ack decorator support for the protocol
```
rolsonquadras, line 44: Do we have an issue for this ?

soluchok: No, we haven’t. I’ll create.

PR #2052, opened 2020-07-28
fix: check VP embedded proof

### LINKED ISSUE #2049: VP embedded proof is not checked

kdimak:
closes #2049

### DIFF
--- a/pkg/doc/verifiable/credential_testsuite.go
+++ b/pkg/doc/verifiable/credential_testsuite.go
@@ -24,3 +24,17 @@ func WithPresNoProofCheck() PresentationOpt {
 		opts.disabledProofCheck = true
 	}
 }
+
+// WithPresRequireVC option enables check for at least one verifiableCredential in the VP.
+func WithPresRequireVC() PresentationOpt {
+	return func(opts *presentationOpts) {
+		opts.requireVC = true
+	}
+}
+
+// WithPresRequireProof option enables check for at least one proof in the VP.
+func WithPresRequireProof() PresentationOpt {
+	return func(opts *presentationOpts) {
+		opts.requireProof = true
+	}
+}
--- a/pkg/doc/verifiable/example_presentation_test.go
+++ b/pkg/doc/verifiable/example_presentation_test.go
@@ -9,8 +9,12 @@ package verifiable_test
 import (
 	"crypto/ed25519"
 	"encoding/json"
+	"errors"
 	"fmt"
 
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/jsonld"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/suite"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/suite/ed25519signature2018"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/verifier"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/util"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/util/signature"
@@ -488,7 +492,7 @@ func ExamplePresentation_MarshalledCredentials() {
 
 	// Decode VP from JWS.
 	// Note that VC-s inside will be decoded as well. If they are JWS, their signature is verified
-	// and thus we need to make sure the public key fetcher can access the
+	// and thus we need to make sure the public key fetcher can retrieve the public key.
 	vp, err = verifiable.ParsePresentation(
 		[]byte(vpJWS),
 		verifiable.WithPresPublicKeyFetcher(func(issuerID, keyID string) (*verifier.PublicKey, error) {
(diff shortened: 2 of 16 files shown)


pkg/doc/verifiable/credential_testsuite.go
```diff
@@ -24,3 +24,17 @@ func WithPresNoProofCheck() PresentationOpt {
 		opts.disabledProofCheck = true
 	}
 }
+
+// WithPresRequireVC option enables check for at least one verifiableCredential in the VP.
+func WithPresRequireVC() PresentationOpt {
```
sudeshrshetty, line 29: may be a better name? We don't need those 'Pres' prefix since they are already 'PresentationOpt' type

pkg/doc/verifiable/test-suite/verifiable_suite_test.go
```diff
@@ -225,7 +225,10 @@ func encodeVCToJSON(vcBytes []byte, testFileName string) {
 func encodeVPToJSON(vcBytes []byte) {
 	// https://www.w3.org/TR/vc-data-model/#presentations-0 states "If present" under verifiableCredential
 	// but the test suite requires the element to be present. Hence, WithPresRequireVC is used in test suite runs.
-	vp, err := verifiable.ParsePresentation(vcBytes, verifiable.WithPresRequireVC())
+	vp, err := verifiable.ParsePresentation(vcBytes,
+		verifiable.WithPresRequireVC(),
+		verifiable.WithPresRequireProof(),
+		verifiable.WithDisabledPresentationProofCheck())
```
llorllale, line 231: We are requiring proof yet disabling the proof check...

pkg/doc/verifiable/credential_testsuite.go
kdimak: This is done to keep consistent naming of everything related to VPs. For VC, it starts from `With...`, for VP the prefix is `WithPres...`.

pkg/doc/verifiable/test-suite/verifiable_suite_test.go
kdimak: We use `WithDisabledPresentationProofCheck()` to disable a check of the signature of the proof. This is applied in the VC test suite as their verification methods are invalid (e.g. DIDs are unresolvable). We apply the same approach for the VC.
As for `WithPresRequireVC()` and  `WithPresRequireProof()` options, they are used to insist on presence of `verifiableCredential` and `proof` fields in the VP. This is a sort of workaround to pass the test cases. In the https://www.w3.org/TR/vc-data-model spec, for example, the presence of `verifiableCredential` is not mandatory.

kdimak:
> See my comment.
> 
> Additionally, I'm concerned that I don't see a clear test case using `ParsePresentation()` that demonstrates it works. Or maybe I'm not seeing it? Please clarify.

Good point! I added the example of VP with linked data proof creation (on the Holder side) and then verification on the Verifier side, `ExamplePresentation_AddLinkedDataProof()`. Also, a follow-up issue #2054 is filed.

pkg/doc/verifiable/test-suite/verifiable_suite_test.go
rolsonquadras: @kdimak - I agree with @llorllale , it's confusing. Few questions would be -  which would take precedence (does it check vp proof or not?) when both are set (`WithPresRequireProof` and `WithDisabledPresentationProofCheck`) ? Does the order matter ? what happens when none of those are set ?

I feel we should have a default behaviour and then one option to override it.

llorllale: @kdimak it now looks like this:

```go
	vp, err := verifiable.ParsePresentation(vcBytes,
		verifiable.WithPresRequireVC(),
		verifiable.WithPresRequireProof(),
		verifiable.WithPresDisabledProofCheck())
```

Trying to understand what you're saying... `WithPresRequireProof()` merely checks the presence of the `proof` attribute, while `WithPresDisabledProofCheck()` disables verification of the `proof` element (if present). Is that it?

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -563,3 +567,187 @@ func ExamplePresentation_MarshalledCredentials() {
...
+		Suite:                   ed25519signature2018.New(suite.WithSigner(issuerSigner)),
+		SignatureRepresentation: verifiable.SignatureJWS,
+		VerificationMethod:      "did:example:123456#key1",
+	}, jsonld.WithDocumentLoader(getJSONLDDocumentLoader()))
+	if err != nil {
+		panic(fmt.Errorf("failed to add linked data proof: %w", err))
+	}
+
+	// 2. ISSUER creates a VP with the VC enclosed.
+	vpStr := `
```
llorllale, line 623: @kdimak in this case, since we already have the VC, isn't the proper way to begin creating a VP with the VC by first calling `vc.Presentation()` ?

llorllale: As a user, I'm trying to understand how to best use this library

pkg/doc/verifiable/example_presentation_test.go
```diff
@@ -563,3 +567,187 @@ func ExamplePresentation_MarshalledCredentials() {
...
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://www.w3.org/2018/credentials/examples/v1"
+  ],
+  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
+  "type": [
+    "VerifiablePresentation",
+    "UniversityDegreeCredential"
```
llorllale, line 632: VP of type UniversityDegreeCredential?

pkg/doc/verifiable/test-suite/verifiable_suite_test.go
rolsonquadras: In that case, the VC just has one option. https://github.com/hyperledger/aries-framework-go/blob/694a5b6c0c217f18ff84f47ddbd9ca598d26455f/pkg/doc/verifiable/credential.go#L589-L593 
Any specific reason to separate proof attribute check and validate the proof in VP only?

kdimak: 1. `WithDisabledProofCheck()` for VCs and `WithPresDisabledProofCheck()` for VPs serve the same purpose: we want to avoid checking crypto primitives (as this check will lead to the error due to unresolvable DIDs in VC test suite data).
2. `WithPresRequireProof()` is used to pass one of the test cases from the VC test suite to have a mandatory `proof` element that [seems incorrect](https://www.w3.org/TR/vc-data-model/#presentations-0). For example, we can parse VP in JWS form, and thus it does not require to have a `proof` element (it's called an embedded proof) since we already verified VP's external proof (a signature from JWT).

pkg/doc/verifiable/example_presentation_test.go
kdimak: Yes, that is correct, we could use `vc.Presentation()`. I think we should use this when we have a parsed VC (using `ParseCredential()` or `ParseUnverifiedCredential()` or constructed manually). From the other hand, if we have VC in JSON form (or in JWT form), then we can construct the VP and then pass one or more credentials in text form using `vp.SetCredentials()`.

pkg/doc/verifiable/example_presentation_test.go
kdimak: Agree, it confuses, I will drop it.

pkg/doc/verifiable/example_presentation_test.go
kdimak: And I will change the example to use `vc.Presentation()`, this is more relevant use case. Thanks @llorllale !

pkg/doc/verifiable/presentation_test.go
```diff
@@ -46,31 +49,7 @@ const validPresentation = `
...
-  "proof": {
-    "type": "Ed25519Signature2018",
-    "created": "2020-01-21T16:44:53+02:00",
-    "proofValue": "eyJhbGciOiJSUzI1NiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..kTCYt5XsITJX1CxPCT8yAV-TVIw5WEuts01mq-pQy7UJiN5mgREEMGlv50aqzpqh4Qq_PbChOMqsLfRoPsnsgxD-WUcX16dUOqV0G_zS245-kronKb78cPktb3rk-BuQy72IFLN25DYuNzVBAh4vGHSrQyHUGlcTwLtjPAnKb78"
-  }
-}
-`
-
-// nolint:lll
-const validEmptyPresentation = `
```
troyronda, line 59: not needed anymore?

pkg/doc/verifiable/presentation_test.go
```diff
@@ -182,18 +153,6 @@ func TestParsePresentation(t *testing.T) {
 		verify(t, vp)
 	})
 
-	t.Run("creates a new Verifiable Presentation from JSON with invalid empty VC structure", func(t *testing.T) {
```
troyronda, line 185: not needed anymore?

pkg/doc/verifiable/presentation_test.go
```diff
@@ -403,20 +380,6 @@ func TestValidateVP_Holder(t *testing.T) {
 	})
 }
 
-func TestValidateVP_Proof(t *testing.T) {
```
troyronda, line 406: not needed anymore?

pkg/doc/verifiable/presentation_test.go
kdimak: I moved `WithPresRequireVC()` and new `WithPresRequireProof()` options to the `credential_testsuite.go` (under `testsuite` tag). But yeah, I did not move the tests. Corrected now - I also added the test for the empty `proof` field. See `TestNewPresentationWithEmptyFields()`.

PR #2068, opened 2020-07-31
remove and replace legacyKMS with LocalKMS

### LINKED ISSUE #752: Move signer to crypto

### LINKED ISSUE #1207: fix hardcode base58

### LINKED ISSUE #1828: Create New Packer
As a follow up to #1798

We need to create a new Packer instance based on the new JWE document and the new LocalKMS instances.

Make this new Packer the default framework one.

### LINKED ISSUE #1993: Generate DID docs with Encryption key as KeyAgreement
As a followup to #986 

DID docs need to store encryption keys to be used by the Packer. The sender needs the recipients keys to Pack() an envelope and the recipient needs the sender key (particularly the sender key id `skid` sent by the sender and its corresponding public key). In order to distinguish verification keys from encryption ones, we need to separate them in the DID document. KeyAgreement seems a viable option. Investigate if this is possible as the agents need the static keys exchange to occur prior to exchanging didComm messages.

Also, part of #986, there is a new storage created for the purpose of storing third party keys in the Authcrypt packer (called ThirdPartyKeysDB). Store the sender key there, before didComm, so Authcrypt can unpack messages on the recipient side.

Storing recipients keys on the sender side in ThirdPartyKeysDB is not necessary as the sender expects to get the keys directly from DID exchange (if the suggestion in the first paragraph above is viable), but a recipient can eventually become a sender in the future. It would be good to store their encryption key (it will be useful when the sender and recipient swap roles).

### LINKED ISSUE #2041: Update PublicKey ID limit to 50  and use JWK fingerprint as the keyID
in order to be fully JWK compliant, key IDs in the KMS must use JWK thumbprint as per https://tools.ietf.org/html/rfc7638#section-3.1

also this means rotating a key should update the Key ID, which means this issue is not valid anymore and should be closed: #1837 

The Key ID must change when rotating a key, the current key rotation's ID update behaviour in the KMS is valid and must not change.

Baha-sk:
This change removes completely LegacyKMS and updates keys with the new LocalKMS keys and KIDs.

Asymmetric key IDs are represented as a JWK thumbprint base64RawURL encoded (no padding).
Symmetric keys are randomly generated.

The only remaining code from LegacyKMS is crypto_box.go file which is moved in this change to LocalKMS to support LegacyPacker.

This change also removes experimental pkg/didcom/packer/jwe/authcrypt packer as it is not used by the framework and was created as a POC.

and other changes...

closes #1828 Create New Packer.
closes #752 Signer interface in LegacyKMS removed. All current and new code should use the Crypto service for signing.
closes #2041 Update Public Key ID length to max 50. Also now Asymetric keys use JWK thumbprint base64URL encoded.
attempted #1993 Generate DID docs with KeyAgreement comprising the Encryption key in JsonWebKey2020 format.
        The above issue cannot be closed due to unclear DID Exchang/DIDComm RFC documentation around verifying signatures using RecipientKeys
        See https://github.com/hyperledger/aries-rfcs/issues/520 for more details.
attempted #1207 fix hardcode base58. Due to LegacyPacker still requiring keys base58 encoded as well as DID docs storing key values base58 
        encoded, this issue could not be completed in this change.


pkg/didcomm/common/service/destination.go
```diff
@@ -54,10 +64,67 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
 
 	// TODO ensure recipient keys are did:key's
 	//  https://github.com/hyperledger/aries-framework-go/issues/1604
+	// above TODO is TBD - for now added KMS encryption key to support new Authcrypt/Anoncrypt
```
llorllale, line 67: a TODO about a TODO....

pkg/didcomm/common/service/destination.go
```diff
@@ -54,10 +64,67 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
...
+		ServiceEndpoint:  didCommService.ServiceEndpoint,
+		RoutingKeys:      didCommService.RoutingKeys,
+		RecipientEncKeys: recipientEncKeys,
 	}, nil
 }
+
+func extractECKey(encKeyJWK *ariesjose.JWK) (*composite.PublicKey, error) {
+	tp, err := encKeyJWK.Thumbprint(sha256.Size)
+	if err != nil {
+		return nil, err
```
llorllale, line 100: please wrap these errors... I've been unfortunate enough to have to hunt down unwrapped errors in this file

pkg/didcomm/common/service/destination.go
```diff
@@ -54,10 +64,67 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
...
+		id, err := computeVerKID([]byte(recKey))
+		if err != nil {
+			return nil, err
+		}
+
+		encKey, encKeyFound := diddoc.LookupEncryptionKey(id, didDoc)
+		if encKeyFound {
+			enc, er := extractECKey(encKey.JSONWebKey())
+			if er != nil {
+				return nil, er
```
llorllale, line 81: please wrap these errors

pkg/didcomm/common/service/destination.go
```diff
@@ -54,10 +64,67 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
...
+	switch pubKey := encKeyJWK.Key.(type) {
+	case *ecdsa.PublicKey:
+		return &composite.PublicKey{
+			KID:   kid,
+			X:     pubKey.X.Bytes(),
+			Y:     pubKey.Y.Bytes(),
+			Curve: pubKey.Curve.Params().Name,
+		}, nil
+	default:
+		return nil, errors.New("")
```
llorllale, line 114: error message

pkg/didcomm/common/service/destination.go
```diff
@@ -19,6 +28,7 @@ type Destination struct {
 	ServiceEndpoint      string
 	RoutingKeys          []string
 	TransportReturnRoute string
+	RecipientEncKeys     []*composite.PublicKey
```
llorllale, line 31: Why isn't `RecipientKeys` deleted?

pkg/didcomm/protocol/didexchange/states_test.go
```diff
@@ -51,69 +52,69 @@ func TestNoopState(t *testing.T) {
...
 // null state can transition to invited state or requested state.
 func TestNullState(t *testing.T) {
-	null := &null{}
-	require.Equal(t, "null", null.Name())
-	require.False(t, null.CanTransitionTo(null))
-	require.True(t, null.CanTransitionTo(&invited{}))
-	require.True(t, null.CanTransitionTo(&requested{}))
-	require.False(t, null.CanTransitionTo(&responded{}))
-	require.False(t, null.CanTransitionTo(&completed{}))
+	nul := &null{}
```
llorllale, line 55: accidental refactors?

troyronda:
@Baha-sk https://lgtm.com/projects/g/hyperledger/aries-framework-go/rev/pr-95263ef7faad27f1315ef659c1550d0c3b54635d

pkg/didcomm/common/service/destination.go
Baha-sk: must have been very late and forgot to add a message..

pkg/didcomm/common/service/destination.go
Baha-sk: `RecipientKeys` contains remain verification keys and are also used by LegacyPacker

pkg/didcomm/protocol/didexchange/states_test.go
Baha-sk: no.. wanted to remove `shadowing warning`

pkg/didcomm/common/service/destination.go
Baha-sk: it says TBD of the above TODO.. not sure if we want to support fetching keys from did:key, currently keys are in-lined and encryption keys are found in KeyAgreement that contains a full JsonWebKey2020 JWK key, it could be considered a did:key, not sure we need anything more.. this is why this TBD is added

pkg/didcomm/common/service/destination.go
Baha-sk: will try to do my best..

pkg/didcomm/common/service/destination.go
llorllale: @Baha-sk 

> it says TBD of the above TODO.. not sure if we want to support fetching keys from did:key, currently keys are in-lined and encryption keys are found in KeyAgreement that contains a full JsonWebKey2020 JWK key, it could be considered a did:key, not sure we need anything more.. this is why this TBD is added

Current [DIDComm service conventions](https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions) <-- this is what we should be following (for now).

Baha-sk: @llorllale This convention only states RecipientKeys, which are supposedly representing verification keys. The new packers will use separate keys for packing/unpacking that will be regarded as encryption keys. The way to allow this separation, we need to use the DID Doc's KeyAgreement method to store the encryption key (similar to `X25519KeyAgreementKey2019` shown in [this example](https://w3c.github.io/did-core/#example-20-key-agreement-property-containing-two-verification-methods) but using JSONWebKey2020 as type, pending approval by the w3c community).

This way, we don't diverge from the [DIDComm service conventions](https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions).

llorllale: @Baha-sk 

> This convention only states RecipientKeys, which are supposedly representing verification keys.

`recipientKeys` is an array of `did:key` that resolve to DID docs that look like [this](https://w3c-ccg.github.io/did-method-key/#example-2-a-did-document-derived-from-a-did-key) (note the `keyAgreement`). Having the keyAgreement key inlined like this (in the form of did:key references) allows a uniform representation across DIDComm usages, whether it's in a DID Doc or in an invitation (ie. no DID doc).

Please follow the service conventions - let's not reinvent our own unless there is a good reason for it.

Baha-sk: @llorllale  this doc creation from did:key implies key conversion from ed25519 to X25519. We aim to separate verification from encryption key (ie no conversion or value relationship between the keys). If this is not the goal, then encryption key might as well be the verification key for ECDSA keys. I'm sure this key update is all about increasing security by separating the two. 

If we want separate enc and ver keys, this also means we can't create (expand) a DID doc from did:key. But the example your showing can only work using key conversion (to create the `KeyAgreement` part of the document)

llorllale: @Baha-sk 

> the example your showing can only work using key conversion (to create the KeyAgreement part of the document to contain the converted X25519 key from the ED25519 ver key).

That is exactly how `did:key` works with Ed25519 keys. 

The point "but why not just multicodec+multiformat the key agreement key?" was raised at the time the decision to use `did:key` in DIDComm was made. The answer is pragmatic: to do so, the DIDComm community would have to invest resources into developing the formats and encoding in a proper standards body. It was decided to just use did:key which has done part of the legwork already.

So in the end we need to use did:key in `recipientKeys` and `routingKeys` to remain compliant with Aries DIDComm.

Baha-sk: this only works for ED25519 keys @llorllale , if we want to use EC verification keys for signature verification and separate (unrelated) EC keys for packing/unpacking then we can't generate a DID doc from `did:key`.

I posted a comment [here](https://github.com/w3c/did-core/issues/185#issuecomment-668658079) to ask the community.

Baha-sk: @llorllale if I understood this [comment](https://github.com/w3c/did-core/issues/185#issuecomment-668668618) correctly, we will need to add did:key support to find `somehow` the `encryption` keys from the `RecipientKeys` list to do do Pack/Unpack. The remaining keys should be used for verification. 

This means the framework's logic in handling this list should change (not all keys are verification keys in the list and therefore should use only a subset representing verification only keys). 

This is beyond the scope of this patch and can be done in a followup change.

pkg/didcomm/common/service/destination.go
Baha-sk: updated

pkg/didcomm/common/service/destination.go
Baha-sk: updated

pkg/didcomm/common/service/destination.go
Baha-sk: I guess in a future PR we will even have RecipientVerKeys as per [this](https://github.com/hyperledger/aries-framework-go/pull/2068#discussion_r465173320) comment.

pkg/didcomm/common/service/destination.go
Baha-sk: updated

pkg/didcomm/common/service/destination.go
llorllale: You mean after [this](https://github.com/hyperledger/aries-framework-go/pull/2068#discussion_r463957199) comment? :)

pkg/client/outofband/client.go
```diff
@@ -327,6 +328,11 @@ func didServiceBlockFunc(p Provider) func() (*did.Service, error) {
 			return nil, errors.New("cast service to Route Service failed")
 		}
 
+		verKey, err := p.KMS().ExportPubKeyBytes(verKeyID)
+		if err != nil {
+			return nil, err
```
llorllale, line 333: please wrap this error

pkg/crypto/tinkcrypto/primitive/composite/ecdh1pu/subtle/ecdh1pu_aes_aead_composite_encrypt.go
```diff
@@ -17,9 +17,13 @@ import (
...
 
-// A256GCM is the default content encryption algorithm value as per
-// the JWA specification: https://tools.ietf.org/html/rfc7518#section-5.1
-const A256GCM = "A256GCM"
+const (
+	// A256GCM is the default content encryption algorithm value as per
+	// the JWA specification: https://tools.ietf.org/html/rfc7518#section-5.1
+	A256GCM = "A256GCM"
+	// DIDComEncType representing the JWE 'Typ' protected type header
+	DIDComEncType = "didcomm-envelope-enc"
```
llorllale, line 25: typo: DIDCom**m**EncType

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_encrypt.go
```diff
@@ -16,9 +16,13 @@ import (
...
 
-// A256GCM is the default content encryption algorithm value as per
-// the JWA specification: https://tools.ietf.org/html/rfc7518#section-5.1
-const A256GCM = "A256GCM"
+const (
+	// A256GCM is the default content encryption algorithm value as per
+	// the JWA specification: https://tools.ietf.org/html/rfc7518#section-5.1
+	A256GCM = "A256GCM"
+	// DIDComEncType representing the JWE 'Typ' protected type header
+	DIDComEncType = "didcomm-envelope-enc"
```
llorllale, line 24: why is this constant duplicated?

pkg/didcomm/packager/packager.go
```diff
@@ -73,8 +77,17 @@ func New(ctx Provider) (*Packager, error) {
...
 func (bp *Packager) addPacker(pack packer.Packer) {
-	if bp.packers[pack.EncodingType()] == nil {
-		bp.packers[pack.EncodingType()] = pack
+	packerID := pack.EncodingType()
+
+	_, ok := pack.(*authcrypt.Packer)
+	if ok {
+		// anoncrypt and authcrypt have the same encoding type
+		// so authcrypt will have an appended suffix
+		packerID += authSuffix
```
llorllale, line 86: I don't see `UnpackMessage()` accounting for this suffix... it's just taking the `Typ` as-is. Is that for a followup PR?

pkg/didcomm/packer/anoncrypt/pack.go
```diff
@@ -40,13 +40,16 @@ type Packer struct {
 
 // New will create an Packer instance to 'AnonCrypt' payloads for a given list of recipients.
 // The returned Packer contains all the information required to pack and unpack payloads.
-func New(ctx packer.Provider, encAlg jose.EncAlg) *Packer {
+func New(ctx packer.Provider, encAlg jose.EncAlg) (*Packer, error) {
 	k := ctx.KMS()
+	if k == nil {
+		return nil, errors.New("anoncrypt: failed to create packer because KMS is empty")
```
llorllale, line 46: why would this happen?

The framework's defaults will inject the KMS. If the user injects a `nil` KMS then it renders the framework largely useless.... in which case I don't mind a panic. Although you can add a check to `WithKMS()` for this.

pkg/didcomm/packer/authcrypt/pack.go
```diff
@@ -51,8 +51,16 @@ type Packer struct {
 // The returned Packer contains all the information required to pack and unpack payloads.
 func New(ctx packer.Provider, encAlg jose.EncAlg) (*Packer, error) {
 	k := ctx.KMS()
+	if k == nil {
+		return nil, errors.New("authcrypt: failed to create packer because KMS is empty")
```
llorllale, line 55: same here, this check doesn't make sense this deep into the framework

pkg/didcomm/packer/legacy/authcrypt/unpack.go
```diff
@@ -147,6 +160,9 @@ func decodeSender(b64Sender string, pk []byte, km legacykms.KeyManager) ([]byte,
 	senderData := base58.Decode(string(senderPub))
 
 	senderPubCurve, err := cryptoutil.PublicEd25519toCurve25519(senderData)
+	if err != nil {
+		return nil, nil, err
```
llorllale, line 164: please wrap error

pkg/didcomm/protocol/mediator/models.go
```diff
@@ -38,7 +38,10 @@ type KeylistUpdate struct {
 // Update route key update message.
 type Update struct {
 	RecipientKey string `json:"recipient_key,omitempty"`
+	RecipientKID string `json:"recipient_kid,omitempty"`
```
llorllale, line 41: this would require updates to Aries RFCs, if really needed

pkg/doc/jose/decrypter.go
```diff
@@ -137,29 +138,33 @@ func (jd *JWEDecrypt) addSenderKey(skid string) error {
...
 
 	encAlg, ok := protectedHeaders.Encryption()
 	if !ok {
-		return nil, "", fmt.Errorf("jwe is missing encryption algorithm 'enc' header")
+		return nil, "", "", fmt.Errorf("jwe is missing encryption algorithm 'enc' header")
 	}
 
+	// TODO go Jose doesn't set/enforce content `typ` in the protected headers when encrypting. When fixed, remove
+	//      the following line and add a check to ensure it's available.
+	encType := "didcomm-envelope-enc" // used by authcrypt/anoncrypt. The jose package is not used for LegacyPacker.
```
llorllale, line 155: there are constants for this elsewhere

pkg/doc/jose/encrypter_decrypter_test.go
```diff
@@ -30,18 +30,20 @@ import (
 	mockstorage "github.com/hyperledger/aries-framework-go/pkg/mock/storage"
 )
 
+const encodingType = "didcomm-envelope-enc"
```
llorllale, line 33: duplicated constant

pkg/doc/verifiable/common.go
```diff
@@ -92,6 +92,8 @@ func (r *DIDKeyResolver) resolvePublicKey(issuerDID, keyID string) (*verifier.Pu
 
 	methods := doc.VerificationMethods()
 	for _, verificationMethods := range methods {
+		// TODO verify which verificationMethod should be checked in order of priority. Looping through the list isn't
```
llorllale, line 95: source?

pkg/didcomm/packager/packager.go
Baha-sk: it's further down below at line [184](https://github.com/hyperledger/aries-framework-go/pull/2068/files/ff7c21b718eeb3fcf2e072a2c390176340e995e6#diff-68bdfb240e9da84a7ae7cdd37c85a8e4R164).

when Pack is called, we know the packer type, so we can check if it's authcrypt, but when unpack is called, we need to discover the packer to use which is done via the envelope's contentType. The comment at line 164 explains the rest.

pkg/didcomm/packer/anoncrypt/pack.go
Baha-sk: Your assumption is that someone initialized the framework. If someone wants the use this `anoncrypt` without initializing the framework, then this is a gentle error to tell the user to setup a kms.

pkg/didcomm/packer/authcrypt/pack.go
Baha-sk: same response as `anoncrypt` above

pkg/didcomm/packer/legacy/authcrypt/unpack.go
Baha-sk: For security reasons, we should not provide hints on any crypto related logic like key conversions, decoding ciphertext, iv, tag, encrypting or decrypting in general, etc.

pkg/didcomm/protocol/mediator/models.go
Baha-sk: indeed.. which triggered me to discuss this on the A WG meeting last week. The same goes with EncKey and EncKID. All these new fields are unused in this patch. I will remove them for now and probably add them once the RFCs are updated.

pkg/doc/verifiable/common.go
Baha-sk: currently this logic simply loops through the list verification methods. This works right now because the list of verifications all have the same key.  And there is only 1 key in each.

I believe verifying proof should use the `authentication` method, right? I don't see this logic anywhere.

pkg/doc/jose/decrypter.go
Baha-sk: > many things going on here but my main concern is the required changes to RFCs. Let's work out a plan on that first before merging this PR.

I would prefer to remove the additional changes required for the new `encryption` keys in didComm (like recipientEncKeys, EncKey and EncKID) and have this pr merged so we can have legacyKMS completely removed. This will unblock people who need a single source of keys.

Baha-sk:
> many things going on here but my main concern is the required changes to RFCs. Let's work out a plan on that first before merging this PR.

I would prefer to remove the new changes related to recipientEncKeys, EncKey and EncKID and do those in a future PR and merge this PR once ready (I'm not far off to finalizing it).

By doing this, this PR's main focus will be on removing legacyKMS from the framework which will unblock you and others who need a single source of keys. It's close to the finish line.

llorllale:
> > many things going on here but my main concern is the required changes to RFCs. Let's work out a plan on that first before merging this PR.
> 
> I would prefer to remove the new changes related to recipientEncKeys, EncKey and EncKID and do those in a future PR and merge this PR once ready (I'm not far off to finalizing it).
> 
> By doing this, this PR's main focus will be on removing legacyKMS from the framework which will unblock you and others who need a single source of keys. It's close to the finish line.

Sounds great!

pkg/client/outofband/client.go
Baha-sk: done

pkg/crypto/tinkcrypto/primitive/composite/ecdh1pu/subtle/ecdh1pu_aes_aead_composite_encrypt.go
Baha-sk: updated

pkg/crypto/tinkcrypto/primitive/composite/ecdhes/subtle/ecdhes_aes_aead_composite_encrypt.go
Baha-sk: moved to common package and removed duplicates

pkg/didcomm/packer/legacy/authcrypt/unpack.go
Baha-sk: done

pkg/doc/jose/decrypter.go
Baha-sk: removed duplicates

pkg/doc/jose/encrypter_decrypter_test.go
Baha-sk: removed

pkg/client/didexchange/client.go
```diff
@@ -304,6 +310,8 @@ func (c *Client) CreateConnection(myDID string, theirDID *did.Doc, options ...Co
 		conn.ServiceEndPoint = destination.ServiceEndpoint
 		conn.RecipientKeys = destination.RecipientKeys
 		conn.RoutingKeys = destination.RoutingKeys
+	} else {
+		return "", fmt.Errorf("create connection: failed to create destination: %w", err)
```
soluchok, line 314: Let’s remove `else` condition.
```
if err !=nil {
return "", fmt.Errorf("create connection: failed to create destination: %w", err)
}
```

pkg/kms/localkms/crypto_box.go
```diff
@@ -0,0 +1,251 @@
...
+//
+// CryptoBox is created by a LegacyKMS, and reads secret keys from the LegacyKMS
+//   for encryption/decryption, so clients do not need to see
+//   the secrets themselves.
+type CryptoBox struct {
+	km *LocalKMS
+}
+
+// NewCryptoBox creates a CryptoBox which provides crypto box encryption using the given KMS's key.
+func NewCryptoBox(w kms.KeyManager) (*CryptoBox, error) {
```
soluchok, line 47: why not receive *LocalKMS structure instead of checking it inside?

pkg/vdri/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,61 @@
...
+// PubKeyFromFingerprint extracts the raw public key from a did:key fingerprint.
+func PubKeyFromFingerprint(fingerprint string) ([]byte, error) {
+	// did:key:MULTIBASE(base58-btc, MULTICODEC(public-key-type, raw-public-key-bytes))
+	// https://w3c-ccg.github.io/did-method-key/#format
+	mc := base58.Decode(fingerprint[1:]) // skip leading "z"
+	if !bytes.Equal(multicodec(ed25519pub), mc[:2]) {
+		return nil, fmt.Errorf("pubKeyFromFingerprint: not supported public key (multicodec code: %#x)", mc[0])
+	}
+
+	return mc[2:], nil
```
soluchok, line 60: Why `2`? Add a comment, please.

Baha-sk: this is from the multicodec spec... the link is in the comment above the code

pkg/client/didexchange/client.go
Baha-sk: `if err == nil` seems weird

I'll remove the condition altogether. I guess the original author mean if CreateDestination was successful, then set the connection properties in the if clause. Otherwise, create a new connection altogether  (done under this code).

@llorllale @soluchok, should the `else` clause be removed fully or should I add `if err !=nil` as suggested by @soluchok ?

pkg/kms/localkms/crypto_box.go
Baha-sk: it can be *LocalKMS... not spending too much thinking about this CryptoBox functionality as it's tied to LegacyPacker and may be removed in the future. But by default a constructor uses interface arguments and implementations can be checked internally, this is what was followed here. 

I guess this argument can be changed to be the implementation type instead because only LegacyPacker uses it.

pkg/client/didexchange/client.go
soluchok: should the else clause be removed fully -I don't know that. My suggestion was based on what we have in this PR.

llorllale: I am the author of that part of the code.

The code is allowing the creation of connections with `theirDID` regardless whether it has a valid `did-communication` service or not.

I cannot recall my thinking at the time I wrote this.

Is it valid to create a connection between, say, a `did:peer` and a `did:key`? Or say two `did:peer` where one has `did-communication` and the other does not? The `service` information would have to be transmitted out of band in the latter case presumably.

I think, in the end, we don't have any use case for this right now. Plus I am unable to defend this logic.

I vote we fix it such that `CreateConnection()` fails if `service.CreateDestination()` returns an error.

Baha-sk: I'll switch the if/else clause to:
```
if err !=nil {
    return "", fmt.Errorf("create connection: failed to create destination: %w", err)
}
conn.ServiceEndPoint = destination.ServiceEndpoint
conn.RecipientKeys = destination.RecipientKeys
conn.RoutingKeys = destination.RoutingKeys
```

Baha-sk: updated

pkg/kms/localkms/crypto_box.go
Baha-sk: We can't change kms.KeyManager to *LocalKMS since this is called in the packager that calls this via a provider. If we do change, then this conversion will move up to the packager code. It's better to do it here rather than duplicating the conversion for each use of NewCryptoBox() even if packager is the only user.

pkg/client/outofband/client.go
```diff
@@ -312,38 +313,45 @@ func didServiceBlockFunc(p Provider) func() (*did.Service, error) {
...
+			return nil, fmt.Errorf("didServiceBlockFunc: failed Coordinate Mediate service: %w", err)
 		}
 
 		routeSvc, ok := s.(mediator.ProtocolService)
 		if !ok {
-			return nil, errors.New("cast service to Route Service failed")
+			return nil, errors.New("didServiceBlockFunc: cast service to Route Service failed")
 		}
 
+		verKey, err := p.KMS().ExportPubKeyBytes(verKeyID)
```
llorllale, line 331: future PR: it'd be really handy to have a `CreateAndExportPubKeyBytes()` function...

Baha-sk: It could be added if there's demand for it in the future.

llorllale:
@Baha-sk the peer DID still has incorrect ids for publicKey.id, publicKey.controller, and authentication:

```json
{
    "@context": [
        "https://w3id.org/did/v1"
    ],  
    "id": "did:peer:1zQmSPZFAQTzXAnN4Ytb6VpKD9bgPBYHpVAJjgnbrUu2vJAw",
    "created": "2020-08-19T07:48:46.635117375-04:00",
    "updated": "2020-08-19T07:48:46.635117375-04:00",
    "authentication": [
        "did:key:z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD#z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD"
    ],  
    "publicKey": [
        {   
            "controller": "did:key:z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD",
            "id": "did:key:z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD#z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD",
            "publicKeyBase58": "CVixSfZin3BSif783NKZC1Tw1AXk8AXRzHjLr7peU2tq",
            "type": "Ed25519VerificationKey2018"
        }   
    ],  
    "service": [
        {   
            "id": "#agent",
            "priority": 0,
            "recipientKeys": [
                "CVixSfZin3BSif783NKZC1Tw1AXk8AXRzHjLr7peU2tq"
            ],  
            "routingKeys": null,
            "serviceEndpoint": "http://example.com/didcomm",
            "type": "did-communication"
        }   
    ]   
} 
```

llorllale:
@Baha-sk we must fix the intermittent signature validation failures... otherwise it's rendered useless and will break downstream integration... what's the plan for this?

Baha-sk:
> @Baha-sk the peer DID still has incorrect ids for publicKey.id, publicKey.controller, and authentication:
> 
> ```json
> {
>     "@context": [
>         "https://w3id.org/did/v1"
>     ],  
>     "id": "did:peer:1zQmSPZFAQTzXAnN4Ytb6VpKD9bgPBYHpVAJjgnbrUu2vJAw",
>     "created": "2020-08-19T07:48:46.635117375-04:00",
>     "updated": "2020-08-19T07:48:46.635117375-04:00",
>     "authentication": [
>         "did:key:z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD#z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD"
>     ],  
>     "publicKey": [
>         {   
>             "controller": "did:key:z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD",
>             "id": "did:key:z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD#z6Mkqwz12upA7afuq9wpiwHQ371vpjobY3mngJeGgPnfPFgD",
>             "publicKeyBase58": "CVixSfZin3BSif783NKZC1Tw1AXk8AXRzHjLr7peU2tq",
>             "type": "Ed25519VerificationKey2018"
>         }   
>     ],  
>     "service": [
>         {   
>             "id": "#agent",
>             "priority": 0,
>             "recipientKeys": [
>                 "CVixSfZin3BSif783NKZC1Tw1AXk8AXRzHjLr7peU2tq"
>             ],  
>             "routingKeys": null,
>             "serviceEndpoint": "http://example.com/didcomm",
>             "type": "did-communication"
>         }   
>     ]   
> } 
> ```

we can't do anything about it until the did:key id representation is updated even in didcomm. I attempted to update them in this PR but it requires more changes which only makes the PR even bigger. It should be done as part of your [did:key issue](https://github.com/hyperledger/aries-framework-go/blob/master/pkg/didcomm/common/service/destination.go#L55)

pkg/didcomm/common/service/destination.go
Baha-sk: comment has been removed

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -486,7 +486,7 @@ func (s *Service) handleCallback(c *callback) (string, error) {
 }
 
 func (s *Service) handleRequestCallback(c *callback) (string, error) {
-	logger.Debugf("input: %+v", c)
+	logger.Debugf("service.handleRequestCallback() input: %+v", c)
```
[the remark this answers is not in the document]
Baha-sk: this change is now reverted

Baha-sk:
> @Baha-sk we must fix the intermittent signature validation failures... otherwise it's rendered useless and will break downstream integration... what's the plan for this?

@llorllale this is now fixed, and reverted PubliceKey.ID to include relative key with `#` prefix.

pkg/client/didexchange/client.go
```diff
@@ -171,17 +179,17 @@ func (c *Client) CreateInvitationWithDID(label, publicDID string) (*Invitation,
...
 	msg, err := service.ParseDIDCommMsgMap(payload)
 	if err != nil {
-		return "", fmt.Errorf("failed to create DIDCommMsg: %w", err)
+		return "", fmt.Errorf("handleInvitation: failed to create DIDCommMsg: %w", err)
 	}
 
 	connectionID, err := c.didexchangeSvc.HandleInbound(msg, "", "")
 	if err != nil {
-		return "", fmt.Errorf("failed from didexchange service handle: %w", err)
+		return "", fmt.Errorf("fhandleInvitation: ailed from didexchange service handle: %w", err)
```
rolsonquadras, line 192: Typo `fhandleInvitation: ailed`.

llorllale:
@Baha-sk there's a buffer overflow warning:

> This pull request **introduces 1 alert** when merging [9f1cc6b](https://github.com/hyperledger/aries-framework-go/commit/9f1cc6b7efa6eb970e4aabb1392c25bd99ae0b88) into [80cb211](https://github.com/hyperledger/aries-framework-go/commit/80cb211c6dd6e5af44b9cfb2204a58979d70c2c0) - [view on LGTM.com](https://lgtm.com/projects/g/hyperledger/aries-framework-go/rev/pr-f04d0b98c50218473f332d554cf55dc19c963b2d)
> 
> **new alerts:**
> 
>     * 1 for Size computation for allocation may overflow

pkg/client/didexchange/client.go
```diff
@@ -115,33 +116,40 @@ func New(ctx provider) (*Client, error) {
...
+		return nil, fmt.Errorf("createInvitation: failed to extract public SigningKey bytes from handle:%w", err)
+	}
+
+	sigPubKeyB58 := base58.Encode(sigPubKey)
+
 	// get the route configs
 	serviceEndpoint, routingKeys, err := mediator.GetRouterConfig(c.routeSvc, c.serviceEndpoint)
 	if err != nil {
-		return nil, fmt.Errorf("create invitation - fetch router config : %w", err)
+		return nil, fmt.Errorf("createInvitation: getRouterConfig: %w", err)
```
rolsonquadras, line 134: Any specific reason to update the error messages ?

pkg/didcomm/common/service/destination.go
```diff
@@ -41,7 +41,7 @@ func GetDestination(did string, vdr vdri.Registry) (*Destination, error) {
 func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
 	didCommService, ok := diddoc.LookupService(didDoc, didCommServiceType)
 	if !ok {
-		return nil, fmt.Errorf("create destination: missing DID doc service")
+		return nil, fmt.Errorf("create destination: create destination: missing DID doc service")
```
rolsonquadras, line 44: `create destination: ` repeated.

pkg/client/didexchange/client.go
Baha-sk: I'll update these..

pkg/client/didexchange/client.go
Baha-sk: `fetch router config ` was repeated by the wrapped error

pkg/didcomm/common/service/destination.go
Baha-sk: I'll remove duplicate..

Baha-sk:
> @Baha-sk there's a buffer overflow warning:
> 
> > This pull request **introduces 1 alert** when merging [9f1cc6b](https://github.com/hyperledger/aries-framework-go/commit/9f1cc6b7efa6eb970e4aabb1392c25bd99ae0b88) into [80cb211](https://github.com/hyperledger/aries-framework-go/commit/80cb211c6dd6e5af44b9cfb2204a58979d70c2c0) - [view on LGTM.com](https://lgtm.com/projects/g/hyperledger/aries-framework-go/rev/pr-f04d0b98c50218473f332d554cf55dc19c963b2d)
> > **new alerts:**
> > ```
> > * 1 for Size computation for allocation may overflow
> > ```

The allocation is for a Copy() from a key `[]byte` into another `[]byte`, a key will never be large enough to cause allocation overflow in an array (unless one day keys reach 1 MB in size). Also this code existed in pkg/vdri/key package and is moved into the new `fingerprint` package in this change.

pkg/client/didexchange/client.go
Baha-sk: done

pkg/didcomm/common/service/destination.go
Baha-sk: done

pkg/didcomm/protocol/mediator/models.go
Baha-sk: They're now removed, we'll deal with full Encryption keys in RecipientKeys (not just Ed25119) in a future PR once didComm/DID Exchange RFCS are updated.

PR #2076, opened 2020-08-04
fix: mysql storage tests are not properly waiting for db readiness

### LINKED ISSUE #2051: mysql storage: tests are not properly waiting for db readiness
https://github.com/hyperledger/aries-framework-go/blob/e4ed522212cb5a4f49eeb06f2c9e882aa389cb60/pkg/storage/mysql/mysqlstore_test.go#L37

The check performed by `checkMySQL()` is insufficient - it needs to also call `db.Ping()` and make sure it succeeds. It should also do an appropriate number of retries.

Afterwards, if `checkMySQL()` fails then the exit code in `TestMain` MUST be something other than `0`.

talwinder50:
closes #2051

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/VictoriaMetrics/fastcache v1.5.7
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/btcsuite/btcutil v1.0.1
+	github.com/cenkalti/backoff/v4 v4.0.2 // indirect
 	github.com/flimzy/diff v0.1.7 // indirect
 	github.com/flimzy/testy v0.1.17 // indirect
 	github.com/go-kivik/couchdb v2.0.0+incompatible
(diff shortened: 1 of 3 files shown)


pkg/storage/mysql/mysqlstore_test.go
```diff
@@ -12,7 +12,9 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/cenkalti/backoff"
```
troyronda, line 17: should be using `github.com/cenkalti/backoff/v4` - this version has go module support.

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/VictoriaMetrics/fastcache v1.5.7
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/btcsuite/btcutil v1.0.1
+	github.com/cenkalti/backoff v2.2.1+incompatible
```
troyronda, line 11: should be using the version that has go module support. see below.

pkg/storage/mysql/mysqlstore_test.go
talwinder50: @troyronda updated

PR #2127, opened 2020-08-19
docs: Issue Credential documentation (godoc)

soluchok:
This PR adds `godoc` for the Issue Credential protocol.

Part of #1790 

### DIFF
--- a/pkg/client/issuecredential/doc.go
+++ b/pkg/client/issuecredential/doc.go
@@ -1,0 +1,80 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package issuecredential provides support for the Issue Credential Protocol 2.0:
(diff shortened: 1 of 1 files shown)


pkg/client/issuecredential/doc.go
```diff
@@ -0,0 +1,80 @@
...
+//          // If Holder is willing to accept the credentials.
+//          client.AcceptCredential(piid, names)
+//          // If Holder is not willing to accept the credentials.
+//          client.DeclineCredential(piid, reason)
+//        }
+//
+//        if event.Message.Type() == presentproof.ProblemReportMsgType {
+//          // Problem report message is triggered to notify client about the error.
+//          // In that case, there is only one option - accept it.
+//          client.AcceptProblemReport(piid)
```
troyronda, line 64: What happens when the problem report is ignored instead of accepted?

soluchok: @troyronda the protocol will not go to the final state. It is applicable to other actions as well.

PR #2132, opened 2020-08-26
feat:  Verifiable - saves participant`s DIDs.

soluchok:
We did not have a possibility to correlate `credentials` or `presentations` that were issued through the protocols with did-connections (participants who were involved in the process of issuing).
This PR fixes it by introducing two new fields in `verifiable.Record` structure. Those fields are not part of credentials/presentations and they do not have any impact on original VC/VP. Fields are being used as `other fields of interest.` according to `Record` structure.

### DIFF
--- a/pkg/didcomm/protocol/middleware/presentproof/middlewares.go
+++ b/pkg/didcomm/protocol/middleware/presentproof/middlewares.go
@@ -19,7 +19,12 @@ import (
 	storeverifiable "github.com/hyperledger/aries-framework-go/pkg/store/verifiable"
 )
 
-const stateNamePresentationReceived = "presentation-received"
+const (
+	stateNamePresentationReceived = "presentation-received"
+	myDIDKey                      = "myDID"
+	theirDIDKey                   = "theirDID"
+	namesKey                      = "names"
+)
 
 // Metadata is an alias to the original Metadata.
 type Metadata presentproof.Metadata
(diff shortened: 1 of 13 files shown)


pkg/didcomm/protocol/middleware/presentproof/middlewares.go
```diff
@@ -56,6 +61,13 @@ func SavePresentation(p Provider) presentproof.Middleware {
 			}
 
 			var names []string
+			var properties = metadata.Properties()
+
+			// nolint: errcheck
+			myDID, _ := properties[myDIDKey].(string)
```
llorllale, line 67: why are we ignoring the error?

test/aries-js-worker/test/issuecredential/issuecredential.js
```diff
@@ -107,7 +109,18 @@ async function issueCredential (mode) {
     })
 
     it("Checks credential", async function () {
-        await getCredential(holder, credentialName)
+        let credential = await getCredential(holder, credentialName)
+        let credentials = await holder.verifiable.getCredentials()
+        for (let i = 0; i < credentials.result.length; i++) {
```
llorllale, line 114: nit: use [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

pkg/didcomm/protocol/middleware/presentproof/middlewares.go
soluchok: It is boolean that says whether we are able to assert value to a string. By default, it will be an empty string. The is no need to handle it.

llorllale: @soluchok correct but, would that not mean there's a bug elsewhere?

soluchok: It can be a bug. I will fix it.

pkg/store/verifiable/store.go
```diff
@@ -30,13 +30,32 @@ const (
...
+
+// WithTheirDID allows specifying TheirDID for credential or presentation that is being issued.
+func WithTheirDID(val string) Opt {
+	return func(o *options) {
+		o.TheirDID = val
+	}
+}
 
 // Store provides interface for storing and managing verifiable credentials.
 type Store interface {
```
llorllale, line 56: future: we should add methods to fetch by DIDs as well

PR #2140, opened 2020-08-27
refactor: revert KID encoding back to base64URL and introduce base58 store wrapper

### LINKED ISSUE #2137: couchdb storage: error saving keys with underscore as a prefix
Example error from the KMS when attempting to save a key set:

```
failed to store keyset: writeToStore: failed to write buffer to store: failed to store data: Bad Request: Only reserved document ids may start with underscore."
```

Baha-sk:
closes #2137

### DIFF
--- a/pkg/storage/base58wrapper/base58wrapper.go
+++ b/pkg/storage/base58wrapper/base58wrapper.go
@@ -1,0 +1,73 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package base58wrapper
+
+import (
+	"github.com/btcsuite/btcutil/base58"
+
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+// package base58wrapper offers provider.Store wrappers for stores with certain constraints on key values such as
+// reserved prefix values (eg key IDs starting with underscore) by wrapping key values in another format supported by
+// the store.
+
+// NewBase58StoreWrapper creates a new Base58StoreWrapper of store.
+func NewBase58StoreWrapper(store storage.Store) storage.Store {
+	return &Base58StoreWrapper{store: store}
+}
+
+// Base58StoreWrapper is a wrapper store that converts key IDs from base64 to base58 encoded values.
+// Keys IDs will be stored in the embedded store base58 encoded while the user of this wrapper store will interact with
+// the wrapper store using key IDs base64 raw (no padding) URL encoded.
+type Base58StoreWrapper struct {
+	store storage.Store
+}
+
+func convert(k string) string {
+	b58k := base58.Encode([]byte(k))
+
+	return b58k
+}
+
+// Put stores the key and the record by converting the key from base64 to base58 encoded value first.
+func (b *Base58StoreWrapper) Put(k string, v []byte) error {
+	b58k := convert(k)
+
+	return b.store.Put(b58k, v)
+}
+
+// Get fetches the record based on key by converting the key ID from base58 to base64 encoded value first.
+func (b *Base58StoreWrapper) Get(k string) ([]byte, error) {
+	b58k := convert(k)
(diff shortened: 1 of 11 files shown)


pkg/storage/base58wrapper/base58wrapper.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Base64ToBase58StoreWrapper is a wrapper store that converts key IDs from base64 to base58 encoded values.
+// Keys IDs will be stored in the embedded store base58 encoded while the user of this wrapper store will interact with
+// the wrapper store using key IDs base64 raw (no padding) URL encoded.
+type Base64ToBase58StoreWrapper struct {
+	store storage.Store
+}
+
+func convert(k string) (string, error) {
+	kidBytes, err := base64.RawURLEncoding.DecodeString(k)
```
sandrask, line 34: Why not convert base64 string to base58 string right away in order to create new id? I am not sure why you are decoding it first here.

pkg/storage/base58wrapper/base58wrapper.go
```diff
@@ -0,0 +1,144 @@
...
+//
+// Args:
+//
+// startKey: Start of the key range, include in the range by converting it from base64 to base58 encoded value first.
+// endKey: End of the key range, not include in the range by converting it from base64 to base58 encoded value first.
+//
+// Returns:
+//
+// StoreIterator: a wrapped iterator for result range.
+func (b *Base64ToBase58StoreWrapper) Iterator(startKey, endKey string) storage.StoreIterator {
```
sandrask, line 74: If iterator is not working yet you should return not implemented error. Applies to all functions that are not working.

pkg/storage/base58wrapper/base58wrapper.go
sandrask: Also this is not going to work unless string passed in is base64 encoded

Baha-sk: I'll update to just base58 encode k directly.

This works because k received is always base64 encoded. The client gets a base64 encoded key IDs always.

pkg/storage/base58wrapper/base58wrapper.go
Baha-sk: I will return nil and add a comment here, it's too bad this interface function doesn't return error.

pkg/storage/base58wrapper/base58wrapper.go
Baha-sk: done

pkg/storage/base58wrapper/base58wrapper.go
Baha-sk: done, and removed  the wrapperIterator.

pkg/storage/base58wrapper/base58wrapper.go
```diff
@@ -0,0 +1,73 @@
...
+
+// NewBase58StoreWrapper creates a new Base64ToBase58StoreWrapper of store.
+func NewBase58StoreWrapper(store storage.Store) storage.Store {
+	return &Base64ToBase58StoreWrapper{store: store}
+}
+
+// Base64ToBase58StoreWrapper is a wrapper store that converts key IDs from base64 to base58 encoded values.
+// Keys IDs will be stored in the embedded store base58 encoded while the user of this wrapper store will interact with
+// the wrapper store using key IDs base64 raw (no padding) URL encoded.
+type Base64ToBase58StoreWrapper struct {
```
sandrask, line 26: Should not mention Base64 in the name, this wrapper takes any string and converts it to base58

Baha-sk: I'll push an update, good catch! Since I don't use base64 decoding anymore, there's no need to mention Base64 in the name.

Baha-sk: renamed

PR #2159, opened 2020-09-03
DID Doc: Fix Marshaling

### LINKED ISSUE #2149: did.Doc should implement json.Marshaler
Users typically expect `json.Marshal(didDoc)` to produce a correctly formatted did document but that is not the case. Users are not typically aware of `did.Doc.JSONBytes()`. 

We should implement `MarshalJSON()` to avoid surprises.

gnarula:
Implemented json.Marshaler and json.Unmarshaler. Also changed
DIDCommMsgMap to use did.ParseDocument.

Closes #2149

### DIFF
--- a/pkg/didcomm/common/service/did_comm_msg.go
+++ b/pkg/didcomm/common/service/did_comm_msg.go
@@ -207,7 +207,7 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
 			if rt2 == reflect.TypeOf(did.Doc{}) {
 				didDoc, err := json.Marshal(v)
 				if err != nil {
-					return nil, err
+					return nil, fmt.Errorf("error remarshaling to json: %w", err)
 				}
 				return did.ParseDocument(didDoc)
 			}
--- a/pkg/didcomm/common/service/did_comm_msg.go
+++ b/pkg/didcomm/common/service/did_comm_msg.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/mitchellh/mapstructure"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -203,6 +204,14 @@ func (m DIDCommMsgMap) Decode(v interface{}) error {
 				return base64.StdEncoding.DecodeString(v.(string))
 			}
 
+			if rt2 == reflect.TypeOf(did.Doc{}) {
+				didDoc, err := json.Marshal(v)
+				if err != nil {
+					return nil, err
```
llorllale, line 210: please wrap this error

llorllale:
@gnarula seems like you've uncovered bugs in the `did:peer` tests

pfeairheller:
I hit this problem trying to parse messages from other frameworks.  What is needed to help move this along?  (yes, I'm volunteering).

llorllale:
@gnarula @pfeairheller apologies for the long delay. I haven't had time to look too much into this, but, after glancing at the tests, my impression is the immediate problem is the setup: `TestComputeDID` expects a method-specific-ID computed from an peer DID doc that was marshalled incorrectly, presumably using `json.Marshal()`. This PR exposed that bug.

A quick way to unblock this PR would be to:

* fix the linter errors
* update the ID in did_test.go#L24 such that it matches the doc under test (L101-126).
* squash comits

Beyond that, in a followup, we should revisit this implementation and ensure alignment with the `did:peer` spec (we should coordinate with the Aries folks) and make sure the tests are being done right.

llorllale:
@pfeairheller what's the source for those `did:peer` DIDs you're trying to parse?

pfeairheller:
@llorllale interop testing with the IBM Verify Creds platform.  From looking at their logs, they are using the nodejs wrapper on indy-sdk.  And it was parsing a DIDDoc embedded inside a `connection` attribute in `didexchange` that errored.

llorllale:
Hi @gnarula 

@pfeairheller was kind enough to submit PR #2340 which as merged.

Closing this PR.

PR #2168, opened 2020-09-10
fix: Fixes unit test failure due to conflict error

soluchok:
Sometimes unit tests are failing due to `conflict` error.
This PR fixes it by generating new keys instead of using constant ones.

### DIFF
--- a/pkg/storage/base58wrapper/base58wrapper_test.go
+++ b/pkg/storage/base58wrapper/base58wrapper_test.go
@@ -16,6 +16,7 @@ import (
 	"testing"
 
 	"github.com/go-kivik/kivik"
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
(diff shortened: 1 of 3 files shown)


aholovko:
Why this `conflict` error happens in the first place? Are we hiding some other issue by switching to the random names?

fqutishat (approved):

soluchok:
> Why this `conflict` error happens in the first place? Are we hiding some other issue by switching to the random names?

Tests usually are copy-pasted (especially for storage). `OpenStore` uses the same names. Because we have one instance of DB (e.g CouchDB) it is easy to get an error like `conflict`.

llorllale (approved):

aholovko:
> > Why this `conflict` error happens in the first place? Are we hiding some other issue by switching to the random names?
> 
> Tests usually are copy-pasted (especially for storage). `OpenStore` uses the same names. Because we have one instance of DB (e.g CouchDB) it is easy to get an error like `conflict`.

Ok, makes sense. I wouldn't try to fix that by introducing random names but if everyone is Okay with that I can live with it as well. :-)

PR #2169, opened 2020-09-10
fix: Fixes unauthorized couchDB

soluchok:
Sometimes BDD tests are failing due to `dial tcp [::1]:11012: connect: connection refused`. See https://github.com/hyperledger/aries-framework-go/runs/1096805345#step:5:2600
The root cause seems to be `failed to start aries agent rest on port [0.0.0.0:11012], failed to initialize framework :  default option initialization failed: can't initialize verifaible store : failed to open vc store: failed to create db: Unauthorized: You are not a server admin.`
This PR tries to fix it by adding retry logic.

### DIFF
--- a/pkg/storage/couchdb/couchdbstore.go
+++ b/pkg/storage/couchdb/couchdbstore.go
@@ -17,12 +17,16 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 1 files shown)


llorllale (changes requested):
Doesn't this check work? https://github.com/hyperledger/aries-framework-go/blob/d012e4885115e87a3fb3ade8b4c7552260dce95d/pkg/storage/couchdb/couchdbstore.go#L67

PR #2175, opened 2020-09-11
fix: Fixes mem store

soluchok:
The current implementation of `mem` storage has a bug in the `Iterator` method.
This bug can be reproduced only when storage is populated with the data.
The issue comes out when we are iterating over keys and the start key doesn't match.
In that case, the start index is equal to `0`. Which means that the iterator will return data starting from 0 index. e.g data[0:end_idx_match].

This PR fixes it by introducing zero index as a -1.
PR includes:
* Bugfix 
* Common test case (for all DBs)
* Test case for `mem` implementation.

### DIFF
--- a/pkg/storage/mem/mem_store.go
+++ b/pkg/storage/mem/mem_store.go
@@ -146,7 +146,7 @@ func (s *memStore) Iterator(start, limit string) storage.StoreIterator {
 	sort.Strings(keys)
 
 	var (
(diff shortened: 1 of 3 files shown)


llorllale:
@soluchok please make the commit's message more descriptive

Baha-sk (approved):

llorllale (approved):

Issue #2170, opened 2020-09-10
mobile: create notification handler interface


ToJen:
**What I'm trying to do**

Create an interface that will handle the notification of events. This interface will be implemented by objects within mobile apps to take advantage of [being able to pass target language objects to Go](https://godoc.org/golang.org/x/mobile/cmd/gobind#hdr-Passing_target_language_objects_to_Go).

Update 

**Expected result**

The changes should allow mobile apps to

- Subscribe to a topic.
- Create implementations of handlers that can be passed into the configuration of an Aries mobile agent.

**Sample code**

```go
// Golang
type Handler interface {
    Handle(topic string, payload []byte) error
}
```

gomobile will generate Java/Objective-C code that will contain an interface named Handler.
Mobile apps can create an implementation of it like so:

``` java
// Java
public class MyHandler extends AriesApi.Handler {
    public void handle(String topic, byte[] message) {
        // do something here
    }
}
```

```objc
// Objective-C
@interface MyHandler: NSObject<ApiHandler>{

}
@end

@implementation MyHandler

- (BOOL) handle: (NSString *)topic message:(NSData *)message
          error:(NSError * _Nullable __autoreleasing *)error {
        // do something here
}

@end
```

And then it'll pass an instance of `MyHandler` to Go by calling a new function called `AriesAgent.Registerhandler`.
```go
// rest implementation will use WebSockets
// command implementation will use a Notifier (github.com/hyperledger/aries-framework-go/pkg/controller/webnotifier)
func (a *Aries) RegisterHandler(h Handler, topic string, payload []byte) string {
    /* ... */
}
```
```java
// Java
            // register handler
            Handler handler = new MyHandler();
            String registrationID = ariesAgent.registerHandler(handler, "didexchange_states");
```
```objc
// Objective-C

    // register handler
    MyHandler *handler = [[MyHandler alloc] init];
    NSString *regID = [ariesAgent registerHandler:handler topics:@"didexchange_states"];
```

**Questions and Concerns**

- [x] What about remote agents i.e. the agents running on external servers that mobile apps can connect to using our SDK
~~- [ ] Lifecycle: TODO (when exactly does the notifier get used internally?)~~
- [x] Unregistering notifiers:
  - [X] By who (the mobile app or the SDK)? -> mobile app

PR #2177, opened 2020-09-11
feat: added the interface for notifiers on mobile devices

ToJen:

**Title:**
Initial interface for mobile notifiers

**Description:**
Part of #2170


cmd/aries-agent-mobile/pkg/api/notifier.go
```diff
@@ -0,0 +1,17 @@
...
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+package api
+
+// Notifier represents a notification dispatcher.
+type Notifier interface {
+
+	// Notify sends the given message to all of the subscribers.
+	Notify(topic string, message []byte) error
```
soluchok, line 13: remove this method `Notify `.
Only `Execute(topic string, message []byte)` method should be here.
You can rename `Execute` to `Notify ` if you want.

cmd/aries-agent-mobile/pkg/api/api.go
```diff
@@ -38,4 +38,7 @@ type AriesController interface {
 
 	// GetKMSController returns an implementation of KMSController
 	GetKMSController() (KMSController, error)
+
+	// RegisterNotifier associates a notifier to a topic
+	RegisterNotifier(n Notifier, topic string, payload []byte) error
```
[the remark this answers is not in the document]
ToJen: I'll remove `payload`.

For topic, type `[]string` won't get parsed by gomobile. The easiest approach will be for the mobile app to send the topics as a JSON array in a `string`.

An alternative is to create a method called `AppendTopic(topic string)` on the `Notifier` interface which will require implementations to have a string array instance variable. To then access this variable, they would also have to implement a `GetTopics() string` method which still has to return a JSON array as a `string` because of gomobile.

cmd/aries-agent-mobile/pkg/api/notifier.go
ToJen: ok, will do. thought so too

cmd/aries-agent-mobile/pkg/api/api.go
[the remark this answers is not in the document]
soluchok: how about that topics should be comma-separated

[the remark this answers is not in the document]
ToJen: ok that's better. i'll add a comment for this

cmd/aries-agent-mobile/pkg/wrappers/config/options.go
```diff
@@ -23,6 +25,8 @@ type Options struct {
 	// not intended to be used by golang code
 	HTTPResolvers     []string
 	OutboundTransport []string
+
+	Notifier api.Notifier
```
soluchok, line 29: the interface should be `command.Notifier`

soluchok: and we need one more implementation. It will be used for SDK (for REST we will use web sockets )

cmd/aries-agent-mobile/pkg/api/notifier.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// GetPayload returns the payload message.
+	GetPayload() []byte
```
soluchok, line 16: we don’t  need this function

soluchok: I just realized that the best name is `Handle`. Please replace Notify with `Handle`.
The `Handle` function plays a callback function role.

soluchok: And `Notifier` will be `Handler`.

soluchok: Make sense?

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
```diff
@@ -38,6 +38,7 @@ import (
 type Aries struct {
 	framework *aries.Aries
 	handlers  map[string]map[string]command.Exec
+	notifiers map[string]api.Notifier
```
soluchok, line 41: we do not need it for now

soluchok:
@ToJen I think we can close it.

ToJen:
closing this to use #2184 instead

PR #2181, opened 2020-09-12
Update golangci lint to 1.31.0

### LINKED ISSUE #2032: Update linter to the latest version (golangci)
https://github.com/golangci/golangci-lint/releases

ken5scal:
**Title:**
Update golangci lint to 1.31.0 from 1.25.1

**Description:**
resolve #2032 

**Summary:**
basically, all lint errors are fixed, except the followings:
- goerr13: It seems to establish an error handling strategy that needs to be done before actually fixing the error. Probably, separating an issue is a good idea.
- noctx: [introduced from v1.28.0](https://github.com/golangci/golangci-lint/commit/f1a46f9861e8208460e45f2b92d2de32bbf8d2ca). just like `goerr113`, the HTTP client strategy should be first established before injecting context to HTTP clients.
- godot: it was already set us ["check-all: false"](https://github.com/hyperledger/aries-framework-go/blob/master/.golangci.yml#L83) . But [according to the issue](https://github.com/golangci/golangci-lint/issues/1188), godot itself changed behavior so that `check-all: false` no longer serves its purpose
- nlreturn: [another linter introduced in v1.31.0](https://github.com/golangci/golangci-lint/commit/6b60cb8a30078fb0b55e7812b45487de288d30cc), the number of no nlreturn was too much to fix.


.golangci.yml
```diff
@@ -79,15 +79,18 @@ linters-settings:
     allow-multiline-assign: true
     allow-case-traling-whitespace: true
     allow-cuddle-declarations: false
-  godot:
-    check-all: false
```
soluchok, line 83: you need to rollback it

.golangci.yml
```diff
@@ -79,15 +79,18 @@ linters-settings:
...
 
 linters:
   enable-all: true
   disable:
     - testpackage
     - maligned
     - prealloc
+    - godot
```
soluchok, line 89: godot should be enabled

ken5scal: https://github.com/hyperledger/aries-framework-go/pull/2181/commits/554a9286a8a0319cecbdf719569e74b09c14d383 

fixed in here

fqutishat (approved):

fqutishat:
@ken5scal can you squash 40 commits ?

ken5scal:
oh, my bad; but thanks for squashing for me

PR #2182, opened 2020-09-12
fix: unit tests depending on docker

llorllale:
Depends on #2180

Unit tests for storages were all ignoring failures (such as CouchDB or MySQL not running). This completely undermined confidence in any of these tests... a real mess.

This patch only ignores these tests when running in GitHub actions using the 'macOS-10.15' virtual environment. Issue #2183 was opened to track the problem of not running these tests in the macOS CI in the first place.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,10 @@ jobs:
 
     - uses: actions/checkout@v2
 
+    - name: Skip tests using docker on macOS (issue-2183)
+      if: matrix.os == 'macOS-10.15'
+      run: echo '::set-env name=SKIP_DOCKER::true'
+
     - name: Run unit test
       timeout-minutes: 15
       run: make unit-test
--- a/pkg/storage/base58wrapper/base58wrapper_test.go
+++ b/pkg/storage/base58wrapper/base58wrapper_test.go
@@ -1,4 +1,4 @@
-// +build !js,!wasm
+// +build !js,!wasm,!ISSUE2183
 
 /*
 Copyright SecureKey Technologies Inc. All Rights Reserved.
@@ -9,13 +9,11 @@ SPDX-License-Identifier: Apache-2.0
 package base58wrapper
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"os"
 	"testing"
 
-	"github.com/go-kivik/kivik"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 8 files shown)


soluchok:
@llorllale I was also thinking about it. But I would suggest using `github.com/ory/dockertest/v3` package. In that case, we don’t need to play with `docker`. It works fine on the local machine (Linux/mac). But it will not work on CI for mac. Since mac (on CI) doesn’t have docker. So, my suggestion is to remove unit tests for mac on CI. What do you think about it?
It will allow us to run the test just hitting the `run` button in `idea`.

@troyronda

llorllale:
@soluchok 

> @llorllale I was also thinking about it. But I would suggest using `github.com/ory/dockertest/v3` package. In that case, we don’t need to play with `docker`. It works fine on the local machine (Linux/mac). But it will not work on CI for mac. Since mac (on CI) doesn’t have docker. So, my suggestion is to remove unit tests for mac on CI. What do you think about it?
> It will allow us to run the test just hitting the `run` button in `idea`. No need to run docker images manually.

* Still depends on docker so it won't work on macOS in the CI (as you said)
* We should _not_ remove unit tests for mac in the CI. Why would we do this? I'm wondering why we don't include windows in there as well...
* The current build script runs each docker container once per run. I think your intention is to have each test package run its own container? Won't that make the tests run longer? I like the idea of compartmentalizing each test package, but also appreciate fast builds... and our build is _dreadfully slow_ ...

soluchok:
@llorllale The current build script runs each docker container once per run. - it runs one docker container per test package. It will not strongly impact on execution time.

llorllale:
@soluchok 

> @llorllale The current build script runs each docker container once per run. - it runs one docker container per test package. It will not strongly impact on execution time.

I believe you meant to say "it will run one docker container per test package". Because currently only one container is ran per test run.

llorllale:
@soluchok let me do some tests to compare the timing diifference

soluchok (approved):

soluchok:
> @soluchok let me do some tests to compare the timing diifference

Sure. Since we are ignoring decker tests for mac on CI I think it makes sense to use`dockertest ` package. I am not saying to use it in this PR. We can create an issue. But, It would simplify our life :)

llorllale:
@soluchok 

>  I am not saying to use it in this PR. We can create an issue.

I tend to agree... I'm wondering why each storage implementation is duplicating tests from the parent package... is it for test coverage?

soluchok:
> @soluchok
> 
> > I am not saying to use it in this PR. We can create an issue.
> 
> I tend to agree... I'm wondering why each storage implementation is duplicating tests from the parent package... is it for test coverage?

Yes, it is for coverage.

llorllale:
@soluchok we should explore the `dockertest` option in a separate issue.

FYI: I did a quick test and ran into an IDE quirk... basically my IDE's test report feature doesn't work for some reason:

image

The tests do pass on the command line.

fqutishat (approved):

pkg/storage/leveldb/leveldb_store_test.go
```diff
@@ -1,4 +1,4 @@
-// +build !js,!wasm
+// +build !js,!wasm,!ISSUEXXX
```
troyronda, line 1: whats ISSUEXXX mean?

llorllale: just noticed I missed it - updated

pkg/storage/providers_wasm_test.go
```diff
@@ -1,4 +1,4 @@
-// +build js,wasm
+// +build js,wasm,!ISSUEXXX
```
troyronda, line 1: ISSUEXXX?

llorllale:
@troyronda @fqutishat added the following as required checks:

* Unit test (macOS-10.15) 
* Unit test wasm

PR #2184, opened 2020-09-14
feat: Mobile notifications support

### LINKED ISSUE #2170: mobile: create notification handler interface

soluchok:
Adds mobile notifications support.
* works on iOS
* works on Android (haven’t checked yet)

The main idea is to use the `Handler` interface with the `Handle` method which plays the role of a callback function.
Usage: The client needs to implement the `Handler` interface. Notifications will be sent through this callback function (`Handle`).

Closes #2170


cmd/aries-agent-mobile/pkg/api/notifier.go
```diff
@@ -0,0 +1,34 @@
+package api
```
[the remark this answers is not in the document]
ToJen: and we should also add a comment at the top that the file is not expected to be used by the mobile app and so gomobile will not compile all the code in it

ToJen (changes requested):

ToJen (approved):
looks good, we just have this extra file: test/bdd/fixtures/agent-rest/couchdb-config/10-single-node.ini

soluchok:
> looks good, we just have this extra file: test/bdd/fixtures/agent-rest/couchdb-config/10-single-node.ini

It was done by make command. I didn't want to create a separate PR for it. So, I just included it in this PR.

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
```diff
@@ -149,6 +171,56 @@ func populateHandlers(commands []command.Handler, pkgMap map[string]map[string]c
...
+				}
+			}
+		}
+		a.mutex.RUnlock()
+	}
+}
+
+// RegisterHandler registers a handler to process incoming notifications from the framework.
+// Handler is implemented by mobile apps.
+func (a *Aries) RegisterHandler(h api.Handler, topics string) string {
```
llorllale, line 193: `topics` can be a var-arg

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
```diff
@@ -42,7 +54,121 @@ func NewAries(opts *config.Options) (*Aries, error) {
...
+func (ar *Aries) startNotificationListener() { // nolint: gocyclo,funlen
+	if ar.WebsocketURL == "" {
+		return
+	}
+
+	conn, _, err := websocket.Dial(context.Background(), ar.WebsocketURL, &websocket.DialOptions{ //nolint:bodyclose
+		HTTPClient: &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true, // nolint: gosec
```
llorllale, line 86: why is this enabled?

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
```diff
@@ -42,7 +54,121 @@ func NewAries(opts *config.Options) (*Aries, error) {
...
+
+			// exit if websocket is closed
+			if websocket.CloseStatus(err) != -1 {
+				return
+			}
+
+			continue
+		}
+
+		ar.mutex.RLock()
```
llorllale, line 120: should defer `RUnlock` here

llorllale:
@soluchok 

> > looks good, we just have this extra file: test/bdd/fixtures/agent-rest/couchdb-config/10-single-node.ini
> 
> It was done by make command. I didn't want to create a separate PR for it. So, I just included it in this PR.

Please remove this file from this PR. PR #2180 had a bug that was later fixed in #2182.

llorllale (changes requested):
Please remove couchdb config file.

Also see my other comments.

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
soluchok: it is a loop. `defer ` will not work

soluchok:
> Please remove couchdb config file.
> 
> Also see my other comments.

@llorllale Removed `couchdb` config file.

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
soluchok: Fixed!

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
soluchok: @llorllale As far as I remember we cannot use variadic arguments.  @ToJen Could you please confirm this?
e.g RegisterHandler(h api.Handler, topics ...string) is not possible to use, @ToJen correct?.

cmd/aries-agent-mobile/pkg/wrappers/rest/aries.go
llorllale: `defer ar.mutex.RUnlock()` will not work because what follows is a loop?

soluchok: it fixed

llorllale (approved):

llorllale:
@soluchok 

> > looks good, we just have this extra file: test/bdd/fixtures/agent-rest/couchdb-config/10-single-node.ini
> 
> It was done by make command. I didn't want to create a separate PR for it. So, I just included it in this PR.

You're right about this file - the couchdb image is setting its access mode to read-only. You can include it in this PR.

cmd/aries-agent-mobile/pkg/wrappers/command/aries.go
ToJen: yes we can only use simple args like `string`

rolsonquadras (approved):
It would be good if we can document the usage of notifiers in [REAME](https://github.com/hyperledger/aries-framework-go/tree/master/cmd/aries-agent-mobile#3-usage) file

ToJen:
@rolsonquadras i will do it in a separate commit

PR #2190, opened 2020-09-17
test: Fixes local KMS writer test

soluchok:
Sometimes the test is failing randomly.
```
    --- FAIL: TestLocalKMSWriter/error_case_-_import_duplicate_keysetID (0.00s)
        localkms_writer_test.go:105: 
            	Error Trace:	localkms_writer_test.go:105
            	Error:      	Should not be: 0x5f
            	Test:       	TestLocalKMSWriter/error_case_-_import_duplicate_keysetID
```

According to the code, we are generating bytes randomly e.g `base64.RawURLEncoding.EncodeToString(random.GetRandomBytes(uint32(keySetIDLength)))`
Then we are checking that the first byte is not equal to `uint8('_')`.
Since `_` (underscore) is a character that can be used for base64 there is no guarantee that `_` will not be the first character.
This is the reason why the test is failing randomly.

Conclusion:
The test is redundant.

### DIFF
--- a/pkg/kms/localkms/localkms_writer_test.go
+++ b/pkg/kms/localkms/localkms_writer_test.go
@@ -101,8 +101,6 @@ func TestLocalKMSWriter(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, len(someKey), n)
 		require.Equal(t, maxKeyIDLen, len(l.KeysetID))
-		// keysetID must not start with _
-		require.NotEqual(t, uint8('_'), l.KeysetID[0])
 		retrievedKey, ok := storeMap[l.KeysetID]
 		require.True(t, ok)
 		require.Equal(t, retrievedKey, someKey)


llorllale:
I am unhappy with the approach taken by @Baha-sk that makes the KMS implementation "storage-aware".

I believe the reason why that test was put there in the first place is precisely to accommodate quirks specific to CouchDB storage.

Now we are applying patches in several places instead of fixing the CouchDB storage implementation.

soluchok:
> I am unhappy with the approach taken by @Baha-sk that makes the KMS implementation "storage-aware".
> 
> I believe the reason why that test was put there in the first place is precisely to accommodate quirks specific to CouchDB storage.
> 
> Now we are applying patches in several places instead of fixing the CouchDB storage implementation.

I have fixed CouchDB. `CouchDB` uses a payload field under the hood.

soluchok:
> I am unhappy with the approach taken by @Baha-sk that makes the KMS implementation "storage-aware".
> 
> I believe the reason why that test was put there in the first place is precisely to accommodate quirks specific to CouchDB storage.
> 
> Now we are applying patches in several places instead of fixing the CouchDB storage implementation.

Yeah, I just checked it `Only reserved document ids may start with underscore.` Anyway, this fix that I did prevents test failure. I will create an issue that describes the problem.

Baha-sk (approved):

Baha-sk:
> I am unhappy with the approach taken by @Baha-sk that makes the KMS implementation "storage-aware".
> 
> I believe the reason why that test was put there in the first place is precisely to accommodate quirks specific to CouchDB storage.
> 
> Now we are applying patches in several places instead of fixing the CouchDB storage implementation.

the kms is not implementation away @llorllale, the kms knows that its keys might have '_' prefix and so uses a store wrapper to prevent this prefix.

fqutishat (approved):

llorllale:
@Baha-sk 

> > I am unhappy with the approach taken by @Baha-sk that makes the KMS implementation "storage-aware".
> > I believe the reason why that test was put there in the first place is precisely to accommodate quirks specific to CouchDB storage.
> > Now we are applying patches in several places instead of fixing the CouchDB storage implementation.
> 
> the kms is not implementation aware @llorllale, the kms knows that its keys might have '_' prefix and so uses a store wrapper to prevent this prefix.

You are missing the point. 

> The KMS knows that its keys might have '_' prefix and so uses a store wrapper to prevent this prefix

... because it is tip-toeing around a quirk of the CouchDB `Store` implementation, no other reason.

PR #2194, opened 2020-09-18
fix: mysql storage provider: No database selected

### LINKED ISSUE #2192: mysql storage provider: `No database selected`
Lesson learned from a downstream `Store` implementation: MySQL's `USE` statement is only good on the connection it was executed. This connection may be recycled over time (`sql.DB` manages a pool of connections), or the query may run on a different connection from the pool.

Reference: https://github.com/trustbloc/edge-core/issues/71

talwinder50:
closes #2192

### DIFF
--- a/pkg/storage/mysql/mysqlstore.go
+++ b/pkg/storage/mysql/mysqlstore.go
@@ -30,7 +30,6 @@ type Provider struct {
 type sqlDBStore struct {
 	db        *sql.DB
 	tableName string
-	dbName    string
 }
 
 type result struct {
@@ -39,12 +38,7 @@ type result struct {
 }
 
 const (
-	blankDBPathErrMsg         = "DB URL for new mySQL DB provider can't be blank"
-	failToCloseProviderErrMsg = "failed to close provider"
-	tablePrefix               = "t_"
-	sqlDBNotFound             = "no rows"
-	createDBQuery             = "CREATE DATABASE IF NOT EXISTS `%s`"
-	useDBQuery                = "USE `%s`"
+	createDBQuery = "CREATE DATABASE IF NOT EXISTS `%s`"
 )
 
 // Option configures the couchdb provider.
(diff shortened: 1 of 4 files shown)


pkg/storage/store_test.go
```diff
@@ -229,10 +228,3 @@ func verifyItr(t *testing.T, itr storage.StoreIterator, count int, prefix string
 	require.Empty(t, itr.Value())
 	require.Error(t, itr.Error())
 }
-
```
talwinder50, line 232: failure while creating table db_prefix_key631219b5-a59f-44b3-94f0-52b52e21c6a6: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-a59f-44b3-94f0-52b52e21c6a6.db_prefix_key631219b5-a59f-44b3-94f0-52b52e21c6a6 . This uuid with new change of database.tableName it keeps failing with sql instance.

llorllale: Quoting the table name with the backtick (`) should do the trick.

[Permitted characters in MySQL object identifiers](https://dev.mysql.com/doc/refman/8.0/en/identifiers.html):

> * Permitted characters in unquoted identifiers:
>    * ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)
>    * Extended: U+0080 .. U+FFFF 
> 
> * Permitted characters in quoted identifiers include the full Unicode Basic Multilingual Plane (BMP), except U+0000:
>   * ASCII: U+0001 .. U+007F
>   * Extended: U+0080 .. U+FFFF 
> * ASCII NUL (U+0000) and supplementary characters (U+10000 and higher) are not permitted in quoted or unquoted identifiers.
> * Identifiers may begin with a digit but unless quoted may not consist solely of digits.
> * Database, table, and column names cannot end with space characters. 
> 
> The identifier quote character is the backtick (`):

llorllale: Please leave the randomized table names in the tests... it allows fast test re-runs without incurring the risk of test failures due to naming conflicts.

pkg/storage/mysql/mysqlstore.go
```diff
@@ -192,18 +192,12 @@ func (s *sqlDBStore) Put(k string, v []byte) error {
...
 
-	// Use query is used to select the created database without this DDL operations are not permitted
-	_, err := s.db.Exec(fmt.Sprintf(useDBQuery, s.dbName))
-	if err != nil {
-		return fmt.Errorf("failed to use db %s: %w", s.dbName, err)
-	}
-
 	// create upsert query to insert the record, checking whether the key is already mapped to a value in the store.
-	createStmt := "INSERT INTO `" + s.tableName + "` VALUES (?, ?) ON DUPLICATE KEY UPDATE value=?"
+	insertStmt := "INSERT INTO " + s.tableName + " VALUES (?, ?) ON DUPLICATE KEY UPDATE value=?"
```
soluchok, line 196: Why did you remove ``` character?

talwinder50: failure while executing insert statement on table prefixdb_test.prefixdb_test: Error 1046: No database selected.
if i dont remove `character, dot doesnt get picked up so thats why i removed. but now hyphen in the uuid key string doesnt work without `

pkg/storage/store_test.go
talwinder50: updated

pkg/storage/mysql/mysqlstore.go
llorllale: @soluchok backticks are already there in `s.tableName`: https://github.com/hyperledger/aries-framework-go/pull/2194/files#diff-5c03fdd2a42c9a160e4a9b056cb9db20R131

llorllale (approved):

soluchok (approved):

PR #2202, opened 2020-09-22
feat: Multiple routers support

### LINKED ISSUE #1076: Register agent with multiple routers

soluchok:
Adds multiple routers support.
The API was changed:
* `Unregister() error` -> `Unregister(connID string) error`
* `GetConnection() (string, error)` -> `GetConnections() ([]string, error)`
* `Config() (*mediator.Config, error)` -> `Config(connID string) (*mediator.Config, error)`

Closes #1076 


pkg/didcomm/protocol/mediator/service.go
```diff
@@ -732,26 +746,34 @@ func (s *Service) setKeyUpdateResponseCh(msgID string, keyUpdateCh chan *Keylist
 	}
 }
 
-func (s *Service) getRouterConnectionID() (string, error) {
-	id, err := s.routeStore.Get(routeConnIDDataKey)
+func (s *Service) connectionExists(connID string) (bool, error) {
```
llorllale, line 749: This should be renamed to `ensureConnectionExists()` and should just return an error in case it doesn't (`ErrRouterNotRegistered`) or any generic IO error that may arise.

pkg/client/outofband/client.go
```diff
@@ -331,25 +335,42 @@ func didServiceBlockFunc(p Provider) func() (*did.Service, error) {
...
+		var services []*did.Service
+
+		for i := range connections {
+			// get the route configs
+			serviceEndpoint, routingKeys, err := mediator.GetRouterConfig(routeSvc, connections[i], p.ServiceEndpoint())
+			if err != nil {
+				return nil, fmt.Errorf("didServiceBlockFunc: create invitation - fetch router config : %w", err)
+			}
+
+			if err = mediator.AddKeyToRouter(routeSvc, connections[i], base58.Encode(verKey)); err != nil {
```
llorllale, line 361: This strategy is hard-coded and is registering the one key created for this OOB invitation/request on all registered routers. This must be discussed. I don't believe this is the correct approach.

A better approach in my mind would be for us to inject the router registration algorithm as a strategy.

llorllale (changes requested):
See my comment on the registration of the key to the router in the OOB client.

pkg/client/outofband/client.go
soluchok: I see that current logic leads us to `chooseTarget` function. It means that we always pick up the first `service`.
Which makes current logic (`loop`) redundant. 
As I understand we can only use one service for `invitation`. 
I will rollback this change. Also, I will provide a possibility to inject `didDocSvcFunc` function. 
Which as I understand will play `router registration algorithm` role.
@llorllale Does it make sense?

ToJen (commented):
mobile section looks alright since it's just renaming things.

pkg/client/outofband/client.go
```diff
@@ -331,25 +331,36 @@ func didServiceBlockFunc(p Provider) func() (*did.Service, error) {
...
+				ID:              uuid.New().String(),
+				Type:            "did-communication",
+				RecipientKeys:   []string{verKeyB58},
+				ServiceEndpoint: p.ServiceEndpoint(),
+			}, nil
+		}
+
 		// get the route configs
-		serviceEndpoint, routingKeys, err := mediator.GetRouterConfig(routeSvc, p.ServiceEndpoint())
+		serviceEndpoint, routingKeys, err := mediator.GetRouterConfig(routeSvc, connections[0], p.ServiceEndpoint())
```
rolsonquadras, line 349: Same comment as DIDExchange client. ie, support for the consumers to specify the routers (could be >1).

test/bdd/features/aries_mediator_e2e_controller.feature
```diff
@@ -45,12 +45,12 @@ Feature: DIDComm Transport between two Agents through DIDComm Routers [REST Bind
...
 
        # Carl registers her Router
-    And   "Carl" unregisters the router
+    And   "Carl" unregisters the router with connection "carl-router-connID"
     And   "Carl" sets connection "carl-router-connID" as the router
     And   "Carl" verifies that the router connection is set to "carl-router-connID"
 
        # Dave registers his Router
-    And   "Dave" unregisters the router
+    And   "Dave" unregisters the router with connection "dave-router-connID"
```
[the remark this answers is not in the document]
soluchok: I will add BDD tests in the next PR

pkg/client/didexchange/client.go
```diff
@@ -123,10 +123,26 @@ func (c *Client) CreateInvitation(label string) (*Invitation, error) {
...
+	}
+
+	var (
+		serviceEndpoint = c.serviceEndpoint
+		routingKeys     []string
+	)
+
+	if len(connections) > 0 {
+		// get the route configs
+		serviceEndpoint, routingKeys, err = mediator.GetRouterConfig(c.routeSvc, connections[0], c.serviceEndpoint)
```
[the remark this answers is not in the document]
soluchok: Sure, I will add it.

pkg/client/outofband/client.go
soluchok: Sure, I will add it as well. It should not be a problem.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -485,17 +485,32 @@ func (ctx *context) getDIDDocAndConnection(pubDID string) (*did.Doc, *Connection
...
-		return nil, nil, fmt.Errorf("did doc - fetch router config: %w", err)
+		return nil, nil, fmt.Errorf("router get connections: %w", err)
+	}
+
+	var services []did.Service
+
+	for i := range connections {
+		s := did.Service{}
+		// get the route configs (pass empty service endpoint, as default service endpoint added in VDRI)
+		s.ServiceEndpoint, s.RoutingKeys, err = mediator.GetRouterConfig(ctx.routeSvc, connections[i], "")
```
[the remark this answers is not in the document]
soluchok: I'll check what I can do here.

pkg/client/didexchange/client.go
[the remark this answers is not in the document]
soluchok: Done! Added possibly to provide `RouterConnectionID`.

pkg/client/outofband/client.go
soluchok: Done! Added possibly to provide `RouterConnectionID`.

pkg/didcomm/protocol/didexchange/states.go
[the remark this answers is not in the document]
soluchok: Done!

pkg/client/outofband/client.go
soluchok: Done! Added possibly to provide the router`s connection.

test/bdd/features/aries_mediator_e2e_controller.feature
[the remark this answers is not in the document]
rolsonquadras: Created an issue https://github.com/hyperledger/aries-framework-go/issues/2205.

pkg/client/didexchange/client.go
```diff
@@ -113,7 +113,7 @@ func New(ctx provider) (*Client, error) {
 // CreateInvitation creates an invitation. New key pair will be generated and base58 encoded public key will be
 // used as basis for invitation. This invitation will be stored so client can cross reference this invitation during
 // did exchange protocol.
-func (c *Client) CreateInvitation(label string) (*Invitation, error) {
+func (c *Client) CreateInvitation(label, routerConnID string) (*Invitation, error) {
```
rolsonquadras, line 116: It would be better to pass it as an option or create new API `CreateInvitation with router`. Having second argument makes passing routerConnID mandatory (either empty or some value) for the consumer, which ideally be shouldn't because framework is deriving the default router.

soluchok: Agree. I did it to understand who is using the function. I will make it optional.

soluchok: Done!

rolsonquadras (approved):

pkg/client/didexchange/client.go
```diff
@@ -35,6 +35,30 @@ const (
...
+	return opts
+}
+
+// Opt represents option function.
+type Opt func(*options)
+
+// WithRouterConnectionID allows you to specify the router connection.
+func WithRouterConnectionID(connID string) Opt {
+	return func(opts *options) {
+		opts.routerConnID = connID
```
llorllale, line 58: @soluchok why only one? `opts.routerConnID` could be an array

soluchok: ok, I will add this.

pkg/client/didexchange/client.go
```diff
@@ -123,10 +149,34 @@ func (c *Client) CreateInvitation(label string) (*Invitation, error) {
...
+	if opts.routerConnID == "" {
+		var connections []string
+
+		connections, err = c.routeSvc.GetConnections()
+		if err != nil {
+			return nil, fmt.Errorf("createInvitation: get connections: %w", err)
+		}
+
+		if len(connections) > 0 {
+			opts.routerConnID = connections[0]
```
llorllale, line 161: This default still makes no sense to me.

If the user has not provided a router connection ID, we shouldn't automatically take the first one.

soluchok: I thought it should be the default behavior. Like it was before.
@rolsonquadras @troyronda ^

pkg/client/didexchange/client.go
llorllale: I was also thinking of the user injecting a _strategy_ for router selection.

Instead of `WithRouterConnectionID` you would have `WithRouterSelection`. Example:

```go
inv, err := client.CreateInvitation(protocols, oob.WithRouterSelection(
    func(svc mediator.ProtocolService) ([]string, error) {
        // return the router connection IDs
    }
))
```

Or something similar. The output doesn't necessarily have to be connection IDs and the input does not necessarily have to be `mediator.ProtocolService`.

The point is we should allow the user to inject _behaviour_, not values.

pkg/client/didexchange/client.go
rolsonquadras: @llorllale 

> This default still makes no sense to me.

Any proposals/thoughts you have in mind for the default ?

pkg/client/didexchange/client.go
soluchok: I don't think it helps. Provide a connection is the easiest way. Also, it works for REST. You can figure out the connection outside. I mean you don’t need a function.

pkg/client/didexchange/client.go
llorllale: I think the user should be responsible for selecting the routers where they want to register their identifiers.

Barring that, next best option is to do randomized selection as default. My concern is mitigated, but not resolved, with a randomized approach.

pkg/client/didexchange/client.go
llorllale: > it works for REST

In that case, ok... make it an array of connection IDs. We should definitely not limit the user to just one router.

soluchok: Sure, I'll do that.

soluchok: @llorllale Added possibility to specify multiple connections. If connections are not specified we will add all connections to the `new did doc`. See `getDIDDocAndConnection` function.

llorllale:
@soluchok @rolsonquadras 

I don't think it's a good idea for the framework to automatically register the user's identifiers on all registered routers without the ability for them to disable this behaviour. This would mean logic like this in the did-exchange service is not OK:

```go
	if len(routerConnections) == 0 {  // routerConnections was provided by the user
		var err error

		routerConnections, err = ctx.routeSvc.GetConnections()  // this override is not OK
		if err != nil {
			return nil, nil, fmt.Errorf("router get connections: %w", err)
		}
	}
```

I think the user should _always_ specify the routers they wish to register their identifiers with. This would mean:

* configuring a default when initializing the framework
* being able to update the default with an existing framework context
* being able to specify different routers for a specific action without updating the default

And the points where these changes would be visible to the user are roughly:

* OOB client (when creating messages, when responding to DIDComm actions)
* did-exchange client (when creating messages, when responding to DIDComm actions)
* option for `aries.New()` (initialize with default)
* method on the `context` (update default)

llorllale:
No question: my proposal above would break downstream projects.

When initializing for the first time users on devices incapable of hosting a stable inbound transport (ie. HTTP) must now:

* init the framework as usual
* did-exchange with router (just like now)
* register router (just like now)
* update framework default (new step)
* save their preferred default router (new step)

Future startups would:

* init the framework as usual
* look to see if they have a preferred default router configured
* update the context's default if they so wish

llorllale:
@troyronda @rolsonquadras I'd like your input on the above please.

troyronda:
> I don't think it's a good idea for the framework to automatically register the user's identifiers on all registered routers

Agree with this doesn't seem like a good idea.

rolsonquadras:
@llorllale 

> @soluchok @rolsonquadras
> 
> I don't think it's a good idea for the framework to automatically register the user's identifiers on all registered routers without the ability for them to disable this behaviour. This would mean logic like this in the did-exchange service is not OK:
> 
> ```go
> 	if len(routerConnections) == 0 {  // routerConnections was provided by the user
> 		var err error
> 
> 		routerConnections, err = ctx.routeSvc.GetConnections()  // this override is not OK
> 		if err != nil {
> 			return nil, nil, fmt.Errorf("router get connections: %w", err)
> 		}
> 	}
> ```
> 

Agree that we shouldn't register against all the routers. I thought this was changed as part of https://github.com/hyperledger/aries-framework-go/pull/2202#discussion_r493663658.

rolsonquadras:
@llorllale 

> No question: my proposal above would break downstream projects.
> 
> When initializing for the first time users on devices incapable of hosting a stable inbound transport (ie. HTTP) must now:
> 
> * init the framework as usual
> * did-exchange with router (just like now)
> * register router (just like now)
> * update framework default (new step)
> * save their preferred default router (new step)
> 
> Subsequent startups would:
> 
> * init the framework as usual
> * look to see if they have a preferred default router configured
> * update the context's default if they so wish

I would be good if we provide option to set the router during framework init. But I guess, in any scenario (in this case what if the user doesn't set the preferred router?), we would end up with framework choosing the router (unless we make it mandatory). At the end, it's a question of where and how framework gonna derive default router connection id.

llorllale:
@rolsonquadras @soluchok 

> I guess, in any scenario (in this case what if the user doesn't set the preferred router?), we would end up with framework choosing the router (unless we make it mandatory)

The framework should not choose any router on its own.

I suspect the initial logic is there in order to make it a bit easier for mobile, endpoint-less agents. However, the framework cannot determine on its own the capabilities of the environment it's being run in.

With that line of thought, I'm now 100% convinced the framework should _only_ add the routing info if the user somehow asked it to.

pkg/client/didexchange/client.go
```diff
@@ -113,7 +143,13 @@ func New(ctx provider) (*Client, error) {
 // CreateInvitation creates an invitation. New key pair will be generated and base58 encoded public key will be
 // used as basis for invitation. This invitation will be stored so client can cross reference this invitation during
 // did exchange protocol.
-func (c *Client) CreateInvitation(label string) (*Invitation, error) {
+func (c *Client) CreateInvitation(label string, args ...Opt) (*Invitation, error) {
+	opts := applyOptions(args...)
+
+	if len(opts.routerConnections) > 1 {
+		return nil, errors.New("only one router connection can be specified for an invitation")
```
llorllale, line 150: `opts.routerConnections` should not be an array

pkg/client/outofband/client.go
```diff
@@ -63,17 +70,26 @@ type Event interface {
...
+	RouterConnections []string
+	Service           []interface{}
+}
+
+func (m *message) RouterConnection() string {
+	if len(m.RouterConnections) == 0 {
+		return ""
+	}
+
+	return m.RouterConnections[0]
```
llorllale, line 85: Unlike didexchange invitations, an OOB invitation's `service` attribute can contain multiple entries, so it makes sense to return all router connection IDs and use them

pkg/controller/rest/outofband/models.go
```diff
@@ -19,10 +19,11 @@ import (
...
 	Body struct {
-		Label    string        `json:"label"`
-		Goal     string        `json:"goal"`
-		GoalCode string        `json:"goal_code"`
-		Service  []interface{} `json:"service"`
+		Label              string        `json:"label"`
+		Goal               string        `json:"goal"`
+		GoalCode           string        `json:"goal_code"`
+		Service            []interface{} `json:"service"`
+		RouterConnectionID string        `json:"router_connection_id"`
```
llorllale, line 26: we can have several here

pkg/controller/rest/outofband/models.go
```diff
@@ -49,11 +50,12 @@ type outofbandCreateRequestResponse struct { // nolint: unused,deadcode
...
-		Goal      string        `json:"goal"`
-		GoalCode  string        `json:"goal_code"`
-		Service   []interface{} `json:"service"`
-		Protocols []string      `json:"protocols"`
+		Label              string        `json:"label"`
+		Goal               string        `json:"goal"`
+		GoalCode           string        `json:"goal_code"`
+		Service            []interface{} `json:"service"`
+		Protocols          []string      `json:"protocols"`
+		RouterConnectionID string        `json:"router_connection_id"`
```
llorllale, line 58: can have several routers

soluchok: you can’t. Create an invitation supports only one destination.

pkg/client/outofband/client.go
soluchok: It leads us to `chooseTarget` function. Which picks up the first one.

llorllale: There is a separation between what the client includes in an OOB invitation, and what the other agent (that may not even be based on this framework) does with it.

`chooseTarget` is what this framework is doing. Another framework may not have the same logic.

pkg/controller/rest/outofband/models.go
llorllale: Answered above.

pkg/client/outofband/client.go
llorllale: We should also provide the user with the ability to choose with `service` entry they would like to use when accepting an OOB invitation.

llorllale: Opened #2209

llorllale: This issue can be deferred for a followup

pkg/client/didexchange/client.go
llorllale: @soluchok this one is pending

pkg/client/outofband/client.go
soluchok: Ok

pkg/client/didexchange/client.go
soluchok: I will fix it!

soluchok: @llorllale Done!

llorllale (approved):

Issue #2199, opened 2020-09-21
Add EDV provider


DRK3:
REST API and EDV models must be decoupled so as to allow the EDV models to be used in other storage providers.

- [x] FormatProvider (done in #2210)
- [x] #2249 
- [x] #2273
- [x] #2208

llorllale:
@DRK3 

**High-level Design**

`storage.Formatter`: formats documents to and from storages in a way that preserves the user's privacy vs. external storage providers

```go
package storage

type Formatter interface {
   FormatPair(k string, v []byte) ([]byte, error)
   ParseValue([]byte) ([]byte, error)
}
```

`edv.Formatter`: implements `storage.Formatter`

`edv.Store`: implements `storage.Store`

We will not have to add any new methods to `storage.Store`. We can delete `storage.Store.Query()`.

DRK3:
Support for the https://github.com/trustbloc/edv ReturnFullDocumentsOnQuery and Batch extensions have also been implemented in #2355/#2356 and #2365.

PR #2207, opened 2020-09-28
feat: SDS REST provider

DRK3:

closes #2207
part of #2199

### DIFF
--- a/pkg/storage/edv/restprovider.go
+++ b/pkg/storage/edv/restprovider.go
@@ -1,0 +1,415 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package edv
+
+import (
+	"bytes"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+const (
+	indexName                  = "StoreName-FriendlyKeyName"
+	indexValueFormat           = "%s-%s"
+	contentTypeApplicationJSON = "application/json"
+	locationHeaderName         = "Location"
+
+	failComputeMACIndexName               = "failed to compute MAC for index name: %w"
+	failComputeMACIndexValue              = "failed to compute index value MAC: %w"
+	failCreateIndexedAttribute            = "failed to create indexed attribute: %w"
+	failComputeBase64EncodedIndexValueMAC = "failed to compute Base64-encoded index value MAC: %w"
+	failCreateDocumentInEDVServer         = "failed to create document in EDV server: %w"
+	failQueryVaultInEDVServer             = "failed to query vault in EDV server: %w"
+	noDocumentMatchingQueryFound          = "no document matching the query was found: %w"
+	failRetrieveDocumentFromEDVServer     = "failed to retrieve document from EDV server: %w"
+
+	failSendGETRequest             = "failed to send GET request: %w"
+	failSendPOSTRequest            = "failed to send POST request: %w"
+	failReadResponseBody           = "failed to read response body: %w"
+	failMarshalQuery               = "failed to marshal query: %w"
+	failResponseFromEDVServer      = "status code %d was returned along with the following message: %s"
+	failUnmarshalDocumentLocations = "failed to unmarshal response bytes into document locations: %w"
+
+	createDocumentRequestLogMsg = "Sending request to create the following document: %s"
+	sendGETRequestLogMsg        = `Sent GET request to %s.
+Response status code: %d
+Response body: %s`
+	sendPOSTRequestLogMsg = `Sent POST request to %s.
+Request body: %s
+
+Response status code: %d
+Response body: %s`
+	failCloseResponseBodyLogMsg = "Failed to close response body: %s"
+)
+
+var (
+	logger                            = log.New("EDV-REST-Provider")
+	errDeleteNotSupported             = errors.New("EDV REST store delete functionality not yet implemented")
+	errMultipleDocumentsMatchingQuery = errors.New("multiple documents matching the query were found." +
+		"This probably indicates an issue with the EDV server's database")
+)
+
+// MACDigester represents a type that can compute MACs.
+type MACDigester interface {
+	ComputeMAC(data []byte, kh interface{}) ([]byte, error)
+}
+
+// MACCrypto is used for computing MACs.
+type MACCrypto struct {
+	kh          interface{}
+	macDigester MACDigester
+}
+
+// ComputeMAC computes a MAC for data using a matching MAC primitive in kh.
+func (m *MACCrypto) ComputeMAC(data string) (string, error) {
+	dataMAC, err := m.macDigester.ComputeMAC([]byte(data), m.kh)
+	return string(dataMAC), err
+}
+
+// NewMACCrypto returns a new instance of a MACCrypto.
+func NewMACCrypto(kh interface{}, macDigester MACDigester) *MACCrypto {
+	return &MACCrypto{
+		kh:          kh,
+		macDigester: macDigester,
+	}
+}
+
+// RESTProvider is a store provider that can be used to store data in a server supporting the
+// data vault HTTPS API as defined in https://identity.foundation/secure-data-store/#data-vault-https-api.
+type RESTProvider struct {
+	vaultID                  string
+	macCrypto                *MACCrypto
+	indexKeyMACBase64Encoded string
+	restClient               restClient
+}
+
+// NewRESTProvider returns a new RESTProvider. edvServerURL is the base URL for the data vault HTTPS API.
+// vaultID is the ID of the vault where this provider will store data. The vault must be created in advance, and since
+// the EDV REST API does not provide a method to check if a vault with a given ID exists, any errors due to a
+// non-existent vault will be deferred until calls are actually made to it in the restStore.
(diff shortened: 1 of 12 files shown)


pkg/storage/sds/restclient/restclient.go
```diff
@@ -0,0 +1,221 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package restclient
```
llorllale, line 7: please rename to just `client`

pkg/storage/sds/models/models.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package models
```
llorllale, line 7: these structs can be promoted upward one level, into the `sds` package

pkg/storage/sds/restclient/restclient.go
```diff
@@ -0,0 +1,221 @@
...
+
+// WithTLSConfig option is for definition of secured HTTP transport using a tls.Config instance.
+func WithTLSConfig(tlsConfig *tls.Config) Option {
+	return func(opts *Client) {
+		opts.httpClient.Transport = &http.Transport{TLSClientConfig: tlsConfig}
+	}
+}
+
+// New returns a new instance of an SDS client.
+func New(sdsServerURL string, opts ...Option) *Client {
```
llorllale, line 52: Please include a link to the SDS spec.

pkg/storage/sds/restclient/restclient.go
DRK3: The reason I named it "restclient" instead of just "client" is because my next step is to add a "full" SDS client that incorporates the crypto ability and the REST stuff, and I was going to make that "full" client type contain a SDS REST client. This allows me to make the files smaller and also allows others to use just the REST client separately from the crypto stuff. Then they can construct Encrypted Documents using their own crypto code. Does that approach make sense?

pkg/storage/sds/models/models.go
DRK3: Done!

pkg/storage/sds/restclient/restclient.go
DRK3: Done! I also included the date and version since the spec page is updated frequently.

pkg/storage/sds/restclient/restclient.go
DRK3: To clarify - I was planning on naming the package with the full SDS client just "client", unless you have a better name.

Or maybe you're thinking of having it all in one package, but different file names? Hmm...

llorllale: I'm not entirely clear what you mean by "full SDS client that incorporates the crypto ability":

* we need a `Store` implementation that stores stuff in the SDS
* encryption is required regardless of the `Store` implementation
* therefore I'm not sure what exact "crypto abilities" this "full SDS client" will have

Let's discuss this to see if I understand the design correctly.

llorllale (changes requested):
Design discussion underway - see https://github.com/hyperledger/aries-framework-go/pull/2207#discussion_r496631879

DRK3:
This PR is on hold for now - it will be reworked after some design changes.

DRK3:
I had to make some changes to the Formatter interface and EncryptedFormatter implementation in order to make this work. Let me explain...

EDV vault IDs are generated and returned by the EDV server upon creation. They are not user-specifiable. In order to allow the EDV REST provider to conform to the Aries Provider interface and have similar semantics to other providers, here's what was done in this PR:
- The `NewRESTProvider` function takes in a `vaultID` corresponding to some vault that's been created previously (outside of Aries). This vault is where everything gets stored.
- When opening a store, the `name` passed in acts as a sort of "prefix" for all documents created within that store. Opening a store doesn't create a new vault on the server.
- Encrypted Document IDs are randomly generated, so we need some way to retrieve their associated document using some known identifier. This is where encrypted indices some in.
- When storing a document within a given store, I create an encrypted index, where the index value is an HMACed version of the store name and the key name concatenated together. This encrypted index is added to the document before storing.
- When retrieving a document, I recreate the original HMACed values based on the store name + key name, and use that to 
 call the Query endpoint on the EDV server. The EDV server will return the actual Encrypted Document ID. I then call the Read Document endpoint using the Encrypted Document ID, and then I will receive the full Encrypted Document back. This is then returned to whoever called the EDV REST Provider's Get method. One use case would be the FormatProvider, which will then use the EncryptedFormatter it was set up with to decrypt the document before sending it back up.

Since the REST provider is the one creating the index now, this code was removed from the EncryptedFormatter. This also means that there's really no reason for the Formatter FormatPair interface method to take in a key and value, so I changed to just `Format([]byte) ([]byte, error)`.

The above strategy will still allow for multiple stores, and each of those stores behaves as you would expect (e.g. you can have the same data in two different stores without them conflicting thanks to the prefix). From the outside, you wouldn't know what's going on. The only requirement is for one single vault (the one passed in) to be created previously. So that vault creation is considered part of the pre-setup, similar to how CouchDB requires a `_users` database to be created in advance. The rest of the code in Aries won't need to be changed - it doesn't need to know that an EDV REST provider is being used.

pkg/storage/edv/restprovider.go
```diff
@@ -0,0 +1,378 @@
...
+	encryptedDocumentBytes, err := r.restClient.ReadDocument(r.vaultID, matchingDocumentIDs[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve document from EDV server: %w", err)
+	}
+
+	return encryptedDocumentBytes, nil
+}
+
+func (r *restStore) Iterator(startKey, endKey string) storage.StoreIterator {
+	panic("implement me")
```
llorllale, line 182: Make sure these panics are removed.

pkg/storage/edv/restprovider.go
```diff
@@ -0,0 +1,378 @@
...
+func WithTLSConfig(tlsConfig *tls.Config) Option {
+	return func(opts *RESTClient) {
+		opts.httpClient.Transport = &http.Transport{TLSClientConfig: tlsConfig}
+	}
+}
+
+// New returns a new instance of an EDV client.
+// This implementation is based on the Secure Data Store 0.1 Unofficial Draft, last updated on 20 September 2020.
+// https://identity.foundation/secure-data-store/
+func New(edvServerURL string, opts ...Option) *RESTClient {
```
llorllale, line 242: s/New/NewRESTClient/

llorllale:
@DRK3 LGTM

pkg/storage/sds/restclient/restclient.go
DRK3: (marking as resolved since this conversation was based on a much older code design. It's not really applicable in the same way anymore)

pkg/storage/edv/restprovider.go
DRK3: Done

pkg/storage/edv/restprovider.go
DRK3: Done

pkg/storage/edv/restprovider.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+// CloseStore is not currently supported by RESTProvider.
+func (r *RESTProvider) CloseStore(string) error {
+	return errCloseStoreNotSupported
+}
+
+// Close is not currently supported by RESTProvider.
+func (r *RESTProvider) Close() error {
+	return errCloseStoreNotSupported
```
llorllale, line 145: This will return an error when calling [`framework.Aries.Close()`](https://github.com/hyperledger/aries-framework-go/blob/ca5e0650014e48021c33f0c96fe23a11123967cf/pkg/framework/aries/framework.go#L301-L324). Have it return `nil` instead. Same with `CloseStore()`.

pkg/storage/edv/restprovider.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+// Not support by restStore. The EDV data vault HTTPS API doesn't support this type of operation,
+// nor is there a feasible way to emulate it. The iterator returned always is always in an error state.
+func (r *restStore) Iterator(string, string) storage.StoreIterator {
+	return &restStoreIterator{}
+}
+
+func (r *restStore) Delete(string) error {
+	panic("implement me")
```
llorllale, line 220: Should return an error here.

Baha-sk (approved):

Baha-sk:
looks good, only needs increased code coverage

pkg/storage/edv/restprovider.go
DRK3: Fixed.

pkg/storage/edv/restprovider.go
DRK3: Fixed. I also added a TODO above the method since Delete is actually in the EDV REST API - we just don't have a use case for it and it's not supported in the implementation I'm testing it against... yet.

https://github.com/hyperledger/aries-framework-go/issues/2286

pkg/storage/edv/restprovider.go
```diff
@@ -0,0 +1,405 @@
...
+	if err != nil {
+		return nil, fmt.Errorf(failQueryVaultInEDVServer, err)
+	}
+
+	if len(matchingDocumentIDs) == 0 {
+		return nil, fmt.Errorf("no documents matching the query were found: %w", storage.ErrDataNotFound)
+	} else if len(matchingDocumentIDs) > 1 {
+		// This should only be possible if the EDV server is not able to maintain the uniqueness property of the
+		// indexedAttribute created in the createIndexedAttribute method.
+		// TODO: Check each of the documents to see if they all have the same content (other than the document ID).
```
llorllale, line 199: issue number?

DRK3: Just added it.

llorllale (approved):

PR #2211, opened 2020-10-01
fix: Mediator - replacing a channel sync with the database sync

soluchok:
Currently, we are using the `go channel` to synchronize whether the `mediate-grant` message was received.
The problem here is that the `mediate-grant` message can be received on the different `node`. Which means we will fail due to timeout. This PR fixes it by saving data to DB.

### DIFF
--- a/pkg/didcomm/protocol/mediator/service.go
+++ b/pkg/didcomm/protocol/mediator/service.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcutil/base58"
+	"github.com/cenkalti/backoff/v4"
 	"github.com/google/uuid"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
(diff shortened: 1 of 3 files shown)


pkg/didcomm/protocol/mediator/service.go
```diff
@@ -549,41 +527,69 @@ func (s *Service) doRegistration(record *connection.Record, req *Request, timeou
...
 	// demonstrates? additionally `ExpiresTime` would need to be migrated to int64
 	req.ExpiresTime = time.Now().UTC().Add(timeout)
 
 	// send message to the router
-	if err := s.outbound.SendToDID(req, record.MyDID, record.TheirDID); err != nil {
+	if err = s.outbound.SendToDID(req, record.MyDID, record.TheirDID); err != nil {
 		return fmt.Errorf("send route request: %w", err)
 	}
 
-	// callback processing (to make this function look like a sync function)
```
rolsonquadras, line 565: Can we have a comment something like this ? It'll help to understand why we are doing this when one someone looks at the code.

rolsonquadras (approved):

llorllale (approved):

sudeshrshetty (approved):

pkg/didcomm/protocol/mediator/service.go
soluchok: Done!

PR #2226, opened 2020-10-12
fix: Couchdb - document update conflict

soluchok:
To update Couchdb data we need revID. If `revID` is empty and we perform an update operation we will get an error `Document update conflict`. It happens because our `Put` operation is not consistent.
e.g 
```
revID = c.getRevID(key)
db.Put(key , {...revID...})
```

This PR fixes the problem of `document update conflict` by adding retry logic.

### DIFF
--- a/pkg/storage/couchdb/couchdbstore.go
+++ b/pkg/storage/couchdb/couchdbstore.go
@@ -17,6 +17,9 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
+
+	"github.com/cenkalti/backoff"
 
 	// The CouchDB driver.
 	_ "github.com/go-kivik/couchdb"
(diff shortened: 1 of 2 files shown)


pkg/storage/couchdb/couchdbstore.go
```diff
@@ -192,21 +195,33 @@ func (c *CouchDBStore) Put(k string, v []byte) error {
...
-		return err
-	}
+	return c.put(k, valueToPut)
+}
 
-	if revID != "" {
-		valueToPut = []byte(`{"_rev":"` + revID + `",` + string(valueToPut[1:]))
-	}
+func (c *CouchDBStore) put(k string, value []byte) error {
+	const maxRetries = 3
```
rolsonquadras, line 202: What's the reason for multiple retries ? If the first update fails, then we should retry with revID only once, right ?

soluchok: Yes, but second also might fail. I think `3` is more than enough. It just makes it more robust.

rolsonquadras: Why would the second attempt fail if we are passing the updateDoc with revID ? Are we trying to solve `concurrent update` issue here (code gets revID, but before it could update the database, there was another update for that record) ?

soluchok: `Are we trying to solve concurrent update` - yes.
It is not under the lock. Here is our code.
```
// here we received an empty revID (means no documents yet in the DB)
revID = c.getRevID(key)
// here we received an error `Document update conflict` because in the same time someone already saved something
db.Put(key , {...revID...})
``

rolsonquadras: Yeah, lock at application layer will not work.

The question is that whether we want to throw error to the user (stale data) or override the updates (like we are doing in this PR).

soluchok: Yeah, I was thinking about it as well. And I decided that the proper fix will be on the database layer.
My decision was based on the consistency between all implementation. Let’s imagine that each implementation will have a similar bug like this. It means that our framework should handle the specific use cases for every implementation. It will be a huge problem. Especially after we move storage implementation to a separate repository.

rolsonquadras: Also, we have mysql/mem implementation where we don't check for the stale data. In a way, this will be consistent at-least. We can revisit `concurrent update` issue in future PRs.

rolsonquadras (approved):

fqutishat (approved):

Issue #2221, opened 2020-10-10
BBS+ signature verifier


kdimak:
- [x] Implement BBS+ signature verifier according to https://w3c-ccg.github.io/ldp-bbs2020.
- [ ] Use it for VC/VP as another linked data proofs suite.

PR #2234, opened 2020-10-13
feat: JSON-LD BbsBlsSignature2020 signature suite

kdimak:

#2221


Baha-sk (approved):

pkg/doc/signature/proof/jws.go
```diff
@@ -143,71 +143,9 @@ func prepareDocumentForJWS(suite signatureSuite, jsonldObject map[string]interfa
...
-	if err != nil {
-		return nil, err
+	contextMap := map[string]interface{}{
+		"@context": "https://w3id.org/security/v2",
 	}
 
 	return jsonld.Default().Compact(docMap, contextMap, opts...)
 }
-
-// cached value from https://w3id.org/security/v2
```
troyronda, line 156: whats the impact of not using cache here?

if internet fetch is needed, we should be preloading.

kdimak: I agree. We should preload `https://w3id.org/security/v2` (and dependent https://w3id.org/security/v1 as well) the same way we pre-load `https://www.w3.org/2018/credentials/v1` for VC/VP.

kdimak: I created a separate issue for that (#2253) in order to not overload current PR.

rolsonquadras (approved):

troyronda (approved):

PR #2252, opened 2020-10-20
feat: reply nested feature in command controller

### LINKED ISSUE #2251: Reply Nested feature in command controller
messenger command controller-> need REST, JS & Mobile support for reply by starting new thread

sudeshrshetty:
- added reply nested (reply by starting new thread) feature in all
command controllers
- added JS bddtests
- closes #2251

### DIFF
--- a/pkg/didcomm/messenger/messenger.go
+++ b/pkg/didcomm/messenger/messenger.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/google/uuid"
 
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
@@ -51,6 +52,8 @@ type Messenger struct {
 	store      storage.Store
 	dispatcher dispatcher.Outbound
 }
+
+var logger = log.New("aries-framework/pkg/didcomm/messenger")
 
 // NewMessenger returns a new instance of the Messenger.
 func NewMessenger(ctx Provider) (*Messenger, error) {
@@ -195,8 +198,8 @@ func (m *Messenger) ReplyTo(msgID string, msg service.DIDCommMsgMap) error {
 // ReplyToNested sends the message by starting a new thread.
 // Do not provide a message with ~thread decorator. It will be rewritten.
 // The function adds ~thread decorator to the message according to the given threadID.
-// NOTE: Given threadID becomes parent threadID.
-func (m *Messenger) ReplyToNested(threadID string, msg service.DIDCommMsgMap, myDID, theirDID string) error {
+// NOTE: Given threadID (from opts or from message record) becomes parent threadID.
+func (m *Messenger) ReplyToNested(msg service.DIDCommMsgMap, opts *service.NestedReplyOpts) error {
 	// fills missing fields
 	fillIfMissing(msg)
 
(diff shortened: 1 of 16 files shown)


fqutishat (approved):

pkg/didcomm/messenger/messenger.go
```diff
@@ -195,19 +198,23 @@ func (m *Messenger) ReplyTo(msgID string, msg service.DIDCommMsgMap) error {
 // ReplyToNested sends the message by starting a new thread.
 // Do not provide a message with ~thread decorator. It will be rewritten.
 // The function adds ~thread decorator to the message according to the given threadID.
-// NOTE: Given threadID becomes parent threadID.
-func (m *Messenger) ReplyToNested(threadID string, msg service.DIDCommMsgMap, myDID, theirDID string) error {
+// NOTE: Given threadID (from opts or from message record) becomes parent threadID.
+func (m *Messenger) ReplyToNested(msg service.DIDCommMsgMap, opts *service.NestedReplyOpts) error {
```
soluchok, line 202: I would suggest instead of using `*service.NestedReplyOpts` we need to use variadic args.
`NestedReplyOpts` should be private. 
Two functions need to be exposed.
e.g `WithMsgID(MsgID)` and with `WithDID(TheirDID,MyDID,ThreadID)`

It will force the client to use the right option without looking into the implementation.
Otherwise, It will confuse what data needs to be populated.
Someone might think about to provide custom `ThreadID` and `MsgID`. e.g
``` 
&NestedReplyOpts {
	ThreadID: "...."
	MsgID:"...."
}
```
The behavior here might surprise :)

soluchok: My point is to create an option function for each use case.

test/aries-js-worker/test/messaging/messaging.js
```diff
@@ -120,6 +120,28 @@ describe("Basic Messaging", function () {
...
+                stop()
+                resolve(msg.payload.message)
+            }, [basicMsgSvcName])
+        })
+
+        assert.equal(incomingMsg["@id"], sampleReplyMsg["@id"])
+        assert.equal(incomingMsg["@type"], sampleReplyMsg["@type"])
+        assert.equal(incomingMsg.content, sampleReplyMsg.content)
+        assert.deepEqual(incomingMsg["~thread"], {
+            "pthid": "1d9a1589-9d7b-4308-9fab-8ee9730720c2"
```
soluchok, line 143: It would be better
```
// first startNotifier
const incoming = new Promise
// do action 
receiver.messaging.reply
// wait for startNotifier response 
incomingMsg await incoming
// checks
assert.equal(incomingMsg["@i

```

pkg/didcomm/messenger/messenger.go
sudeshrshetty: I created one, but later reverted back while pushing the code,
 Since it is an interface, we have to export opts struct & resolve function as well so that implementing types can refer them from implementation.

Still, I get your point I can update it. 
In above approach we can also mix and match options.
For example: in this implementation 'TheirDID,MyDID,ThreadID'. User can pass `MsgID` & `MyDID,TheirDID` options, missing `ThreadID` will be pre-filled by record read by MsgID.

test/aries-js-worker/test/messaging/messaging.js
sudeshrshetty: I didn't get you, it is same variation of previous step. Can you elobarate?

sudeshrshetty: fixed

soluchok (approved):

rolsonquadras (approved):

sudeshrshetty:
retriggered

PR #2259, opened 2020-10-20
feat: flexible cache for jsonld.Processor DocumentLoader

### LINKED ISSUE #2253: flexible cache for jsonld.Processor DocumentLoader
We need to be able to apply extra cache (like the cache of `https://w3id.org/security/v2`) in some cases on the fly in the `signature` package. Having `jsonld.WithDocumentLoader` passed as a functional argument, we should develop a mechanism to somehow append the cache to the possibly passed function argument (if not passed, a new `DocumentLoader` should be created). Ideas:
- implement the appending mechanism in `jsonld.WithDocumentLoader()` or
- add another functional option to pass cache items (url + schema contents) directly.

kdimak:
Added `jsonld. WithDocumentLoaderCache()` which can be used to pass cached JSON-LD contexts as pairs of URLs and JSON-LD context content (e.g. as string or `io.Reader`).
It can be used in combination with actively used `jsonld. WithDocumentLoader()`.

Current logic:
- if neither `DocumentLoader` nor `DocumentLoaderCache` is defined - do not set `DocumentLoader`JSON-LD option;
- if `DocumentLoader` is not defined - create it using `ld.NewCachingDocumentLoader()`;
- otherwise, check if `DocumentLoader` is caching one, if not, cteate a wrapping caching loader.
- apply cached contexts from `DocumentLoaderCache` to the cached document loader.

Checked that the tests from `signature` package can be run off-line (so there is no dependency on any online available JSON-LD contexts). This is achieved by passing the cached document loader and context maps in the tests.

closes #2253

### DIFF
--- a/pkg/doc/signature/jsonld/processor.go
+++ b/pkg/doc/signature/jsonld/processor.go
@@ -7,8 +7,11 @@ SPDX-License-Identifier: Apache-2.0
 package jsonld
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"strings"
 
 	"github.com/piprate/json-gold/ld"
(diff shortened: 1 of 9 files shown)


pkg/doc/signature/proof/jws.go
```diff
@@ -129,6 +129,8 @@ func prepareDocumentForJWS(suite signatureSuite, jsonldObject map[string]interfa
 	doc := GetCopyWithoutProof(jsonldObject)
 
 	if suite.CompactProof() {
+		opts = append(opts, jsonld.WithDocumentLoaderCache(jsonldCache))
```
kdimak, line 132: @troyronda Here we apply extra cached contexts (e.g. security v2 context) and still use input opts. Input opts can contain additional caches or custom context loader with defined cached context maps. All of them (cache) will be utilized when making JSON-LD operations like canonization or compaction.

pkg/doc/signature/jsonld/processor.go
```diff
@@ -27,42 +28,56 @@ var logger = log.New("aries-framework/json-ld-processor")
...
 // WithDocumentLoader option is for passing custom JSON-LD document loader.
 func WithDocumentLoader(loader ld.DocumentLoader) ProcessorOpts {
-	return func(opts *normalizeOpts) {
+	return func(opts *processorOpts) {
 		opts.documentLoader = loader
 	}
 }
 
+// WithDocumentLoaderCache option is for passing cached contexts to be used by JSON-LD context document loader.
+func WithDocumentLoaderCache(cache map[string]interface{}) ProcessorOpts {
```
rolsonquadras, line 58: I read it in the PR desc that string is supported as the value for cache map. Can we mention the supported types for the cache value in the godoc as well?

rolsonquadras (approved):

sandrask (approved):

fqutishat (approved):

PR #2263, opened 2020-10-21
feat: Enforce EDV document ID format in EDVFormatProvider

### LINKED ISSUE #2228: Query method for in-memory store

DRK3:

also closes #2228

### DIFF
--- a/pkg/didcomm/protocol/didexchange/service_test.go
+++ b/pkg/didcomm/protocol/didexchange/service_test.go
@@ -716,7 +716,7 @@ type mockStore struct {
 	put    func(string, []byte) error
 	get    func(string) ([]byte, error)
 	delete func(string) error
-	query  func(string) (storage.StoreIterator, error)
+	query  func(string, string) (storage.StoreIterator, error)
 }
 
 // Put stores the key and the record.
@@ -734,8 +734,8 @@ func (m *mockStore) Delete(k string) error {
 	return m.delete(k)
 }
 
-func (m *mockStore) Query(query string) (storage.StoreIterator, error) {
-	return m.query(query)
+func (m *mockStore) Query(name, value string) (storage.StoreIterator, error) {
+	return m.query(name, value)
 }
 
 // Search returns storage iterator.
--- a/pkg/storage/edv/formatprovider/formatprovider.go
+++ b/pkg/storage/edv/formatprovider/formatprovider.go
@@ -7,49 +7,80 @@ SPDX-License-Identifier: Apache-2.0
 package formatprovider
 
 import (
+	"crypto/rand"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+
+	"github.com/btcsuite/btcutil/base58"
 
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/storage/edv"
 )
 
 const (
+	failComputeMACIndexName        = "failed to compute MAC for index name: %w"
 	failOpenUnderlyingStore        = "failed to open underlying store in FormatProvider: %w"
 	failCloseUnderlyingStore       = "failed to close underlying store in FormatProvider: %w"
 	failCloseAllUnderlyingStores   = "failed to close all underlying stores in FormatProvider: %w"
+	failGenerateEDVCompatibleID    = "failed to generate EDV compatible ID: %w"
+	failCreateIndexedAttribute     = "failed to create indexed attribute: %w"
+	failToComputeMACIndexValue     = "failed to compute MAC for index value: %w"
 	failEncryptStructuredDocument  = "failed to encrypt structured document into a encrypted document: %w"
 	failMarshalEncryptedDocument   = "failed to marshal encrypted document into bytes: %w"
 	failPutInUnderlyingStore       = "failed to put encrypted document in underlying store in formatStore: %w"
-	failGetFromUnderlyingStore     = "failed to get encrypted document bytes from underlying store in formatStore: %w"
 	failUnmarshalEncryptedDocument = "failed to unmarshal encrypted document bytes into encrypted document struct: %w"
 	failDecryptEncryptedDocument   = "failed to decrypt encrypted document into a structured document: %w"
 	failDeleteInUnderlyingStore    = "failed to delete key-value pair in underlying store in formatStore: %w"
 	failQueryUnderlyingStore       = "failed to query underlying store in formatStore: %w"
-	payloadKey                     = "payload"
+
+	payloadKey  = "payload"
+	keyIndexKey = "indexKey"
 )
 
 var (
 	errPayloadKeyMissing = errors.New(`the structured document content did not contain the ` +
 		`expected "payload" key`)
-	errPayloadNotAssertableAsByteArray = errors.New("unable to assert the payload value as a []byte")
+	errPayloadNotAssertableAsString = errors.New("unable to assert the payload value as a string")
 )
 
 type marshalFunc func(interface{}) ([]byte, error)
 
+// MACDigester represents a type that can compute MACs.
+type MACDigester interface {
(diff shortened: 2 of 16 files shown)


DRK3:
An explanation of why this is needed:

While working on the EDV REST provider, I realized an issue with the approach I was taking. The spec says that the ID in a structured document and encrypted document MUST be a Base58-encoded 128-bit random value. This PR provides a solution to that problem while still allowing the client to use a "regular" non-Base58-encoded 128-bit random value.

Here's how it works:

- The EDVFormatProvider is initialized with a key to be used for HMAC operations.
- When storing documents, an EDV-compatible ID is randomly generated. This is the ID that is used for the underlying provider. The "user-provided" key (the one that's passed into the EDVFormatProvider) is hashed using an HMAC function. The hashed value is set as an EDV encrypted index
- When retrieving documents, the "user-provided" key is hashed. This hashed value is then used to query the underlying provider for the original document.

This generic indexing+querying approach should work with the upcoming EDV REST provider and any other providers that support querying.

@sandrask @troyronda And anyone else who reviews this: let me know what you think of this approach.

pkg/storage/edv/formatprovider/formatprovider.go
```diff
@@ -38,10 +43,24 @@ var (
 
 type marshalFunc func(interface{}) ([]byte, error)
 
+type MACComputer interface {
```
DRK3, line 46: I realize this name is a little bit funny. Let me know if you have any suggestions.

llorllale (approved):
Approved. Messy workaround necessary - request to remove this document ID constraint from the spec was filed a while ago: https://github.com/decentralized-identity/secure-data-store/issues/112

pkg/storage/edv/formatprovider/formatprovider.go
Baha-sk: `MACDigester` is better as we are usually talking about a message digest when referring to Mac.

pkg/didcomm/protocol/didexchange/service_test.go
```diff
@@ -734,7 +734,7 @@ func (m *mockStore) Delete(k string) error {
 	return m.delete(k)
 }
 
-func (m *mockStore) Query(query string) (storage.StoreIterator, error) {
+func (m *mockStore) Query(indexKey string, indexValue string) (storage.StoreIterator, error) {
 	return m.query(query)
```
Baha-sk, line 738: `query` argument is not refactored following the signature change above

DRK3: Fixed

pkg/storage/edv/formatprovider/formatprovider.go
DRK3: Done!

pkg/storage/edv/formatprovider/formatprovider.go
```diff
@@ -7,81 +7,122 @@ SPDX-License-Identifier: Apache-2.0
...
+func (m *MACCrypto) ComputeMAC(data string) (string, error) {
+	dataMAC, err := m.macDigester.ComputeMAC([]byte(data), m.kh)
+	return string(dataMAC), err
+}
+
 // DocumentProcessor represents a type that can encrypt and decrypt between
 // Structured Documents and Encrypted Documents.
 type DocumentProcessor interface {
-	Encrypt(*edv.StructuredDocument) (*edv.EncryptedDocument, error)
+	Encrypt(*edv.StructuredDocument, []edv.IndexedAttributeCollection) (*edv.EncryptedDocument, error)
```
llorllale, line 68: Just noticed these issues:

* DocumentProcessor is inside the `edv/formatprovider` package instead of being a first-class `storage` citizen (like we discussed previously). To recall: the "document processor" encrypts all artifacts into a standard form and would be reused with other storage implementations
* It looks like the documentprocessor does not encrypt the indexed attribute collection. Does it really need to be an argument? It's not even returned from `Decrypt()`, so there is an asymmetry here

DRK3: - The DocumentProcessor is used by the EDVFormatProvider, which allows the EDV format to be reused with any other storage implementation. There's no need for a "standard form".
- The indexed attributes are hashed before being passed in to the DocumentProcessor. This is what allows the documents to be queried. This allows us to still get documents based on a key per the usual provider method signature, so from the outside, an EDVFormatProvider (with whatever underlying provider you want) still works the same as any other provider. StructuredDocuments don't have encrypted indices, so they aren't returned.

DRK3: Now, I could create some sort of even more general DocumentProcessor class that just assumes JWE as what's being used for encryption. Or I could make it even more general and have it deal with just []byte. What do you think?

DRK3: Just realized one potentially confusing thing: the EDV spec calls them "Encrypted Indices", although they're actually HMACed.

DRK3: Note that the EDVFormatProvider will still require the more specific DocumentProcessor that deals with Structured and Encrypted Documents per the spec

llorllale: @DRK3 

> Now, I could create some sort of even more general DocumentProcessor class that just assumes JWE as what's being used for encryption. Or I could make it even more general and have it deal with just []byte. What do you think?

Last time we agreed we'd use the EDV format across all stores. The format would be pluggable, thus we'd have an "edv format" and potentially other formats. All would adhere to some common interface. This interface would a first-class citizen under `storage` or somewhere not far from it.

I guess we can regroup on monday.

llorllale (changes requested):
Let's first discuss this PR's direction as indicated [here](https://github.com/hyperledger/aries-framework-go/pull/2263#discussion_r511470829)

pkg/storage/edv/formatprovider/formatprovider.go
llorllale: Output of monday's session will be a GitHub issue with the high-level tasks and design

llorllale: Wrote down general direction here: https://github.com/hyperledger/aries-framework-go/issues/2199#issuecomment-716617512

DRK3: Created an issue and added a ToDo above the FormatProvider: https://github.com/hyperledger/aries-framework-go/issues/2273

pkg/storage/edv/formatprovider/formatprovider.go
```diff
@@ -101,22 +145,34 @@ func (p FormatProvider) Close() error {
...
-	marshal marshalFunc
+	underlyingStore          storage.Store
+	documentProcessor        DocumentProcessor
+	macCrypto                *MACCrypto
+	indexKeyMACBase64Encoded string
+	marshal                  marshalFunc
+	randomBytesFunc          generateRandomBytesFunc
 }
 
 func (s formatStore) Put(k string, v []byte) error {
```
troyronda, line 156: Why isn't formatStore a pointer receiver (i.e., *formatStore)?

when in doubt, use a pointer receiver.

DRK3: Yes, that's a mistake. I had originally fixed it in this PR (you'll see it if you go back a few force-commits), but since I'll be moving this entire file per https://github.com/hyperledger/aries-framework-go/pull/2263#discussion_r512089148, I figured that it would make more sense to keep this PR small and fix it in the next one. I'll make a note in that issue to fix it so that I don't miss it.

pkg/storage/edv/formatprovider/formatprovider.go
DRK3: My next PR will be to generalize this FormatProvider. This will involve moving it out of the `edv` folder and moving logic around to different places

llorllale (approved):

sandrask (approved):

PR #2279, opened 2020-10-28
feat: Generalize format provider

### LINKED ISSUE #2273: Generalize EDV format provider
Generalize edv.FormatProvider to be a more general format provider that can allow for other encrypted data formats.

See https://github.com/hyperledger/aries-framework-go/issues/2199#issuecomment-716617512.

Also, when moving the formatProvider, ensure that pointer receivers are used. This was missed in the original PR that added edv.FormatProvider.

DRK3:
Also removed Query method from the storage.Store interface as it will no longer be needed with this new design.

closes #2273


pkg/storage/format/formatprovider.go
```diff
@@ -0,0 +1,131 @@
...
+// Formatter represents a type that can encrypt and decrypt between
+// Structured Documents and Encrypted Documents.
+type Formatter interface {
+	FormatPair(k string, v []byte) ([]byte, error)
+	ParseValue([]byte) ([]byte, error)
+}
+
+// Provider is an encrypted storage provider that uses EDV document models
+// as defined in https://identity.foundation/secure-data-store/#data-model.
+// TODO (#2273): Generalize this to allow for other encrypted formats.
```
llorllale, line 34: TODO no longer applicable

pkg/storage/format/formatprovider.go
```diff
@@ -0,0 +1,131 @@
...
+	failCloseAllUnderlyingStores = "failed to close all underlying stores: %w"
+	failFormatValue              = "failed to format value: %w"
+	failPutInUnderlyingStore     = "failed to put encrypted document in underlying store: %w"
+	failParseFormattedValue      = "failed to parse formatted value: %w"
+	failDeleteInUnderlyingStore  = "failed to delete key-value pair in underlying store: %w"
+)
+
+// Formatter represents a type that can encrypt and decrypt between
+// Structured Documents and Encrypted Documents.
+type Formatter interface {
```
llorllale, line 27: Move this one level up into `storage` package

pkg/storage/format/formatprovider.go
```diff
@@ -0,0 +1,131 @@
...
+	FormatPair(k string, v []byte) ([]byte, error)
+	ParseValue([]byte) ([]byte, error)
+}
+
+// Provider is an encrypted storage provider that uses EDV document models
+// as defined in https://identity.foundation/secure-data-store/#data-model.
+// TODO (#2273): Generalize this to allow for other encrypted formats.
+type Provider struct {
```
llorllale, line 35: I forgot to include this in the original [comment](https://github.com/hyperledger/aries-framework-go/issues/2199#issuecomment-716617512), now updated:

> `storage.FormattedStore`: implements `storage.Store` and is composed of an underlying store and a `storage.Formatter`. It formats all data before saving it to the underlying store.

So this provider needs to be generalized and moved up a level into the `storage` package.

pkg/storage/formattedstore.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Formatter represents a type that can encrypt and decrypt between
+// Structured Documents and Encrypted Documents.
+type Formatter interface {
+	FormatPair(k string, v []byte) ([]byte, error)
+	ParseValue([]byte) ([]byte, error)
+}
+
+// FormattedProvider is a storage provider that allows for data to be formatted in an underlying provider.
+type FormattedProvider struct {
```
troyronda, line 33: I think the fact that we are creating encrypted/confidential stores could imply a stronger terminology than `Formatted` and FormattedProvider.

pkg/storage/store.go
```diff
@@ -58,9 +55,6 @@ type Store interface {
 
 	// Delete will delete a record with k key
 	Delete(k string) error
-
-	// Query searches all stored key-value pairs and returns all values that satisfy the query.
-	Query(name string, value string) (StoreIterator, error)
```
troyronda, line 63: The PR description should also mention that Query has been removed and the rationale.

pkg/storage/format/formatprovider.go
DRK3: Fixed

pkg/storage/format/formatprovider.go
DRK3: Done.

pkg/storage/format/formatprovider.go
DRK3: Done. Note that I had to move the FormattedStore tests into their own package (formattedstore_test.go, still the same folder) in order to prevent a circular dependency between the memstore package and the store.storage package.

pkg/storage/formattedstore.go
DRK3: I renamed the concrete EDV Formatter to EncryptedFormatter. However, the Formatter interface and the FormattedProvider type are general enough to be easily used for something that isn't encryption, which is why I didn't give them a more specific name. It just so happens that using it for an encrypted format is a concrete use case that I'm adding.

pkg/storage/store.go
DRK3: Done.

pkg/storage/edv/encryptedformatter.go
```diff
@@ -0,0 +1,226 @@
...
+func NewMACCrypto(kh interface{}, macDigester MACDigester) *MACCrypto {
+	return &MACCrypto{
+		kh:          kh,
+		macDigester: macDigester,
+	}
+}
+
+// EncryptedFormatter uses Aries crypto to encrypt and decrypt between
+// Structured Documents and Encrypted Documents.
+type EncryptedFormatter struct {
```
Baha-sk, line 74: The `Formatter` api description mentions the following:
```Formatter represents a type that can encrypt and decrypt between ...```

`EncryptedFormatter` is generic if we assume other `Formatter` implementations will also do encryption/decryption, I would rename this specific type to `JWEFormatter`

sandrask (approved):

pkg/storage/edv/encryptedformatter.go
Baha-sk: I would also rectify the description to mention:
```uses JWE to encrypt and decrypt between ...```

DRK3:
@Baha-sk Originally I had added the Query() method as a way to workaround one of the limitations of the EDV spec, which is that encrypted documents have randomly generated IDs, so the key passed into the FormatProvider would not be the one actually used in the store. The key passed into the FormatProvider would be HMACed in order to create an encrypted index, and we could use this to query the store later to figure out which document matches up with the key passed into the FormatProvider.

With this new design, the FormatProvider is no longer strictly for EDV. So the HMACing was moved to the DocumentProcessor (now called the EncryptedFormatter) and that querying behaviour can be done only by the upcoming EDV REST provider since it's really the only provider that needs this workaround. Other providers like memstore, CouchDB, etc. can simply store the key passed in to the FormatProvider and the value (the marshalled encrypted document) directly. The encrypted document is still a valid EDV encrypted document per the spec and could be migrated to an EDV REST server later without modification.

pkg/storage/edv/encryptedformatter.go
DRK3: The description I had on the `Formatter` interface was incorrect... I've updated it.

The `Formatter` represents a type that can convert data from one representation into another. `edv.EncryptedFormatter` is an implementation of this that does plaintext<-->EDV encrypted documents.

Does this clear things up? I think your original suggestion won't apply anymore.

Baha-sk (approved):

troyronda (approved):

pkg/storage/edv/encryptedformatter.go
DRK3: Even though locally I updated the comment above the Formatter interface, it looks like I somehow forgot to push it up. I'll fix it in my next PR.

PR #2280, opened 2020-10-28
fix: panic: cannot parse did document with `@base`

### LINKED ISSUE #2133: panic: cannot parse did document with `@base`
<details><summary>Test Case</summary>

```go
func TestBaseContext(t *testing.T) {
	test := `{
    "@context": [
        "https://www.w3.org/ns/did/v1",
        {
            "@base": "did:sidetree:EiBFsUlzmZ3zJtSFeQKwJNtngjmB51ehMWWDuptf9b4Bag"
        }
    ],
    "authentication": [
        "#signingKey"
    ],
    "id": "did:sidetree:EiBFsUlzmZ3zJtSFeQKwJNtngjmB51ehMWWDuptf9b4Bag",
    "publicKey": [
        {
            "controller": "",
            "id": "#signingKey",
            "publicKeyJwk": {
                "crv": "secp256k1",
                "kty": "EC",
                "x": "y9kzrVBqXx24sVMEQQk4Ce-0bqZ1i5TwxllWCkzA7wU",
                "y": "c2JHxQqUUtyWILAIicmpBGC47PgmJT465tPocG2q18k"
            },
            "type": "EcdsaSecp256k1VerificationKey2019"
        }
    ],
    "service": [
        {
            "id": "#serviceEndpointId123",
            "serviceEndpoint": "https://www.url.com",
            "type": "someType"
        }
    ]
}`
	_, err := ParseDocument([]byte(test))
	require.NoError(t, err)
}
```

</details>

<details><summary>Stacktrace</summary>

```
panic: interface conversion: interface {} is map[string]interface {}, not string [recovered]
	panic: interface conversion: interface {} is map[string]interface {}, not string

goroutine 19 [running]:
testing.tRunner.func1.1(0x932380, 0xc00024ca80)
	/usr/local/go/src/testing/testing.go:940 +0x2f5
testing.tRunner.func1(0xc00014d560)
	/usr/local/go/src/testing/testing.go:943 +0x3f9
panic(0x932380, 0xc00024ca80)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/hyperledger/aries-framework-go/pkg/doc/did.(*rawDoc).ParseContext(0xc000264000, 0x0, 0x0, 0xc00013afd0)
	/home/llorllale/dev/hyperledger/aries-framework-go/pkg/doc/did/doc.go:600 +0x2fe
github.com/hyperledger/aries-framework-go/pkg/doc/did.(*rawDoc).schemaLoader(0xc000264000, 0x36e, 0x380)
	/home/llorllale/dev/hyperledger/aries-framework-go/pkg/doc/did/doc.go:614 +0x2f
github.com/hyperledger/aries-framework-go/pkg/doc/did.ParseDocument(0xc000168a80, 0x36e, 0x380, 0xc000168a80, 0x36e, 0x380)
	/home/llorllale/dev/hyperledger/aries-framework-go/pkg/doc/did/doc.go:261 +0x15f
github.com/hyperledger/aries-framework-go/pkg/doc/did.TestBaseContext(0xc00014d560)
	/home/llorllale/dev/hyperledger/aries-framework-go/pkg/doc/did/doc_test.go:1453 +0x65
testing.tRunner(0xc00014d560, 0xb0b670)
	/usr/local/go/src/testing/testing.go:991 +0xdc
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1042 +0x357
```

</details>

fqutishat:
closes #2133


llorllale:
As I mentioned in chat:

@fqutishat  @troyronda  @sandrask  can you guys clarify the intended direction to support the multiple core representations of a did document?

If a sequence of bytes is received that contain a JSON-like object with @context and a @base then it is very clearly a JSON-LD representation. It would thus be best suited for a json-ld parser that would then map the parsed result into the core DID model.

llorllale:
This PR does not take into account content-type and is pushing JSON-LD specific stuff into the existing `did.ParseDocument` function

fqutishat:
> This PR does not take into account content-type and is pushing JSON-LD specific stuff into the existing `did.ParseDocument` function

Yes, Can be done in different PR

llorllale:
Created #2281 as a followup to the bigger story of refactoring our representation, production, and consumption of DID documents.

sandrask:
> Created #2281 as a followup to the bigger story of refactoring our representation, production, and consumption of DID documents.

I agree with @llorllale that we should look into JSON-LD and that it can be done later on with different PR #2281

sandrask (approved):

pkg/doc/did/doc.go
```diff
@@ -104,6 +104,7 @@ func Parse(did string) (*DID, error) {
 // Doc DID Document definition.
 type Doc struct {
 	Context              []string
+	Base                 string
```
troyronda, line 107: Does Base need to be exposed? @fqutishat @sandrask 

Instead suggest using a function that returns an absolute URI for a DID URI property.

fqutishat: @troyronda no need to expose it

pkg/doc/did/doc.go
```diff
@@ -714,49 +801,48 @@ func (doc *Doc) JSONBytes() ([]byte, error) {
...
-		Proof:                populateRawProofs(context, doc.Proof),
-		Updated:              doc.Updated,
+		Context: doc.Context, ID: doc.ID, PublicKey: publicKeys,
+		Authentication: auths, AssertionMethod: assertionMethods, CapabilityDelegation: capabilityDelegations,
+		CapabilityInvocation: capabilityInvocations, KeyAgreement: keyAgreements,
+		Service: populateRawServices(doc.Service, doc.processingMeta.base), Created: doc.Created,
+		Proof: populateRawProofs(context, doc.processingMeta.base, doc.Proof), Updated: doc.Updated,
+	}
+
+	if doc.processingMeta.base != "" {
```
troyronda, line 844: @fqutishat DID Core specifies the default transformation behavior for relative URIs into absolute URIs. I would think we would follow this behavior when base isn't present.

See https://www.w3.org/TR/did-core/#relative-did-urls

troyronda: In essence, I would think baseURI isn't empty but rather populated either due to JSON-LD base property or defaulting to the DID URI.

pkg/doc/did/doc.go
```diff
@@ -115,6 +115,12 @@ type Doc struct {
 	Created              *time.Time
 	Updated              *time.Time
 	Proof                []Proof
+	processingMeta       processingMeta
+}
+
+// processingMeta include info how to process the doc.
+type processingMeta struct {
+	base string
```
troyronda, line 123: more generically this is the baseURI.

pkg/doc/did/doc.go
fqutishat: I didn't change the behaviour when base not exist it's the same

pkg/doc/did/doc.go
```diff
@@ -410,6 +440,36 @@ func populateServices(rawServices []map[string]interface{}) []Service {
 	return services
 }
 
+func populateRecipientKeys(recipientKeys []string, base string) []string {
```
kdimak, line 443: Maybe create one method `populateKeys()` and use it for the recipient and routing keys?

pkg/doc/did/doc.go
```diff
@@ -597,26 +670,40 @@ func decodePublicKeyJwk(jwkMap map[string]interface{}, publicKey *PublicKey) err
...
+func (r *rawDoc) ParseContext() ([]string, string) {
 	switch ctx := r.Context.(type) {
 	case []interface{}:
 		var context []string
+
+		var base string
+
 		for _, v := range ctx {
-			context = append(context, v.(string))
+			valueOfString, ok := v.(string)
```
kdimak, line 681: Extract the code into a separate method? Use type switch?

pkg/doc/did/doc.go
```diff
@@ -767,6 +853,21 @@ func (doc *Doc) JSONBytes() ([]byte, error) {
 	return byteDoc, nil
 }
 
+func addDocBase(doc *Doc) []interface{} {
```
kdimak, line 856: Rename to `contextWithBase()`?

pkg/doc/did/schema.go
```diff
@@ -22,8 +22,15 @@ const (
         }
       ],
       "additionalItems": {
-        "type": "string",
-        "format": "uri"
+        "oneOf": [
```
kdimak, line 25: For verifiable credentials, we have the following schema definition for `@context`:
```
"@context": {
  "oneOf": [
    {
      "type": "string",
      "const": "https://www.w3.org/2018/credentials/v1"
    },
    {
      "type": "array",
      "items": [
        {
          "type": "string",
          "const": "https://www.w3.org/2018/credentials/v1"
        }
      ],
      "uniqueItems": true,
      "additionalItems": {
        "oneOf": [
          {
            "type": "object"
          },
          {
            "type": "string"
          }
        ]
      }
    }
  ]
}
```

There is one test case in VC test suite where the context is defined as follows: 
```
"@context": [
  "https://www.w3.org/2018/credentials/examples/v1",
  {
    "image": { "@id": "schema:image", "@type": "@id" }
  }
]
```
So: we should support any objects in `@context`

kdimak (approved):

pkg/doc/did/schema.go
troyronda: Correct - we should be ignoring unknown properties when processing as JSON.

We also need a followup for proper JSON-LD processing as @llorllale mentioned above.

pkg/doc/did/doc.go
llorllale: @troyronda @fqutishat 

> defaulting to the DID URI

[Relative URL dereferencing](https://www.w3.org/TR/did-core/#relative-did-urls) is defined over the DID document model, not over a core representation. After you parse the representation, whether it is in JSON-LD or just JSON, you end up with the DID document model wherein some of the references may still be relative. It is here we must take the base URI as the DID subject.

We may choose to opt out of having relative references after parsing the did document into the abstract model for [reasons] but I just wanted to make that distinction. Hopefully it helps.

pkg/doc/did/doc.go

pkg/doc/did/doc.go

pkg/doc/did/doc.go

pkg/doc/did/schema.go
fqutishat: done

rolsonquadras (approved):

pkg/doc/did/doc.go
troyronda: @llorllale

llorllale: @fqutishat is there a test that shows how the document is serialized to bytes?

llorllale: Regardless, I consider the followup #2281 as the story to track to refactor all this.

llorllale: FYI just tested by adding `validDocWithBase` to `TestJSONConversion`. ParseDocument and JSONBytes are symmetrical and the serialization adds `@base`

fqutishat: Done

llorllale (approved):

pkg/doc/did/doc.go
troyronda: >and the serialization adds `@base`

@llorllale @fqutishat I assume that we aren't adding the base property when serializing peer DID documents right?

llorllale: This PR is adding it.

troyronda: @fqutishat @llorllale  We shouldn't be implicitly adding base property. This property isn't a default property for DID methods.

fqutishat: I'm not adding that

fqutishat: if @base is exist in DID document I will add it

fqutishat: if not exist before parsing I will not add it

fqutishat: this test will confirm TestJSONConversion

fqutishat: https://github.com/hyperledger/aries-framework-go/compare/87947039894ac90b9da3ef3dbc8ca4f2f21020cf..e86b70555480fba40df163f17f38ccc6d80a0525

llorllale: @troyronda 

> @fqutishat @llorllale We shouldn't be implicitly adding base property when we output DID documents. This property isn't a default property for DID methods.

Yes. Followups will be done to rectify this as part of #2281

Issue #2246, opened 2020-10-19
Support for DID Web method


rolsonquadras:
Add support for DID Web method in framework - https://github.com/w3c-ccg/did-method-web. Need to add `web` implementation [here](https://github.com/hyperledger/aries-framework-go/tree/master/pkg/vdr).

rolsonquadras:
@amodkala

PR #2288, opened 2020-11-02
feat: #2246 Support for DID Web Method

amodkala:
Adds ability to resolve did:web dids, addresses issue #2246 

Signed off by: Amod Kala <amod.kala@securekey.com>

### DIFF
--- a/pkg/doc/did/doc.go
+++ b/pkg/doc/did/doc.go
@@ -79,7 +79,7 @@ func (d *DID) String() string {
 func Parse(did string) (*DID, error) {
 	// I could not find a good ABNF parser :(
 	const idchar = `a-zA-Z0-9-_\.`
-	regex := fmt.Sprintf(`^did:[a-z0-9]+:(:+|[:%s]+)*[%s]+$`, idchar, idchar)
+	regex := fmt.Sprintf(`^did:[a-z0-9]+:(:+|[:%s]+)*[%%:%s]+[^:]$`, idchar, idchar)
 
 	r, err := regexp.Compile(regex)
 	if err != nil {
--- a/pkg/vdr/web/creator.go
+++ b/pkg/vdr/web/creator.go
@@ -1,0 +1,19 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package web
+
+import (
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
+)
+
+// Build creates a did:web diddoc (unsupported at the moment).
+func (v *VDR) Build(pubKey *vdr.PubKey, opts ...vdr.DocOpts) (*did.Doc, error) {
+	return nil, fmt.Errorf("error building did:web did doc --> build not supported in http binding vdr")
+}
--- a/pkg/vdr/web/did.go
+++ b/pkg/vdr/web/did.go
@@ -1,0 +1,48 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package web
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+const (
+	defaultPath  = "/.well-known/doc.json"
+	documentPath = "/doc.json"
+)
+
+// parseDIDWeb consumes a did:web identifier and returns the URL location of the did Doc.
+func parseDIDWeb(id string) (string, string, error) {
+	var address, host string
+
+	parsedDID, err := did.Parse(id)
+	if err != nil {
+		return address, host, fmt.Errorf("invalid did, does not conform to generic did standard --> %w", err)
+	}
+
+	pathComponents := strings.Split(parsedDID.MethodSpecificID, ":")
+
+	pathComponents[0], err = url.QueryUnescape(pathComponents[0])
+	if err != nil {
+		return address, host, fmt.Errorf("error parsing did:web did")
+	}
+
+	host = strings.Split(pathComponents[0], ":")[0]
+
+	switch len(pathComponents) {
+	case 1:
+		address = "https://" + pathComponents[0] + defaultPath
+	default:
+		address = "https://" + strings.Join(pathComponents, "/") + documentPath
+	}
+
+	return address, host, nil
+}
--- a/pkg/vdr/web/resolver.go
+++ b/pkg/vdr/web/resolver.go
@@ -1,0 +1,70 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package web
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
+)
+
+var logger = log.New("aries-framework/pkg/vdr/web")
+
+// Read resolves a did:web did.
+func (v *VDR) Read(didID string, opts ...vdr.ResolveOpts) (*did.Doc, error) {
+	// apply resolve opts
+	docOpts := &vdr.ResolveDIDOpts{
+		HTTPClient: &http.Client{},
+	}
+
+	for _, opt := range opts {
+		opt(docOpts)
+	}
+
+	address, host, err := parseDIDWeb(didID)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> could not parse did:web did --> %w", err)
+	}
+
+	resp, err := docOpts.HTTPClient.Get(address)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> http request unsuccessful --> %w", err)
+	}
+
+	for _, i := range resp.TLS.PeerCertificates {
+		err = (*i).VerifyHostname(host)
+		if err != nil {
+			return nil, fmt.Errorf("error resolving did:web did --> identifier does not match TLS host --> %w", err)
+		}
+	}
+
+	defer closeResponseBody(resp.Body)
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> error reading http response body: %s --> %w", body, err)
+	}
+
+	doc, err := did.ParseDocument(body)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> error parsing did doc --> %w", err)
+	}
+
+	return doc, nil
+}
+
+func closeResponseBody(respBody io.Closer) {
+	e := respBody.Close()
+	if e != nil {
+		logger.Errorf("Failed to close response body: %v", e)
+	}
+}
--- a/pkg/vdr/web/resolver_test.go
+++ b/pkg/vdr/web/resolver_test.go
@@ -1,0 +1,132 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package web
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	urlapi "net/url"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	didapi "github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
+)
+
+const (
+	prefix = "did:web:"
+
+	validURL                = "www.example.org"
+	validURLWithPath        = "www.example.org/user/example"
+	validDID                = prefix + validURL
+	validDIDWithPath        = prefix + "www.example.org:user:example"
+	validDIDWithHost        = prefix + "localhost%3A8080"
+	validDIDWithHostAndPath = prefix + "localhost%3A8080:user:example"
+
+	invalidDIDNoMethod = "did:" + validURL
+	invalidDIDNoPrefix = validURL
+
+	validDoc = `{
+  		"@context": ["https://w3id.org/did/v1"],
+  		"id": "did:web:www.example.org"
+	}`
+
+	invalidDoc = `{}`
+)
+
+func TestParseDID(t *testing.T) {
+	t.Run("test parse did success", func(t *testing.T) {
+		address, host, err := parseDIDWeb(validDID)
+		require.NoError(t, err)
+		require.Equal(t, "https://"+validURL+defaultPath, address)
+		require.Equal(t, validURL, host)
+		address, host, err = parseDIDWeb(validDIDWithPath)
+		require.NoError(t, err)
+		require.Equal(t, "https://"+validURLWithPath+documentPath, address)
+		require.Equal(t, validURL, host)
+		address, host, err = parseDIDWeb(validDIDWithHost)
+		require.NoError(t, err)
+		require.Equal(t, "https://localhost:8080/.well-known/doc.json", address)
+		require.Equal(t, "localhost", host)
+		address, host, err = parseDIDWeb(validDIDWithHostAndPath)
+		require.NoError(t, err)
+		require.Equal(t, "https://localhost:8080/user/example/doc.json", address)
+		require.Equal(t, "localhost", host)
+	})
+
+	t.Run("test parse did failure", func(t *testing.T) {
+		v := New()
+		doc, err := v.Read(invalidDIDNoMethod)
+		require.Error(t, err)
+		require.Nil(t, doc)
+		require.Contains(t, err.Error(), "does not conform to generic did standard")
+		doc, err = v.Read(invalidDIDNoPrefix)
+		require.Error(t, err)
+		require.Nil(t, doc)
+		require.Contains(t, err.Error(), "does not conform to generic did standard")
+	})
+}
+
+func TestResolveDID(t *testing.T) {
+	t.Run("test resolve did with request failure", func(t *testing.T) {
+		s := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, err := w.Write([]byte(invalidDoc))
+			require.NoError(t, err)
+		}))
+		defer s.Close()
+		did := fmt.Sprintf("did:web:%s", urlapi.QueryEscape(strings.TrimPrefix(s.URL, "https://")))
+		v := New()
+		doc, err := v.Read(did)
+		require.Nil(t, doc)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "http request unsuccessful")
+	})
+	t.Run("test resolve did with invalid doc format failure", func(t *testing.T) {
+		s := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, err := w.Write([]byte(invalidDoc))
+			require.NoError(t, err)
+		}))
+		defer s.Close()
+		did := fmt.Sprintf("did:web:%s", urlapi.QueryEscape(strings.TrimPrefix(s.URL, "https://")))
+		v := New()
+		doc, err := v.Read(did, vdr.WithHTTPClient(s.Client()))
+		require.Nil(t, doc)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "error parsing did doc")
+	})
+	t.Run("test resolve did success", func(t *testing.T) {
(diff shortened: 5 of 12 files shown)


pkg/vdr/web/did.go
```diff
@@ -0,0 +1,47 @@
...
+	var urlString string
+
+	// parses did to make sure it conforms to generic did standard
+	_, err := did.Parse(id)
+	if err != nil {
+		return urlString, fmt.Errorf("invalid did, does not conform to generic did standard --> %w", err)
+	}
+
+	parts := strings.Split(id, ":")
+	if parts[0] != "did" || parts[1] != "web" {
```
llorllale, line 32: You can use the result of `did.Parse()`, which is a [`did.DID`](https://github.com/hyperledger/aries-framework-go/blob/e86b70555480fba40df163f17f38ccc6d80a0525/pkg/doc/did/doc.go#L65-L70):

```go
// DID is parsed according to the generic syntax: https://w3c.github.io/did-core/#generic-did-syntax
type DID struct {
	Scheme           string // Scheme is always "did"
	Method           string // Method is the specific DID methods
	MethodSpecificID string // MethodSpecificID is the unique ID computed or assigned by the DID method
}
```

pkg/vdr/web/did_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+const (
+	prefix = "did:web:"
+
+	validURL         = "www.securekey.com"
```
llorllale, line 18: please remove this and use a fictitious url like `www.example.org`

pkg/vdr/web/did_test.go
```diff
@@ -0,0 +1,55 @@
...
+	validURLWithPath = "www.securekey.com/user/amod"
```
llorllale, line 19: please remove this and use a fictitious url like `www.example.org`

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,43 @@
...
+	// TODO actually use the ResolveOpts
+
+	// parse doc url from did
+	url, err := parseDIDWeb(didID)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> could not resolve did:web did --> %w", err)
+	}
+
+	// send get request to retrieve did doc
+	resp, err := http.Get(url)
```
llorllale, line 29: We need to set the TLS config on the client for this outbound call.

llorllale: [Example](https://github.com/hyperledger/aries-framework-go/blob/4f10ba10077a9b0ef2e7a9531682549609cda2a4/test/bdd/pkg/util/util.go#L25-L31)

llorllale: Would need to add a new `vdr.ResolveOpts` option: https://github.com/hyperledger/aries-framework-go/blob/ca5e0650014e48021c33f0c96fe23a11123967cf/pkg/framework/aries/api/vdr/vdr.go#L65-L64

amodkala: Do you mean I should send the request from a https enabled client? I can add that but is there a preferred package or method that we use that I should use as a guideline?

llorllale (changes requested):
See my comments about the test URLs

amodkala:
addressed changes requested to test urls and tidied up the did parsing code, working on making the http client more robust now as per your recommendations

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
llorllale, line 1: please exclude IDE-specific files from the commit. Feel free to add `.vscode/` to the root `.gitignore` file.

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		// nolint: errcheck
+		defer resp.Body.Close()
+	default:
+		resp, err = docOpts.HTTPClient.Do(req)
+		if err != nil {
+			return nil, fmt.Errorf("error resolving did:web did --> error with custom get request to %s --> %w", url, err)
+		}
+		// nolint: errcheck
+		defer resp.Body.Close()
```
sandrask, line 54: You can call this line outside of switch statement (it is common in both cases) so there will be only one //nolint: errcheck. We usually don't disable error checks  and you have three of them in this function.

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,68 @@
...
+	defaultClient := &http.Client{}
+	// parse doc url from did
+	url, err := parseDIDWeb(didID)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> could not parse did:web did --> %w", err)
+	}
+
+	// send get request to retrieve did doc
+	// nolint: errcheck
+	req, _ := http.NewRequest("GET", url, nil)
```
sandrask, line 36: You should check for error here

pkg/vdr/web/did.go
```diff
@@ -0,0 +1,45 @@
...
+func parseDIDWeb(id string) (string, error) {
+	var urlString string
+
+	// parses did to make sure it conforms to generic did standard
+	parsedDID, err := did.Parse(id)
+	if err != nil {
+		return urlString, fmt.Errorf("invalid did, does not conform to generic did standard --> %w", err)
+	}
+
+	if parsedDID.Method != "web" {
```
sandrask, line 31: How is this even possible if your vdr is meant for "web" method?  It will not get invoked unless it is "web".

pkg/vdr/web/resolver_test.go
```diff
@@ -0,0 +1,89 @@
...
+	invalidDIDNoMethod = "did:" + validURL
+	invalidDIDNoPrefix = validURL
+)
+
+func TestParseDID(t *testing.T) {
+	t.Run("test parse did success", func(t *testing.T) {
+		url, err := parseDIDWeb(validDID)
+		require.NoError(t, err)
+		require.Equal(t, "https://"+validURL+defaultPath, url)
+		fmt.Println(url)
```
sandrask, line 37: Remove Println - other places too

pkg/vdr/web/resolver_test.go
```diff
@@ -0,0 +1,89 @@
...
+		url, err = parseDIDWeb(invalidDIDNoPrefix)
+		require.Error(t, err)
+		require.Empty(t, url)
+		require.Contains(t, err.Error(), "does not conform to generic did standard")
+	})
+}
+
+func TestResolveDID(t *testing.T) {
+	t.Run("test resolve did with invalid did failure", func(t *testing.T) {
+		v, err := New()
```
sandrask, line 62: this is not a valid test case - see my comment above

pkg/vdr/web/did.go
amodkala: I've made the changes you mentioned, I guess I was still a little unclear on the role of vdr in aries so I'll read into it more and keep that in mind for future changes : )

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+// Read resolves a did:web did.
+func (v *VDR) Read(didID string, opts ...vdr.ResolveOpts) (*did.Doc, error) {
+	// apply resolve opts
+	docOpts := &vdr.ResolveDIDOpts{}
+
+	for _, opt := range opts {
+		opt(docOpts)
+	}
+
+	var customClient bool
```
sandrask, line 27: If you choose string instead of bool here then you don't need second switch for error message in sendResolveRequest() -  you can use clientType directly in your error message.

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+
+	var client *http.Client
+
+	switch docOpts.HTTPClient {
+	case nil:
+		client = &http.Client{}
+		customClient = false
```
sandrask, line 34: clientType = "default"

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+	default:
+		client = docOpts.HTTPClient
+		customClient = true
```
sandrask, line 37: clientType = "custom"

pkg/vdr/web/resolver.go
amodkala: good call😁

pkg/vdr/web/creator.go
```diff
@@ -0,0 +1,19 @@
...
+package web
+
+import (
+	"fmt"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
+)
+
+// Build creates a did:web diddoc.
```
llorllale, line 16: Please update godoc indicating this VDR does not support creating DIDs.

pkg/vdr/web/resolver.go
llorllale: I would not bother with this.

I would initialize `docOpts` with `&http.Client{}` and then let the user override if they want. There is no requirement to distinguish the client used in the logs.

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, fmt.Errorf("error resolving did:web did --> could not parse did:web did --> %w", err)
+	}
+
+	return sendResolveRequest(client, url, customClient)
+}
+
+func sendResolveRequest(client *http.Client, url string, customClient bool) (*did.Doc, error) {
+	resp, err := client.Get(url)
+	if err != nil {
+		switch customClient {
```
llorllale, line 51: this code is unnecessarily complicated - it's enough to log error without having to distinguish between default and custom clients.

llorllale: Please include the url in the log message.

pkg/vdr/web/resolver_test.go
```diff
@@ -0,0 +1,74 @@
...
+		doc, err := v.Read("did:web:www.example123.org")
+		require.Nil(t, doc)
+		require.Contains(t, err.Error(), "http request with default client unsuccessful")
+	})
+	t.Run("test resolve did with custom client failure", func(t *testing.T) {
+		v, err := New()
+		require.NoError(t, err)
+		doc, err := v.Read("did:web:www.example123.org", vdr.WithHTTPClient(&http.Client{}))
+		require.Nil(t, doc)
+		require.Contains(t, err.Error(), "http request with custom client unsuccessful")
```
llorllale, line 65: Please check `require.Error(t, err)` before checking the error message. We avoid a potential panic this way.

pkg/vdr/web/resolver_test.go
```diff
@@ -0,0 +1,74 @@
...
+		require.Empty(t, url)
+		require.Contains(t, err.Error(), "does not conform to generic did standard")
+		url, err = parseDIDWeb(invalidDIDNoPrefix)
+		require.Error(t, err)
+		require.Empty(t, url)
+		require.Contains(t, err.Error(), "does not conform to generic did standard")
+	})
+}
+
+func TestResolveDID(t *testing.T) {
```
llorllale, line 52: There are no positive tests, only negative tests. We must have positive tests. Do you want to add those in this PR or in a followup?

For positive tests, consider using `httptest.Server`.

pkg/vdr/web/resolver.go
sandrask: I agree that it is not necessary to distinguish between default and custom client, but it looks like he really wanted that.

pkg/vdr/web/resolver_test.go
```diff
@@ -0,0 +1,112 @@
...
+	t.Run("test resolve did with path success", func(t *testing.T) {
+		s := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, err := w.Write([]byte(validDoc))
+			require.NoError(t, err)
+		}))
+		defer s.Close()
+		did := fmt.Sprintf("did:web:%s:user:johnsmith", strings.TrimPrefix(s.URL, "https://"))
+		v := New()
+		_, err := v.Read(did, vdr.WithHTTPClient(s.Client()))
+		require.Nil(t, err)
```
sandrask, line 110: you should check document as well, not just error

pkg/vdr/web/did.go
```diff
@@ -0,0 +1,45 @@
...
+
+	pathComponents := strings.Split(parsedDID.MethodSpecificID, ":")
+
+	if len(pathComponents) == 1 {
+		urlString = "https://" + pathComponents[0] + defaultPath
+		return urlString, nil
+	}
+
+	switch pathComponents[0] {
+	case "localhost", "127.0.0.1":
```
sandrask, line 38: Is this code added just for testing purposes? If so you should elaborate why because we usually don't add extra code to support testing.
Is this valid scenario for "web" VDR?

amodkala: yes it's just for testing, web dids shouldn't be originating from any site with a host:port url ~really~ as any organization trusted enough to have did:web would likely have a proper url

amodkala: from the spec: "The method specific identifier MUST match the common name used in the SSL/TLS certificate, and it MUST NOT include IP addresses or port numbers. Directories and subdirectories MAY optionally be included, delimited by colons rather than slashes. " https://w3c-ccg.github.io/did-method-web/#method-specific-identifier

pkg/vdr/web/resolver_test.go
amodkala: `doc, err := v.Read(didID, vdr.WithHTTPClient(s.Client()))
		require.Nil(t, err)
		expectedDoc, err := did.ParseDocument([]byte(validDoc))
		require.Nil(t, err)
		require.Equal(t, expectedDoc, doc)`
does this look good? (given validDoc is what's being returned by server)

pkg/vdr/web/did.go
amodkala: ``
	switch pathComponents[0] {
	// This is for testing purposes ONLY and should not be used in production
	case "localhost", "127.0.0.1":
		urlString = "https://" + strings.Join(pathComponents[0:2], ":") + "/" + strings.Join(pathComponents[2:], "/")

		log.New("aries-framework/pkg/vdr/web").Warnf("using an ip address or port number does not conform to spec")
``

would something like this be acceptable ? @sandrask

llorllale: This is not acceptable.

Yes, `127.0.0.1` is not allowed by the spec, but `localhost` is.

Also, ports will be supported. There is a proposed solution to base64url-encode the host portion: https://github.com/w3c-ccg/did-method-web/issues/7#issuecomment-623485232. At least one implementation has adopted the proposal: https://github.com/ProjectProvenance/proof-points/pull/42/files.

@amodkala let's adopt the proposal - assume ports are allowed and the host:port portion is base64url-encoded.

llorllale: @amodkala my apologies - somehow I mixed up url-encoding with base64url-encoding this morning. The proposal for the `did:web` ID to have the `host:port` portion and all path components url-encoded. See the issue I linked to above for more details.

amodkala:
apologies, forgot to push these changes yesterday😬

amodkala:
oops broke a unit test, one second

amodkala:
fix on the way -_-

amodkala:
rebased, ready for review !! @llorllale @sandrask

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> error reading http response body: %s --> %w", body, err)
+	}
+
+	defer func() {
+		if resp.Body.Close() != nil {
+			panic("couldn't close http response body")
```
sandrask, line 46: you should not panic - just log warning with err message from body.Close()

amodkala: should I use the logger we have built in to aries instead of a plain fmt.Println for example?

sandrask: Check this out: https://github.com/hyperledger/aries-framework-go/blob/d7623b88ae7e87e4a63f5a0e3662f0d2314a0528/pkg/vdr/httpbinding/vdr.go#L108

amodkala:
@llorllale @sandrask

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,63 @@
...
+	doc, err := did.ParseDocument(body)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> error parsing did doc --> %w", err)
+	}
+
+	return doc, nil
+}
+
+func closeResponseBody(respBody io.Closer) {
+	logger := log.New("aries-framework/pkg/vdr/web")
```
sandrask, line 57: I would still declare logger globally so next time you need it - it will be ready to use.

sandrask (approved):

pkg/doc/did/doc.go
```diff
@@ -79,7 +79,7 @@ func (d *DID) String() string {
 func Parse(did string) (*DID, error) {
 	// I could not find a good ABNF parser :(
 	const idchar = `a-zA-Z0-9-_\.`
-	regex := fmt.Sprintf(`^did:[a-z0-9]+:(:+|[:%s]+)*[%s]+$`, idchar, idchar)
+	regex := fmt.Sprintf(`^did:[a-z0-9]+:(:+|[:%s]+)*[%%:%s]+[^:]$`, idchar, idchar)
```
llorllale, line 82: what did you fix here?

llorllale: was it due to the url-encoding?

pkg/vdr/web/did.go
llorllale: Just realized url-encoding the path components makes it non-compliant with the generic did-core syntax

pkg/vdr/web/resolver.go
amodkala: ah okay, I hadn't seen this line in the lint file

  exclude:
    # Allow package logger variables (for now)
    - \`logger\` is a global variable

pkg/doc/did/doc.go
amodkala: yes that was precisely it, however as you mentioned in https://github.com/w3c-ccg/did-method-web/issues/7#issuecomment-730434663 this isn't exactly valid for the time being, looking into workarounds as we speak.

pkg/vdr/web/did.go
llorllale: Followup: #2336

pkg/doc/did/doc.go
llorllale: Followup: #2336 

We should fix `did.Parse()`

pkg/vdr/web/did.go
```diff
@@ -0,0 +1,46 @@
...
+func parseDIDWeb(id string) (string, error) {
+	var urlString string
+
+	parsedDID, err := did.Parse(id)
+	if err != nil {
+		return urlString, fmt.Errorf("invalid did, does not conform to generic did standard --> %w", err)
+	}
+
+	pathComponents := strings.Split(parsedDID.MethodSpecificID, ":")
+	pathComponents[0], err = url.QueryUnescape(pathComponents[0])
```
llorllale, line 32: We should actually do this in `did.Parse()` (#2336).

pkg/vdr/web/resolver.go
```diff
@@ -0,0 +1,63 @@
...
+	for _, opt := range opts {
+		opt(docOpts)
+	}
+
+	url, err := parseDIDWeb(didID)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving did:web did --> could not parse did:web did --> %w", err)
+	}
+
+	resp, err := docOpts.HTTPClient.Get(url)
```
llorllale, line 36: If the http client is configured properly, it will validate the certificate against the configured root authorities. This is good.

Is this also ensuring this part of the `did:web` spec (important):

> The method specific identifier MUST match the common name used in the SSL/TLS certificate, and it MUST NOT include IP addresses or port numbers.

So for instance if you resolve `did:web:acme.org` and the server's certificate's common name is "Faber College" - does it fail or not? This might be a followup PR - can you confirm?

amodkala: After some digging through the http and tls packages I don't think we explicitly check for that at the moment. Would this just require a comparison between the two (common cert name and method specific identifier)?

llorllale: @amodkala

> Would this just require a comparison between the two (common cert name and method specific identifier)?

Roughly, yes.

llorllale: @amodkala are we updating this PR any time soon? Or would we rather add this verification in a followup?

pkg/vdr/web/resolver_test.go
```diff
@@ -0,0 +1,147 @@
...
+		defer s.Close()
+		did := fmt.Sprintf("did:web:%s", urlapi.QueryEscape(strings.TrimPrefix(s.URL, "https://")))
+		v := New()
+		doc, err := v.Read(did, vdr.WithHTTPClient(s.Client()))
+		require.Nil(t, doc)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "error parsing did doc")
+	})
+	/*
+		t.Run("test resolve did with certificate mismatch failure", func(t *testing.T) {
```
llorllale, line 105: @amodkala can we enable this test?

llorllale (approved):

PR #2310, opened 2020-11-12
feat: EDV REST provider iterator method

### LINKED ISSUE #2309: EDV REST provider Iterator method
It's currently unimplemented.

Implementation steps:
1. When storing a document add an encrypted index to it based on the store name
2. When calling the Iterator method, do a query for all documents that have that store name index.
3. Decrypt the encrypted indices on all the documents that were returned.
4. Now you can find which documents have a friendly key that resides within the iterator range.

DRK3:
Also fixed the formatStore's Iterator method which before this change wasn't unformatting the values received from the underlying store's iterator, which was not the intended behaviour.

closes #2309


pkg/storage/formattedstore/formattedstore.go
```diff
@@ -4,10 +4,13 @@ Copyright SecureKey Technologies Inc. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
```
DRK3, line 4: Moved this file here to avoid a circular dependency with the `mem` package.

pkg/storage/edv/encryptedformatter.go
```diff
@@ -58,9 +58,10 @@ func NewEncryptedFormatter(jweEncrypter jose.Encrypter, jweDecrypter jose.Decryp
 }
 
 // Format encrypts v into encrypted document format.
-func (f *EncryptedFormatter) Format(v []byte) ([]byte, error) {
+func (f *EncryptedFormatter) Format(k string, v []byte) ([]byte, error) {
```
llorllale, line 61: does it make sense to rename this back to `FormatPair`?

DRK3: Yes, I think it does. Just renamed it.

pkg/storage/edv/restprovider.go
```diff
@@ -213,70 +233,139 @@ func (r *restStore) Get(k string) ([]byte, error) {
...
+	storeIndexValueMAC, err := r.macCrypto.ComputeMAC(r.name)
 	if err != nil {
-		return nil, fmt.Errorf(failComputeBase64EncodedIndexValueMAC, err)
+		return nil, fmt.Errorf(failComputeMACStoreIndexValue, err)
 	}
 
-	indexedAttribute := IndexedAttribute{
-		Name:   r.indexKeyMACBase64Encoded,
-		Value:  indexValueMACBase64Encoded,
+	base64.URLEncoding.EncodeToString([]byte(storeIndexValueMAC))
```
llorllale, line 307: what about the return value?

DRK3: Good catch. Fixed!

fqutishat (approved):

pkg/storage/formattedstore/formattedstore.go
```diff
@@ -112,8 +115,29 @@ func (s *formatStore) Get(k string) ([]byte, error) {
...
+
+	errFormattedIterator := formattedIterator.Error()
+	if errFormattedIterator != nil {
+		return mem.NewMemIterator(nil,
+			fmt.Errorf("failed to get iterator from underlying store: %w", errFormattedIterator))
+	}
+
+	var batch [][]string
+
+	for formattedIterator.Next() {
```
bstasyszyn, line 129: This will load all values into memory. Would it be better for the MemIterator to hold on the the formattedIterator and retrieve the next value when memIterator.Next() is invoked, and then release the formattedIterator when MemIterator.Release() is called.

DRK3: The code has been updated since you originally commented. With the new way it works, the values must be deformatted in order to apply the startKey-endKey filter, so I don't think it's possible to avoid loading all the values in memory... but let me know if you think it's still possible.

pkg/storage/formattedstore/formattedstore.go
```diff
@@ -0,0 +1,178 @@
...
+				continue
+			}
+
+			if key >= startKey && key < endKey {
+				batch = append(batch, []string{key, string(value)})
+			}
+		}
+	}
+
+	formattedIterator.Release()
```
DRK3, line 166: I would appreciate an extra set of eyes on this filter logic. My goal was to ensure that the filtering logic was consistent with the mem store version, however the major difference here is that the way I've implemented it doesn't require sorting all the keys. I think the logic is kind of messy, though. Is there a better way to do it?

@fqutishat, @soluchok , I saw you were both involved in Iterator method work in the past, can you take a look?

Also note that I added a bunch of tests for this, but I used the same test style as the existing mem store test so I could compare my results to ensure consistency.

DRK3:
@llorllale This PR has been updated per our discussions with the team earlier.

Some changes to note:

- The multiple, redundant decryption is gone. The REST Provider once again no longer has any encryption/decryption ability.

- The Formatter's `ParseValue` method has been changed to `ParsePair` to accommodate the new design.

- Now that the FormatProvider can do that filter logic, it's possible to have redundant filtering if the underlying provider's Iterator() method also does it. I added a boolean to the FormatProvider to allow the filtering to be optional, so if the caller knows that the underlying provider will do the filtering then you don't have to waste cycles doing a redundant filter. I think that with the larger storage rework we were discussing I should be able to avoid having this weird boolean, so I left a TODO. Also, you'll see a bunch of tests I added to check that behaviour across an underlying mem store and REST provider. I wanted the REST provider tests specifically to make sure the behaviour of the filtering is consistent.

- I moved the mock EDV server under a new package in the mock folder since I use it across two different testing files now.

- I moved the EDV models to a separate folder nested under the EDV folder to avoid a circular dependency due to the above change.

- I added a few more iterator tests to the mem iterator test file in order to help me better test consistency with the FormatProvider's Iterator filtering.

pkg/storage/edv/restprovider.go
```diff
@@ -213,70 +228,171 @@ func (r *restStore) Get(k string) ([]byte, error) {
...
-func (r *restStore) Iterator(string, string) storage.StoreIterator {
-	return &restStoreIterator{}
+// Iterator returns all documents within the store. It does not support start and end key filtering.
+func (r *restStore) Iterator(_, _ string) storage.StoreIterator {
+	allDocumentLocations, err := r.getAllDocumentLocations()
+	if err != nil {
+		return &restStoreIterator{err: fmt.Errorf(failGetAllDocumentLocations, err)}
+	}
+
+	allDocuments, err := r.getAllDocuments(allDocumentLocations)
```
bstasyszyn, line 238: I don't know what the typical number of documents this will return. If it's a small number then it's fine to load all documents into memory. If it turns out to be in the order of hundreds/thousands then I would rework this logic to possibly load all of the document IDs into memory and then retrieve each document on demand during iteration.

bstasyszyn (approved):

pkg/storage/edv/restprovider.go
fqutishat: @bstasyszyn this solution will change @DRK3 create task for refactor the interface

DRK3: @bstasyszyn @fqutishat Here's the task: #2315 

I also added a comment to address @bstasyszyn's comment in that issue.

PR #2314, opened 2020-11-13
feat: remoteKMS client implementation

### LINKED ISSUE #2313: create remoteKMS client implementation
will be followed up with a remoteCrypto implementation..

Baha-sk:
closes #2313

### DIFF
--- a/pkg/crypto/tinkcrypto/primitive/composite/api/composite_encrypt.go
+++ b/pkg/crypto/tinkcrypto/primitive/composite/api/composite_encrypt.go
@@ -9,11 +9,12 @@ package api
 // package api provides the composite primitive interfaces. These will be mainly used as the crypto primitives for
 // building protected JWE messages.
 
-// CompositeEncrypt will encrypt a `plaintext` using AEAD primitive and ECDH-ES key wrapping by recipient
-// It returns the resulting serialized JWE []byte. This type is used mainly for repudiation requests where the sender
-// identity remains unknown to the recipient in a serialized EncryptedData envelope (used mainly to build JWE messages).
+// CompositeEncrypt will encrypt a `plaintext` using AEAD primitive (with ECDH-ES cek key wrapping by recipient executed
+// externally). It returns the resulting serialized JWE []byte. This type is used mainly for repudiation requests where
+// the sender identity remains unknown to the recipient in a serialized EncryptedData envelope (used mainly to build JWE
+// messages).
 type CompositeEncrypt interface {
-	// Encrypt operation: encrypts plaintext with aad represented as the list of recipient's corresponding public keys
-	// Returns resulting EncryptedData wrapping ciphertext and the recipients protected keys or error if failed.
+	// Encrypt operation: encrypts plaintext with aad for a given cek.
+	// Returns resulting EncryptedData wrapping ciphertext content or error if failed.
 	Encrypt(plainText, aad []byte) ([]byte, error)
 }
--- a/pkg/kms/webkms/remotekms.go
+++ b/pkg/kms/webkms/remotekms.go
@@ -1,0 +1,302 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package webkms
+
+import (
+	"bytes"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
+)
+
+const (
+	// Namespace of the remoteKMS local config storage.
+	Namespace = "remotekmsdb"
+
+	// KeystoreURLField representing the user's keystore URL field in the remoteKMS local storage.
+	KeystoreURLField = "keystoreurl"
+
+	// ControllerField representing the user's Controller field in the remoteKMS local storage.
+	ControllerField = "controller"
+
+	// SecretField representing the user's secret field in the remoteKMS local storage.
+	SecretField = "secret"
+
+	createKeystoreEndpoint = "{serverEndpoint}/kms/keystores"
+
+	// ContentType is remoteKMS http content-type.
+	ContentType = "application/json"
+
+	// LocationHeader is remoteKMS http header set by the key server (usually to identify a keystore or key url).
+	LocationHeader = "Location"
+
+	// KeyBytesHeader is remoteKMS http header set by the key server (to contain an base64URL encoded public key).
+	KeyBytesHeader = "KeyBytes"
+)
+
+var logger = log.New("aries-framework/kms/webkms")
+
+type createKeystoreReq struct {
+	Controller string `json:"controller,omitempty"`
+	Passphrase string `json:"passphrase,omitempty"`
+}
+
+type createKeyReq struct {
+	KeyType    string `json:"keytype,omitempty"`
+	Passphrase string `json:"passphrase,omitempty"`
+}
+
+type exportKeyReq struct {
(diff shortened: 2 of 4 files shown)


pkg/crypto/tinkcrypto/primitive/composite/api/composite_encrypt.go
```diff
@@ -9,11 +9,12 @@ package api
...
-// It returns the resulting serialized JWE []byte. This type is used mainly for repudiation requests where the sender
-// identity remains unknown to the recipient in a serialized EncryptedData envelope (used mainly to build JWE messages).
+// CompositeEncrypt will encrypt a `plaintext` using AEAD primitive (with ECDH-ES cek key wrapping by recipient executed
+// externally). It returns the resulting serialized JWE []byte. This type is used mainly for repudiation requests where
+// the sender identity remains unknown to the recipient in a serialized EncryptedData envelope (used mainly to build JWE
+// messages).
 type CompositeEncrypt interface {
-	// Encrypt operation: encrypts plaintext with aad represented as the list of recipient's corresponding public keys
-	// Returns resulting EncryptedData wrapping ciphertext and the recipients protected keys or error if failed.
+	// Encrypt operation: encrypts plaintext with aad for a givent cek.
```
llorllale, line 17: nit: typo

pkg/kms/remotekms/remotekms.go
```diff
@@ -0,0 +1,266 @@
...
+//   - keyserverURL representing the key server url
+//   - controller is the identifier of the keystore owner (usually a DID key ID)
+//   - secret represents the configStore secret for locking/unlocking the keystore
+// Returns:
+//  - keystore URL (if successful)
+//  - error (if error encountered)
+func CreateKeyStore(store storage.Store, httpClient *http.Client, keyserverURL, controller,
+	secret string) (string, error) {
+	destination := strings.ReplaceAll(createKeystoreEndpoint, "{serverEndpoint}", keyserverURL)
+	req := []byte(fmt.Sprintf(createKeystoreReq, controller, secret))
```
rolsonquadras, line 108: Can we use Go struct to define the req/response model and use json marshal ? It would be hard to maintain with string formatting to create the json requests.

pkg/kms/remotekms/remotekms.go
```diff
@@ -0,0 +1,266 @@
...
+//   - configStore is where the keystoreURL, controller and secret will be stored
+//   - httpClient used to POST the request
+//   - keyserverURL representing the key server url
+//   - controller is the identifier of the keystore owner (usually a DID key ID)
+//   - secret represents the configStore secret for locking/unlocking the keystore
+// Returns:
+//  - keystore URL (if successful)
+//  - error (if error encountered)
+func CreateKeyStore(store storage.Store, httpClient *http.Client, keyserverURL, controller,
+	secret string) (string, error) {
```
llorllale, line 106: nit: secret is usually modeled as a `[]byte`. Downstream projects will have the user's device create the secret, not the user themselves.

pkg/crypto/tinkcrypto/primitive/composite/api/composite_encrypt.go
Baha-sk: will update

pkg/kms/remotekms/remotekms.go
Baha-sk: sure.. the same goes for the key server @aholovko ;-)

I will convert createKeyStoreReq to be json objects and pass them directly

pkg/kms/remotekms/remotekms.go
Baha-sk: I'll convert to []byte

pkg/crypto/tinkcrypto/primitive/composite/api/composite_encrypt.go
Baha-sk: updated

pkg/kms/remotekms/remotekms.go
Baha-sk: done

pkg/kms/remotekms/remotekms.go
Baha-sk: done, the secret stored in the config store is now base64 url encoded

pkg/kms/webkms/remotekms.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	return kid, pubKey, nil
+}
+
+// PubKeyBytesToHandle is not implemented in remoteKMS.
+func (r *RemoteKMS) PubKeyBytesToHandle(pubKey []byte, kt kms.KeyType) (interface{}, error) {
+	return nil, errors.New("function PubKeyBytesToHandle is not implemented in remoteKMS")
```
llorllale, line 272: `PubKeyBytesToHandle` is being used by the zcap primitives: https://github.com/trustbloc/edge-core/blob/8631b9765a2ef8aea7d2280e2c130ef820d2c78a/pkg/zcapld/middleware.go#L112-L111

The use case is dereferencing a `did:key` URL as a verificationMethod and using `crypto.Crypto` to verify, and it needs a key handle.

The `did:key` URL is not equal to the KMS' internal KID.

pkg/kms/webkms/remotekms.go
```diff
@@ -0,0 +1,297 @@
...
+// Returns:
+//  - marshalled public key []byte
+//  - error if it fails to export the public key bytes
+func (r *RemoteKMS) ExportPubKeyBytes(keyID string) ([]byte, error) {
+	keyURL, err := r.buildKIDURL(keyID)
+	if err != nil {
+		return nil, err
+	}
+
+	destination := keyURL + "/keys/export"
```
aholovko, line 222: endpoint is going to be "/keys/" + keyID + "/export"

pkg/kms/webkms/remotekms.go
Baha-sk: Implementation of this function should be done in a separate pr, it needs more thinking, ie: how to resolve `did:key` URL to key server URL..

also, do we want to tie remoteKMS to Tink api, I tried to avoid this coupling as much as possible.. to be discussed

pkg/kms/webkms/remotekms.go
Baha-sk: will update..

pkg/kms/webkms/remotekms.go
aholovko: I assume that we are going to use `ExportPubKeyBytes` and then "manually" build `did:key` from the public key bytes

aholovko: probably `ExportPubKeyBytes` can be implemented as matching pubkey to the one on remote KMS and then returning handle of that key that in case of remote KMS is key URL

Baha-sk: `ExportPubKeyBytes` should just return raw public key bytes, it should not manage converting its format.

Also if you notice, `Create()` function doesn't return the Tink `*keyset.Handle` in the second return statement, it returns the key URL as defined by the key server instead.

Similarly, IMHO, to create a `*keyset.Handle` should be managed outside of remoteKMS

pkg/kms/webkms/remotekms.go
Baha-sk: done

pkg/kms/webkms/remotekms.go
Baha-sk: >probably ExportPubKeyBytes can be implemented as matching pubkey to the one on remote KMS and then returning handle of that key that in case of remote KMS is key URL

@aholovko correct, and the crypto service would be a `remoteCrypto` instance that can take the key URL as the key handle argument in `crypto.Verify()`

aholovko (approved):

fqutishat (approved):

rolsonquadras (approved):

llorllale (approved):

Issue #2307, opened 2020-11-10
Create RemoteKMS/RemoteCrypto implementations


Baha-sk:
to support a remote key server managing keys and executing crypto primitives for an agent with limited crypto capabilities, we need to create remote implementations of the kms.Keymanager and crypto.Crypto interfaces.

These remote implementations will include an http client that can do REST calls to a remote key server (following WebKMS spec)

PR #2322, opened 2020-11-17
feat: add HeadersOpt optional function to remoteKMS New()

Baha-sk:
This change introduced a new optional function passed to remotekms.New()
This new option allows calling a function to potentially call an
authorization server and setup auth info as http headers. This function
should be created by the user of the framework. The absence of this option
means the remoteKMS is not setting any authorization info in the headers.

This change also removed 'secret' and 'controller' http headers as these
should not be dictated by the framework.

It also includes some minor refactoring like placing CreateKeystore()
above New() as a keystore is required before creating a new remoteKMS instance.

part of #2307

### DIFF
--- a/pkg/kms/webkms/headers_opts.go
+++ b/pkg/kms/webkms/headers_opts.go
@@ -1,0 +1,34 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package webkms
+
+import "net/http"
+
+// addHeaders function supports adding custom http headers.
+type addHeaders func(req *http.Request) (*http.Header, error)
+
+type headersOpts struct {
+	headersFunc addHeaders
+}
+
+// NewOpt creates a new empty additional http headers option.
+// Not to be used directly. It's intended for implementations of remoteKMS.
+// Use WithHeaders() option function below instead.
+func NewOpt() *headersOpts { // nolint // unexported type doesn't need to be used outside of remotekms package
+	return &headersOpts{}
+}
+
+// HeadersOpt are the remoteKMS headers option.
+type HeadersOpt func(opts *headersOpts)
+
+// WithHeaders option is for setting additional http request headers (since it's a function, it can call a remote
+// authorization server to fetch the necessary info needed in these headers).
+func WithHeaders(addHeadersFunc addHeaders) HeadersOpt {
+	return func(opts *headersOpts) {
+		opts.headersFunc = addHeadersFunc
+	}
+}
--- a/pkg/kms/webkms/remotekms.go
+++ b/pkg/kms/webkms/remotekms.go
@@ -9,30 +9,19 @@ package webkms
 import (
 	"bytes"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/kms"
-	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
 const (
-	// Namespace of the remoteKMS local config storage.
-	Namespace = "remotekmsdb"
-
-	// KeystoreURLField representing the user's keystore URL field in the remoteKMS local storage.
-	KeystoreURLField = "keystoreurl"
-
-	// ControllerField representing the user's Controller field in the remoteKMS local storage.
-	ControllerField = "controller"
-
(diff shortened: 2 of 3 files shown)


rolsonquadras (approved):

pkg/kms/webkms/remotekms.go
```diff
@@ -41,64 +43,46 @@ const (
 	// LocationHeader is remoteKMS http header set by the key server (usually to identify a keystore or key url).
 	LocationHeader = "Location"
 
-	// KeyBytesHeader is remoteKMS http header set by the key server (to contain an base64URL encoded public key).
-	KeyBytesHeader = "KeyBytes"
+	// SecretHeader is the keystore secret (share) in the key server.
+	SecretHeader = "Secret"
+
+	// InvokerHeader is remoteKMS http request invoker.
+	InvokerHeader = "Invoker"
```
llorllale, line 50: This header is not needed. The invoker is picked by the service from the `keyId` in `Signature` header.

llorllale: The invoker is always the sender of the HTTP message, who authenticates the message with a signature computed using the `keyID`.

aholovko: I am trying to understand why at all we should specify headers here. Shouldn't this be done by the one who configures the framework instance?

Baha-sk: yes, I know, using the optional function for adding headers (that will invoke the authz key server for getting the cap key) will call the the auth service with keyID and set the resulting `Signature` header

I added it because when creating a keystore, we explicitly set the `controller` in the call so we can store the controller in the local remoteKMS config storage for future calls where the controller is used as the invoker (for preventive measures in addition to the keyID signatures).

pkg/kms/webkms/remotekms.go
```diff
@@ -116,17 +100,24 @@ func New(provider storage.Provider, client *http.Client, marshaller marshalFunc)
 func CreateKeyStore(store storage.Store, httpClient *http.Client, keyserverURL, controller string,
```
aholovko, line 100: `CreateKeyStore` is not a part of `kms.KeyManager` API. Should it be here?

aholovko, line 100: `CreateKeyStore` is not a part of `kms.KeyManager` API. Should it be here?

pkg/kms/webkms/headers_opts.go
```diff
@@ -0,0 +1,34 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package webkms
+
+import "net/http"
+
+// addHeaders function supports adding http headers to req for a given key server defined in ksConfig.
+type addHeaders func(req *http.Request, ksConfig KeyServerCfg) (*http.Header, error)
```
fqutishat, line 12: what is  need for ksConfig ?

pkg/kms/webkms/remotekms.go
llorllale: > (for preventive measures in addition to the keyID signatures).

The server's zcap-ld authz will not use this `invoker` header to determine the invoker of the capability. Instead, the `keyID` parameter of the `Signature` header will be used.

pkg/kms/webkms/remotekms.go
Baha-sk: it's part of webkms api though.. it's kept here to maintain webkms interoperability even though it's not part of kms.KeyManager API

pkg/kms/webkms/headers_opts.go
Baha-sk: ksConfig should provide the auth server connection config info (url, secret, controller) necessary to obtain a request authorization (eg: zcap signature).

pkg/kms/webkms/remotekms.go
Baha-sk: I will remove controller header

pkg/kms/webkms/headers_opts.go
Baha-sk: thinking a bit more on this, the framework should not be aware of this info, I'll remove this argument from the options, it is needed for integration by the client who needs to authenticate with the key server, so it should be done outside the framework.

pkg/kms/webkms/remotekms.go
Baha-sk: Updated. Now `secret` and `controller` headers are not managed by remoteKMS anymore (instead a user must provider the optional headers function that should set any auhorization info needed in the request)

pkg/kms/webkms/headers_opts.go
Baha-sk: second argument is now removed

fqutishat (approved):

PR #2325, opened 2020-11-18
feat: client for messaging

### LINKED ISSUE #2323: Add client pkg for messaging service features
- add client for messaging services so that it can be reused by other commands for sending messages

sudeshrshetty:
- added client package for messaging features so that any command
controller can reuse basic messaging feature to avoid code duplication
and re-work.
- closes #2323

### DIFF
--- a/pkg/client/messaging/client.go
+++ b/pkg/client/messaging/client.go
@@ -1,0 +1,248 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package messaging
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/btcsuite/btcutil/base58"
+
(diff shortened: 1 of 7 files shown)


pkg/client/messaging/client.go
```diff
@@ -0,0 +1,248 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package messaging
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/mr-tron/base58"
```
troyronda, line 13: are we using a different base58 package. I thought we were using "github.com/btcsuite/btcutil/base58"

sudeshrshetty: fixed it, auto-import caused this issue in my editor.

fqutishat (approved):

talwinder50 (approved):

rolsonquadras (approved):

PR #2329, opened 2020-11-18
feat: reconnect all & create-invitation in mediator command

### LINKED ISSUE #2328: reconnect all and create-invitation features in mediator commands

sudeshrshetty:
- also added REST & JS bindings
- closes #2328

TODO: mobile bindings

### DIFF
--- a/pkg/controller/command/mediator/command.go
+++ b/pkg/controller/command/mediator/command.go
@@ -14,11 +14,13 @@ import (
 
 	"github.com/hyperledger/aries-framework-go/pkg/client/mediator"
 	"github.com/hyperledger/aries-framework-go/pkg/client/messagepickup"
+	"github.com/hyperledger/aries-framework-go/pkg/client/outofband"
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/controller/command"
 	"github.com/hyperledger/aries-framework-go/pkg/controller/internal/cmdutil"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/internal/logutil"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
 )
 
 var logger = log.New("aries-framework/command/route")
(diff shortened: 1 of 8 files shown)


fqutishat (approved):

pkg/controller/command/mediator/command.go
```diff
@@ -244,6 +269,74 @@ func (o *Command) Reconnect(rw io.Writer, req io.Reader) command.Error {
...
+
+	logutil.LogDebug(logger, CommandName, ReconnectAllCommandMethod, successString)
+
+	return nil
+}
+
+// CreateInvitation creates out-of-band invitation from one of the mediator connections.
+func (o *Command) CreateInvitation(rw io.Writer, req io.Reader) command.Error {
+	connections, err := o.routeClient.GetConnections()
+	if err != nil {
```
soluchok, line 301: I see that this function only adds a random connection. I don't think that we need a new function for that.
I would suggest modifying the existing function (outOfBand.CreateInvitation) to support random `router connection`. 
To be honest, I don't like this random pick up at all. I think it should be part of edge-agent refactoring. After we set up the router we have a connection, so we can use it everywhere we need it.

@sudeshrshetty @llorllale @rolsonquadras ^

sudeshrshetty: Nothing wrong in adding a command to reduce number of lines or reduce repeated logic for clients. This is moved from agent-sdk.

if you pass all connections then out of band client picks first one always, which is not what we need for router.

As comment says, this command is for clients who can directly create invitation without passing any connection details, like `mediator.createInvitation()`.  Mediator will pickup one random connection (current webwallet logic).

If someone wants to create invitation for particular connection, he doesn't need this command.

llorllale: I agree the randomization of the router selection should be exposed in the clients and not implemented here.

llorllale: We should not be reimplementing this logic everywhere.

sudeshrshetty: removed `mediator.createInvitation()`

soluchok (approved):

rolsonquadras (approved):

Issue #2315, opened 2020-11-13
Rework storage interface to allow for index use


DRK3:
This will allow for improved performance in several providers including the EDV REST provider.

This refactoring may also involve making an even more basic store interface that represents the EDV storage features (the "lowest common denominator"/most restrictive one.)

DRK3:
With the new design, also address https://github.com/hyperledger/aries-framework-go/pull/2310#discussion_r523426982

DRK3:
Also address https://github.com/hyperledger/aries-framework-go/pull/2330#discussion_r526526574

PR #2330, opened 2020-11-18
feat: EDV REST provider can now update documents

DRK3:
- When calling the EDV REST provider's Put(k, v) method with a key that already exists in the store, it now updates the underlying document with v.
- Updated log messages to be on one line.
- Moved getAllDocuments down so that it's in the same place as the other unexported methods.

### DIFF
--- a/pkg/storage/edv/restprovider.go
+++ b/pkg/storage/edv/restprovider.go
@@ -31,13 +31,16 @@ const (
 	contentTypeApplicationJSON = "application/json"
 	locationHeaderName         = "Location"
 
+	failCheckForExistingEDVDocument                  = "failed to check for existing EDV document: %w"
+	failStoreEDVDocument                             = "failed to store EDV document: %w"
 	failComputeMACStoreIndexName                     = "failed to compute MAC for the store index name: %w"
 	failComputeMACStoreIndexValue                    = "failed to compute MAC for the store index value: %w"
 	failComputeMACStoreAndKeyIndexName               = "failed to compute MAC for the store+key index name: %w"
 	failComputeMACStoreAndKeyIndexValue              = "failed to compute MAC for the store+key index value: %w"
-	failCreateIndexedAttribute                       = "failed to create indexed attribute: %w"
+	failCreateIndexedAttributes                      = "failed to create indexed attributes: %w"
 	failComputeBase64EncodedStoreAndKeyIndexValueMAC = "failed to compute Base64-encoded store+key index value MAC: %w"
 	failCreateDocumentInEDVServer                    = "failed to create document in EDV server: %w"
+	failUpdateDocumentInEDVServer                    = "failed to update existing document in EDV server: %w"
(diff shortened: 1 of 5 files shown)


pkg/storage/edv/restprovider.go
```diff
@@ -179,24 +175,20 @@ type restStore struct {
...
+	// See if this key is already being used in an encrypted index tag in an existing document.
+	existingEDVDocumentID, err := r.retrieveEDVDocumentID(k)
+	if err != nil && !errors.Is(err, storage.ErrDataNotFound) {
+		return fmt.Errorf(failCheckForExistingEDVDocument, err)
 	}
 
-	indexedAttributeCollection, err := r.createIndexedAttributes(k)
+	// If existingEDVDocumentID was set, then this means that there is already an existing document that
+	// well get updated.
+	err = r.createEDVDocument(k, v, existingEDVDocumentID)
```
Baha-sk, line 189: trying to understand the logic here since `errors.Is(err, storage.ErrDataNotFound)` is not explicitly covered...

if `existingEDVDocumentID` is empty.. it's going to create a new edv document, otherwise it will fetch the existing document?

DRK3: @Baha-sk Yes, exactly.

Baha-sk: I think `r.createEDVDocument` is a bit confusing... it should be createOrUpdateEDVDocument ... or better do 2 functions, one create and one update... and add the `if existingEDVDocumentID == ""` here instead

DRK3: Actually, you're almost correct. If existingEDVDocumentID is empty, then it's going to update that document instead of creating a new one (the old document is never actually fetched though)

Baha-sk: I think `createOrUpdateEDVDocument` is a more accurate name.. but it's private.. so it should be fine

Baha-sk (approved):

sandrask (approved):

pkg/storage/edv/restprovider.go
DRK3: Actually that's what I originally did - 2 separate functions. The problem was a lot of code duplication and it would have been multiple tests too... I struggled with the name a bit as well, as I was trying to keep it short as well based on previous feedback I got from @sandrask... What do you think about the name? Is there maybe some other nice, succinct way of describing this?

sandrask: createOrUpdateEDVDocument is short and clear

fqutishat (approved):

pkg/storage/edv/restprovider.go
DRK3: OK - I'll fix this later in #2315. I'll be doing a big storage refactor there anyway. I linked to the comment here in that issue so I won't forget to do it.

PR #2343, opened 2020-11-22
fix: json tags in remote kms/crypto

aholovko:
Fixes/updates json tags in remotekms and remotecrypto requests/responses.

### DIFF
--- a/pkg/kms/webkms/remotekms.go
+++ b/pkg/kms/webkms/remotekms.go
@@ -35,16 +35,16 @@ const (
 var logger = log.New("aries-framework/kms/webkms")
 
 type createKeystoreReq struct {
-	Controller         string `json:"controller,omitempty"`
-	OperationalVaultID string `json:"pperationalvaultid,omitempty"`
+	Controller string `json:"controller,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/kms/webkms/remotekms.go
```diff
@@ -36,15 +36,15 @@ var logger = log.New("aries-framework/kms/webkms")
 
 type createKeystoreReq struct {
 	Controller         string `json:"controller,omitempty"`
-	OperationalVaultID string `json:"pperationalvaultid,omitempty"`
+	OperationalVaultID string `json:"operationalVaultID,omitempty"`
```
troyronda, line 39: i am confused where the term operational is coming from for a generic remote kms.

aholovko: operational keys are the user's "working" keys in opposed to the "service" keys used for supporting tasks (e.g., delegate/recipient keys)

I agree that this is too implementation-specific, so it makes sense to rename it to a more generic term, like "edvVaultID" or just "vaultID".

fqutishat (approved):

PR #2349, opened 2020-11-27
feat: Add cache provider to edv rest

### LINKED ISSUE #2348: Add cache provider to edv rest

fqutishat:
closes #2348

### DIFF
--- a/pkg/storage/formattedstore/formattedstore.go
+++ b/pkg/storage/formattedstore/formattedstore.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/storage"
 	"github.com/hyperledger/aries-framework-go/pkg/storage/mem"
 )
@@ -27,15 +28,28 @@ const (
 	failGetIteratorFromUnderlyingStore = "failed to get iterator from underlying store: %w"
 )
 
+var logger = log.New("formatted-store")
+
 // Formatter represents a type that can convert data between two formats.
 type Formatter interface {
 	FormatPair(k string, v []byte) ([]byte, error)
 	ParsePair([]byte) (k string, v []byte, err error)
 }
 
+// Option configures the formatted store.
+type Option func(opts *FormattedProvider)
+
+// WithCacheProvider option is for using caching provider.
+func WithCacheProvider(cacheProvider storage.Provider) Option {
+	return func(opts *FormattedProvider) {
+		opts.cacheProvider = cacheProvider
+	}
+}
+
 // FormattedProvider is a storage provider that allows for data to be formatted in an underlying provider.
 type FormattedProvider struct {
 	provider              storage.Provider
+	cacheProvider         storage.Provider
 	formatter             Formatter
 	skipIteratorFiltering bool
 }
@@ -47,12 +61,18 @@ type FormattedProvider struct {
 // Iterator(startKey, endKey string) method, then set skipIteratorFiltering to true to avoid redundant
 // filtering in FormattedProvider.
 func NewFormattedProvider(provider storage.Provider, formatter Formatter,
-	skipIteratorFiltering bool) *FormattedProvider {
-	return &FormattedProvider{
+	skipIteratorFiltering bool, opts ...Option) *FormattedProvider {
+	formattedProvider := &FormattedProvider{
 		provider:              provider,
 		formatter:             formatter,
 		skipIteratorFiltering: skipIteratorFiltering,
 	}
+
+	for _, opt := range opts {
+		opt(formattedProvider)
+	}
+
+	return formattedProvider
 }
 
 // OpenStore opens a store in the underlying provider with the given name and returns a handle to it.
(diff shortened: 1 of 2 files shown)


pkg/storage/formattedstore/formattedstore.go
```diff
@@ -122,14 +168,31 @@ func (s *formatStore) Get(k string) ([]byte, error) {
...
+	if s.cacheStore != nil {
+		r := s.cacheStore.Iterator(startKey, endKey)
+		if r.Error() == nil {
+			return r
+		}
+
+		logger.Debugf("failed to get startKey %s endKey %s from cache store", startKey, endKey)
+	}
+
 	formattedIterator := s.store.Iterator(startKey, endKey)
```
DRK3, line 196: Need to use `r` down here if the cacheStore exists

pkg/storage/formattedstore/formattedstore.go
```diff
@@ -122,14 +168,31 @@ func (s *formatStore) Get(k string) ([]byte, error) {
...
 // TODO (#2315) While doing the storage interface rework, see if the skipIteratorFiltering switch is still
 //  needed.
 func (s *formatStore) Iterator(startKey, endKey string) storage.StoreIterator {
+	if s.cacheStore != nil {
+		r := s.cacheStore.Iterator(startKey, endKey)
+		if r.Error() == nil {
+			return r
+		}
+
+		logger.Debugf("failed to get startKey %s endKey %s from cache store", startKey, endKey)
```
DRK3, line 193: I think you mean something like: "failed to get iterator from cache store".

fqutishat: yes

pkg/storage/formattedstore/formattedstore.go
```diff
@@ -122,14 +168,31 @@ func (s *formatStore) Get(k string) ([]byte, error) {
...
 
 // The "don't filter" switch is here since it's possible for the underlying storage to have already done
 // the filtering, in which case it would be redundant to do it here too.
 // TODO (#2315) While doing the storage interface rework, see if the skipIteratorFiltering switch is still
 //  needed.
 func (s *formatStore) Iterator(startKey, endKey string) storage.StoreIterator {
+	if s.cacheStore != nil {
+		r := s.cacheStore.Iterator(startKey, endKey)
+		if r.Error() == nil {
+			return r
```
DRK3, line 190: Can't return `r` immediately - need to run it through the loop below

fqutishat: no loop needed

DRK3: Wait - my comment is only true if the cache store is storing encrypted data. If the cache store contains the decrypted stuff already, then I think you're ok.

pkg/storage/formattedstore/formattedstore.go
fqutishat: done

pkg/storage/formattedstore/formattedstore.go
DRK3: If the data is already formatted and you know the filtering has already been done, then you can disregard my comment

fqutishat: this logic is doing that https://github.com/hyperledger/aries-framework-go/pull/2349/files#diff-8c872e6990542239384a04aec51b7f81b13913192843fad32b1f20f7c4afd06bR189

pkg/storage/formattedstore/formattedstore.go
DRK3, line 190: Currently the cacheStore can be any type of provider. I know it would be silly, but if you made the cacheStore another EDV REST provider, then this will not work correctly. The EDV REST store's iterator method ignored the startKey and endKey and always returns all document.

(Even if you ignore this weird edge case, you should have a comment with a warning or something like that)

fqutishat: @DRK3 this interface will change correct ?

DRK3: @fqutishat Eventually, yeah. I want to get rid of that "skipIteratorFiltering" check (see the TODO above this method).

rolsonquadras (approved):

pkg/storage/formattedstore/formattedstore.go
```diff
@@ -108,10 +138,26 @@ func (s *formatStore) Put(k string, v []byte) error {
...
+			return err
+		}
+	}
+
 	return nil
 }
 
 func (s *formatStore) Get(k string) ([]byte, error) {
+	if s.cacheStore != nil {
+		v, err := s.cacheStore.Get(k)
```
DRK3, line 153: If the data isn't here in the cache, you'll need to check the non-cache provider

DRK3: (if you're using a mem provider, this will fail if/when the aries agent gets instantiated again)

DRK3: (discussed this offline - this choice is left to the user)

DRK3 (approved):

Baha-sk (approved):

bstasyszyn (approved):

pkg/storage/formattedstore/formattedstore.go
```diff
@@ -27,15 +28,28 @@ const (
 	failGetIteratorFromUnderlyingStore = "failed to get iterator from underlying store: %w"
 )
 
+var logger = log.New("EDV-REST-Provider")
```
troyronda, line 31: do we use mix caps in log names?

troyronda (approved):

pkg/storage/formattedstore/formattedstore.go
fqutishat: I change it

fqutishat: done

PR #2365, opened 2020-12-01
feat: EDV REST provider batch support

DRK3:

### DIFF
--- a/pkg/storage/edv/restprovider.go
+++ b/pkg/storage/edv/restprovider.go
@@ -40,6 +40,7 @@ const (
 	failCreateIndexedAttributes                      = "failed to create indexed attributes: %w"
 	failComputeBase64EncodedStoreIndexValueMAC       = "failed to compute Base64-encoded store index value MAC: %w"
 	failComputeBase64EncodedStoreAndKeyIndexValueMAC = "failed to compute Base64-encoded store+key index value MAC: %w"
+	failAddEncryptedIndices                          = "failed to add encrypted indices to encrypted document: %w"
 	failCreateDocumentInEDVServer                    = "failed to create document in EDV server: %w"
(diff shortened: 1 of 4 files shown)


pkg/storage/edv/restprovider.go
```diff
@@ -206,6 +208,38 @@ func (r *RESTProvider) Close() error {
...
+	encryptedDocument.IndexedAttributeCollections = indexedAttributeCollection
+
+	return &encryptedDocument, nil
+}
+
+func (r *RESTProvider) RetrieveEDVDocumentID(k string, restStore *restStore) (string, error) {
+	return restStore.retrieveEDVDocumentID(k)
+}
+
+func (r *RESTProvider) Batch(batch *models.Batch) error {
```
bstasyszyn, line 234: Should there be a return value that contains the response of each update, e.g. CouchDB returns the following information for each item in the batch:

```
	ID     string
	Error  string
	Reason string
	Ok     bool
	Rev    string
```

DRK3: The EDV server will return an array of responses, one for each item in the batch. Each individual response follows the spec for the [Create Document, Update Document and Delete Document endpoints](https://identity.foundation/confidential-storage/#http-api). In the case of an error, then the response will contain the specific reason. The `error` returned from this RESTProvider Batch method will contain all of this information... however, all Aries cares about is whether there it succeeded or failed... if it failed, then the full error will get printed out.

The lower level restClient will actually return the array of responses in case someone wants to look at them. (But for the Aries provider, we have no use for this information)

DRK3: (Note: the full error will contain the responses for each item in the batch, so that way we can figure out which item failed.)

bstasyszyn (approved):

sandrask (approved):

fqutishat (approved):

PR #2372, opened 2020-12-03
feat: switch batch write to explicit flush

fqutishat:

### DIFF
--- a/pkg/storage/formattedstore/batchwrite.go
+++ b/pkg/storage/formattedstore/batchwrite.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/hyperledger/aries-framework-go/pkg/storage/edv/models"
 )
 
@@ -27,14 +29,17 @@ type batchStore interface {
 
 // BatchService is a batch service.
 type BatchService struct {
-	sizeLimit int
-	timeLimit time.Duration
-	sync.RWMutex
-	values       models.Batch
-	futureValues map[string]*batch
-	formatter    Formatter
-	provider     batchProvider // TODO remove batch provider after refactoring edv batch
-	addBatchChan chan *batch
+	mutex            sync.RWMutex
+	futureValues     map[string]*batch
+	futureValuesLock sync.RWMutex
+	formatter        Formatter
+	provider         batchProvider // TODO remove batch provider after refactoring edv batch
+	current          *pendingBatch
+}
+
+type processBatch struct {
+	id             string
+	vaultOperation *models.VaultOperation
 }
 
 type batch struct {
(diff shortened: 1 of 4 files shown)


pkg/storage/formattedstore/batchwrite.go
```diff
@@ -27,78 +27,83 @@ type batchStore interface {
...
+		for addBatch := range b.addBatchChan {
+			if addBatch.isReady {
+				b.readyChan <- struct{}{}
+				return
+			}
+
+			if addBatch.isDeleted {
+				b.Lock()
+
+				id, err := b.formatter.GenerateEDVCompatibleID(addBatch.keyID)
```
rolsonquadras, line 66: If this batch service/formatter is specific to edv, then better to move this to /storage/edv rather than here (/storage/formattedstore) and inject the batch service (probably a future PR).

rolsonquadras (approved):

pkg/storage/formattedstore/batchwrite.go
fqutishat: this will be part of refactoring storage interface

fqutishat: storage interface need flush method

fqutishat: will be done in v0.1.6

bstasyszyn (approved):

sandrask (approved):

pkg/storage/formattedstore/batchwrite.go
```diff
@@ -125,57 +130,58 @@ func (b *BatchService) Get(k string) ([]byte, error) {
...
 	b.futureValues[k] = addBatch
-	b.Unlock()
+	b.futureValuesLock.Unlock()
 
 	b.addBatchChan <- addBatch
 }
 
-func (b *BatchService) flush() error {
+// Flush data.
+func (b *BatchService) Flush() error {
```
troyronda, line 159: We may want to think more about the func name (not blocking this PR).

pkg/storage/formattedstore/batchwrite.go
```diff
@@ -125,57 +130,58 @@ func (b *BatchService) Get(k string) ([]byte, error) {
...
+	b.futureValuesLock.Lock()
 	b.futureValues[k] = addBatch
-	b.Unlock()
+	b.futureValuesLock.Unlock()
 
 	b.addBatchChan <- addBatch
 }
 
-func (b *BatchService) flush() error {
+// Flush data.
```
troyronda, line 158: How does the caller know when the data has _finished_ writing? If the application goes away (e.g., browser is closed - memory and go channels are gone), how does the caller recover when the application comes back?

troyronda (approved):

pkg/storage/formattedstore/batchwrite.go
fqutishat: @troyronda the caller need to flush data when before finishing the flow

DRK3 (approved):

pkg/storage/formattedstore/batchwrite.go
bstasyszyn: I've seen 'Sync' being used before.

pkg/storage/formattedstore/batchwrite.go
```diff
@@ -125,57 +91,161 @@ func (b *BatchService) Get(k string) ([]byte, error) {
...
+
+			b.Unlock()
+
+			go func(addBatch *batch, b *BatchService, processID string) {
+				logger.Infof("start format pair for put k %s", addBatch.keyID)
+
+				formattedValue, err := b.formatter.FormatPair(addBatch.keyID, addBatch.value)
+				if err != nil {
+					logger.Errorf("failed to format pair for k=%s: %w", addBatch.keyID, err)
+					b.Unlock()
```
bstasyszyn, line 225: Don't need this here since it's inside the Go routine.

pkg/storage/formattedstore/batchwrite.go
```diff
@@ -125,57 +91,161 @@ func (b *BatchService) Get(k string) ([]byte, error) {
...
+
+					return
+				}
+
+				encryptedDocument, err := addBatch.addEncryptedIndicesFunc(addBatch.keyID, formattedValue)
+				if err != nil {
+					logger.Errorf("failed to add encrypted indices k=%s: %w", addBatch.keyID, err)
+					b.Unlock()
```
bstasyszyn, line 233: Don't need this here since it's inside the Go routine.

PR #2378, opened 2020-12-05
feat: add batch size to batch write

fqutishat:

### DIFF
--- a/pkg/storage/formattedstore/batchwrite.go
+++ b/pkg/storage/formattedstore/batchwrite.go
@@ -35,6 +35,7 @@ type BatchService struct {
 	formatter        Formatter
 	provider         batchProvider // TODO remove batch provider after refactoring edv batch
 	current          *pendingBatch
+	batchSizeLimit   int
 }
 
 type processBatch struct {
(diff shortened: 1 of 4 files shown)


DRK3 (approved):

bstasyszyn (approved):

pkg/storage/formattedstore/batchwrite.go
```diff
@@ -35,6 +35,7 @@ type BatchService struct {
 	formatter        Formatter
 	provider         batchProvider // TODO remove batch provider after refactoring edv batch
 	current          *pendingBatch
+	batchSizeLimit   int
```
troyronda, line 38: We could also additionally consider an optional max batch time in a followup.

pkg/storage/formattedstore/batchwrite.go
```diff
@@ -148,10 +164,13 @@ func newPendingBatch(provider batchProvider, formatter Formatter) *pendingBatch
 	return &pendingBatch{
 		provider:  provider,
 		formatter: formatter,
+		size:      0,
```
troyronda, line 167: Does 0 mean no maximum or disable batch?

Should also have a way to indicate no maximum.

troyronda (approved):

pkg/storage/formattedstore/batchwrite.go
fqutishat: yes

pkg/storage/formattedstore/batchwrite.go
fqutishat: I will add case for no maximum

fqutishat: done

PR #2404, opened 2020-12-15
feat: Presentation exchange - adds entity validation

soluchok:
Adds entity validation for presentation definition.


pkg/doc/presexch/api.go
```diff
@@ -257,10 +157,10 @@ func (p *PresentationDefinition) evalSubmissionRequirements(matched map[string]*
 	return nil
 }
 
-func (p *PresentationDefinition) inputDescriptor(id string) *InputDescriptor {
-	for i := range p.InputDescriptors {
-		if p.InputDescriptors[i].ID == id {
-			return p.InputDescriptors[i]
+func (pd *PresentationDefinition) inputDescriptor(id string) *InputDescriptor {
+	for i := range pd.InputDescriptors {
```
llorllale, line 161: https://github.com/decentralized-identity/presentation-exchange/issues/163

llorllale (approved):

pkg/doc/presexch/definition.go
```diff
@@ -0,0 +1,361 @@
...
+	MaxLength        int                    `json:"maxLength,omitempty"`
+	ExclusiveMinimum StrOrInt               `json:"exclusiveMinimum,omitempty"`
+	ExclusiveMaximum StrOrInt               `json:"exclusiveMaximum,omitempty"`
+	Const            StrOrInt               `json:"const,omitempty"`
+	Enum             []StrOrInt             `json:"enum,omitempty"`
+	Not              map[string]interface{} `json:"not,omitempty"`
+}
+
+// IsValid validates presentation definition.
+func (pd *PresentationDefinition) IsValid() error {
```
sudeshrshetty, line 143: Why we are not using JSON schema validator??

pkg/doc/presexch/api.go
```diff
@@ -163,7 +63,7 @@ func WithJSONLDDocumentLoader(l ld.DocumentLoader) MatchOption {
 }
 
 // Match returns the credentials matched against the InputDescriptors ids.
-func (p *PresentationDefinition) Match(vp *verifiable.Presentation, // nolint:gocyclo,funlen
+func (pd *PresentationDefinition) Match(vp *verifiable.Presentation, // nolint:gocyclo,funlen
```
sudeshrshetty, line 66: what if multiple credentials matched??

pkg/doc/presexch/definition.go
sudeshrshetty: something like [this](https://github.com/trustbloc/edge-agent/blob/master/cmd/wallet-web/src/pages/chapi/wallet/common/presentationDefSchema.js#L12-L168)

pkg/doc/presexch/definition.go
```diff
@@ -0,0 +1,361 @@
...
+		if err := pd.SubmissionRequirements[i].IsValid(); err != nil {
+			return fmt.Errorf("submission_requirements.%d: %w", i, err)
+		}
+	}
+
+	return nil
+}
+
+// IsValid validates submission requirement.
+func (sr *SubmissionRequirement) IsValid() error { // nolint: gocyclo
```
sudeshrshetty, line 176: Couldn't see [validation](https://identity.foundation/presentation-exchange/#pick-rule) on min/max values for `pick` rule

pkg/doc/presexch/definition.go
soluchok: I am using schema validator only for tests. It will require marshaling every time you want to validate definitions.
https://github.com/hyperledger/aries-framework-go/pull/2404/files#diff-1a239b0b79a42a8274b0eb84d4be2f453f4b22cf56e0559eb6c65871d8f190feR577

pkg/doc/presexch/api.go
soluchok: It is an old function. Will be rewritten soon.

pkg/doc/presexch/definition.go
soluchok: It is not related to definitions. It will be used when I will filter the credentials.

soluchok: Current validation is only about checking whether the definitions were provided properly (equivalent to schema validator with some additional checks).

pkg/doc/presexch/definition.go
sudeshrshetty: In this object, are you getting presentation definition as objects? we follow JSON schema validation pattern in other entities like DIDDoc. It is easier to maintain, faster and caller gets exact validation errors. Are you using the one in RFC for schema validation in tests? They don't cover all validation rules.

pkg/doc/presexch/definition.go
sudeshrshetty: you can find them [here](https://identity.foundation/presentation-exchange/#pick-rule), if they are wrong you can catch this error early to save some cycles.

pkg/doc/presexch/definition.go
soluchok: `Are you using the one in RFC for schema validation in tests?`  - yes
`They don't cover all validation rules.` - true
I can use `schema validator` but as I said it will marshall the structure one more time. @troyronda @llorllale 
See https://github.com/xeipuuv/gojsonschema/blob/master/jsonLoader.go#L304

sudeshrshetty: you can reuse the one we wrote [here](https://github.com/hyperledger/aries-framework-go/pull/2404#discussion_r544408482). You can create a follow-up task for that for now.

soluchok: @sudeshrshetty It does not follow the RFC

soluchok: Done. I am using a schema validator.

sudeshrshetty: The one I wrote in wallet follows old version of RFC.

sudeshrshetty: are all validation rules covered in schema you added?? or there gonna be one more PR for that?

soluchok: I added from RFC if they change schema we need to change it.

sudeshrshetty: well, then we gonna miss some validations since RFC schema is not complete. We can use RFC schema for tests like you used earlier and advanced one with validation for realtime validating presentation definition.

llorllale: @soluchok can we rename this method to `ValidateSchema()`? I don't know what `IsValid()` would mean even when it is a valid JSON object for the schema.

llorllale: @sudeshrshetty 

> well, then we gonna miss some validations since RFC schema is not complete. We can use RFC schema for tests like you used earlier and advanced one with validation for realtime validating presentation definition.

Can you push PRs to the P-E repo to fix the schema?

soluchok: @llorllale renamed to `ValidateSchema()`. Opened a PR https://github.com/decentralized-identity/presentation-exchange/pull/164

sudeshrshetty: @llorllale sure I can, that sounds like a good idea

sudeshrshetty (approved):
Please remember to create a follow up issues of finding multiple VC match for given input descriptor.

PR #2407, opened 2020-12-17
 feat: Adds the possibility to create VP based on the definition

soluchok:
Adds the possibility to create VP based on the definition.

### DIFF
--- a/pkg/doc/presexch/definition.go
+++ b/pkg/doc/presexch/definition.go
@@ -8,9 +8,12 @@ package presexch
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/xeipuuv/gojsonschema"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
 )
 
 const (
@@ -23,6 +26,12 @@ const (
 	Required Preference = "required"
 	// Preferred predicate`s value.
 	Preferred Preference = "preferred"
+)
+
+// nolint: gochecknoglobals
+var (
+	defaultVPContext = []string{"https://www.w3.org/2018/credentials/v1"}
+	defaultVPType    = []string{"VerifiablePresentation", "PresentationSubmission"}
 )
 
 type (
@@ -102,6 +111,14 @@ type Schema struct {
 	Required bool   `json:"required,omitempty"`
 }
 
+func (s *Schema) String() string {
+	if s.Required {
+		return "required:" + s.URI
+	}
+
+	return s.URI
+}
+
 // Holder describes Constraints`s  holder object.
 type Holder struct {
 	FieldID   []string    `json:"field_id,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/doc/presexch/definition.go
```diff
@@ -166,3 +168,90 @@ func (pd *PresentationDefinition) ValidateSchema() error {
...
+
+// CreateVP creates verifiable presentation.
+func (pd *PresentationDefinition) CreateVP(credentials ...*verifiable.Credential) (*verifiable.Presentation, error) {
+	if err := pd.ValidateSchema(); err != nil {
+		return nil, err
+	}
+
+	if len(pd.SubmissionRequirements) > 0 {
+		// TODO: handle submission requirements
+		return nil, errors.New("not supported yet")
```
llorllale, line 180:
```suggestion
		return nil, errors.New("submission requirements not supported yet")
```

pkg/doc/presexch/definition.go
```diff
@@ -166,3 +168,90 @@ func (pd *PresentationDefinition) ValidateSchema() error {
...
+	}
+
+	var result []*verifiable.Credential
+
+	for _, descriptor := range pd.InputDescriptors {
+		filtered := filterSchema(descriptor.Schema, credentials)
+		if len(filtered) == 0 {
+			return nil, errors.New("credentials do not satisfy the requirements")
```
llorllale, line 188:
```suggestion
			return nil, fmt.Errorf("credentials do not satisfy requirements with schema %s", descriptor.Schema)
```

pkg/doc/presexch/definition.go
```diff
@@ -166,3 +168,90 @@ func (pd *PresentationDefinition) ValidateSchema() error {
...
+	}
+
+	vp := &verifiable.Presentation{}
+
+	err := vp.SetCredentials(credentialsToInterface(merge(result))...)
+	if err != nil {
+		return nil, err
+	}
+
+	return vp, nil
```
llorllale, line 201: This VP is left without context, schema, etc. You'd have to manually set them on line 194.

(we could really use builder APIs for VPs....)

pkg/doc/presexch/definition_test.go
```diff
@@ -39,6 +41,286 @@ func TestPresentationDefinition_IsValid(t *testing.T) {
 	})
 }
 
+func TestPresentationDefinition_CreateVP(t *testing.T) {
```
llorllale, line 44: We should include tests that check whether the output VP is formed correctly (should pass a parsing check). This is related to by comment [here](https://github.com/hyperledger/aries-framework-go/pull/2407/files#r545164335).

pkg/doc/presexch/definition.go
soluchok: VP will be improved. The final result must be a valid VP. We do not have builder APIs :(

pkg/doc/presexch/definition_test.go
soluchok: I will add this check

pkg/doc/presexch/definition.go
sudeshrshetty: I don't `presentation_submission` in Verifiable Presentation you generated, is it going to be added in future?

soluchok: @sudeshrshetty yes, it will be added soon

sudeshrshetty (approved):

pkg/doc/presexch/definition_test.go
soluchok: @llorllale added `context` and `type`. Also added check in tests whether the VP is parsable.

llorllale (approved):

PR #2413, opened 2020-12-22
feat: New storage interface

DRK3:
This new interface is defined separately from the existing interface for now. Once new implementations of the storage providers are complete, then the full switchover will happen.


bstasyszyn (approved):

sandrask (approved):

pkg/newstorage/store.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package storage
```
troyronda, line 7: You should name the parent folder storage. To make this work, you need a grandparent folder with a different name.

DRK3: Updated the package name to match the folder name.

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+
+// ErrStoreNotFound is returned when a store is not found.
+var ErrStoreNotFound = errors.New("store not found")
+
+// ErrDataNotFound is returned when data is not found.
+var ErrDataNotFound = errors.New("data not found")
+
+// StoreConfiguration represents the configuration of a store.
+type StoreConfiguration struct {
+	TagNames []string
```
troyronda, line 19: It would be good to comment on the relationship of this property with the Tag struct on L40.

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+func WithPageSize(size int) QueryOption {
+	return func(opts *QueryOptions) {
+		opts.pageSize = size
+	}
+}
+
+// Tag represents a Name + Value pair that can be associated with a k,v pair for querying later.
+type Tag struct {
+	Name  string
+	Value string
```
troyronda, line 42: The rationale behind `value` should be documented.

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Provider represents a storage provider.
+type Provider interface {
+	// OpenStore opens a store with the given name and returns a handle.
+	// If the store has never been opened before, then it is created.
+	OpenStore(name string) (Store, error)
+
+	// SetStoreConfig sets the configuration on a store.
+	// The store must be created prior to calling this method.
+	SetStoreConfig(name string, config StoreConfiguration) error
```
troyronda, line 60: It is debatable whether this should be generally config or specifically just the tags. I suppose we are doing this for future possibilities (but it would be nice to comment on if we have any envisioned future config properties beyond tags).

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// GetStoreConfig gets the current store configuration.
+	// The store must be created prior to calling this method.
+	GetStoreConfig(name string) (StoreConfiguration, error)
+
+	// CloseStore closes the store with the given name.
+	// This does not delete any data in the store, and the store can be reopened again.
+	CloseStore(name string) error
```
troyronda, line 68: We might still want to discuss if CloseStore is in the right place.

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+	// Query returns all data that satisfies the expression. Expression format: TagName:TagValue.
+	// If TagValue is not provided, then all data associated with the TagName will be returned.
+	// For now, expression can only be a single tag Name + Value pair.
+	// If no options are provided, then defaults will be used.
+	Query(expression string, options ...QueryOption) (Iterator, error)
+
+	// Delete deletes the key + value pair (and all tags) associated with key.
+	Delete(key string) error
+
+	// Batch performs multiple Put and/or Delete operations in order.
```
troyronda, line 100: Should comment on what happens in case of operation failure(s).

How does the function indicate which operations failed (and why)?

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+	// Put stores the key + value pair along with the (optional) tags.
+	Put(key string, value []byte, tags ...Tag) error
+
+	// Get fetches the value associated with the given key.
+	Get(key string) ([]byte, error)
+
+	// Get fetches all tags associated with the given key.
+	GetTags(k string) ([]Tag, error)
+
+	// GetBulk fetches the values associated with the given keys.
```
troyronda, line 86: Similar comment https://github.com/hyperledger/aries-framework-go/pull/2413/files#r547979499

pkg/newstorage/newstorage.go
```diff
@@ -0,0 +1,122 @@
...
+	// Depending on the implementation, this method may be faster than repeated Put and/or Delete calls.
+	Batch(operations []Operation) error
+}
+
+// Iterator represents an iterator that can be used to iterate over stored key-value pairs.
+type Iterator interface {
+	// Next moves the pointer to the next value in the iterator. It returns false if the iterator is exhausted.
+	Next() (bool, error)
+
+	// Release releases associated resources. Release should always result in success
```
troyronda, line 110: We should document the rationale for the error return value.

troyronda (approved):

pkg/newstorage/newstorage.go
DRK3: Done in my new PR: https://github.com/hyperledger/aries-framework-go/pull/2426/files#diff-98db716f6d6bedd511a5f903dc6d785b6f60df130787f50612d64804a1f8195fR46

pkg/newstorage/newstorage.go
DRK3: Done in my new PR: https://github.com/hyperledger/aries-framework-go/pull/2426/files#diff-98db716f6d6bedd511a5f903dc6d785b6f60df130787f50612d64804a1f8195fR49

pkg/newstorage/newstorage.go
DRK3: I don't actually have any envisioned future config properties in mind. Do you think I should change it to just `TagNames`?

pkg/newstorage/newstorage.go
DRK3: The reason I kept CloseStore in the Provider is because the CouchDB provider implementation (and possibly others) holds on to opened stores as a sort of "cache", so in order to keep that behaviour I had to keep CloseStore in the provider. The other reason I kept it here was for symmetry with the OpenStore method.

pkg/newstorage/newstorage.go
DRK3: See https://github.com/hyperledger/aries-framework-go/pull/2426/files#diff-98db716f6d6bedd511a5f903dc6d785b6f60df130787f50612d64804a1f8195fR13 in my new PR

pkg/newstorage/newstorage.go
DRK3: See https://github.com/hyperledger/aries-framework-go/pull/2426/files#diff-98db716f6d6bedd511a5f903dc6d785b6f60df130787f50612d64804a1f8195fR13 in my new PR.

pkg/newstorage/newstorage.go
DRK3: I've removed the "should always result in success" which should mean that a rationale isn't needed anymore: https://github.com/hyperledger/aries-framework-go/pull/2426/files#diff-98db716f6d6bedd511a5f903dc6d785b6f60df130787f50612d64804a1f8195fR119

pkg/newstorage/newstorage.go
troyronda: The idea of a `Close` is to release resources. I'm not really understanding what `CloseStore` is doing in your case.

If resources need to be released, it would make sense to release from the handle. Perhaps I'm missing something?

DRK3: This is what I'm referring to: https://github.com/hyperledger/aries-framework-go-ext/blob/main/component/storage/couchdb/store.go#L119

I feel like the CouchDB provider doesn't really need to hold on to a list of opened store objects. Do you think it's okay to not have that "cache"? (not exactly a cache... more like a shortcut in case you try to open an already existing store)

PR #2418, opened 2020-12-28
feat: Presentation definition - adds  limit disclosure support

soluchok:
Adds limit disclosure support.


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/gorilla/mux v1.7.3
 	github.com/jinzhu/copier v0.0.0-20190924061706-b57f9002281a
+	github.com/kawamuray/jsonpath v0.0.0-20201211160320-7483bafabd7e
```
troyronda, line 21: Is there a better maintained implementation?

soluchok: No, I haven't found any library that has functionality as I need.

sudeshrshetty: in RFC, they recommended to use `https://github.com/PaesslerAG/jsonpath`, any reason for choosing this one over the one recommended?

soluchok: I am using it as well `https://github.com/PaesslerAG/jsonpath`. The problem when we have `limit_disclosure=true` we need to set the value. To et the value we need an original path (to recreate the same structure).
See this use case https://github.com/hyperledger/aries-framework-go/pull/2418/files#diff-1a239b0b79a42a8274b0eb84d4be2f453f4b22cf56e0559eb6c65871d8f190feR326

sudeshrshetty (approved):

fqutishat (approved):

PR #2426, opened 2021-01-04
feat: Improvements to new storage interface

DRK3:
- Improved documentation.
- Moved CloseStore method from Provider to Store.
- Introduced custom error type for bulk methods so the caller can determine which operation failed.

### DIFF
--- a/pkg/newstorage/newstorage.go
+++ b/pkg/newstorage/newstorage.go
@@ -8,6 +8,12 @@ package newstorage
 
 import "errors"
 
+// MultiError represents the errors that occurred during a bulk operation.
+type MultiError interface {
+	error
+	Errors() []error // Errors returns the error objects for all operations.
+}
+
 // ErrStoreNotFound is returned when a store is not found.
 var ErrStoreNotFound = errors.New("store not found")
 
@@ -16,7 +22,7 @@ var ErrDataNotFound = errors.New("data not found")
 
 // StoreConfiguration represents the configuration of a store.
 type StoreConfiguration struct {
-	TagNames []string
+	TagNames []string // A list of Tag names that key + value pairs in this store can be associated with.
 }
 
 // QueryOptions represents various options for Query calls in a store.
@@ -36,9 +42,13 @@ func WithPageSize(size int) QueryOption {
 	}
 }
 
-// Tag represents a Name + Value pair that can be associated with a k,v pair for querying later.
+// Tag represents a Name + Value pair that can be associated with a key + value pair for querying later.
 type Tag struct {
-	Name  string
+	// Name can be used to tag a given key + value pair as belonging to a group.
+	// Tag Names are static values that the store must be configured with (see TagNames in StoreConfiguration).
+	Name string
+	// Value can be used to indicate some optional metadata associated with a given key + value pair + tag name.
+	// Unlike Tag Names, Tag Values are dynamic and are not specified during store creation.
 	Value string
 }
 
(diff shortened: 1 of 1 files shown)


pkg/newstorage/newstorage.go
```diff
@@ -8,6 +8,12 @@ package newstorage
 
 import "errors"
 
+// BulkOperationError represents one or more errors that occurred during a bulk operation.
+type BulkOperationError interface {
+	error
```
DRK3, line 13: BulkOperationError.Error() would return a summary of all errors

pkg/newstorage/newstorage.go
```diff
@@ -99,16 +109,15 @@ type Store interface {
...
 
-// Iterator represents an iterator that can be used to iterate over stored key-value pairs.
+// Iterator allows for iteration over a collection of key-value pairs.
 type Iterator interface {
 	// Next moves the pointer to the next value in the iterator. It returns false if the iterator is exhausted.
 	Next() (bool, error)
 
-	// Release releases associated resources. Release should always result in success
-	// and can be called multiple times without causing an error.
+	// Release releases associated resources.
```
DRK3, line 120: It was weird that the method said that it should always result in success despite it still being able to return an error. I've simplified it and removed that wording. Most of our implementations can throw errors during Release(), so it seems like we still need an error return value.

pkg/newstorage/newstorage.go
```diff
@@ -84,9 +94,9 @@ type Store interface {
 	GetTags(k string) ([]Tag, error)
 
 	// GetBulk fetches the values associated with the given keys.
-	// If a key doesn't exist, then a nil []byte is returned for that value and is not considered an error.
+	// If a key doesn't exist, then a nil []byte is returned for that value. It is not considered an error.
 	// Depending on the implementation, this method may be faster than calling Get for each key individually.
-	GetBulk(keys ...string) ([][]byte, error)
```
troyronda, line 89: You should return a generic error type from functions.

pkg/newstorage/newstorage.go
```diff
@@ -99,16 +109,15 @@ type Store interface {
 
 	// Batch performs multiple Put and/or Delete operations in order.
 	// Depending on the implementation, this method may be faster than repeated Put and/or Delete calls.
-	Batch(operations []Operation) error
```
troyronda, line 102: You should return a generic error type from functions.

pkg/newstorage/newstorage.go
```diff
@@ -8,6 +8,12 @@ package newstorage
 
 import "errors"
 
+// BulkOperationError represents one or more errors that occurred during a bulk operation.
+type BulkOperationError interface {
+	error
+	Errors() map[int]error // Errors returns a mapping of failed operation indices to their associated errors.
```
troyronda, line 14: why a map[int] instead of a slice?

pkg/newstorage/newstorage.go
```diff
@@ -8,6 +8,12 @@ package newstorage
 
 import "errors"
 
+// BulkOperationError represents one or more errors that occurred during a bulk operation.
+type BulkOperationError interface {
```
troyronda, line 12: also this pattern is more generally the multierror.

pkg/newstorage/newstorage.go
DRK3: My thinking is that a caller might want the ability to know exactly which operation failed.

troyronda:
We also need to decide what to do with CloseStore (from prior comments).

pkg/newstorage/newstorage.go
DRK3: Changed to slice.

pkg/newstorage/newstorage.go
DRK3: Changed to MultiError.

DRK3:
> We also need to decide what to do with CloseStore (from prior comments).

I've moved it to the Store interface after some discussion

troyronda (approved):

PR #2432, opened 2021-01-06
feat: add pbkdf2 masterlock + ref NIST in docs

Baha-sk:

### DIFF
--- a/pkg/secretlock/local/local_secret_lock.go
+++ b/pkg/secretlock/local/local_secret_lock.go
@@ -22,7 +22,10 @@ import (
 // in a local file or an environment variable prior to using this service.
 //
 // The user has the option to encrypt the master key using hkdf.NewMasterLock(passphrase, hash func(), salt)
-// found in the sub package masterlock/hkdf.
+// found in the sub package masterlock/hkdf. There's also the option of using pbkdf2.NewMasterLock() instead of hkdf
+// which is located under masterlock/pbkdf2.
+//
+// This lock services uses the NIST approved AES-GCM 256 bit encryption as per NIST SP 800-38D.
 //
 // The master key must be stored (encrypted with a MasterLock or not encrypted) either in a file or in
 // an environment variable.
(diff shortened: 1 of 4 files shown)


pkg/secretlock/local/local_secret_lock.go
```diff
@@ -22,7 +22,10 @@ import (
 // in a local file or an environment variable prior to using this service.
 //
 // The user has the option to encrypt the master key using hkdf.NewMasterLock(passphrase, hash func(), salt)
-// found in the sub package masterlock/hkdf.
+// found in the sub package masterlock/hkdf. There's also the option of using pbkdf2.NewMasterLock() instead of hkdf
```
troyronda, line 25: when should someone use hkdf vs pbkdf2?

troyronda (approved):

pkg/secretlock/local/local_secret_lock.go
Baha-sk: the default is set to HKDF in the framework (this PR doesn't change this behaviour, it only adds pbkdf2 as a new locker option)

both use a hashing function and the framework obviously defaults to SHA256 with 32 byte long inputs.

we would use pbkdf2 when we want to derive ascii based (user input) passwords, and try to set a higher iteration count to reduce brute force attacks, I've tested with 8k and it seems to not impact the kdf much (4 ms vs 2 ms using 2k whereas 64 k takes about 34 ms on my machine)

HKDF is normally used for a generated passphrase where user input is not required. It makes sense in the framework as we don't expect user input to be used as a pass (it would be a randomly generated []byte stored somewhere, the user doesn't need to input it directly)

PBKDF2 is normally used for real user entered passwords. But both work for KDF.

Usually as long as the passphrase is randomly generated and is not a simple word, HDKF is fine in our case since our passwords are randomly generated strings. For simpler passwords (like plain English words) then PBKDF2 with higher iterations count is recommended.

PR #2437, opened 2021-01-08
docs: Presentation Exchange - adds samples

soluchok:
Adds Presentation Exchange samples.

### DIFF
--- a/pkg/doc/presexch/README.md
+++ b/pkg/doc/presexch/README.md
@@ -1,0 +1,104 @@
+# Presentation Exchange
+
+This document shows how to use [Presentation Exchange](https://identity.foundation/presentation-exchange) by examples.
+Code examples can be found [here](example_test.go).
+
+1. This example demonstrates `predicate` and `limit_disclosure` usage.
+   The `presentation definition` below requires field `age` to be greater or equal to 18.
+   Also, we have `limit_disclosure=true` which requires that output data is limited to the entries specified in the `fields` property.
+   The `predicate` means that the result should be expressed as a boolean value.
+```json
+{
+  "id": "31e2f0f1-6b70-411d-b239-56aed5321884",
+  "purpose": "To sell you a drink we need to know that you are an adult.",
+  "input_descriptors": [
+    {
+      "id": "867bfe7a-5b91-46b2-9ba4-70028b8d9cc8",
+      "purpose": "Your age should be greater or equal to 18.",
+      "schema": [
+        {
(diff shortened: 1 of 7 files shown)


pkg/doc/presexch/README.md
```diff
@@ -0,0 +1,103 @@
...
+      "id": "2dc74354-e965-4883-be5e-bfec48bf60c7",
+      "issuer": "",
+      "type": "VerifiableCredential"
+    }
+  ]
+}
+```
+
+As you can see the VP has a credential without `first_name` and `last_name` (because of `limit_disclosure`).
+Also, instead of `age`, we have a boolean value (because of `predicate`).
```
troyronda, line 103: are you providing any examples of using the presentation exchange code in aries-framework-go?

soluchok: Done!

troyronda: Can we add a link from this document to the code examples?

troyronda (approved):

pkg/doc/presexch/README.md
```diff
@@ -0,0 +1,104 @@
...
+  "type": [
+    "VerifiablePresentation",
+    "PresentationSubmission"
+  ],
+  "verifiableCredential": [
+    {
+      "@context": [
+        "https://www.w3.org/2018/credentials/v1"
+      ],
+      "age": true,
```
llorllale, line 88: @soluchok 

This is a different claim than the one originally issued (by the Issuer). This new claim is not covered by cryptographic proof from the Issuer; it is unverifiable.

How will this work in practice?

pkg/doc/presexch/README.md
```diff
@@ -0,0 +1,104 @@
+# Presentation Exchange
+
+This document shows how to use [Presentation Exchange](https://identity.foundation/presentation-exchange) by examples.
+Code examples can be found [here](example_test.go).
+
+1. This example demonstrates `predicate` and `limit_disclosure` usage.
+   The `presentation definition` below requires field `age` to be greater or equal to 18.
+   Also, we have `limit_disclosure=true` which requires that output data is limited to the entries specified in the `fields` property.
+   The `predicate` means that the result should be expressed as a boolean value.
```
llorllale, line 9: @soluchok 

> The `predicate` means that the result should be expressed as a boolean value.

This is a misunderstanding of the spec. The predicate is just criteria that the credential submitted in the presentation must satisfy. It is _not_ a transformation function.

Issue #1806, opened 2020-05-08
Create New Chacha Key private/public key Managers for ECDH-1PU Tink key type


Baha-sk:
And support ECDH public key export of the newly added key templates.

Issue #1684, opened 2020-04-23
Add Chacha JWE Content encryption


Baha-sk:
The new JWE encryption supports P-256 content encryption only. 

Add support for Chacha JWE Content encryption as well.

Requires: 
 - [x] Rename ECDHES Tink key types to include curve names #1804
 - [x] Create new Chacha key protos for ECDHES Tink templates #1805
 - [x] Create New Chacha Key private/public key Managers for ECDHES Tink key type #1806
 - [x] Add chacha key templates #1637
 - [ ] Add new ECDH KMS key types #2447
 - [x] Add support for X25519 key wrapping in ECDH Tink key #2445

ensure this issue is closed when closing this issue:  #815

Issue #1637, opened 2020-04-16
Add chacha key templates 


Baha-sk:
The new Crypto Tink templates for Anoncrypt adds NIST Curve encryption (AES-GCM P-265) support.

Add (X)Chacha20Poly1305 Curve encryption and crypto key templates as well.

Issue #815, opened 2019-11-14
Generic JWE support for envelopes


troyronda:
JWE packer should support ECDH key agreement for curves:
- X25519
- NIST P-256/secp256r1

For key encryption algorithms, we should support:
- ECDH-ES+A256KW
- ECDH-1PU+A256KW

> When performing an ECDH key agreement between a static private key
   and any untrusted public key, care should be taken to ensure that the public key is a valid point on the same curve as the private key. ([IETF ECDH-1PU RFC](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-02#section-4))

For content encryption, we should support a similar set to what is being discussed in related DIF issues:

- XChaCha20-Poly1305 (XC20P)
- ~~ChaCha20-Poly1305 (C20P)~~
- 256-bit AES-GCM (to allow for a FIPS compatability mode)

For the ChaCha modes, we should also compare the related issue JWE format to what we have implemented and update the Aries issue (and implementation).

IANA Notes:
- C20P and XC20P do not currently have IANA registrations.
- ECDH-1PU does not have an IANA registration but is defined in an [IETF draft (Network WG)](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-02). 

#857 lists some additional algorithms that were contemplated.

PR #2438, opened 2021-01-08
feat: add XChacha key to ECDH AEAD Tink key

Baha-sk:
This is a first change to add XChacha20Poly1305 keys to ECDH Tink keys to support XChacha AEAD encryption and key wrapping.

It includes adding new Tink key managers in the Tink keys package to support keys for:
1. NIST P curves KW - AES/XChacha20Poly1305 content encryption
2. X25519 KW - AES/XChacha20Poly1305 content encryption

This change also includes new key templates to support the creation of these ECDH keys.

Future changes will include supporting Key exports of these ECDH keys, adding new KMS key types, add XChacha Key wrapping and finally testing the new key with the JWE packers.

part of #1637, #1806, #1684, #815


pkg/crypto/tinkcrypto/primitive/composite/ecdh/ecdh_x25519_private_key_manager.go
```diff
@@ -1,9 +0,0 @@
-/*
-Copyright SecureKey Technologies Inc. All Rights Reserved.
-
-SPDX-License-Identifier: Apache-2.0
-*/
-
-package ecdh
-
-// TODO implement
```
troyronda, line 9: we still need x25519

pkg/crypto/tinkcrypto/primitive/composite/ecdh/ecdh_xchacha_aead_private_key_manager.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return nil, errInvalidECDHXChachaPrivateKeyFormat
+	}
+
+	err = validateKeyXChachaFormat(keyFormat.Params)
+	if err != nil {
+		return nil, errInvalidECDHXChachaPrivateKeyFormat
+	}
+
+	pub, pvt, err := ed25519.GenerateKey(rand.Reader)
```
troyronda, line 94: I don't understand the usage of ed25519.

troyronda, line 94: I also don't understand generating an Ed25519/X25519 key for key wrapping.

pkg/crypto/tinkcrypto/primitive/composite/ecdh/ecdh_xchacha_aead_private_key_manager.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("ecdh_xchacha_private_key_manager: GenerateECDHKeyPair failed: %w", err)
+	}
+
+	x25519Pub, err := cryptoutil.PublicEd25519toCurve25519(pub)
+	if err != nil {
+		return nil, fmt.Errorf("ecdh_xchacha_private_key_manager: Convert to X25519 pub key failed: %w", err)
+	}
+
+	x25519Pvt, err := cryptoutil.SecretEd25519toCurve25519(pvt)
```
troyronda, line 104: Is there no function to just generate X25519 directly?

Baha-sk: I haven't found one, I guess since it's a random []byte sequence, the closest I've seen is [HChacha()](https://github.com/golang/crypto/blob/master/chacha20/chacha_generic.go#L344), but it creates a key from another key and a nonce which is used during Chacha primitive execution.

pkg/crypto/tinkcrypto/primitive/composite/ecdh/ecdh_xchacha_aead_private_key_manager.go
Baha-sk: this function returns  key pair.

what is needed for key wrapping (ECDH) is a public/private key pair representing a recipient

troyronda: How do I use P curves with ChaCha? Why is the X25519 logic in this key manager instead of its own key manager?

proto/tink/ecdh_aead.proto
```diff
@@ -82,7 +82,7 @@ message EcdhAeadPublicKey {
 }
 
 // EcdhAeadPrivateKey represents ECDHDecrypt primitive.
-// key_type: type.hyperledger.org/hyperledger.aries.crypto.tink.EcdhAeadPrivateKey
+// key_type: type.hyperledger.org/hyperledger.aries.crypto.tink.EcdhAesAeadPrivateKey
 message EcdhAeadPrivateKey {
```
troyronda, line 86: Are you renaming these structures to include Aes?

troyronda:
We will need additional key managers to handle the other valid combinations:
- P Curves + ChaCha
- X25519 + AES

proto/tink/ecdh_aead.proto
Baha-sk: with the additional key managers that I need to add (P Curves + ChaCha and X25519 + AES) it doesn't make sense to create new proto but for keys, I will revert the proto bufs and use `EcdhAeadPublicKey ` and `EcdhAeadPrivateKey` in all the key managers

Baha-sk: the proto buf has been reverted. It's used for both NIST P curves and X25519 keys.

Baha-sk:
> We will need additional key managers and types to handle the other valid combinations:
> 
> * P Curves + ChaCha
> * X25519 + AES

done

fqutishat (approved):

PR #2454, opened 2021-01-15
feat: Presentation Exchange - BBS+ support

soluchok:
Adds BBS+ support for PE.

### DIFF
--- a/pkg/doc/presexch/definition.go
+++ b/pkg/doc/presexch/definition.go
@@ -299,7 +299,8 @@ func makeRequirement(requirements []*SubmissionRequirement, descriptors []*Input
 }
 
 // CreateVP creates verifiable presentation.
-func (pd *PresentationDefinition) CreateVP(credentials ...*verifiable.Credential) (*verifiable.Presentation, error) {
+func (pd *PresentationDefinition) CreateVP(credentials []*verifiable.Credential,
+	opts ...verifiable.CredentialOpt) (*verifiable.Presentation, error) {
 	if err := pd.ValidateSchema(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/doc/presexch/definition.go
```diff
@@ -298,8 +298,14 @@ func makeRequirement(requirements []*SubmissionRequirement, descriptors []*Input
 	return req, nil
 }
 
+// CredentialEntity contains the necessary information for processing the credential.
+type CredentialEntity struct {
```
llorllale, line 302: Can we rename this to "CredentialCandidate" to align a bit more with the text in the P-E spec? The term "candidate input" is used in section 5.3 but I think "CredentialCandidate" might be a better fit in this context.

soluchok: Sure.

pkg/doc/presexch/definition.go
```diff
@@ -298,8 +298,14 @@ func makeRequirement(requirements []*SubmissionRequirement, descriptors []*Input
 	return req, nil
 }
 
+// CredentialCandidate contains the necessary information for processing the credential.
+type CredentialCandidate struct {
+	Opts       []verifiable.CredentialOpt
```
llorllale, line 303: I don't think we need to specify options for every credential. Can you explain why we do?

All `verifiable.CredentialOpts` should apply equally to all credentials. For example the JSONLD document loader configured in the unit test.

llorllale: I would probably refactor as: `CreateVP(credentials []*verifiable.Credential, options ...verifiable.CredentialOpt)`

soluchok: ok, I will rollback it. I did it bc I don't know what else might be needed :) working with structure is easier comparing to function arguments

llorllale: > comparing to function arguments

Ehm - to be clear, I was showing an example of how I would refactor the existing `CreateVP` function in my comment above. I was not suggesting functional arguments for `CreateVP`

soluchok: Done!

fqutishat (approved):

llorllale (approved):

PR #2477, opened 2021-01-21
feat: Caching support for new formatted storage provider

### LINKED ISSUE #2470: Caching support for new formatted storage provider implementation
The old formatted storage provider implementation supported a cache. Add this to the new one too.

DRK3:

closes #2470

### DIFF
--- a/component/newstorage/formattedstore/formattedstore.go
+++ b/component/newstorage/formattedstore/formattedstore.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/newstorage"
 )
 
@@ -23,6 +24,8 @@ const (
 	failDeformat             = `failed to deformat %s "%s" returned from the underlying store: %w`
 	failQueryUnderlyingStore = "failed to query underlying store: %w"
 )
+
+var logger = log.New("formattedstore")
 
 var errInvalidQueryExpressionFormat = errors.New("invalid expression format. " +
 	"it must be in the following format: TagName:TagValue")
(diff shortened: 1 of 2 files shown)


component/newstorage/formattedstore/formattedstore.go
```diff
@@ -100,6 +140,16 @@ func (f *FormattedProvider) SetStoreConfig(name string, config newstorage.StoreC
 // The store must be created prior to calling this method.
 // If the store cannot be found, then an error wrapping ErrStoreNotFound will be returned.
 func (f *FormattedProvider) GetStoreConfig(name string) (newstorage.StoreConfiguration, error) {
+	if f.cacheProvider != nil {
+		config, err := f.cacheProvider.GetStoreConfig(name)
+		if err == nil {
+			return config, nil
+		}
+
+		logger.Debugf(`cache miss: failed to get store config for store "%s" from cache storage provider: %s`,
```
Baha-sk, line 149: no need for the message to look like an error (it would confuse ops), just a simple `cache miss: fetching data from real store` suffice.

also, you can check the type of error returned by the cache (if err != notFoundErr then it's a valid cache miss error, otherwise, print a real error?)

bstasyszyn (commented):
I'm thinking that the provider and store shouldn't know anything about the cache. The caching store should be just another implementation of the provider/store interfaces, wrapping the actual store.

sandrask:
> I'm thinking that the provider and store shouldn't know anything about the cache. The caching store should be just another implementation of the provider/store interfaces, wrapping the actual store.

I agree - then there would be no need for all of the following checks:  if f.cacheProvider != nil

component/newstorage/formattedstore/formattedstore.go
DRK3: You're right. FIxed!

DRK3:
@bstasyszyn @sandrask Great idea. I agree. However, for the sake of time, would it be acceptable to do this refactoring after I've completed with my other high-priority storage tasks? While it's not too complicated, getting all the unit tests done is fairly tedious and takes quite a bit of time. I need to make sure I get the Aries code migrated over to using the new storage interface before the next milestone.

sandrask:
> @bstasyszyn @sandrask Great idea. I agree. However, for the sake of time, would it be acceptable to do this refactoring after I've completed with my other high-priority storage tasks? While it's not too complicated, getting all the unit tests done is fairly tedious and takes quite a bit of time. I need to make sure I get the Aries code migrated over to using the new storage interface before the next milestone.

I am ok for this to be fixed later (please create issue)

sandrask (approved):

DRK3:
@sandrask @bstasyszyn Created the issue: #2479

Baha-sk (approved):

bstasyszyn (approved):

fqutishat (approved):

PR #2478, opened 2021-01-21
feat: Add support for X25519 key wrapping in ECDH Tink key

### LINKED ISSUE #2445: Add support for X25519 key wrapping in ECDH Tink key
Include support for X25519 key wrapping similar to NIST approved curves ECDH-ES key wrapping.

Baha-sk:
This change adds support for both X25519 and NIST P key wrapping
via the use of the ECDH key types and the Crypto service's
WrapKey() and UnwrapKey() functions.

This change also removes redundant ECDH key templates who
were originally intended to include KW primitive logic but
is not necessary since KW is executed outside of the Tink
key.

closes #2445


pkg/crypto/api.go
```diff
@@ -42,7 +42,8 @@ type Crypto interface {
 
 	// WrapKey will execute key wrapping of cek using apu, apv and recipient public key 'recPubKey'.
 	// 'opts' allows setting the option sender key handle using WithSender() option. It allows ECDH-1PU key wrapping
-	// (aka Authcrypt). The absence of this option uses ECDH-ES key wrapping (aka Anoncrypt).
+	// (aka Authcrypt). The absence of this option uses ECDH-ES key wrapping (aka Anoncrypt). Another option that can
+	// be used is WithXC20PKW() to instruct the WrapKey to use XC20P key wrapping instead of the default A256GCM.
```
llorllale, line 46:
```suggestion
	// be used is WithXC20PKW() to encrypt the ephemeral key (`epk`) with XC20P instead of the default A256GCM.
```

pkg/crypto/api.go
```diff
@@ -51,7 +52,8 @@ type Crypto interface {
 
 	// UnwrapKey unwraps a key in recWK using recipient private key kh.
 	// 'opts' allows setting the option sender key handle using WithSender() option. It allows ECDH-1PU key unwrapping
-	// (aka Authcrypt). The absence of this option uses ECDH-ES key unwrapping (aka Anoncrypt).
+	// (aka Authcrypt). The absence of this option uses ECDH-ES key unwrapping (aka Anoncrypt). There is no need to
+	// use WithXC20PKW() for UnwrapKey since the function will use the wrapping algorithm based on recWK.Alg.
```
llorllale, line 56: > 	// UnwrapKey unwraps a key in recWK using recipient private key kh.
	// 'opts' allows setting the option sender key handle using WithSender() option. It allows ECDH-1PU key unwrapping
	// (aka Authcrypt). The absence of this option uses ECDH-ES key unwrapping (aka Anoncrypt).

It is strange to require options to instruct `UnwrapKey` on which algorithm to use to do the unwrapping. There is sufficient information in the recipient structure to accomplish this.

Is there really a need for `WithSender()` and other options here?

pkg/crypto/tinkcrypto/crypto.go
```diff
@@ -221,34 +226,34 @@ func (t *Crypto) WrapKey(cek, apu, apv []byte, recPubKey *cryptoapi.PublicKey,
...
 }
 
 // UnwrapKey unwraps a key in recWK using ECDH (ES or 1PU) with recipient private key kh.
-// The optional 'wrapKeyOpts' specifies the sender kh for 1PU key unwrapping.
-// Note, if the option was used in WrapKey(), then it must be set here as well for a successful unwrapping.
 // This function is used with the following parameters:
-//  - Key Unwrapping: ECDH-ES (no options)/ECDH-1PU (using crypto.WithSender() option) over A256KW as
-// 		per https://tools.ietf.org/html/rfc7518#appendix-A.2
-//  - KDF: Concat KDF as per https://tools.ietf.org/html/rfc7518#section-4.6
+//  - Key Unwrapping: `ECDH-ES` (no options) or `ECDH-1PU` (using crypto.WithSender() option in wrapKeyOpts) over either
```
llorllale, line 239: Like I said above, I don't understand why we need to pass in options for the algorithms to use when all this information is in the recipient header.

pkg/crypto/api.go
Baha-sk: to be more specific: it wrap the cek using XC20P instead of A256GCM, the ephemeral key (epk) is used to along with the recipient key (and sender for 1PU) to derived the wrapping key (kek) used to wrap the cek.

WithXC20PKW() option is for the wrapping (encryption) alg used to wrap the cek, not the epk

pkg/crypto/api.go
Baha-sk: `WithSender()` option is used for both WrapKey() and UnwrapKey() to instruct the functions to use ECDH-1PU key derivation instead of ECDH-ES.

for WrapKey(), the option must contain the *keyset.Handle to the private key (since the sender owns the key).
for UnwrapKey(), it must contain the crypto.PublicKey of the sender (since unwrapping uses the sender public key only)

**to answer the question:** why this option is needed for UnwrapKey() since `There is sufficient information in the recipient structure to accomplish this`?
**answer:** the recipient structure (or JWE recipient headers) does not contain the sender public key. As per the [ECDH-1PU draft](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03), the only information about the sender in the envelope is the `skid` header. This is to avoid sending the sender key information within the envelope. It means the recipient should resolve this `skid` from somewhere to fetch the sender public key. I've contacted the draft author about this and asked him the following question:
```
The `skid` header field mentioned is for the JWE envelope level (not the recipient level), is this right? Also, I think the application should be able to resolve this id to retrieve the sender’s public key and do the 1PU key derivation. Is this correct? In other words, the spec does not enforce this key resolution, it’s rather up to application to solve this key resolution. I just want to double check with you as the field has no mention of this.
```
his answer was:
```
Yes, exactly. The “skid” header is the direct analogue of the existing “kid” header. It would go in the JWE Protected Header, which is the same for all recipients. How the application resolves the skid to a key is deliberately not specified and should be defined by particular applications.

You may ask why I did not define “sjwk”, “sjku”, “sx5c” etc headers to mirror all the various per-recipient headers for resolving a key? The reason is that I actually believe that having different headers for all these cases was a mistake. It encourages implementations to automatically handle all of these cases, but this is not safe in general - e.g., you generally shouldn’t trust a key in a “jwk” header, and you definitely shouldn’t blindly make HTTP requests to a “jku” header. By having a single “skid” header, whose contents is opaque and not specified, applications must agree between themselves what a safe way to resolve the key material is.
```

this is the only reason why I've added [thirdPartyKS](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/didcomm/packer/authcrypt/pack.go#L45) in the authcrypt JWE packer. It must be pre-populated, on the recipient side, with the sender's public key prior to doing authcrypt.Unpack().

pkg/crypto/tinkcrypto/crypto.go
Baha-sk: see my answer to the previous question above ..

pkg/crypto/api.go
llorllale: My bad - I got confused and thought the `skid` was included in each recipient's header. I also forgot that `Crypto` is in a layer lower than JWEs.

llorllale (approved):

fqutishat (approved):

PR #2483, opened 2021-01-24
feat: Batched storage provider implementing new storage interface

### LINKED ISSUE #2471: Batch service support for new formatted storage provider implementation
The new formatted storage provider implementation has an explicit batch method, but the old formatted provider implementation supported the ability to batch `Put`s and `Delete`s automatically. Add this optional feature to the new one, and make sure it takes advantage of the new Batch method that's now in the interface.

DRK3:
- This new storage provider wraps around any other storage provider and does automatic batching of data.
- Updated new storage provider interface to support functionality needed for clients to force flush data. New GetOpenStores method allows a client to flush all stores at once.
- Also made a small adjustment to the new storage interface documentation to make it explicit that empty store names are not supported. Added a common unit test to ensure this behaviour is consistent across store implementations.
- Updated in-memory storage implementation to conform to the previous point.
- Moved mock storage provider from formatted store test file to a separate package since it's now being used by multiple test files in different packages.

closes #2471 by allowing a formatted provider to be passed in as the underlying provider to this new batched storage provider introduced in this PR.


component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,360 @@
...
+	err := p.underlyingProvider.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close underlying provider: %w", err)
+	}
+
+	return nil
+}
+
+// Flush forces all open stores to flush their data immediately.
+func (p *Provider) Flush() error {
```
fqutishat, line 136: the flush should not be part of provider

fqutishat: @troyronda @bstasyszyn  @sandrask

DRK3: The reason this is there is for edge-agent in Trustbloc - it needs that ability to force a flush at times (which you discovered during the last milestone). It needs to flush all the stores. If it's not a part of the Provider, I'm not sure how I can do that.

fqutishat: is this method part of Provider interface ?

fqutishat: @DRK3

DRK3: @fqutishat No.

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,380 @@
...
+
+// Provider is a newstorage.Provider that allows for data to be automatically batched.
+// It acts as a wrapper around another storage provider (typically, one that would benefit from batching).
+// TODO (#2484): Support batching across multiple stores for storage providers that inherently support this
+//  (like the EDV REST provider, which uses a single underlying database (vault) for all of its stores).
+type Provider struct {
+	underlyingProvider newstorage.Provider
+	openStores         map[string]*store
+	batchSizeLimit     int
+	sync.RWMutex
```
bstasyszyn, line 35: This should probably be a private member, otherwise you're exposing the Lock/RLock functions to clients.

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,380 @@
...
+	return config, nil
+}
+
+// GetOpenStores returns all currently open stores.
+func (p *Provider) GetOpenStores() []newstorage.Store {
+	openStores := make([]newstorage.Store, len(p.openStores))
+
+	var counter int
+
+	for _, openStore := range p.openStores {
```
bstasyszyn, line 114: Need a read lock.

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,380 @@
...
+	err := p.underlyingProvider.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close underlying provider: %w", err)
+	}
+
+	return nil
+}
+
+// Flush forces all open stores to flush their data immediately.
+func (p *Provider) Flush() error {
```
bstasyszyn, line 149: You're still exposing Flush on the provider?

component/newstorage/batchedstore/batchedstore.go
DRK3: @fqutishat It's part of the interface now.

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,380 @@
...
+
+	return nil
+}
+
+// TODO (#2485): Check data in current batch before resorting to a flush.
+func (s *store) Get(key string) ([]byte, error) {
+	s.Lock()
+	defer s.Unlock()
+
+	if len(s.currentBatch) > 0 {
```
bstasyszyn, line 216: I would put this check inside of Flush since it's repeated in multiple function. Also, you're acquiring a write lock each time you check currentBatch. The check should be done using a read lock and then use a write lock when modifying the value.

fqutishat (approved):

component/newstorage/formattedstore/formattedstore.go
```diff
@@ -175,6 +195,20 @@ func (f *FormattedProvider) GetStoreConfig(name string) (newstorage.StoreConfigu
 	return newstorage.StoreConfiguration{TagNames: tagNames}, nil
 }
 
+// GetOpenStores returns all currently open stores.
+func (f *FormattedProvider) GetOpenStores() []newstorage.Store {
+	openStores := make([]newstorage.Store, len(f.openStores))
+
+	var counter int
+
+	for _, openStore := range f.openStores {
```
bstasyszyn, line 204: Need a read lock.

component/newstorage/mem/mem.go
```diff
@@ -92,15 +96,25 @@ func (p *Provider) GetStoreConfig(name string) (newstorage.StoreConfiguration, e
...
-// Close closes all stores created under this store provider. All data will be deleted.
-func (p *Provider) Close() error {
-	for _, memStore := range p.dbs {
-		err := memStore.Close()
-		if err != nil {
-			return fmt.Errorf("failed to close a mem store: %w", err)
-		}
+// GetOpenStores returns all currently open stores.
+func (p *Provider) GetOpenStores() []newstorage.Store {
+	openStores := make([]newstorage.Store, len(p.dbs))
```
bstasyszyn, line 101: Read lock

component/newstorage/mem/mem.go
```diff
@@ -92,15 +96,25 @@ func (p *Provider) GetStoreConfig(name string) (newstorage.StoreConfiguration, e
...
+
+	var counter int
+
+	for _, db := range p.dbs {
```
bstasyszyn, line 105: You can do this:

```
for i, db := range p.dbs{
  openStores[i] = db
}

```

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

component/newstorage/batchedstore/batchedstore.go
DRK3: Oops - I forgot to remove that. Done.

component/newstorage/mem/mem.go
DRK3: `dbs` is a map, not an array, so this won't work.

component/newstorage/formattedstore/formattedstore.go
DRK3: Done

component/newstorage/mem/mem.go
DRK3: Done

component/newstorage/batchedstore/batchedstore.go
DRK3: Done.

component/newstorage/batchedstore/batchedstore.go
bstasyszyn: I would make this a read lock: RLock()

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,361 @@
...
+
+	s.currentBatch = append(s.currentBatch, newstorage.Operation{
+		Key:   key,
+		Value: value,
+		Tags:  tags,
+	})
+
+	s.Unlock()
+
+	s.RLock()
```
bstasyszyn, line 186: Don't need the extra read lock - just move the assignment inside the write lock above.

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,361 @@
...
+
+	s.currentBatch = append(s.currentBatch, newstorage.Operation{
+		Key:   key,
+		Value: nil,
+	})
+
+	s.Unlock()
+
+	s.RLock()
+	currentBatchLength := len(s.currentBatch)
```
bstasyszyn, line 272: Move assignment before Unlock above

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,361 @@
...
+		s.currentBatch = append(s.currentBatch, newstorage.Operation{
+			Key:   operation.Key,
+			Value: operation.Value,
+			Tags:  operation.Tags,
+		})
+
+		s.Unlock()
+
+		s.RLock()
+		currentBatchLength := len(s.currentBatch)
```
bstasyszyn, line 304: Move assignment before Unlock above

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,361 @@
...
+	s.Lock()
+
+	var timeToFlush time.Duration
+
+	if len(s.currentBatch) > 0 {
+		startFlushTime := time.Now()
+
+		err := s.underlyingStore.Batch(s.currentBatch)
+		if err != nil {
+			return fmt.Errorf("failure while executing batched operations: %w", err)
```
bstasyszyn, line 348: Need to unlock - or use defer Unlock() above.

component/newstorage/formattedstore/formattedstore.go
```diff
@@ -175,6 +195,23 @@ func (f *FormattedProvider) GetStoreConfig(name string) (newstorage.StoreConfigu
 	return newstorage.StoreConfiguration{TagNames: tagNames}, nil
 }
 
+// GetOpenStores returns all currently open stores.
+func (f *FormattedProvider) GetOpenStores() []newstorage.Store {
+	openStores := make([]newstorage.Store, len(f.openStores))
+
+	var counter int
+
+	f.lock.Lock()
```
bstasyszyn, line 204: Lock should be at the top since you're referencing f.openStores. Also, this should be a read lock RLock().

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,361 @@
...
+				return fmt.Errorf(failFlush, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (s *store) Close() error {
+	s.Lock()
```
bstasyszyn, line 319: You don't need a lock around the close

component/newstorage/mem/mem.go
bstasyszyn: Lock needs to be at the top since you're doing len(p.dbs).

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

component/newstorage/mem/mem.go
DRK3: Done. Also fixed in formattedstore.go where I made the same mistake

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

DRK3:
https://github.com/hyperledger/aries-framework-go/pull/2483#discussion_r565344028 Done

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

component/newstorage/batchedstore/batchedstore.go
DRK3: Oops. Done. I decided to use an unlock instead of defer so that the logging at the bottom of the function won't unnecessarily delay the release of the lock.

component/newstorage/formattedstore/formattedstore.go
DRK3: Done.

DRK3: Also fixed the same mistake I made in batchedstore.go

component/newstorage/batchedstore/batchedstore.go
DRK3: Done

DRK3:
https://github.com/hyperledger/aries-framework-go/pull/2483#discussion_r565344028 done

bstasyszyn (approved):

sandrask (approved):

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,352 @@
...
+	for _, operation := range operations {
+		if operation.Key == "" {
+			return errors.New("an operation's key was empty")
+		}
+	}
+
+	for _, operation := range operations {
+		s.Lock()
+
+		s.currentBatch = append(s.currentBatch, newstorage.Operation{
```
Moopli, line 291: Curious why you copy the operation here

DRK3, line 291: @Moopli Thanks for catching this - I don't know why I copied the operation object manually like this. I should just be appending `operation` here. I will fix it now

DRK3: That was an oversight! It's fixed now.

component/newstorage/batchedstore/batchedstore.go
```diff
@@ -0,0 +1,348 @@
...
+
+	err = s.underlyingStore.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close underlying store: %w", err)
+	}
+
+	return nil
+}
+
+func (s *store) Flush() error {
```
Moopli, line 324: @DRK3 @bstasyszyn multiple concurrent `Put()`, `Delete()`, and `Batch()` calls can cause `Flush()` to be executed multiple times in parallel, and if another `Put()` or `Batch()` partially fills `s.currentBatch` after one of those Flushes empties it, then the next `Flush()` to enter its critical section would flush `s.currentBatch` before it's been filled again, when the intention behind the call to `Flush()` was to flush a full buffer. 

It's not a correctness issue, all concurrent calls to `Put()`, `Delete()` and `Batch()` will have each of their operations flushed exactly once, so I'm not sure it's a concern, but it will occasionally spread operations out across extraneous flushes. Might be a performance concern depending on how we use it.

DRK3: @Moopli Thanks for catching this. Also as you pointed out in a separate chat, it would be possible for log statements to be inappropriately printed out when there are no operations to flush.

I just pushed a fix to this PR that should address that. Let me know how it looks @Moopli @bstasyszyn

Issue #857, opened 2019-11-22
Generic JWE support for envelopes (future)


troyronda:
Extended from #815 

JWE packer should support the public key formats that are allowed in DID documents:

- curve25519
- sec256k1 (maybe future - 2nd priority)
- RSA (maybe future - 3rd priority)

For key encryption algorithms, we should support:
- ECDH-ES+A256KW
- ECDH-SS+A256KW ?

TBD (maybe future):
- [ECDH-SS+XC20PKW & ECDH-SS+C20PKW](https://github.com/hyperledger/aries-rfcs/issues/133#issuecomment-522009615) (TBD: we will need to see what happens with this spec and the Aries issue).

PR #2486, opened 2021-01-25
feat: add KMS key type for X25519 key

### LINKED ISSUE #475: Separate all crypto implementation details from Pack/Unpack

### LINKED ISSUE #596: Refactor Key structure in KMS (from KeyPair to KeySet)

### LINKED ISSUE #815: Generic JWE support for envelopes

### LINKED ISSUE #1684: Add Chacha JWE Content encryption

### LINKED ISSUE #2447: Add new ECDH KMS key types
Currently the ECDH KMS key types in the framework are limited to NIST P curved keys as KW + AES content encryption.
With the addition of XChacha ECDH keys, update the KMS key types to support X25519 KW mixed with either AES or XChacha20Poly1305 content encryptions

Baha-sk:
This is the last change about X25519 keys for ECDH KW.
It updates the old ECDH key types into the new type names
and add the X25519 key type as well.

It includes JWE encryption/decryption updates to support
XC20P content encryption along with recipients kw using both
NIST P curved keys and X25519 keys.

Also part of this change is the removal of remnant code
from legacyKMS which was removed from the framework last
year.

closes #2447
closes #1684
closes #815
closes #475
closes #596
also part of #857


pkg/didcomm/packer/authcrypt/pack_test.go
```diff
@@ -79,8 +79,8 @@ func TestAuthryptPackerUsingKeysWithDifferentCurvesSuccess(t *testing.T) {
 	_, recipientsKey1, keyHandles1 := createRecipients(t, k, 1)
 	// since authcrypt does ECDH kw using the sender key, the recipient keys must be on the same curve as the sender's.
```
DRK3, line 80: Looks like you need to update this comment now

pkg/doc/util/jwkkid/kid_creator.go
```diff
@@ -133,3 +142,52 @@ func getCurve(kt kms.KeyType) elliptic.Curve {
...
+	thumbprint := thumbprintX25519(jwk)
+
+	return base64.RawURLEncoding.EncodeToString(thumbprint), nil
+}
+
+func buildX25519JWK(keyBytes []byte) (string, error) {
+	const x25519ThumbprintTemplate = `{"crv":"X25519","kty":"OKP",x":"%s"}`
+
+	if len(keyBytes) > cryptoutil.Curve25519KeySize {
+		return "", errors.New("buildX25519JWK: invalid ECDH X25519 key")
```
DRK3, line 177: Consider returning more specific error here based on the if statement condition

pkg/kms/localkms/pubkey_writer.go
```diff
@@ -97,7 +98,8 @@ func write(w io.Writer, msg *tinkpb.Keyset) error {
 func writePubKey(w io.Writer, key *tinkpb.Keyset_Key) (bool, error) {
 	var marshaledRawPubKey []byte
 
-	// TODO add other key types than the ones below and other than ecdhAESPublicKeyTypeURL
+	// TODO add other key types than the ones below and other than nistPECDHKWPublicKeyTypeURL and
```
DRK3, line 101: Issue #?

DRK3 (approved):

DRK3:
Left minor comments only

pkg/kms/localkms/pubkey_writer.go
Baha-sk: there's no specific issue # for this, BBS+ is the next new key type to be added in the kms, I'll remove this comment when BBS+ key type is added to the kms (and the same will be added when the Bitcoin key type is added later)

pkg/doc/util/jwkkid/kid_creator.go
Baha-sk: This error will technically never occur since the KMS creates valid sized keys, this condition is added as a precautionary measure.
+
The error will not be reused anywhere else, it's not worth creating a constant error for it.

pkg/didcomm/packer/authcrypt/pack_test.go
Baha-sk: sure.. I'll update..

Baha-sk:
This PR requires additional updates in doc Jose encrypter/decrypter to account for X25519 support for packing (recipient key derivation using X25519 keys, not only NIST P curved keys). I will be pushing an update soon.

pkg/doc/util/jwkkid/kid_creator.go
DRK3: I didn't mean to create a constant error for it, I just meant that the error message could be more specific than "invalid ECDH X25519 key".

Baha-sk: crypto logs should not provide hints on why an operation fails, this is why I don't log the root cause of the error.

pkg/didcomm/packer/authcrypt/pack_test.go
Baha-sk: updated

pkg/doc/util/jwkkid/kid_creator.go
llorllale: > crypto logs should not provide hints on why an operation fails, this is why I don't log the root cause of the error.

You are though because `buildX25519JWK: invalid ECDH X25519 key` is a unique string in the code base in this PR and the code is open source.

Baha-sk: >You are though because buildX25519JWK: invalid ECDH X25519 key is a unique string in the code base in this PR and the code is open source.

what do you suggest the log string should be @llorllale ? I can remove `ECDH X25519` if it makes the logs more obscure.
We need to be able to trace the function for ops troubleshooting at least.

llorllale: I am fine with the log string as-is. I just wanted to point out that it is in fact still an oracle.

fqutishat (approved):

llorllale (approved):

Issue #1833, opened 2020-05-14
Proof shouldn't override/hardcode jsonld contexts


sudeshrshetty:
It is json ld doc author's responsibility to pass proof related contexts. Aries shouldn't override/hardcode contexts.

Currently "JsonWebSignature2020" is treated as invalid RDF even though [proper context](https://trustbloc.github.io/context/vc/credentials-v1.jsonld) is available in VC.
Because aries overrides it and add its own context which doesn't have JsonWebSignature2020.

https://github.com/hyperledger/aries-framework-go/blob/b077c8b62a512a89a9c8caa7ca95ee397f458380/pkg/doc/signature/proof/jws.go#L109

kdimak:
It was implemented in accordance with the reference implementation by Digitalbazaar.

https://github.com/digitalbazaar/jsonld-signatures/blob/521fe3a56e60d4210a45acaab9dcaf64acad739c/lib/ProofSet.js#L304

If we want to be compatible with their implementation, we should leave the context substitution as is by default.

llorllale:
@kdimak @sudeshrshetty @fqutishat I'm not sure about the implications on interop, but here is an example of an JSON-LD document for which downstream projects are verifying proofs using `verifier.DocumentVerifier`:

<details><summary>Example</summary>

```json
{
    "@context": "https://w3id.org/security/v2",
    "id": "http://www.example.org/foo/documents/a3480d17-df7f-449f-9480-e2c35e20a865",
    "allowedAction": ["read", "write"],
    "invocationTarget": {
        "ID": "http://www.example.org/foo/documents/a3480d17-df7f-449f-9480-e2c35e20a865",
        "Type": "urn:edv:document"
    },
    "proof": [{
        "created": "2020-12-04T15:28:14.673975717-05:00",
        "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..6OfIULug35ZmoU7lysChVpD6sjYfV71UwxqIZ8u0woYSIzRtzCo3MsZJw6cGIZMEaMssnQyRqIzo8B0yHEL2Dw",
        "nonce": "da7CcJahAdFG0GXN-JnS2f2mywcFNtaLyXtGVqku2DwVwUaJbGpUQjhlNi5kDbS4ZMi2cNhEN5ac6LponS-C9w",
        "proofPurpose": "capabilityDelegation",
        "type": "Ed25519Signature2018",
[...]

PR #2487, opened 2021-01-26
refactor: caching in jsonld processor

aholovko:
This PR is a proposal on refactoring caching functionality in jsold processor.

There are 2 major issues that I have with the current approach:

* it doesn't save from the repetitive code

Using `WithDocumentLoaderCache` might look like a robust solution but in practice, it requires a lot of boilerplate code (examples in Trustbloc projects: [here](https://github.com/trustbloc/kms/blob/main/cmd/kms-rest/startcmd/lddocument_loader.go), [here](https://github.com/trustbloc/edv/blob/main/pkg/auth/zcapld/context.go), and similar should be added to [edge-agent](https://github.com/trustbloc/edge-agent)).

* problems with API consistency

Cache is preloaded in one case and is not in another: [here](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/signature/proof/jws.go#L123) and [here](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/signature/proof/jws.go#L132). It's also not necessarily clear that under the hood framework adds some additional opts to ones that you have specified explicitly. There is also hardcoded usage of `ld.NewCachingDocumentLoader`. ~The latter is a problem because I might want to use `ld.NewCachingDocumentLoader` with preloaded contexts and be able to pass it to the processor using `WithDocumentLoader(...)` - the current implementation will wrap it with one more `ld.NewCachingDocumentLoader`.~  UPD: existing code has guards against such scenario [here](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/signature/jsonld/processor.go#L423)

The current PR simplifies logic and adds more consistency to API. `ld.NewCachingDocumentLoader` with predefined contexts in the cache is used as a default. You can override it to any loader you want with options. I also moved `cache.go` from `proof` package to `jsonld` - for me it looks more like a processor concern.


llorllale:
Does this fix #1833?

aholovko:
> Does this fix #1833?

It looks like the mentioned issue describes two problems:

* handling of "JsonWebSignature2020" context

targeted by this [PR](https://github.com/hyperledger/aries-framework-go/pull/1834)

* panic when the loader with pre-loaded contexts is used

I am trying to reproduce this panic with the old code. And then check it with the updated one (but I think that the problem should go).

@llorllale plz correct me if I didn't get it right.

aholovko:
During the discussion with @kdimak, we decided that the caching functionality requires more consideration and also should account for other issues like the one mentioned by George. I am closing the current PR as it doesn't bring enough benefits to justify the risk of breaking the core stuff.

PR #2491, opened 2021-02-01
feat: presexch: copy @context and type from doc to BBS+ JSON-LD frame

### LINKED ISSUE #2488: presexch: copy "@context" and "type" from doc to BBS+ JSON-LD frame
When building BBS+ JSON-LD on VP creation at Presentation Definition, we need to copy `@context` and `type` fields from the input VC document if we apply JSON-LD framing on the particular JSON-LD level. Otherwise, we get still a valid JSON-LD frame document but its appliance can lead to an invalid output VP document.

kdimak:
For every field from constraints, do not just set `@explicit` in the reveal doc, but also copy special fields like `type` and `context` from the source VC document to the reveal document if present. Also, we need to copy the parent special fields if present as well.

closes #2488

### DIFF
--- a/pkg/doc/presexch/definition.go
+++ b/pkg/doc/presexch/definition.go
@@ -608,7 +608,7 @@ func createNewCredential(constraints *Constraints, src, limitedCred []byte,
 	var (
 		BBSSupport          = hasBBS(credential)
 		modifiedByPredicate bool
-		explicitPaths       []string
+		explicitPaths       = make(map[string]bool)
 	)
 
 	for _, f := range constraints.Fields {
@@ -648,8 +648,8 @@ func createNewCredential(constraints *Constraints, src, limitedCred []byte,
 
 			if constraints.LimitDisclosure && BBSSupport {
 				chunks := strings.Split(path[0], ".")
-				chunks[len(chunks)-1] = "@explicit"
-				explicitPaths = append(explicitPaths, strings.Join(chunks, "."))
+				explicitPath := strings.Join(chunks[:len(chunks)-1], ".")
+				explicitPaths[explicitPath] = true
 			}
 
 			limitedCred, err = sjson.SetBytes(limitedCred, path[0], val)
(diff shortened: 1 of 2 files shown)


pkg/doc/presexch/definition.go
```diff
@@ -663,25 +663,55 @@ func createNewCredential(constraints *Constraints, src, limitedCred []byte,
...
 
-	doc["@explicit"] = true
-
 	return credential.GenerateBBSSelectiveDisclosure(doc, []byte(uuid.New().String()), opts...)
 }
 
+func enhanceRevealDoc(explicitPaths map[string]bool, limitedCred, vcBytes []byte) ([]byte, error) {
+	var err error
+
+	for path := range explicitPaths {
```
soluchok, line 682: We are missing the part when the path is few levels deep.
Let's say we have path `subject.user.address`.
The current implementation sets values only for the full path.
e.g 
```
subject.user.address.@explicit
subject.user.address.type
subject.user.address.@context
```

In that case, we are missing everything between.
e.g
```
# root
@explicit
type
@context

# first level
subject.@explicit
subject.@type
subject.@context

# second level
subject.user.@explicit
subject.user.@type
subject.user.@context
```

pkg/doc/presexch/definition.go
```diff
@@ -663,23 +663,79 @@ func createNewCredential(constraints *Constraints, src, limitedCred []byte,
...
 
-	return credential.GenerateBBSSelectiveDisclosure(doc, []byte(uuid.New().String()), opts...)
+		specialFieldValue := gjson.GetBytes(vcBytes, specialFieldPath)
+		if specialFieldValue.Type == gjson.Null {
+			continue
+		}
+
+		limitedCred, err = sjson.SetBytes(limitedCred, specialFieldPath, specialFieldValue.Value())
+		if err != nil {
+			return nil, err
```
Baha-sk, line 734: can these err types be wrapped?

Baha-sk (approved):

soluchok (approved):

pkg/doc/presexch/definition.go
kdimak: I just followed the error handling pattern in this source file...

pkg/doc/presexch/definition.go
kdimak: Fixed!

fqutishat (approved):

PR #2498, opened 2021-02-03
fix: json tags

llorllale:
* fix: json tags on did.Service
* fix: json tag on outofband.Invitation

### DIFF
--- a/pkg/didcomm/protocol/outofband/models.go
+++ b/pkg/didcomm/protocol/outofband/models.go
@@ -14,7 +14,7 @@ type Request struct {
 	Type     string                  `json:"@type"`
 	Label    string                  `json:"label,omitempty"`
 	Goal     string                  `json:"goal,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/didcomm/protocol/outofband/models.go
```diff
@@ -27,5 +27,5 @@ type Invitation struct {
 	Goal      string        `json:"goal,omitempty"`
 	GoalCode  string        `json:"goal-code,omitempty"`
```
Moopli, line 28: Shouldn't this be json tag `goal_code`
https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md#messages

llorllale: Fixed

fqutishat (approved):

soluchok (approved):

Baha-sk (approved):

Moopli (approved):

PR #2501, opened 2021-02-03
feat: EDV REST Provider implementation of new storage interface

### LINKED ISSUE #2468: EDV REST provider implementation of new storage interface
EDV REST Provider implementation of https://github.com/hyperledger/aries-framework-go/blob/master/pkg/newstorage/newstorage.go

DRK3:
The EDV encryption format can also used in other store types via the formatted store.

closes #2468


component/newstorage/batchedstore/batchedstore_test.go
```diff
@@ -7,76 +7,19 @@ SPDX-License-Identifier: Apache-2.0
...
-		formattedTags[i] = newstorage.Tag{
-			Name:  base64.StdEncoding.EncodeToString([]byte(tag.Name)),
-			Value: base64.StdEncoding.EncodeToString([]byte(tag.Value)),
-		}
-	}
-
-	return formattedTags, nil
-}
-
-func (b *base64Formatter) DeformatKey(formattedKey string) (string, error) {
```
DRK3, line 48: Moved these to a separate package since they're shared across a few packages now

component/newstorage/edv/models.go
```diff
@@ -0,0 +1,96 @@
...
+const (
+	// upsertDocumentVaultOperation represents an upsert operation to be performed in a batch.
+	upsertDocumentVaultOperation = "upsert"
+	// deleteDocumentVaultOperation represents a delete operation to be performed in a batch.
+	deleteDocumentVaultOperation = "delete"
+)
+
+// vaultOperation represents an upsert or delete operation to be performed in a vault.
+// This is currently non-standard and should only be used with an EDV server that supports it.
+type vaultOperation struct {
```
DRK3, line 92: The format of this vault operation will probably change when (if?) a batch endpoint gets added to the EDV spec.

component/newstorage/edv/restprovider.go
```diff
@@ -0,0 +1,620 @@
...
+
+	returnFullDocumentsOnQuery    bool
+	batchEndpointExtensionEnabled bool
+}
+
+// NewRESTProvider returns a new RESTProvider. edvServerURL is the base URL for the EDV server.
+// vaultID is the ID of the vault where this provider will store data. The vault must be created in advance, and since
+// the EDV REST API does not provide a method to check if a vault with a given ID exists, any errors due to a
+// non-existent vault will be deferred until calls are actually made to it in the store.
+func NewRESTProvider(edvServerURL, vaultID string, formatter *EncryptedFormatter, options ...Option) *RESTProvider {
```
DRK3, line 86: One major difference between this new RESTProvider implementation and the old one is that this new one now directly takes in an EncryptedFormatter instead of expecting to be used as an underlying provider in a format provider. This has made the code much cleaner - before the RESTProvider was somewhat confusing since the formatted provider and RESTProvider both needed MAC crypto objects passed in, which I always thought was kind of weird. Now if you want to use an EDV server, you just create a REST provider with an Encrypted Formatter.

The Encrypted Formatter can still be used with a formatted store if someone wanted to use the EDV data model with some other type of storage provider. (allowing for this use case was one of my original requirements, so I needed to maintain that functionality)

component/newstorage/formattedstore/formattedstore.go
```diff
@@ -20,25 +21,27 @@ const (
...
 
-var errInvalidQueryExpressionFormat = errors.New("invalid expression format. " +
-	"it must be in the following format: TagName:TagValue")
+	errEmptyKey                     = errors.New("key cannot be empty")
+	errInvalidQueryExpressionFormat = errors.New("invalid expression format. " +
+		"it must be in the following format: TagName:TagValue")
+)
 
 // Formatter represents a type that can convert data between two formats.
 type Formatter interface {
```
DRK3, line 40: I had to change the Formatter interface in order to accommodate the EDV Encrypted Formatter, which is more "restrictive" due to the fact that key and tag formatting is not reversible (as they use hashing functions). You'll find that some of the format provider functions have been changed to accommodate this. I left some comments in the code explaining why certain things here work the way they do. The whole idea here is to allow someone to use the Encrypted Formatted with any store, not just the EDV REST provider.. You'll find unit tests demonstrating the use of an Encrypted Formatter + mem-store combo.

(while you could theoretically pass the EDV REST provider in as the underlying store to a formatted provider, there's no reason to, as the REST provider already requires the encrypted formatter... so you would end up with double formatting/deformatting operations)

scripts/check_unit.sh
```diff
@@ -21,7 +21,39 @@ if [ -f profile.out ]; then
 fi
 }
 
-# Running aries-framework-go unit test
+# First argument is the exit code.
```
DRK3, line 24: Unfortunately I couldn't get https://github.com/ory/dockertest to work properly with the EDV image. No matter what I tried to do, it refused to map the port I set it to so that it could be reached from the running go tests. Not sure why this is, since I was able to get CouchDB's port mapped just fine.

So, I had to resort to starting them in the scripts. All this code basically ensures that:

1. The containers are removed after the test is finished
2. If the containers are already running, they are killed automatically
3. Unexpected errors aren't swallowed

fqutishat (approved):

bstasyszyn (approved):

PR #2505, opened 2021-02-04
chore: Use proper revision for new storage interface module imports

DRK3:
- Use proper revision for new storage interface module imports.
- Also removed replace directives.
These changes are needed to make these new storage implementations importable.


DRK3:
I ran `go mod tidy` in each of the go modules, which is why you see so many line deletions

bstasyszyn (approved):

fqutishat (approved):

PR #2508, opened 2021-02-04
fix: Incorrect checks in new in-memory iterator implementation

DRK3:

### DIFF
--- a/component/newstorage/mem/mem.go
+++ b/component/newstorage/mem/mem.go
@@ -354,6 +354,7 @@ type memIterator struct {
 // Next moves the pointer to the next entry in the iterator. It returns false if the iterator is exhausted.
 func (m *memIterator) Next() (bool, error) {
 	if len(m.dbEntries) == m.currentIndex || len(m.dbEntries) == 0 {
+		m.dbEntries = nil
 		return false, nil
 	}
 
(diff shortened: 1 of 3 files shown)


bstasyszyn (approved):

Baha-sk (approved):

component/newstorage/mem/mem.go
```diff
@@ -354,6 +354,7 @@ type memIterator struct {
 // Next moves the pointer to the next entry in the iterator. It returns false if the iterator is exhausted.
 func (m *memIterator) Next() (bool, error) {
 	if len(m.dbEntries) == m.currentIndex || len(m.dbEntries) == 0 {
+		m.dbEntries = nil
```
sudeshrshetty, line 357: could be potential race condition situation

bstasyszyn: Shouldn't the iterator be 'owned' by the caller of 'Query'? In which case the iterator doesn't need to be thread-safe.

DRK3: Should be fixed now... let me know if you see any other issues

bstasyszyn (approved):

component/newstorage/mem/mem.go
DRK3: > Shouldn't the iterator be 'owned' by the caller of 'Query'? In which case the iterator doesn't need to be thread-safe.

I would think so as well... while technically possible to create the problematic situation, I don't see how it would happen normally.

sudeshrshetty (approved):

fqutishat (approved):

PR #2519, opened 2021-02-06
feat: Additional common storage tests

DRK3:


DRK3:
I took these tests from the Aries-Framework-Go-Ext CouchDB implementation I did recently - they were very general tests and really belonged here instead. After this PR I will update Ext to use this common test suite and remove the redundant tests.

troyronda (approved):

PR #2523, opened 2021-02-09
feat: add integrity headers in JWE/DIDComm

### LINKED ISSUE #2513: DIDComm: Integrity checks for JWE Anoncrypt and Authcrypt packers
The packers are missing the recipients integrity checks as per [RFC 334](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope)

Need to update AAD and recipients' APU/APV fields to be compliant with the RFC

Baha-sk:
This changes adds JWE integrity checks to help verify the recipients and/or the sender.
It also fixes the encrypter's AEAD's AAD content for single recipient envelopes where it used to contain
only the two main JWE Protected headers (enc and typ). It now includes the single recipient's
merged headers into the protected header (ie: epk, apu, apv, typ, enc, kid and alg).

closes #2513

### DIFF
--- a/pkg/doc/jose/decrypter.go
+++ b/pkg/doc/jose/decrypter.go
@@ -8,7 +8,6 @@ package jose
 
 import (
 	"crypto/ecdsa"
-	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -93,6 +92,16 @@ func (jd *JWEDecrypt) Decrypt(jwe *JSONWebEncryption) ([]byte, error) {
 		return nil, fmt.Errorf("jwedecrypt: %w", err)
 	}
 
+	if len(recWK) == 1 {
+		// ensure EPK is marshalled the same way as during encryption since it is merged into ProtectHeaders.
+		marshalledEPK, err := convertRecEPKToMarshalledJWK(recWK[0])
+		if err != nil {
+			return nil, fmt.Errorf("jwedecrypt: %w", err)
+		}
+
+		jwe.ProtectedHeaders["epk"] = json.RawMessage(marshalledEPK)
+	}
+
 	return jd.decryptJWE(jwe, cek)
 }
 
@@ -157,19 +166,6 @@ func (jd *JWEDecrypt) decryptJWE(jwe *JSONWebEncryption, cek []byte) ([]byte, er
 		return nil, err
 	}
 
-	if len(jwe.Recipients) == 1 {
-		authData = []byte(jwe.OrigProtectedHders)
-		if lenAAD := len(aadBytes); lenAAD > 0 {
-			authData = append(authData, '.')
-
-			encLen := base64.RawURLEncoding.EncodedLen(lenAAD)
-			aadEncoded := make([]byte, encLen)
-
-			base64.RawURLEncoding.Encode(aadEncoded, aadBytes)
-			authData = append(authData, aadEncoded...)
-		}
-	}
-
 	return decPrimitive.Decrypt(encryptedData, authData)
 }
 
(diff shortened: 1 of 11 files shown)


pkg/doc/jose/decrypter.go
```diff
@@ -37,6 +36,8 @@ type JWEDecrypt struct {
 	store  storage.Store
 	crypto cryptoapi.Crypto
 	kms    kms.KeyManager
+	// ForGoJoseInterop similar to JWEEncrypt, adds extra encoding for APU/APV for interop with Go Jose.
```
troyronda, line 39: I'm not sure about this.

pkg/doc/jose/decrypter.go
```diff
@@ -93,6 +94,16 @@ func (jd *JWEDecrypt) Decrypt(jwe *JSONWebEncryption) ([]byte, error) {
 		return nil, fmt.Errorf("jwedecrypt: %w", err)
 	}
 
+	if len(recWK) == 1 {
+		// ensure EPK is marshalled the same way as during encryption since it is merged into ProtectHeaders.
```
troyronda, line 98: https://tools.ietf.org/html/rfc7518#section-4.6.1

> The "epk" (ephemeral public key) value created by the originator for
>    the use in key agreement algorithms.  This key is represented as a
>    JSON Web Key [JWK] public key value.  It MUST contain only public key
>    parameters and SHOULD contain only the minimum JWK parameters
>    necessary to represent the key

I believe the implication is that use should not be in the epk header since it is not part of the minimum parameters to represent the key.

pkg/doc/jose/encrypter.go
```diff
@@ -232,35 +242,42 @@ func (je *JWEEncrypt) wrapCEKForRecipients(cek, apu, apv, aad []byte,
 	return recipientsWK, singleRecipientAAD, nil
 }
 
-func (je *JWEEncrypt) getWrapKeyOpts() (string, []cryptoapi.WrapKeyOpts) {
-	var wrapOpts []cryptoapi.WrapKeyOpts
+func (je *JWEEncrypt) encodeAPUAPVForGoJose(kek *cryptoapi.RecipientWrappedKey) {
+	if je.ForGoJoseInterop {
+		if len(kek.APU) > 0 {
```
troyronda, line 247: what breaks in go-jose when this isn't done?

pkg/doc/jose/encrypter.go
```diff
@@ -232,35 +242,42 @@ func (je *JWEEncrypt) wrapCEKForRecipients(cek, apu, apv, aad []byte,
 	return recipientsWK, singleRecipientAAD, nil
 }
 
-func (je *JWEEncrypt) getWrapKeyOpts() (string, []cryptoapi.WrapKeyOpts) {
-	var wrapOpts []cryptoapi.WrapKeyOpts
+func (je *JWEEncrypt) encodeAPUAPVForGoJose(kek *cryptoapi.RecipientWrappedKey) {
+	if je.ForGoJoseInterop {
```
troyronda, line 246: what breaks in go-jose when this isn't done?

troyronda: My guess is that the base64url encoding needs to be fixed in AFG.

pkg/doc/jose/decrypter.go
Baha-sk: removed flag, JWE encryption/decryption is compliant with go-jose

pkg/doc/jose/decrypter.go
Baha-sk: removed `use` JWK header in `epk`

pkg/doc/jose/encrypter.go
Baha-sk: yes, APU/APV must be base64URL encoded, in this case they're KID or epk.x which are already base64URL values, they require an additional encoding to be compliant with their definition as per the [JWA](https://tools.ietf.org/html/rfc7518#section-4.6.1.2) ietf spec.

fqutishat (approved):

PR #2533, opened 2021-02-11
fix: wrong Ed25519 KID value

### LINKED ISSUE #2530: fix Ed25519 KID value
Currently KIDs are based on the public key's JWK thumbprint.

There's a [bug](https://github.com/square/go-jose/pull/324) for Ed25519 JWK thumbprints in go-jose.

Fix this manually in the framework as the go-jose fix might take a while to be merged in their Github repo.

Baha-sk:
This change fixes go-jose's Ed25519 JWK Thumbprint value used in the KMS KID creation.

It also includes a check for emptiness of the key bytes with calling CreateKID() and
fixed related unit tests in the framework.

Finally the webkms bdd tests are commented out since the remote kms server requires
an udpate. They will be uncommented in a future revision when the server image tag
is updated with a revision of aries-framework-go that contains this change.

closes #2530

### DIFF
--- a/pkg/doc/util/jwkkid/kid_creator.go
+++ b/pkg/doc/util/jwkkid/kid_creator.go
@@ -34,14 +34,25 @@ var errInvalidKeyType = errors.New("key type is not supported")
 //  - base64 raw (no padding) URL encoded KID
 //  - error in case of error
 func CreateKID(keyBytes []byte, kt kms.KeyType) (string, error) {
-	// X25519 JWK is not supported by go jose, manually build it and build its resulting KID.
-	if kt == kms.X25519ECDHKWType {
+	if len(keyBytes) == 0 {
+		return "", errors.New("createKID: empty key")
+	}
+
+	switch kt {
+	case kms.X25519ECDHKWType: // X25519 JWK is not supported by go jose, manually build it and build its resulting KID.
 		x25519KID, err := createX25519KID(keyBytes)
 		if err != nil {
 			return "", fmt.Errorf("createKID: %w", err)
 		}
 
 		return x25519KID, nil
+	case kms.ED25519Type: // go-jose JWK thumbprint of Ed25519 has a bug, manually build it and build its resulting KID.
+		ed25519KID, err := createED25519KID(keyBytes)
+		if err != nil {
+			return "", fmt.Errorf("createKID: %w", err)
+		}
+
+		return ed25519KID, nil
 	}
 
 	jwk, err := buildJWK(keyBytes, kt)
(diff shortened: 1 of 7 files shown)


pkg/doc/util/jwkkid/kid_creator.go
```diff
@@ -69,11 +80,13 @@ func buildJWK(keyBytes []byte, kt kms.KeyType) (*jose.JWK, error) {
 		if err != nil {
 			return nil, fmt.Errorf("buildJWK: failed to build JWK from ecdsa DER key: %w", err)
 		}
-	case kms.ED25519Type:
-		jwk, err = jose.JWKFromPublicKey(ed25519.PublicKey(keyBytes))
-		if err != nil {
-			return nil, fmt.Errorf("buildJWK: failed to build JWK from ed25519 key: %w", err)
-		}
+		// TODO remove `case kms.ED25519Type` in CreateKID() and uncomment below case when go-jose fixes Ed25519
```
troyronda, line 83: What is go Jose issue number

pkg/doc/util/jwkkid/kid_creator_test.go
```diff
@@ -32,9 +36,23 @@ func TestCreateKID(t *testing.T) {
...
+	jwk, err := jose.JWKFromPublicKey(pubKey)
+	require.NoError(t, err)
+
+	goJoseTP, err := jwk.Thumbprint(crypto.SHA256)
+	require.NoError(t, err)
+
+	goJoseKID := base64.RawURLEncoding.EncodeToString(goJoseTP)
+	require.NotEqual(t, kid, goJoseKID)
+
+	// now build bad thumbprint and ensure it matches goJoseTP
```
troyronda, line 50: Where is a test showing we are building the correct kid

pkg/doc/util/jwkkid/kid_creator.go
Baha-sk: there's no issue in go-jose repo, only a pending pr mentioned in the AFG issue #2530

pkg/doc/util/jwkkid/kid_creator_test.go
Baha-sk: according to the pending pr, the fix is here https://github.com/square/go-jose/pull/324/files#diff-adb413cfea72e48e90449f9604b1472d63b22160f9c1230bb22517ece6a0fa89R334

In AFG, we build a correct JWK marshalled string then do the same as go-jose. I have not searched if there's an official source of JWK Thumbprint. Unfortunately IETF [RFC 7638](https://tools.ietf.org/html/rfc7638) has no mention of JWK's of Ed25519 keys and contains only an RSA example, so I take it go-jose's format is valid.

fqutishat (approved):

pkg/doc/util/jwkkid/kid_creator_test.go
troyronda: We need a test that demonstrates the correct behavior, not just that it isn't the wrong one.

Baha-sk: added `TestCreateKIDFromFixedKey` further down at line 122

llorllale: * JWK thumbprint algo: https://tools.ietf.org/html/rfc7638#section-3
* JWK of public Ed25519 key: https://tools.ietf.org/html/rfc8037#appendix-A.2

llorllale (approved):

pkg/doc/util/jwkkid/kid_creator_test.go
Baha-sk: I'll update the positive KID  test with https://tools.ietf.org/html/rfc8037#appendix-A.2 @llorllale , thanks

Baha-sk: updated, Ed25519 JWK Thumbprint test now matches [rfc 8037 A.3](https://tools.ietf.org/html/rfc8037#appendix-A.3) example

Issue #1347, opened 2020-02-25
js indexdb to collect store names from higher level packages


sudeshrshetty:
more info: https://github.com/hyperledger/aries-framework-go/pull/1346#discussion_r384045565

PR #2539, opened 2021-02-12
feat: IndexedDB tag support

DRK3:


component/storage/indexeddb/indexeddb.go
```diff
@@ -79,18 +96,64 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
 		return nil, err
 	}
 
-	return &store{name: name, db: p.stores[name]}, nil
+	return p.stores[name], nil
 }
 
-// SetStoreConfig is currently not implemented. It always returns a nil error so that IndexedDB can still be used as
-// a cache provider in FormatProvider for the time being.
+// SetStoreConfig sets the configuration on a store.
```
DRK3, line 102: This "indexing" implementation is intended to be a temporary solution while I work on updating aries-framework-go and TrustBloc.

In order to properly update everything in aries-framework-go to use the new storage interface without breaking the aries-js-worker I need the new IndexedDB implementation to support tags+querying.

The problem is that the proper way to do indexing (using IndexedDB's built in indexing support) requires setting the indexes up on store creation... but the awkward thing is that we need to "pre-create" some stores per #1347. So I need to know all the stores Aries uses and their indexes up front. But I don't know what those indexes are until I do the big task of updating aries-framework-go to use the new storage interface... a bit of a chicken and egg problem.

So, for now, the "indexing" done in this PR is a sort of emulated approach that allows me to update aries-framework-go and test it along with our TrustBloc stuff. The "real" indexing approach will also take a lot longer to implement, and it's quite important that I get aries-framework-go updated to the new storage interface earlier rather than later so I can discover issues early on.

I've created a follow up issue to improve this implementation.

DRK3: (this isn't the only way to approach this problem, but I believe it's easier than the alternative which is trying to locally do `replaces` and then sort of working backwards/forwards creating PRs...)

component/storage/indexeddb/indexeddb.go
```diff
@@ -79,18 +96,64 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
...
+
+	err = store.Put(storeConfigKey, configBytes)
+	if err != nil {
+		return fmt.Errorf("failed to put store store configuration: %w", err)
+	}
+
+	// Create the tag map if it doesn't exist already.
+	_, err = store.Get(tagMapKey)
+	if errors.Is(err, storage.ErrDataNotFound) {
+		err = store.Put(tagMapKey, []byte(""))
```
llorllale, line 125: Is the value here supposed to be fixed in a future PR?

component/storage/indexeddb/indexeddb.go
```diff
@@ -79,18 +96,64 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
...
+	return p.stores[name], nil
 }
 
-// SetStoreConfig is currently not implemented. It always returns a nil error so that IndexedDB can still be used as
-// a cache provider in FormatProvider for the time being.
+// SetStoreConfig sets the configuration on a store.
+// TODO (#2540): Use proper IndexedDB indexing instead of the "Tag Map" once aries-framework-go is updated to use the
+// new storage interface. The tag names used in the predefined stores (see getStoreNames()) will need to be passed in
+// somehow.
 func (p *Provider) SetStoreConfig(name string, config storage.StoreConfiguration) error {
```
llorllale, line 106: I am out of the loop here.

Is this method to be invoked by the framework or the user? (my guess is the framework)

component/storage/indexeddb/indexeddb.go
llorllale: Shouldn't this be the part where the indexes are created?

component/storage/indexeddb/indexeddb.go
llorllale: Is this method supposed to be called only once per store in its lifetime?

component/storage/indexeddb/indexeddb.go
```diff
@@ -143,6 +204,20 @@ func (s *store) Put(key string, value []byte, _ ...storage.Tag) error {
 	m["key"] = key
 	m["value"] = string(value)
 
+	if len(tags) > 0 {
```
llorllale, line 207: do these tags have to match the store's index configuration?

component/storage/indexeddb/indexeddb.go
```diff
@@ -172,19 +247,78 @@ func (s *store) Get(key string) ([]byte, error) {
...
+
+		matchingDatabaseKeys, err :=
+			s.getDatabaseKeysMatchingTagNameAndValue(tagMap, expressionTagName, expressionTagValue)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get database keys matching tag name and value: %w", err)
+		}
+
+		return &iterator{keys: matchingDatabaseKeys, store: s}, nil
+	default:
+		return nil, errInvalidQueryExpressionFormat
```
llorllale, line 320: can you wrap this error and include the offending expression?

component/storage/indexeddb/indexeddb.go
DRK3: In this PR, the initial value for tagMap doesn't really matter. I just create this tagMap once so I don't have to check every time I do a Put. Subsequent Puts will update this tagMap.

This PR doesn't use the real IndexedDB indexing mechanism, which would make this code look a lot different. All I'm doing here is emulating the expected behaviour using the more basic put/get methods.

component/storage/indexeddb/indexeddb.go
DRK3: > Is this method to be invoked by the framework or the user? (my guess is the framework)

Framework does it.

> Is this method supposed to be called only once per store in its lifetime?

It'll be called right after a store is created (that needs to use tags).

component/storage/indexeddb/indexeddb.go
DRK3: For this "emulated" implementation, they do not. The store configuration in this PR is really just here to ensure consistent behaviour across store implementations (the common tests I've created mandate a certain minimum set of behaviours). When I do the real one, the tags will need to match the store's config.

component/storage/indexeddb/indexeddb.go
DRK3: Done!

component/storage/indexeddb/indexeddb.go
llorllale: > Framework does it.

How can a user assign arbitrary tags?

llorllale (approved):

fqutishat (approved):

component/storage/indexeddb/indexeddb.go
DRK3: > > Framework does it.
> 
> How can a user assign arbitrary tags?

Yep - although the framework will be using specific tags for is purposes. Someone extending the framework may have their own uses for tags.

DRK3: > > Framework does it.
> 
> How can a user assign arbitrary tags?

Sorry, just realized you were asking "how".

Tags are assigned via "Put" calls to associate some data as a part of some larger "grouping". Does that answer your question?

Issue #2532, opened 2021-02-10
LevelDB implementation of new storage interface


DRK3:
Once this is done, add support for LevelDB back in to aries-agent-rest (was removed in #2543)

DRK3:
LevelDB support for aries-agent-rest was added back in in #2607

PR #2543, opened 2021-02-18
feat: Switch to new storage interface

### LINKED ISSUE #2315: Rework storage interface to allow for index use

### LINKED ISSUE #2490: EDV REST provider stores don't namespace corrrectly
Due to optimizations made just before the v0.1.5 release with encrypted document ID generations EDV REST stores don't namespace correctly anymore - all stores "bleed" into each other. e.g. A put in one store is a put in any store.

The new storage implementation of the EDV REST provider/ encrypted formatter will fix this.

DRK3:
- New storage interface is now used everywhere.
- Removed old storage implementations.
- Removed LevelDB support from aries-agent-rest. It will be added back again once there's a new LevelDB storage provider implementing the new storage interface.

closes #2315
closes #2490


DRK3:
FYI: @pfeairheller

cmd/aries-agent-mobile/pkg/api/storage.go
```diff
@@ -6,64 +6,95 @@ SPDX-License-Identifier: Apache-2.0
...
-	Key() []byte
-
-	// Value returns the value of the current key/value pair, or nil if done.
-	// The caller should not modify the contents of the returned slice, and
-	// its contents may change on the next call to any 'seeks method'.
-	Value() []byte
+import spi "github.com/hyperledger/aries-framework-go/spi/storage"
+
+// StorageProvider represents a storage provider.
+type StorageProvider interface {
```
troyronda, line 12: Isn't this defined in spi, imported above?

cmd/aries-agent-mobile/pkg/api/storage.go
```diff
@@ -6,64 +6,95 @@ SPDX-License-Identifier: Apache-2.0
...
+	// GetOpenStores returns all currently open stores.
+	GetOpenStores() []Store
+
+	// Close closes all stores created under this store provider.
+	// For persistent store implementations, this does not delete any data in the stores.
+	Close() error
 }
 
-// Store is the storage interface.
+// Store represents a storage database.
```
troyronda, line 38: Isn't this defined in spi, imported above?

cmd/aries-agent-mobile/pkg/api/storage.go
```diff
@@ -6,64 +6,95 @@ SPDX-License-Identifier: Apache-2.0
...
+	// any data in the underlying databases.
+	Close() error
 }
 
-// StorageProvider defines storage interface.
-type StorageProvider interface {
-	// OpenStore opens a store with given name space and returns the handle
-	OpenStore(name string) (Store, error)
+// StoreIterator allows for iteration over a collection of entries in a store.
+type StoreIterator interface {
```
troyronda, line 85: Isn't this defined in spi?

cmd/aries-agent-rest/startcmd/start.go
```diff
@@ -55,7 +54,7 @@ const (
 	databaseTypeEnvKey        = "ARIESD_DATABASE_TYPE"
 	databaseTypeFlagShorthand = "q"
 	databaseTypeFlagUsage     = "The type of database to use for everything except key storage. " +
-		"Supported options: mem, leveldb. " +
+		"Supported options: mem. " +
```
llorllale, line 57: Supporting only `mem` isn't good. We should import `-ext` and keep levelDB as default. There should be no problem importing `-ext` from _this_ module.

pkg/didcomm/protocol/presentproof/service.go
```diff
@@ -484,25 +489,42 @@ func (s *Service) deleteTransitionalPayload(id string) error {
...
-		fmt.Sprintf(transitionalPayloadKey, ""),
-		fmt.Sprintf(transitionalPayloadKey, storage.EndKeySuffix),
-	)
-	defer records.Release()
+	records, err := s.store.Query(transitionalPayloadKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query store: %w", err)
+	}
+
+	defer func() {
```
llorllale, line 497: we could use a static helper function to make it a bit easier and less repetitive...

```go
defer storage.Close(records, logger)  // logger should be optional
```

pkg/didcomm/protocol/presentproof/service.go
```diff
@@ -484,25 +489,42 @@ func (s *Service) deleteTransitionalPayload(id string) error {
...
+		errClose := records.Close()
+		if errClose != nil {
+			logger.Errorf("failed to close records iterator: %s", errClose.Error())
+		}
+	}()
 
 	var actions []Action
 
-	for records.Next() {
+	more, err := records.Next()
```
llorllale, line 506: I think the API's error-handling design can be improved significantly: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_eliminate_error_handling_by_eliminating_errors

cmd/aries-agent-rest/startcmd/start.go
DRK3: While I agree, a LevelDB implementation of the new storage interface was moved out of this milestone, so I don't have one to use.

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: My understanding from talking to @soluchok was that this duplication of interfaces is necessary for the mobile bindings. Perhaps @soluchok can elaborate?

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: My understanding from talking to @soluchok was that this duplication of interfaces is necessary for the mobile bindings. Perhaps @soluchok can elaborate?

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: My understanding from talking to @soluchok was that this duplication of interfaces is necessary for the mobile bindings. Perhaps @soluchok can elaborate?

cmd/aries-agent-mobile/pkg/api/storage.go
troyronda: When you determine if it's needed, it might be an idea to add a comment to this file for future readers.

DRK3: @soluchok, If you let me know the reasoning I can add a comment

pkg/didcomm/protocol/presentproof/service.go
DRK3: I have about a half dozen storage implementations of the current API. Can you take a look at them and let me know what specifically you're thinking? I'm not really sure what action I should take based on that link.

cmd/aries-agent-rest/startcmd/start.go
DRK3: @troyronda @fqutishat ^^

pkg/didcomm/protocol/presentproof/service.go
llorllale: It's not a blocker. But the API's design could use improvements. Like the link says - stop returning so many errors that users have to handle.

DRK3: @llorllale I agree that the constant error returning is for sure a pain... The problem is that it's hard to get away from, since each storage implementation works differently. Some generate potential errors on `Next()`, some generate potential errors on `Value()`, some generate potential errors on several of those methods, etc... Apart from doing deferred error handling with an `Error()` method in the iterator interface (which is how the old interface worked), I'm not sure what else can be done. I was avoiding that type of deferred error handling since it made it very easy to swallow errors and skip checking them...

On a separate note - one of my requirements was to avoid breaking the storage interface multiple times, so if there's an improvement to be made to the interface then now is the time to do it.

cmd/aries-agent-mobile/pkg/api/storage.go
soluchok: The main reason to have one more interface was related to the `Release` method. `Release` is a reserved word for ios.
It was not possible to use the original interface. This is why it was wrapped like this `Release() { s.StoreIterator.Free() }`. Now, when we do not have `Release` in the iterator interface it might work. @DRK3 It needs to be checked.

See https://github.com/trustbloc/agent-sdk/pull/49#discussion_r520000674

pkg/didcomm/protocol/presentproof/service.go
DRK3: Where would the static helper function go? An `internal` file perhaps?

pkg/didcomm/protocol/presentproof/service.go
llorllale: @DRK3 the general strategy is to just defer these errors and the user checks `itr.Error() != nil` after use.

`itr.Next()` should return false if there are no more records _or_ there is an error.

The whole thing will collapse to just:

```go
	records, err := s.store.Query(transitionalPayloadKey)
	if err != nil {
		return nil, fmt.Errorf("failed to query store: %w", err)
	}

        defer ....

        for records.Next() {
             record, err := records.Value()
             if err != nil {
                   // handle error
             }

             // handle record
        }

        if itr.Error() != nil {
             // handle error
        }
```

llorllale: I feel confident all implementations can behave this way.

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: @soluchok Can you check and get back to me? Otherwise I'll have to continue with the PR as-is following the current pattern, since I don't want to break the mobile bindings.

pkg/didcomm/protocol/presentproof/service.go
DRK3: @llorllale Indeed I could do that. But then errors can be easily missed, since nothing forces a check on the `Error()` method. It seems wrong to make an error ignorable (hence why we have linters to check for that normally, but if we rely on the developer to call `Error()` then the linter can't enforce that)...

llorllale: @DRK3 I was responding to what seemed to be the meat of your counterargument:

> The problem is that it's hard to get away from, since each storage implementation works differently. Some generate potential errors on Next(), some generate potential errors on Value(), some generate potential errors on several of those methods, etc... Apart from doing deferred error handling with an Error() method in the iterator interface (which is how the old interface worked), I'm not sure what else can be done.

Regarding what is "preferable" vs what is not - that is opinion-based and you can run a poll if you wish. My opinion is it should be simplified not just because it looks cleaner, but also (important) - **it implies less test cases in our unit tests**, which translates into more productivity.

cmd/aries-agent-mobile/pkg/api/storage.go
soluchok: @DRK3 you can check it by building mobile builds. If it is possible to inject storage from the iOS and Android then it works. 
How do you know that your code works? Have you built it (for iOS and Android)?
I don’t think it will work `Put(key string, value []byte, tags ...spi.Tag) error` bc of `...spi.Tag`.
Same here `Query(expression string, options ...spi.QueryOption)` bc of `options ...spi.QueryOption`.

soluchok: https://pkg.go.dev/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions

pkg/didcomm/protocol/presentproof/service.go
llorllale: Not `internal` - downstream packages won't be able to use it.

I would place it right in `storage`.

pkg/didcomm/protocol/presentproof/service.go
DRK3: So if the `Next()` call doesn't return an error, then by my count there's one less error check... that would be one fewer test then, I believe...not a huge improvement, but yes I agree that would be nicer.

In terms of productivity... I would argue that missing an error check (resulting in potential bugs) would be very bad for productivity... perhaps much worse than the pain of adding one more unit test.

In terms of running a poll - I'll leave that up to you to pursue, since the current design was the one already reviewed & approved by everyone months ago. I'm happy to change it if everyone prefers the tradeoffs associated with an explicit Error() method. Changing the current design will also require updating all the storage implementations... but if we're going to do it, I think it's now or never.

llorllale: > So if the Next() call doesn't return an error, then by my count there's one less error check... that would be one fewer test then, I believe.

There are 3 less errors that users have to handle in the snippet.

llorllale: +1 more if you consider the utility function for closing the iterator.

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: @soluchok How do I build for mobile?

Also, according to the link you sent, it sounds like everything in the new storage package should be supported, including `spi.Tag` and `spi.QueryOption`. So now I'm wondering if we need to have this extra interface at all.

soluchok: I don’t think that this will work `tags ...spi.Tag` and `options ...spi.QueryOption`.

`How do I build for mobile?` - we have a readme https://github.com/hyperledger/aries-framework-go/tree/main/cmd/aries-agent-mobile

Baha-sk (commented):
lgtm

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: @soluchok Why won't it work? Do all structs also have to be defined in that file as well? The documentation you linked to seems to suggest that all the types I use are supported.

soluchok: I don’t know how `tags ...spi.Tag` expression can be converted to java (probably it will work somehow). Just make sure your code is compatible. The current code is not compilable.

```
github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile
gobind/gobind
# gobind/gobind
gobind/go_apimain.go:2797:10: cannot use (*proxyapi_Store)(res_0_ref) (type *proxyapi_Store) as type "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".Store in assignment:
	*proxyapi_Store does not implement "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".Store (missing Batch method)
gobind/go_configmain.go:171:7: cannot use (*proxyapi_StorageProvider)(_v_ref) (type *proxyapi_StorageProvider) as type "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".StorageProvider in assignment:
	*proxyapi_StorageProvider does not implement "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".StorageProvider (missing GetOpenStores method)
gobind/go_storagemain.go:34:7: cannot use (*proxyapi_StorageProvider)(_v_ref) (type *proxyapi_StorageProvider) as type "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".StorageProvider in assignment:
	*proxyapi_StorageProvider does not implement "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".StorageProvider (missing GetOpenStores method)
gobind/go_storagemain.go:84:14: cannot use (*proxyapi_StorageProvider)(_param_p0_ref) (type *proxyapi_StorageProvider) as type "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".StorageProvider in assignment:
	*proxyapi_StorageProvider does not implement "github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/api".StorageProvider (missing GetOpenStores method)
gomobile: go build -v -ldflags -s -w -buildmode=c-shared -o=/var/folders/3l/pr52sz3j3t19m1mqgn7mjyz00000gn/T/gomobile-work-135526515/android/src/main/jniLibs/armeabi-v7a/libgojni.so ./gobind failed: exit status 2

make: *** [bindings-android] Error 1
```

pkg/didcomm/protocol/presentproof/service.go
DRK3: @llorllale Yes, you're right, two less errors actually.

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: @soluchok OK, I will get Android Studio and Xcode set up and try making those bindings.

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: @troyronda After talking to @soluchok, I now better understand the reason we need these additional interfaces. I've added comments to the files explaining it for future readers.

cmd/aries-agent-mobile/pkg/api/storage.go
DRK3: @soluchok I've fixed up the mobile storage interface and wrapper to support the new storage interface, and the bindings seem to be generated correctly now. Can you let me know if the new code looks ok to you?

pkg/didcomm/protocol/presentproof/service.go
DRK3: Done - since spi/storage is in a different module, I had to do it in a separate commit... please see https://github.com/hyperledger/aries-framework-go/pull/2566

DRK3: Good idea - this should also help reduce the number of lines needed to be unit tested

DRK3: I've updated this PR to use that new helper method

pkg/didcomm/protocol/introduce/service.go
```diff
@@ -523,7 +542,7 @@ func (s *Service) saveTransitionalPayload(id string, data transitionalPayload) e
 		return fmt.Errorf("marshal transitional payload: %w", err)
 	}
 
-	return s.store.Put(fmt.Sprintf(transitionalPayloadKey, id), src)
+	return s.store.Put(fmt.Sprintf(transitionalPayloadKey, id), src, storage.Tag{Name: transitionalPayloadKey})
```
DRK3, line 545: General note... it's possible that many of the tag names I'm using could be shortened. For now, I just used the existing constants defined in these files - some of which may have extraneous info. I didn't want to risk changing more stuff than I had to... (and I already had to change a ton of files).

On a similar note, some of the keys that are being used could probably be simplified - before the long key names were used in order to allow for filtering via the old Iterator method, since that relied on using certain key suffixes/prefixes.

I think this is a task for the future...

fqutishat (approved):

pkg/didcomm/protocol/mediator/service_test.go
```diff
@@ -1432,29 +1428,6 @@ func TestGetConnections(t *testing.T) {
 		require.NoError(t, err)
 		require.Empty(t, connID)
 	})
-
-	t.Run("test get connection - db error", func(t *testing.T) {
```
troyronda, line 1436: removed test?

cmd/aries-agent-rest/startcmd/start.go
troyronda: @DRK3 Need a TODO to restore.

pkg/didcomm/protocol/mediator/service_test.go
DRK3: The method that it was testing doesn't exit anymore

cmd/aries-agent-rest/startcmd/start.go
DRK3: @troyronda https://github.com/hyperledger/aries-framework-go/issues/2532

cmd/aries-agent-mobile/pkg/wrappers/storage/storage.go
```diff
@@ -8,47 +8,366 @@ SPDX-License-Identifier: Apache-2.0
...
+// Close closes all stores created under this store provider.
+// For persistent mobile binding store implementations, this does not delete any data in the underlying databases.
+func (p *Provider) Close() error {
+	p.lock.RLock()
+
+	openStoresSnapshot := make([]*store, len(p.openStores))
+
+	var counter int
+
+	for _, openStore := range p.openStores {
```
sudeshrshetty, line 146: Why can't it be like this?

```
for counter, openStore := range p.openStores {
		openStoresSnapshot[counter] = openStore
	}
	```

File: cmd/aries-agent-mobile/pkg/wrappers/storage/storage.go [hunk 10]
```diff
@@ -8,47 +8,366 @@ SPDX-License-Identifier: Apache-2.0
...
+		openStoresSnapshot[counter] = openStore
+		counter++
+	}
+	p.lock.RUnlock()
+
+	for _, openStore := range openStoresSnapshot {
```
sudeshrshetty, line 152: Don't you need a write lock here??

cmd/aries-agent-mobile/pkg/wrappers/storage/storage.go
DRK3: p.openStores is a map, not an array. `Range` doesn't return an index for maps

cmd/aries-agent-mobile/pkg/wrappers/storage/storage.go
```diff
@@ -8,47 +8,366 @@ SPDX-License-Identifier: Apache-2.0
...
+	mobileBindingStore api.Store
+	name               string
+	close              closer
+}
+
+func (s *store) Put(key string, value []byte, tags ...spi.Tag) error {
+	var tagsBytes []byte
+
+	if len(tags) > 0 {
+		var err error
```
sudeshrshetty, line 179: `var err error` can be moved out of this block

DRK3: Indeed it could...is there an advantage to doing so?

DRK3: Done in #2567!

cmd/aries-agent-mobile/pkg/wrappers/storage/storage.go
DRK3: I don't believe so. So each store implementation will have their own way of "closing" a store. For many, it's simply a no-op. For others, they may terminate a connection. If a store needs a write lock, they'll have to implement one inside their own methods...

Let me know if there's a specific scenario you can think of that would cause issues.

PR #2550, opened 2021-02-19
feat: KMS support for BBS+ keys: create Tink key managers

### LINKED ISSUE #2546: KMS support for BBS+ keys: create Tink key managers
part of #2295

The KMS requires new Tink key managers to be executed in the kms.

This issue is to create these key managers (Signer and Verifier BBS key managers), register them in Tink and create their supporting subtle primitives.

Baha-sk:
closes #2546


pkg/crypto/tinkcrypto/primitive/bbs/subtle/bls12381g2_verifier.go
```diff
@@ -0,0 +1,51 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package subtle
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/doc/bbs/bbs12381g2pub"
```
troyronda, line 10: I assume there is a todo to move these primitives?

Baha-sk: not really, the subtle package uses these primitives directly

Baha-sk: unless we move the content of `"github.com/hyperledger/aries-framework-go/pkg/doc/bbs/bbs12381g2pub"` to `"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/bbs/subtle/"` and make the core primitive internal (hidden within subtle package)

pkg/crypto/tinkcrypto/primitive/bbs/api/verifier.go
```diff
@@ -0,0 +1,28 @@
...
+	// returns:
+	// 		error in case of errors or nil if signature proof verification was successful
+	VerifyProof(messages [][]byte, proof, nonce []byte) error
+
+	// DeriveProof will create a BBS+ signature proof for a list of revealed messages using BBS signature
+	// (can be built using a Signer's Sign() call) and the verifier's public key.
+	// returns:
+	// 		signature proof in []byte
+	//		error in case of errors
+	DeriveProof(messages [][]byte, signature, nonce []byte, revealedIndexes []int) ([]byte, error)
```
troyronda, line 27: DeriveProof doesn't seem to be a Verifier activity.

Baha-sk: DeriveProof uses the public key of the signer, similarly to Verify and VerifyProof.

You're right about the comment, it's the Signer's public key, not the verifier's. I'll update it.

Baha-sk: DeriveProof can be moved to the Signer interface, I added to the verifier to interface because it uses the public key, not the private key like Sign().

Let me know if it makes more sense to move it to the Signer interface.

pkg/crypto/tinkcrypto/primitive/bbs/subtle/bls12381g2_verifier.go
troyronda: In any case, I think it would make sense for crypto primitives to be in the crypto folder hierarchy rather than the doc folder hierarchy. Can be a followup.

pkg/crypto/tinkcrypto/primitive/bbs/api/verifier.go
troyronda: I wonder if there should be an additional interface if it doesn't quite fit in either?
@Baha-sk

pkg/crypto/tinkcrypto/primitive/bbs/subtle/bls12381g2_verifier.go
Baha-sk: Sure, we can move them to crypto in a follow-up issue.

pkg/crypto/tinkcrypto/primitive/bbs/api/verifier.go
Baha-sk: The interface is used (returned) by the key manager's Primitive() call, since we have BBS Signer and Verifier key managers, we only need 2 interfaces types, creating a third type doesn't make sense.

troyronda (approved):

fqutishat (approved):

PR #2552, opened 2021-02-20
feat: VC wallet client - KMS client &  open/close

sudeshrshetty:
VC wallet client can be initialized with following key manager options.

`passphrase` : if client chooses to use passphrase for localkms.
`secret lock service`: if client chooses to use its own secret lock service.
`keyServerURL`: if client chooses to user remotekms (webkms).

**wallet.Open() :** accepts one of the below params to instantiate key
manager client and unlock wallet for use.

`auth token`: auth token to be part of header for remote kms calls
to key server.
`passphrase` : if client chooses to use passphrase for localkms.
secret lock service: if client chooses to use its own secret lock
service.
`expiration`: time duration for which token has to be active

**wallet.Close()**: expires token and removes key manager client instance.

### DIFF
--- a/pkg/client/vcwallet/client.go
+++ b/pkg/client/vcwallet/client.go
@@ -9,25 +9,138 @@ package vcwallet
 import (
 	"encoding/json"
 	"fmt"
+	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/secretlock"
+	"github.com/hyperledger/aries-framework-go/pkg/storage"
 )
 
 // provider contains dependencies for the verifiable credential wallet client
 // and is typically created by using aries.Context().
 type provider interface {
-	// TODO to be added #2433
+	StorageProvider() storage.Provider
+}
+
+// kmsOpts contains options for creating verifiable credential wallet client.
+type kmsOpts struct {
+	// local kms options
+	secretLockSvc secretlock.Service
+	passphrase    string
+
+	// remote(web) kms options
+	keyServerURL string
+}
+
+// KeyManagerOptions is option for verifiable credential wallet client key manager.
+type KeyManagerOptions func(opts *kmsOpts)
+
+// WithSecretLockService option, when provided then wallet client will use local kms for key operations.
+func WithSecretLockService(svc secretlock.Service) KeyManagerOptions {
+	return func(opts *kmsOpts) {
+		opts.secretLockSvc = svc
+	}
+}
+
+// WithPassphrase option to provide passphrase for local kms for key operations.
+func WithPassphrase(passphrase string) KeyManagerOptions {
+	return func(opts *kmsOpts) {
+		opts.passphrase = passphrase
+	}
+}
+
+// WithKeyServerURL option, when provided then wallet client will use remote kms for key operations.
+// This option will be ignore if provided with 'WithSecretLockService' option.
+func WithKeyServerURL(url string) KeyManagerOptions {
+	return func(opts *kmsOpts) {
+		opts.keyServerURL = url
+	}
 }
 
 // Client enable access to verifiable credential wallet features.
 type Client struct {
 	// ID of wallet content owner
 	userID string
+
+	// wallet profile
+	profile *profile
+
+	// storage provider
+	storeProvider storage.Provider
 }
 
 // New returns new verifiable credential wallet client for given user.
-func New(userID string, ctx provider) *Client {
-	// TODO initialize providers for stores, VDR, KMS #2433
-	// TODO create user profile if not already created #2433
-	return &Client{userID: userID}
+// returns error if wallet profile is not found.
+// To create a new wallet profile, use `CreateProfile()`.
+// To update an existing profile, use `UpdateProfile()`.
+func New(userID string, ctx provider) (*Client, error) {
+	store, err := newProfileStore(ctx.StorageProvider())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get store to fetch VC wallet profile info: %w", err)
+	}
+
+	profile, err := store.get(userID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get VC wallet profile: %w", err)
+	}
+
+	return &Client{userID: userID, profile: profile, storeProvider: ctx.StorageProvider()}, nil
+}
+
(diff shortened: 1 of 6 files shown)


pkg/client/vcwallet/client.go
```diff
@@ -9,25 +9,139 @@ package vcwallet
...
+
+// WithSecretLockService option, when provided then wallet client will use local kms for key operations.
+func WithSecretLockService(svc secretlock.Service) KeyManagerOptions {
+	return func(opts *kmsOpts) {
+		opts.secretLockSvc = svc
+	}
+}
+
+// WithPassPhrase option to provide passphrase for local kms for key operations.
+func WithPassPhrase(passphrase string) KeyManagerOptions {
```
troyronda, line 45: Passphrase is one word.

sudeshrshetty: fixed

pkg/client/vcwallet/client.go
```diff
@@ -9,25 +9,139 @@ package vcwallet
...
+
+	err = store.save(profile, update)
+	if err != nil {
+		return nil, fmt.Errorf("failed to save VC wallet profile: %w", err)
+	}
+
+	return &Client{userID: userID, profile: profile, storeProvider: ctx.StorageProvider()}, nil
+}
+
+// Open unlocks wallet client's key manager instance and returns a token for subsequent use of wallet features.
```
troyronda, line 129: Open vs Get is confusing.

sudeshrshetty: package level functions:
`vcwallet.Get()` to get already created wallet.
`vcwallet.Create()` to get wallet client by creating new profile.
`vcwallet.Update()` to get wallet client by updating existing profile.

Open() to be called on wallet client instance.
`walletInstance.Open()`

pkg/client/vcwallet/client.go
```diff
@@ -9,25 +9,139 @@ package vcwallet
...
+	profile *profile
+
+	// storage provider
+	storeProvider storage.Provider
+}
+
+// Get returns new verifiable credential wallet client for given user.
+// returns error if wallet profile is not found.
+// To create a new client with new wallet profile, use `New()`.
+func Get(userID string, ctx provider) (*Client, error) {
```
troyronda, line 74: This is traditionally the New function (new client).

pkg/client/vcwallet/client.go
```diff
@@ -9,25 +9,139 @@ package vcwallet
...
+	}
+
+	return &Client{userID: userID, profile: profile, storeProvider: ctx.StorageProvider()}, nil
+}
+
+// Create creates and returns new verifiable credential wallet client by creating new wallet profile for given user.
+// returns error if wallet profile is already created.
+// Use `Get()` for already created VC wallet profile.
+// Use `Update()` for replacing an already created VC wallet profile.
+func Create(userID string, ctx provider, options ...KeyManagerOptions) (*Client, error) {
```
troyronda, line 92: It might be a bit confusing to have New/Get,Create,Update all return clients.

I wonder if Create/Update should be CreateProfile and UpdateProfile and they don't return clients?

troyronda: note: I'm also not entirely sure if we should be mixing `Create`/`Update` with returning a `Client`.

troyronda: If these functions names / roles are more clear, it is easier to use the normal `New` method name for the client.

pkg/client/vcwallet/client.go
```diff
@@ -9,25 +9,139 @@ package vcwallet
...
+}
+
+// Open unlocks wallet client's key manager instance and returns a token for subsequent use of wallet features.
+//
+//	Args:
+//		- auth : auth token in case of remotekms or passphrase in case of localkms.
+//		- secretLockSvc: secret lock service for localkms if you choose not to provide passphrase.
+//		- tokenExpiry : (optional, default: 10 * time.minute) time duration after which issued token will expiry.
+//
+//	Returns token with expiry that can be used for subsequent use of wallet features.
```
troyronda, line 136: I think currently this package is being structured as the Service-side of the vcwallet.

You could probably imagine another package that hides the details of this string when calling service functions.

fqutishat (approved):

pkg/client/vcwallet/client.go
```diff
@@ -9,25 +9,139 @@ package vcwallet
...
 type provider interface {
-	// TODO to be added #2433
+	StorageProvider() storage.Provider
+}
+
+// kmsOpts contains options for creating verifiable credential wallet client.
+type kmsOpts struct {
+	// local kms options
+	secretLockSvc secretlock.Service
+	passphrase    string
```
Baha-sk, line 28: local and remote kms are mutually exclusive, it's important to check if either one of these options is set, not both and not none.

sudeshrshetty: yes they are mutually exclusive, I have defined in option comments about the precedence. Which one takes priority when all options are supplied.

Baha-sk (approved):

pkg/client/vcwallet/client.go
sudeshrshetty: I agree, to be done as part of decoupling token & cache #2556

pkg/client/vcwallet/client.go
sudeshrshetty: fixing it now, will be pushed shortly.

pkg/client/vcwallet/client.go
sudeshrshetty: fixed as part of https://github.com/hyperledger/aries-framework-go/pull/2552#discussion_r579813776

pkg/client/vcwallet/client.go
sudeshrshetty: fixed

pkg/client/vcwallet/client.go
sudeshrshetty: Looks good to me, 
call `CreateProfile()` once and use `New()` whenever you want client.

fixed: 

- `CreateProfile/UpdateProfile` with returning error
- `New` returning client

troyronda (approved):

PR #2570, opened 2021-02-25
refactor: move bbs12381g2pub to pkg/crypto

Baha-sk:


.golangci.yml
```diff
@@ -11,7 +11,7 @@ run:
   tests: true
   build-tags: [""]
   skip-dirs:
-    - pkg/doc/bbs/bbs12381g2pub/internal/kilic/bls12-381
+    - pkg/crypto/bbs12381g2pub/internal/kilic/bls12-381
```
troyronda, line 14: would it make sense to have a folder between crypto and bbs12381g2pub

/pkg/crypto/.../bbs12381g2pub

Baha-sk: ok, I'll add `bbs` in between just like the original path

Baha-sk: done

Baha-sk: renamed `bbs` to `primitive`

DRK3 (approved):

troyronda (approved):

PR #2583, opened 2021-03-02
feat: IndexedDB provider - create tag map on store open

DRK3:
- The IndexedDB provider now creates a tag map when opening a store, so it no longer requires a call to SetStoreConfig to use tags + querying (it never really needed to in the first place).
- Removed some redundant unit tests.

### DIFF
--- a/component/storage/indexeddb/indexeddb.go
+++ b/component/storage/indexeddb/indexeddb.go
@@ -45,6 +45,9 @@ const (
 		"it must be in the following format: TagName:TagValue"
 )
 
+// TODO (#2540): Use proper IndexedDB indexing instead of the "Tag Map" once aries-framework-go is updated to use the
+// new storage interface. The tag names used in the predefined stores (see getStoreNames()) will need to be passed in
+// somehow.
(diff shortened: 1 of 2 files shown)


component/storage/indexeddb/indexeddb.go
```diff
@@ -96,14 +99,26 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
 		return nil, err
 	}
 
+	newlyCreatedStore := p.stores[name]
+
+	// Create the tag map if it doesn't exist already.
+	_, err = p.stores[name].Get(tagMapKey)
+	if errors.Is(err, storage.ErrDataNotFound) {
+		err = newlyCreatedStore.Put(tagMapKey, []byte("{}"))
```
sudeshrshetty, line 107: how about empty bytes??

DRK3: It has to be unmarshallable to a `tagMapping`. Empty bytes cause a failure, while `{}` allows it to marshal to an empty `tagMapping` (which is desired).

fqutishat (approved):

sudeshrshetty (approved):

PR #2588, opened 2021-03-02
perf: optimize storage usage in didexchange

### LINKED ISSUE #2586: Optimize storage usage in didexchange protocol

aholovko:
This PR improves performance of didexchange protocol by reducing the number of expensive calls for saving/retrieving DIDs.

Key changes:

* Fetching DIDs was moved from [packager.UnpackMessage](https://github.com/hyperledger/aries-framework-go/blob/6d90017383c45e76f72c71254821435959c5f4f9/pkg/didcomm/packager/packager.go#L199) to inbound message handler ([here](https://github.com/hyperledger/aries-framework-go/blob/cc43b0aff6bf417daf7e6304091b7d1fbd13d2fe/pkg/framework/context/context.go#L183))

  There are 2 reasons for that:
  1. At the point when a message is unpacked, we don't know yet what service will process the message and can't optimize accordingly (in case of didexchange there is no need for expensive DID retrievals).
  2. It's not the responsibility of `packager.UnpackMessage` to retrieve DIDs, so moving that logic away should improve separation of concerns.

* `DIDConnectionStore` is now an injectable framework component that defaults to `did.NewConnectionStore` and is backed by storage and VDR that the framework is initialized with (done as a part of the optimizations above)

* Improvements in didexchange protocol:

  * `myDID` is not persisted on every saving of connection record (depending on the underlying storage saving `myDID` might be very expensive)
  * clear separation between transient storage (`connectionRecorder` and permanent DID connection storage (`connectionStore`)

Closes #2586


fqutishat (approved):

pkg/framework/context/context.go
```diff
@@ -158,16 +162,33 @@ func (p *Provider) tryToHandle(svc service.InboundHandler, msg service.DIDCommMs
...
 // InboundMessageHandler return an inbound message handler.
-func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
-	return func(message []byte, myDID, theirDID string) error {
-		msg, err := service.ParseDIDCommMsgMap(message)
+func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler { //nolint:gocyclo
+	connectionStore, err := did.NewConnectionStore(p)
+	if err != nil {
+		return func(*transport.Envelope) error { return err }
+	}
+
```
Baha-sk, line 170: as discussed, even though this function is called at the start of the handler only, it is better (more correct) to reference an existing connection store instead of creating a new one here.

pkg/didcomm/packager/packager.go
```diff
@@ -195,22 +185,5 @@ func (bp *Packager) UnpackMessage(encMessage []byte) (*transport.Envelope, error
...
-
-	// ignore error - at beginning of DIDExchange, you might be about to generate a DID
-	myDID, err := bp.connectionStore.GetDID(base58.Encode(envelope.ToKey))
-	if errors.Is(err, did.ErrNotFound) {
-	} else if err != nil {
-		return nil, fmt.Errorf("failed to get my did: %w", err)
-	}
-
-	envelope.ToDID = myDID
-	envelope.FromDID = theirDID
```
llorllale, line 213: I am confused by the removal of both `myDID` and `theirDID`.

Are we losing information here or is it being supplied somewhere else? How are users of the framework affected by this removal?

aholovko: these properties were used just to carry information from [here](https://github.com/hyperledger/aries-framework-go/blob/d24bb59ffab793f895032ad9356251bb1b2092a3/pkg/didcomm/packager/packager.go#L212) to [here](https://github.com/hyperledger/aries-framework-go/blob/d24bb59ffab793f895032ad9356251bb1b2092a3/pkg/didcomm/transport/ws/pool.go#L111), it was updated as retrieving DIDs should not be the responsibility of `packager.UnpackMessage()`

llorllale: How are users of the framework affected by this removal?

aholovko: * in the framework these properties were used only in the transport layer (http and ws) and introduced when support for inbound/outbound messaging was added ([commit](https://github.com/hyperledger/aries-framework-go/commit/5d321d4181ffb8eedf4ddb43364ee1c75fea28dc)) - it looks more like some sort of workaround or quick "solution" - logic for DID retrieval is moving now to the message handler - and it will act as an orchestrator (reasonable role for it)
* I think that using properties of `transport.Envelope` outside of the framework should be considered as an error (this is low-level details of our implementation rather than declared contract from our side)

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -83,16 +92,18 @@ type stateMachineMsg struct {
 type Service struct {
 	service.Action
 	service.Message
-	ctx             *context
-	callbackChannel chan *message
-	connectionStore *connectionStore
+	ctx                *context
+	callbackChannel    chan *message
+	connectionRecorder *connection.Recorder
+	connectionStore    didstore.ConnectionStore
```
Baha-sk, line 98: since a new `connectionRecoder` is created, I would move `connectionStore` into it.

`Service` doesn't need to handle 2 struct types for connection, 1 is enough.

I would also make this call directly from `connectionRecoder` as well: https://github.com/hyperledger/aries-framework-go/pull/2588/files#diff-a4b94fca66b9a726621a9536a1083e4c60cc89a64ad65094608a7c42a25ca702R333


this way the service only handles 1 struct for connections only

aholovko: this is exactly what we had before and what I was trying to avoid - problem with having a single struct for these different things is that it becomes confused which one is used - and they are very different: connectionRecorder is "lightweight" transient storage and connectionStore is expensive one that can be backed by heavy thing like EDV

Baha-sk: but the service doesn't need to be aware of the what/where the connectionStore lies. The connectionRecorder handles this struct on behalf of the service and can easily change it without the service's knowledge. This separation of concern helps keep the service cleaner and requires less maintenance.

Baha-sk: another option would be to create a `connectionManager` that has a `connectionRecorder` and `connectionStore` and have the service include a `connectionManager` instead

aholovko: I have mixed feelings about that. With the previous structure
```
type connectionStore struct {
	*connection.Recorder
	*did.ConnectionStore
}
```
it was not immediately clear whether `c.SaveConnectionRecord` belongs to "cheap" `connection.Recorder` or to "expensive" `did.ConnectionStore`. It makes sense for that sort of information to be explicit. In my opinion, this is what service should NOT hide.

aholovko: Guys, do you have any thoughts here?^

Moopli (approved):

Moopli:
I think this is the right approach, but I do think there's further followup that can be done (in another PR):

I noticed that `ConnectionRecorder` uses the standard `context.StorageProvider()`, same as `ConnectionStore`, so I imagine this refactor will enable optimization but won't offer much of a speedup on its own.

I think `ConnectionRecorder` could use some further refactoring - it's currently trying to serve a couple of purposes, and they conflict a little bit for the purpose of optimization:
- DIDexchange stores "incomplete" connection records, which other protocols don't want to see.
- DIDexchange stores persistent connection records, which other protocols load occasionally.
- Out-of-band saves invitations and invitation-requests created by the client, which didexchange loads.

Baha-sk (approved):

llorllale:
@aholovko is this ready?

aholovko:
> @aholovko is this ready?

I would like to double check these changes in the downstream project (sandbox)

pkg/didcomm/packager/packager.go
llorllale: @aholovko 

> these properties were used just to carry information from [here](https://github.com/hyperledger/aries-framework-go/blob/d24bb59ffab793f895032ad9356251bb1b2092a3/pkg/didcomm/packager/packager.go#L212) to [here](https://github.com/hyperledger/aries-framework-go/blob/d24bb59ffab793f895032ad9356251bb1b2092a3/pkg/didcomm/transport/ws/pool.go#L111), it was updated as retrieving DIDs should not be the responsibility of `packager.UnpackMessage()`

`myDID` and `theirDID` travel further than the code you pointed to: https://github.com/hyperledger/aries-framework-go/blob/d24bb59ffab793f895032ad9356251bb1b2092a3/pkg/didcomm/transport/ws/pool.go#L111

The end up being included in `DIDCommAction` events, eg. [`introduce` protocol](https://github.com/hyperledger/aries-framework-go/blob/f1c7f7aa39eca010c0c14cdb8ea7e0287db52368/pkg/didcomm/protocol/introduce/properties.go#L25-L32), making this a breaking change to the API.

llorllale: Here is the issuer adapter using these two properties during the issuance flow:

https://github.com/trustbloc/edge-adapter/blob/2835b01720871ae3a74d88ab4247aafd6d14d26e/pkg/restapi/issuer/operation/operations.go#L1362-L1379

llorllale (changes requested):
Concerns about breaking the API without defining a clear transition plan for consumers.

pkg/didcomm/packager/packager.go
aholovko: @llorllale Introduce protocol initializes both `myDID` and `theirDID` properties of `metaData` structure using input parameters of [HandleInbound](https://github.com/hyperledger/aries-framework-go/blob/6c1b27989cc2d17adaac330504b2ddcb440d913e/pkg/didcomm/protocol/introduce/service.go#L296) method. We still pass those data to the `HandleInbound` method - it's just fetched now in [InboundMessageHandler](https://github.com/hyperledger/aries-framework-go/blob/6c1b27989cc2d17adaac330504b2ddcb440d913e/pkg/framework/context/context.go#L183) and not in `UnpackMessage` of packager.

aholovko:
> I noticed that `ConnectionRecorder` uses the standard `context.StorageProvider()`, same as `ConnectionStore`, so I imagine this refactor will enable optimization but won't offer much of a speedup on its own.

`ConnectionRecorder` uses both `ProtocolStateStorageProvider()` and `StorageProvider()` ([code](https://github.com/hyperledger/aries-framework-go/blob/6c1b27989cc2d17adaac330504b2ddcb440d913e/pkg/store/connection/connection_lookup.go#L87)). Transient state is handled by `ProtocolStateStorage`, while cases where permanent storage is needed by "regular" `StorageProvider()`.

@Moopli I totally agree with you regarding

> I think `ConnectionRecorder` could use some further refactoring - it's currently trying to serve a couple of purposes, and they conflict a little bit for the purpose of optimization

aholovko:
> Concerns about breaking the API without defining a clear transition plan for consumers.

@llorllale plz check https://github.com/hyperledger/aries-framework-go/pull/2588#discussion_r594551643

pkg/didcomm/packager/packager.go
llorllale: Fair.

llorllale (approved):

pkg/didcomm/protocol/didexchange/service.go
Baha-sk: that's fine to keep it as is, I'm ok with this for now.

llorllale (changes requested):
Looks good, but we're holding off until downstream projects stabilize.

llorllale (approved):

PR #2594, opened 2021-03-03
feat: add BBS and NISTP did key creation

### LINKED ISSUE #2319: did:key - support P256
The `did:key` VDR should support the P-256 curve. [`JsonWebKey2020`](https://www.w3.org/TR/did-spec-registries/#jsonwebkey2020) can be used as the verification method's `type`.

More on JsonWebKey2020: https://w3c-ccg.github.io/lds-jws2020/#json-web-key-2020

Note: method-specific ID considerations: https://w3c-ccg.github.io/did-method-key/#p-256

Here's a test vector for P-256: https://github.com/w3c-ccg/did-method-key/blob/master/test-vectors/nist-curves.json#L2

Baha-sk:
This change introduces NIST P ECDSA and BBS did:key creation
And did:key resolution for ECDSA keys.

closes #2319

### DIFF
--- a/pkg/vdr/fingerprint/fingerprint.go
+++ b/pkg/vdr/fingerprint/fingerprint.go
@@ -18,14 +18,30 @@ import (
 )
 
 const (
+	// X25519PubKeyMultiCodec for Curve25519 public key in multicodec table.
 	// source: https://github.com/multiformats/multicodec/blob/master/table.csv.
-	ed25519pub = 0xed // Ed25519 public key in multicodec table
+	X25519PubKeyMultiCodec = 0xec
+	// ED25519PubKeyMultiCodec for Ed25519 public key in multicodec table.
+	ED25519PubKeyMultiCodec = 0xed
+	// BLS12381g2PubKeyMultiCodec for BLS12-381 G2 public key in multicodec table.
+	BLS12381g2PubKeyMultiCodec = 0xeb
+	// P256PubKeyMultiCodec for NIST P-256 public key in multicodec table.
+	P256PubKeyMultiCodec = 0x1200
+	// P384PubKeyMultiCodec for NIST P-384 public key in multicodec table.
+	P384PubKeyMultiCodec = 0x1201
+	// P521PubKeyMultiCodec for NIST P-521 public key in multicodec table.
+	P521PubKeyMultiCodec = 0x1202
 )
 
-// CreateDIDKey creates a did:key ID using the multicodec key fingerprint as per the did:key format spec found at:
-// https://w3c-ccg.github.io/did-method-key/#format.
+// CreateDIDKey calls CreateDIDKeyByCode with Ed25519 key code.
 func CreateDIDKey(pubKey []byte) (string, string) {
-	methodID := KeyFingerprint(ed25519pub, pubKey)
+	return CreateDIDKeyByCode(ED25519PubKeyMultiCodec, pubKey)
+}
+
+// CreateDIDKeyByCode creates a did:key ID using the multicodec key fingerprint as per the did:key format spec found at:
+// https://w3c-ccg.github.io/did-method-key/#format.
+func CreateDIDKeyByCode(code uint64, pubKey []byte) (string, string) {
+	methodID := KeyFingerprint(code, pubKey)
 	didKey := fmt.Sprintf("did:key:%s", methodID)
 	keyID := fmt.Sprintf("%s#%s", didKey, methodID)
 
@@ -76,6 +92,9 @@ func PubKeyFromFingerprint(fingerprint string) ([]byte, uint64, error) {
 }
 
 // PubKeyFromDIDKey parses the did:key DID and returns the key's raw value.
+// note: for NIST P ECDSA keys, the raw value does not have the compression point.
+//	In order to use elliptic.Unmarshal() with the raw value, the uncompressed point ([]byte{4}) must be prepended.
+//	see https://github.com/golang/go/blob/master/src/crypto/elliptic/elliptic.go#L319.
 func PubKeyFromDIDKey(didKey string) ([]byte, error) {
 	id, err := did.Parse(didKey)
 	if err != nil {
@@ -95,7 +114,8 @@ func PubKeyFromDIDKey(didKey string) ([]byte, error) {
 	}
 
 	switch code {
-	case ed25519pub:
+	case X25519PubKeyMultiCodec, ED25519PubKeyMultiCodec, BLS12381g2PubKeyMultiCodec,
+		P256PubKeyMultiCodec, P384PubKeyMultiCodec, P521PubKeyMultiCodec:
 		break
 	default:
 		return nil, fmt.Errorf("unsupported key multicodec code [0x%x]", code)
--- a/pkg/vdr/key/creator.go
+++ b/pkg/vdr/key/creator.go
@@ -7,6 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package key
 
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/x509"
+	"errors"
 	"fmt"
 	"time"
 
@@ -22,16 +26,12 @@ const (
 	ed25519VerificationKey2018 = "Ed25519VerificationKey2018"
 	x25519KeyAgreementKey2019  = "X25519KeyAgreementKey2019"
 	bls12381G2Key2020          = "Bls12381G2Key2020"
+	jsonWebKey2020             = "JsonWebKey2020"
 )
 
-const (
-	// source: https://github.com/multiformats/multicodec/blob/master/table.csv.
-	x25519pub     = 0xec // Curve25519 public key in multicodec table
-	ed25519pub    = 0xed // Ed25519 public key in multicodec table
-	bls12381g2pub = 0xeb // BLS12-381 G2 public key in multicodec table
-)
-
-// Create new DID document.
+// Create new DID document for didDoc.
+// Either didDoc must contain non-empty VerificationMethod[] or opts must contain KeyType value of kms.KeyType to create
+// a new key and a corresponding *VerificationMethod entry.
 func (v *VDR) Create(keyManager kms.KeyManager, didDoc *did.Doc,
 	opts ...vdrapi.DIDMethodOption) (*did.DocResolution, error) {
 	createDIDOpts := &vdrapi.DIDMethodOpts{Values: make(map[string]interface{})}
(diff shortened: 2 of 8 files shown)


pkg/framework/aries/api/vdr/vdr.go
```diff
@@ -98,3 +99,10 @@ func WithOption(name string, value interface{}) DIDMethodOption {
 		didMethodOpts.Values[name] = value
 	}
 }
+
+// WithKeyType adds key type option for did method.
+func WithKeyType(kt kms.KeyType) DIDMethodOption {
```
fqutishat, line 104: please remove this option and use WithOption need to keep VDR interface generic

fqutishat (changes requested):

pkg/vdr/fingerprint/fingerprint.go
```diff
@@ -19,13 +19,22 @@ import (
 
 const (
 	// source: https://github.com/multiformats/multicodec/blob/master/table.csv.
-	ed25519pub = 0xed // Ed25519 public key in multicodec table
+	ed25519pub    = 0xed   // Ed25519 public key in multicodec table
+	bls12381g2pub = 0xeb   // BLS12-381 G2 public key in multicodec table
+	p256pub       = 0x1200 // NIST P-256 public key in multicodec table
+	p384pub       = 0x1201 // NIST P-384 public key in multicodec table
+	p521pub       = 0x1202 // NIST P-521 public key in multicodec table
```
soluchok, line 26: Constants should be exported. e.g `CreateDIDKeyByCode(P256pub,key)`

pkg/framework/aries/api/vdr/vdr.go
llorllale: I think this option should be removed altogether to make this interface less confusing.

The method already accepts a `*did.Doc`. I'm already using _that_ to set the verification methods I need in downstream projects. I don't need an additional way to specify a key type.

pkg/vdr/key/creator.go
```diff
@@ -88,10 +97,82 @@ func (v *VDR) Create(keyManager kms.KeyManager, didDoc *did.Doc,
...
+		vmType = bls12381G2Key2020
+	default:
+		keyType = kms.ED25519Type
+		vmType = ed25519VerificationKey2018
+	}
+
+	return keyType, vmType
+}
+
+func getKeyCode(keyType kms.KeyType, verificationMethod *did.VerificationMethod) (uint64, error) {
```
llorllale, line 139: Note: the reverse is also needed... translate a verificationMethod into a `kms.KeyType`: https://github.com/llorllale/edge-service/blob/44ebaad708cd1bc2094b1d70ac72813cfefd8f1c/pkg/restapi/csh/operation/zcapld/zcapld.go#L205-L237

pkg/vdr/key/creator.go
```diff
@@ -88,10 +97,82 @@ func (v *VDR) Create(keyManager kms.KeyManager, didDoc *did.Doc,
 	return &did.DocResolution{DIDDocument: createDoc(publicKey, keyAgr, didKey)}, nil
 }
 
+func getKeyTypeAndVerMethodType(kt kms.KeyType) (kms.KeyType, string) {
```
llorllale, line 100: I don't think you meant to return the `kms.KeyType`

pkg/framework/aries/api/vdr/vdr.go
Baha-sk: @llorllale your case assumes the key value is in the `*did.Doc`, there's a condition in `Create()` that checks if the didDoc.VerificationMethod[] is empty, it will then create a new key from the kms, this is where the key type comes in handy, to create a new key for that keyType option. I will remote that option and use WithOption instead.

pkg/vdr/key/creator.go
```diff
@@ -88,10 +97,82 @@ func (v *VDR) Create(keyManager kms.KeyManager, didDoc *did.Doc,
...
+	var (
+		keyType kms.KeyType
+		vmType  string
+	)
+
+	switch kt {
```
llorllale, line 106: Probably better to have a map.

pkg/framework/aries/api/vdr/vdr.go
llorllale: Yes, I'm aware of the condition. I'm still confused by this interface.

At this moment, `did:key` is the only method I can think of where it makes more sense for the framework to automatically set the verification methods on the resulting `did.Doc`. This doesn't apply to any other method. So we are adding an option that creates ambiguity on the interface, all to support a single DID method.

This option isn't needed - that's why it was never added. The verification methods are being manually set with `did.Doc`.

llorllale: Speaking of which - I think the `key.KeyManager` should also be removed from the method args in the future.

llorllale: We should provide a uniform way to create DID documents - this PR is taking us down an even more fractured path.

pkg/vdr/key/creator.go
```diff
@@ -40,38 +45,42 @@ func (v *VDR) Create(keyManager kms.KeyManager, didDoc *did.Doc,
 		opt(createDIDOpts)
 	}
 
+	keyType, vmType := getKeyTypeAndVerMethodType(createDIDOpts.KeyType)
```
llorllale, line 48: This code is confusing.

If the user does not provide the key type option, it defaults to `Ed25519`, regardless of the verification method type they have set in `did.Doc`.

pkg/vdr/key/creator.go
```diff
@@ -88,10 +97,82 @@ func (v *VDR) Create(keyManager kms.KeyManager, didDoc *did.Doc,
...
+func getKeyCode(keyType kms.KeyType, verificationMethod *did.VerificationMethod) (uint64, error) {
+	var keyCode uint64
+
+	switch verificationMethod.Type {
+	case ed25519VerificationKey2018:
+		keyCode = ed25519Pub
+	case bls12381G2Key2020:
+		keyCode = bls12381g2Pub
+	case jsonWebKey2020:
+		switch keyType {
```
llorllale, line 148: For more cleaner code, the `keyCode` should be derived from the values of the verificationMethod.

pkg/vdr/key/creator.go
Baha-sk: fyi, JWS signing works with keys encoded in IEEE_P1363 format, verifying with ASN.1 DER key format don't work with JWS signatures, this is why we recommend using DER ECDSA key types.

Baha-sk: also my change allows for creating ECDSA  keys in both formats, the logic from your link only supports one format (DER).

this is why doing a switch over the `keyType` is needed

pkg/vdr/key/creator.go
Baha-sk: sure, I can use the key type directly

pkg/vdr/key/creator.go
llorllale: Yes, the link was provided to show that this kind of conversion is needed. Improvements are welcome.

pkg/framework/aries/api/vdr/vdr.go
Baha-sk: are you saying this DID:key creator doesn't need to handle kms key creation? (ie VerificaionMethod[] will never be empty?) I'm happy to remove this condition and kms test cases, it will make my life a lot easier!

pkg/vdr/key/creator.go
Baha-sk: if we don't need to create a **kms** key in `CreateDIDKey()` then this call is pointless

I will remove it and remove kms key creation as well

pkg/framework/aries/api/vdr/vdr.go
llorllale: No - `did:key` does not need to automatically create keys. They should be provided to it like all other methods.

Baha-sk: in case where DID:Key creation doesn't have a full DID document (with empty `[]verificationMethod`) the `Create()` function now accepts a general VDR option (using `WithOption()`) registered as an option key `KeyType` to create a new KMS key and a new `*VerificationMethod` entry to be added to the DIDDoc.

If the option is not passed, then the original behaviour with a DID document having a non empty `[]VerificationMethod` is required.

The custom `WithKeyType()` option has been removed.

pkg/vdr/fingerprint/fingerprint.go
Baha-sk: contents are now exported

pkg/vdr/key/creator.go
Baha-sk: the logic of getKeyCode() is now enhanced:
* if keyType option was passed, the key code will be determined by keyType
* else the VerificationMethod.Value containing the raw key is used to determine the key code based on the raw key's length (for JSONWebKey2020 keys assuming they're only ECDSA keys for now, this can be enhanced later to support other keys as JSONWebKey2020, the challenge is we can't determine the key type from the raw key value alone so for now we assume JSONWebKey2020 are ECDSA keys).

The reverse is available via `fingerprint.PubKeyFromFingerPrint(fingerprint string) ([]byte, uint64, error) ` [here](https://github.com/hyperledger/aries-framework-go/pull/2594/files#diff-db294652c6ab1f742e49369b6ad733d1f4a3312dc0072e08bd8aed2ae43a92f8R71)

pkg/vdr/key/creator.go
Baha-sk: the function now is simplified and renamed to `getVerMethodType`. It returns the VerificationMethod.Type value only

pkg/vdr/key/creator.go
Baha-sk: a new function called `getVerMethodType` is added and has a map for verification method type

pkg/vdr/key/creator.go
Baha-sk: after careful thought, I ended up not removing the kms key creation, to keep the original behaviour of creating a new key and a new *VerificationMethod if the didDoc has an empty list of []VerificationMethod

pkg/vdr/key/creator.go
Baha-sk: yes, `keyCode` is derived from the values of verificationMethod, but if keyType option was used (ie creating a new kms key and VerificatonMethod entry) then keyType is more specific. See my previous comment [above](https://github.com/hyperledger/aries-framework-go/pull/2594#discussion_r587078042)

soluchok (approved):

fqutishat (approved):

pkg/vdr/key/creator.go
llorllale: > else the VerificationMethod.Value containing the raw key is used to determine the key code based on the raw key's length (for JSONWebKey2020 keys assuming they're only ECDSA keys for now, this can be enhanced later to support other keys as JSONWebKey2020, the challenge is we can't determine the key type from the raw key value alone so for now we assume JSONWebKey2020 are ECDSA keys).

JsonWebKey2020 is just a JWK and has the `crv` attribute... [example](https://w3c-ccg.github.io/lds-jws2020/#example-1). To determine whether the use is signature or key agreement, you would check the verification method relationship (basically, all verification method relationships are for signing except `keyAgreement`).

llorllale: See [this table](https://w3c-ccg.github.io/lds-jws2020/#jose-conformance).

llorllale (approved):

pkg/vdr/key/creator.go
Baha-sk: Agreed, the logic needs to be enhanced by fetching the JWK of that verification method (`verificationMethod.JSONWebKey()`), with this we can detect the raw key type and multicodec key code. 

This can be done in a followup change.

PR #2602, opened 2021-03-04
feat: BBS+ BDD tests for PE

soluchok:

### DIFF
--- a/pkg/controller/command/verifiable/command.go
+++ b/pkg/controller/command/verifiable/command.go
@@ -156,14 +156,21 @@ type kmsSigner struct {
 	bbs       bool
 }
 
-func newKMSSigner(keyManager kms.KeyManager, c ariescrypto.Crypto, creator string) (*kmsSigner, error) {
-	// creator will contain didID#keyID
-	idSplit := strings.Split(creator, "#")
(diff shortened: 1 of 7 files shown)


fqutishat (approved):

sudeshrshetty (approved):

rolsonquadras (approved):

pkg/controller/command/verifiable/command.go
```diff
@@ -156,14 +156,21 @@ type kmsSigner struct {
...
 	}
 
-	keyHandler, err := keyManager.Get(idSplit[1])
+	idSplit := strings.Split(opts.VerificationMethod, "#")
+	if len(idSplit) == creatorParts {
+		return idSplit[1]
+	}
+
+	return ""
+}
```
Baha-sk, line 170: is this logic still relevant? this is not a kms KID value

aholovko (approved):

Baha-sk (approved):

pkg/controller/command/verifiable/command.go
soluchok: I don't want to remove it because it will break compatibility (especially before release). Some services might depend on it.

PR #2604, opened 2021-03-05
feat: LevelDB implementation of new Aries storage interfaces

### LINKED ISSUE #2532: LevelDB implementation of new storage interface

DRK3:

closes #2532


component/storage/leveldb/leveldb.go
```diff
@@ -0,0 +1,503 @@
...
+// Provider leveldb implementation of storage.Provider interface.
+type Provider struct {
+	dbPath string
+	dbs    map[string]*store
+	lock   sync.RWMutex
+}
+
+type closer func(storeName string)
+
+type tagMapping map[string]map[string]struct{} // map[TagName](Set of database Keys)
```
DRK3, line 44: Note: LevelDB has no built in indexing mechanism, so the tag/indexing stuff has to be handled in the provider/store implementation directly.

fqutishat (approved):

component/storage/leveldb/leveldb.go
```diff
@@ -0,0 +1,503 @@
...
+
+	for _, openStore := range openStoresSnapshot {
+		err := openStore.Close()
+		if err != nil {
+			return fmt.Errorf(`failed to close open store with name "%s": %w`, openStore.name, err)
+		}
+	}
+
+	return nil
+}
```
soluchok, line 147:
```
func (p *Provider) Close() error {
	p.lock.Lock()
        deffer p.lock.Unlock()

	for _, openStore := range p.dbs {
		if err := openStore.Close(); err != nil {
			return fmt.Errorf(`failed to close open store with name "%s": %w`, openStore.name, err)
		}
	}

	return nil
}
```

component/storage/leveldb/leveldb.go
```diff
@@ -0,0 +1,503 @@
...
+
+// Put stores the key and the record.
+func (s *store) Put(key string, value []byte, tags ...storage.Tag) error {
+	if key == "" {
+		return errors.New("key cannot be blank")
+	}
+
+	if value == nil {
+		return errors.New("value cannot be nil")
+	}
```
soluchok, line 210: Why value cannot be blank/nil? Should be removed.

component/storage/leveldb/leveldb_test.go
```diff
@@ -0,0 +1,207 @@
...
+		t.Fatalf("Failed to create leveldb directory: %s", err)
+	}
+
+	return dbPath, func() {
+		err := os.RemoveAll(dbPath)
+		if err != nil {
+			t.Fatalf("Failed to clear leveldb directory: %s", err)
+		}
+	}
+}
```
soluchok, line 35: Better to use `t.Cleanup(func() {})`

component/storage/leveldb/leveldb.go
llorllale: I haven't looked closely at all this storage work, but does this imply then that each individual node in a cluster has a different view of `tagMapping`? If so then that's not good and sounds like it should be stored in LevelDB.

component/storage/leveldb/leveldb.go
DRK3: There's a couple reasons...
1. For the Batch method defined in the interface, a `nil` value is special and indicates a deletion (See https://github.com/hyperledger/aries-framework-go/blob/9009c86eef13eaee1b7f7d686ba74f4c4f68f20c/spi/storage/storage.go#L66). So a "nil" value has special meaning. If you were to replace a `Put` call with a `Batch` call, the behaviour should be the same. Imagine if you were storing a nil value before, and then suddenly it becomes a deletion... So for consistency, I think it's best if you can't store a `nil` value. I already defined in the interface that this is an error (https://github.com/hyperledger/aries-framework-go/blob/9009c86eef13eaee1b7f7d686ba74f4c4f68f20c/spi/storage/storage.go#L100) before as well.
2. I have common storage tests ensuring that a nil value causes an error already (https://github.com/hyperledger/aries-framework-go/blob/9009c86eef13eaee1b7f7d686ba74f4c4f68f20c/test/component/storage/storage.go#L407), and those are used here, so to keep them passing I have to maintain it.

component/storage/leveldb/leveldb.go
DRK3: Done!

component/storage/leveldb/leveldb_test.go
DRK3: Done!

component/storage/leveldb/leveldb.go
DRK3: Actually, I can't do this - the reason is that each store's close call calls into Provider.removeStore, so that results in a deadlock. The reason for the openStoresSnapshot thing is to work around this. (I do a similar thing in most other providers)

PR #2620, opened 2021-03-09
fix: create local doc cache loader

### LINKED ISSUE #1833: Proof shouldn't override/hardcode jsonld contexts

Baha-sk:
This change creates a local JSON-LD Document Cache loader using a cache of type *sync.Map
to avoid 'error on cache writes' panics.

closes #1833 for the fatal error mentioned in the comment
Also related and closes the discussion about the same panic topic in #2487


pkg/doc/signature/jsonld/processor_test.go
```diff
@@ -685,6 +711,25 @@ const jsonVCWithProperContexts = `{
     ]
 }`
 
+// nolint
+const jsonVCWithProperContexts2 = `{
```
Baha-sk, line 715: example picked from @llorllale's comment [here](https://github.com/hyperledger/aries-framework-go/issues/1833#issuecomment-739317499)

fqutishat (approved):

pkg/doc/signature/jsonld/processor_test.go
```diff
@@ -1607,3 +1665,222 @@ func TestTransformBlankNodes(t *testing.T) {
...
+			//	name:   "canonizing sample document with extra io.Reader cached dummy context",
+			//	doc:    jsonLDMultipleInvalidRDFs,
+			//	result: canonizedSampleVP_extraContext,
+			//	opts: []ProcessorOpts{
+			//		WithRemoveAllInvalidRDF(), WithExternalContext("http://localhost:8652/dummy.jsonld"),
+			//		WithDocumentLoaderCache(map[string]interface{}{
+			//			"http://localhost:8652/dummy.jsonld": strings.NewReader(extraJSONLDContext),
+			//		}),
+			//	},
+			// },
```
soluchok, line 1791: why it is commented out?

soluchok (approved):

pkg/doc/signature/jsonld/processor.go
```diff
@@ -506,34 +507,43 @@ func (p *Processor) normalizeFilteredDataset(view string) (string, error) {
...
-	for k, v := range cache {
-		cachingLoader.AddDocument(k, v)
+	if cache != nil {
+		cache.Range(func(key, value interface{}) bool {
+			k, ok := key.(string)
+			if !ok {
+				return false
+			}
+
+			cachingLoader.AddDocument(k, value)
```
aholovko, line 528: wouldn't this add the same documents again and again to `cachingLoader`?

aholovko (changes requested):

pkg/doc/signature/jsonld/processor.go
Baha-sk: it's adding from cache into cacheLoader, just like the original logic, it's just the way looping through cache is different because of its `*sync.Map` type

pkg/doc/signature/jsonld/processor_test.go
Baha-sk: Because the buffer in `strings.NewReader()` is not thread safe.
One would need to use a thread safe Reader implementation for this to work. Example: https://gist.github.com/jmackie/11570bdcd8a4c10d72619a5e1f21c5f8

I can probably add a note in the DocumentLoader constructor about this. But this is an edge case with a seldom use. Users need to be aware to avoid using strings.NewReader() as a sync.Map value.

pkg/doc/jsonld/document_loader.go
```diff
@@ -0,0 +1,55 @@
...
+	"errors"
+	"sync"
+
+	"github.com/piprate/json-gold/ld"
+)
+
+// CachingDocumentLoader is similar to json-gold's CachingDocumentLoader but uses cache as *sync.Map instead.
+type CachingDocumentLoader struct {
+	nextLoader ld.DocumentLoader
+	cache      *sync.Map
```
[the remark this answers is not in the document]
Baha-sk: I'll update my pr with standard map and a RWMutex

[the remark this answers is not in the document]
Baha-sk: replaced

pkg/doc/signature/jsonld/processor.go
```diff
@@ -38,7 +39,7 @@ type processorOpts struct {
 	validateRDF         bool
 	documentLoader      ld.DocumentLoader
 	externalContexts    []string
-	documentLoaderCache map[string]interface{}
+	documentLoaderCache *sync.Map
```
Baha-sk, line 42: I missed this change.. will update in a bit..

Baha-sk: updated

pkg/doc/jsonld/document_loader.go
```diff
@@ -0,0 +1,73 @@
...
+		if cachedDoc, ok := doc.(*ld.RemoteDocument); ok {
+			return cachedDoc, nil
+		}
+
+		return nil, errors.New("invalid document entry")
+	}
+
+	cdl.rwMutex.RUnlock()
+
+	docFromLoader, err := cdl.nextLoader.LoadDocument(u)
```
llorllale, line 48: Looks like `nextLoader` must also be thread-safe.

This will be an issue when people want to add extra contexts in downstream projects and they understandably forget about this bug, such as when they use the [`DocumentVerifier`](https://github.com/hyperledger/aries-framework-go/blob/fea319ee4b47b0a81ceea9f2e6d96e3902c5d1f1/pkg/doc/signature/verifier/verifier.go#L70-L79) directly (like zcaps do).

The better solution is to have the user prepare the contexts themselves in an `ld.DocumentLoader` and be done with this. No need for locking or any of this extra complexity.

PR #2630, opened 2021-03-10
chore: presexch - expose JSONSchema, add JSONLD context

llorllale:

### DIFF
--- a/pkg/doc/presexch/schema.go
+++ b/pkg/doc/presexch/schema.go
@@ -6,7 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 
 package presexch
 
-const definitionSchema = `
+// DefinitionJSONSchema is the JSONSchema definition for PresentationDefinition.
+// TODO - which commit version of the spec was this copied from?
+const DefinitionJSONSchema = `
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "definitions": {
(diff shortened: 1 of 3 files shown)


pkg/doc/presexch/schema.go
```diff
@@ -230,3 +232,21 @@ const definitionSchema = `
         }
     }
 }`
+
+// PresentationSubmissionJSONLDContext based con this commit:
+// https://github.com/decentralized-identity/presentation-exchange/commit/980ee781976fa0a6a247807e879f7d7d0e7e8e0a.
+const PresentationSubmissionJSONLDContext = `{
```
soluchok, line 238: we already have it here `https://github.com/hyperledger/aries-framework-go/blob/809d27ce5111714a1a23c62648cec9d19fae13bd/pkg/doc/presexch/definition.go#L75`

soluchok (approved):

pkg/doc/presexch/schema.go
llorllale: oh :)

llorllale: Substituted `presentationSubmissionContext` for the new exposed one.

Also substituted use of `presentationSubmissionContextURI` with the one already exposed.

troyronda (approved):

PR #2650, opened 2021-03-17
feat: add DIDComm supported envelope types

### LINKED ISSUE #2646: add DIDcomm V1/V2 `cty` JWE header value when packing with the new (JWE) packers
Make sure the `cty` header is added to the JWE headers with a value matching DIDComm's plaintext header values:
v1: [application/json;flavor=didcomm-msg](https://github.com/hyperledger/aries-rfcs/tree/master/features/0044-didcomm-file-and-mime-types#didcomm-v1-messages-dm)

v2: [application/didcomm-plain+json](https://identity.foundation/didcomm-messaging/spec/#didcomm-plaintext-messages)

Also ensure the JWE `typ` header reflect the JWE media type for DIDcomm (v2): [application/didcomm-encrypted+json](https://identity.foundation/didcomm-messaging/spec/#didcomm-encrypted-message)

Baha-sk:
closes #2646

### DIFF
--- a/pkg/didcomm/packer/anoncrypt/pack.go
+++ b/pkg/didcomm/packer/anoncrypt/pack.go
@@ -32,8 +32,6 @@ import (
 // messages anonymously between parties with message repudiation, ie the sender identity is not revealed (and therefore
 // not authenticated) to the recipient(s).
 
-const encodingType = "didcomm-envelope-enc"
-
 var logger = log.New("aries-framework/pkg/didcomm/packer/anoncrypt")
 
 // Packer represents an Anoncrypt Pack/Unpacker that outputs/reads Aries envelopes.
@@ -63,10 +61,10 @@ func New(ctx packer.Provider, encAlg jose.EncAlg) (*Packer, error) {
 	}, nil
 }
 
-// Pack will encode the payload argument
-// Using the protocol defined by the Anoncrypt message of Aries RFC 0334
-// Anoncrypt ignores the sender argument, it's added to meet the Packer interface.
-func (p *Packer) Pack(payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, error) {
+// Pack will encode the payload argument using the protocol defined by the Anoncrypt message of Aries RFC 0334.
+// Anoncrypt ignores the sender argument, it's added to meet the Packer interface. It uses DIDComm typ V2 header value
+// (default envelope 'typ' protected header).
+func (p *Packer) Pack(contentType string, payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, error) {
 	if len(recipientsPubKeys) == 0 {
 		return nil, fmt.Errorf("anoncrypt Pack: empty recipientsPubKeys")
 	}
@@ -76,7 +74,8 @@ func (p *Packer) Pack(payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, er
 		return nil, fmt.Errorf("anoncrypt Pack: failed to convert recipient keys: %w", err)
 	}
 
-	jweEncrypter, err := jose.NewJWEEncrypt(p.encAlg, encodingType, "", nil, recECKeys, p.cryptoService)
+	jweEncrypter, err := jose.NewJWEEncrypt(p.encAlg, p.EncodingType(), contentType, "",
+		nil, recECKeys, p.cryptoService)
 	if err != nil {
 		return nil, fmt.Errorf("anoncrypt Pack: failed to new JWEEncrypt instance: %w", err)
 	}
@@ -85,6 +84,13 @@ func (p *Packer) Pack(payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, er
 	if err != nil {
 		return nil, fmt.Errorf("anoncrypt Pack: failed to encrypt payload: %w", err)
 	}
+
+	mPh, err := json.Marshal(jwe.ProtectedHeaders)
+	if err != nil {
+		return nil, fmt.Errorf("anoncrypt Pack: %w", err)
+	}
+
+	logger.Debugf("protected headers: %s", mPh)
 
 	var s string
 
@@ -137,9 +143,9 @@ func unmarshalRecipientKeys(keys [][]byte) ([]*cryptoapi.PublicKey, []byte, erro
 
 // Unpack will decode the envelope using a standard format.
 func (p *Packer) Unpack(envelope []byte) (*transport.Envelope, error) {
-	jwe, err := jose.Deserialize(string(envelope))
+	jwe, cty, err := deserializeEnvelope(envelope)
 	if err != nil {
-		return nil, fmt.Errorf("anoncrypt Unpack: failed to deserialize JWE message: %w", err)
+		return nil, err
 	}
 
 	for i := range jwe.Recipients {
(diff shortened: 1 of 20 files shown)


pkg/didcomm/packer/anoncrypt/pack.go
```diff
@@ -101,6 +107,26 @@ func (p *Packer) Pack(payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, er
...
+	}
+
+	typ := pOpts.TypHeader()
+	if typ == "" {
+		typ = packer.EncodingTypeV2
+	}
+
+	cty := pOpts.CTYHeader()
+	if cty == "" {
+		cty = packer.ContentEncodingTypeV2
```
llorllale, line 124: This is incorrect - default cannot be DIDComm V2 media types since they are not part of the core group in AIP 2.0

pkg/didcomm/packer/anoncrypt/pack.go
```diff
@@ -101,6 +107,26 @@ func (p *Packer) Pack(payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, er
...
+func jweTypesFromOpts(opts []packer.Opts) (string, string) {
+	pOpts := packer.NewOpt()
+
+	for _, opt := range opts {
+		opt(pOpts)
+	}
+
+	typ := pOpts.TypHeader()
+	if typ == "" {
+		typ = packer.EncodingTypeV2
```
llorllale, line 119: DIDComm V2 cannot be default while not being part of the core group in AIP 2.0

pkg/didcomm/packer/packer_opts.go
```diff
@@ -0,0 +1,46 @@
...
+	return pk.typHeader
+}
+
+// CTYHeader gets the 'cty' header to be used when packing a message.
+func (pk *packerOpts) CTYHeader() string {
+	return pk.ctyHeader
+}
+
+// WithTypHeader sets the `typ` header when packing.
+func WithTypHeader(typ string) Opts {
```
llorllale, line 35: This PR only uses these two options in unit tests. I assume there will be a followup to expose this as an option to `aries.New()` somehow.

llorllale (changes requested):
See comments

pkg/didcomm/packer/packer_opts.go
Baha-sk: currently I decided to have these options exposed on Pack() calls directly, ie  not passing these headers to the packer's New() calls

this means the user has the option to set the `typ` and `cty` protected header when packing which saves the user from creating different packers with different values, one can simply call `Pack(... , packer.WitTypHeader(customTyp), packer.WithCTYHeader(customCTY))`.

there's no need to configure this info at the framework level

pkg/didcomm/packer/anoncrypt/pack.go
Baha-sk: this anoncrypt implementation is not going to be supported by AIP 2.0 since their anoncrypt version is different:
* this implementation uses ECDH-ES (no sender info used at all)
* AIP 2.0's version of anoncypt uses a fresh DID with a fresh sender key (this is technically our AFG's authcrypt implementation with a new random sender key).

But to answer your concern, since JWE packers (both anoncrypt/authcrypt) are currently not used AFG DIDComm, there's no point in making V1 the default, we'll support V2 from the get go when DIDComm is configured to use these packers by default.

fqutishat (approved):

pkg/didcomm/packer/packer_opts.go
Baha-sk: options are removed

pkg/didcomm/packer/api.go
```diff
@@ -13,6 +13,22 @@ import (
 	"github.com/hyperledger/aries-framework-go/spi/storage"
 )
 
+const (
+	// EncodingTypeV2 is the default JWE `typ` protected header value as per:
```
troyronda, line 17: where is EncodingTypeV1?

pkg/didcomm/packer/api.go
```diff
@@ -13,6 +13,22 @@ import (
 	"github.com/hyperledger/aries-framework-go/spi/storage"
 )
 
+const (
```
troyronda, line 16: where is EncodingTypeV1 defined?

pkg/didcomm/packer/api.go
```diff
@@ -13,6 +13,22 @@ import (
...
+	// EncodingTypeV2 is the default JWE `typ` protected header value as per:
+	// https://identity.foundation/didcomm-messaging/spec/#didcomm-encrypted-message
+	// and
+	//nolint:lll
+	// https://github.com/hyperledger/aries-rfcs/blob/master/features/0044-didcomm-file-and-mime-types/README.md#detecting-didcomm-versions
+	// for DIDComm compliance.
+	EncodingTypeV2 = "application/didcomm-encrypted+json"
```
troyronda, line 23: You might want to call this `EnvelopeEncodingTypeV2`

pkg/doc/jose/common.go
```diff
@@ -87,8 +87,10 @@ const (
 	A256GCMALG = "A256GCM"
 	// XC20PALG represented XChacha20Poly1305 content encryption algorithm value.
 	XC20PALG = "XC20P"
-	// DIDCommEncType representing the JWE 'Typ' protected type header.
-	DIDCommEncType = "didcomm-envelope-enc"
+	// DIDCommEncType representing the JWE 'Typ' protected type header for DIDComm V2 (default).
```
troyronda, line 90: This should be moved to the test file.

pkg/doc/jose/common.go
```diff
@@ -87,8 +87,10 @@ const (
...
+	DIDCommEncType = "application/didcomm-encrypted+json"
+	// DIDCommContentEncodingType represent the JWE `Cty` protected type header for DIDComm V2 (default).
```
troyronda, line 92: This should be moved to the test file.

pkg/didcomm/packer/anoncrypt/pack.go
```diff
@@ -63,20 +61,27 @@ func New(ctx packer.Provider, encAlg jose.EncAlg) (*Packer, error) {
...
-func (p *Packer) Pack(payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, error) {
+// Pack will encode the payload argument using the protocol defined by the Anoncrypt message of Aries RFC 0334.
+// Anoncrypt ignores the sender argument, it's added to meet the Packer interface. It uses DIDComm typ and cty V2 values
+// (default envelope type and content type headers).
+func (p *Packer) Pack(contentType string, payload, _ []byte, recipientsPubKeys [][]byte) ([]byte, error) {
 	if len(recipientsPubKeys) == 0 {
 		return nil, fmt.Errorf("anoncrypt Pack: empty recipientsPubKeys")
 	}
 
+	switch contentType {
```
troyronda, line 72: I don't think contentType is the responsibility of the packer except for passing it through.

pkg/didcomm/packer/api.go
Baha-sk: it was already defined in [LegacyPacker](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/didcomm/packer/legacy/authcrypt/authcrypt.go#L24)

pkg/didcomm/packer/api.go
Baha-sk: ok..

pkg/didcomm/packer/anoncrypt/pack.go
Baha-sk: ok, I'll remove this switch statement

pkg/didcomm/packer/api.go
Baha-sk: updated

pkg/doc/jose/common.go
Baha-sk: moved to test file

pkg/doc/jose/common.go
Baha-sk: moved to test file

pkg/didcomm/packer/anoncrypt/pack.go
Baha-sk: switch statement removed, now city is pass through in the packer and JWEEncrypter, if it is set, it will be added to the JWE protect headers.

pkg/didcomm/packer/anoncrypt/pack.go
Baha-sk: ContentEncodingType (`cty`) is passed to the packers, there's no more default, this code is outdated

llorllale (approved):

PR #2652, opened 2021-03-18
feat: VC wallet contents - add/get/remove

### LINKED ISSUE #2651: VC Wallet Content store - Add/Remove/Get

sudeshrshetty:
- fixed: UUID for dbnamespacing instead of user ID.
- added add/get/remove features for credential, collection, diddocument,
metadata, connection data models.
- closes #2651

### DIFF
--- a/pkg/client/vcwallet/client_test.go
+++ b/pkg/client/vcwallet/client_test.go
@@ -8,6 +8,7 @@ package vcwallet
 
 import (
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"testing"
 
@@ -98,21 +99,42 @@ func TestCreate(t *testing.T) {
 		require.Error(t, err)
 		require.Empty(t, wallet)
 	})
+
+	t.Run("test create new wallet failure - create content store error", func(t *testing.T) {
+		mockctx := newMockProvider()
+		mockctx.storeProvider = &mockStorageProvider{
+			MockStoreProvider: mockstorage.NewMockStoreProvider(),
+			failure:           fmt.Errorf(sampleClientErr),
+		}
+
+		err := CreateProfile(sampleUserID, mockctx, WithKeyServerURL(sampleKeyServerURL))
+		require.NoError(t, err)
+
+		wallet, err := New(sampleUserID, mockctx)
+		require.Error(t, err)
+		require.Empty(t, wallet)
+		require.Contains(t, err.Error(), "failed to get wallet content store:")
+	})
 }
 
 func TestUpdate(t *testing.T) {
 	t.Run("test update wallet client using local kms passphrase", func(t *testing.T) {
 		mockctx := newMockProvider()
+		createSampleProfile(t, mockctx)
+
 		err := UpdateProfile(sampleUserID, mockctx, WithPassphrase(samplePassPhrase))
 		require.NoError(t, err)
 
 		wallet, err := New(sampleUserID, mockctx)
 		require.NoError(t, err)
 		require.NotEmpty(t, wallet)
+		require.NotEmpty(t, wallet.profile.MasterLockCipher)
 	})
 
 	t.Run("test update wallet client using local kms secret lock service", func(t *testing.T) {
 		mockctx := newMockProvider()
+		createSampleProfile(t, mockctx)
+
 		err := UpdateProfile(sampleUserID, mockctx, WithSecretLockService(&secretlock.MockSecretLock{}))
 		require.NoError(t, err)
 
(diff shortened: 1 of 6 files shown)


pkg/client/vcwallet/client_test.go
```diff
@@ -372,10 +414,20 @@ func TestClient_Get(t *testing.T) {
...
+	require.NotEmpty(t, content)
+
+	err = vcWalletClient.Remove(Metadata, "did:example:123456789abcdefghi")
+	require.NoError(t, err)
+
+	content, err = vcWalletClient.Get(Metadata, "did:example:123456789abcdefghi")
+	require.Empty(t, content)
 	require.Error(t, err)
-	require.EqualError(t, err, toBeImplementedErr)
+	require.Contains(t, err.Error(), "data not found")
```
DRK3, line 430: Consider using `errors.Is(spi.ErrDataNotFound)` (if possible)

sudeshrshetty: this is unit test and this is "contains" check

pkg/client/vcwallet/contents.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// newContentStore returns new wallet content store instance.
+func newContentStore(p storage.Provider, pr *profile) (*contentStore, error) {
+	store, err := p.OpenStore(pr.ID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create store for user '%s' : %w", pr.User, err)
+	}
+
+	// storage.StoreConfiguration{TagNames: []string{connIDKeyPrefix}}
```
DRK3, line 75: Should this comment be here?

sudeshrshetty: will remove it

sudeshrshetty: done

rolsonquadras (approved):

pkg/client/vcwallet/client_test.go
DRK3: You can also check for it by using `errors.Is(spi.ErrDataNotFound)`, which is an even better check since it will avoid false positives and will ensure that even if the error message changes, that the test will still pass

DRK3: You don't need to check the whole error this way either. You don't want some other "data not found" string to accidentally pass it.

pkg/client/vcwallet/contents_test.go
```diff
@@ -0,0 +1,265 @@
...
+	})
+}
+
+type mockStorageProvider struct {
+	*mockstorage.MockStoreProvider
+	config  storage.StoreConfiguration
+	failure error
+}
+
+// SetStoreConfig always return a nil error.
```
DRK3, line 251: Looks like this comment isn't accurate

pkg/client/vcwallet/contents_test.go
```diff
@@ -0,0 +1,265 @@
...
+func (s *mockStorageProvider) SetStoreConfig(name string, config storage.StoreConfiguration) error {
+	s.config = config
+
+	return s.failure
+}
+
+// GetStoreConfig is not implemented.
```
DRK3, line 258: This comment also looks inaccurate

pkg/client/vcwallet/profile.go
```diff
@@ -25,6 +27,9 @@ var ErrProfileNotFound = errors.New("profile does not exist")
 
 // profile of VC wallet contains wallet specific settings of wallet user to be remembered.
 type profile struct {
+	// ID unique identifier assigned to this wallet profile.
+	ID string
+
 	// User ID of the wallet profile user.
 	User string
```
DRK3, line 34: I'm confused about the difference between `ID` and `User`. Based on the comments they sort of sound like the same thing? Maybe you can clarify the comments

DRK3 (changes requested):

pkg/client/vcwallet/contents_test.go
sudeshrshetty: fixed

pkg/client/vcwallet/contents_test.go
sudeshrshetty: fixed

pkg/client/vcwallet/profile.go
sudeshrshetty: assigning UUID for each user profile which will be used for db namespacing, based on comments from previous PR

DRK3: You should add it to the comment so future readers know - if I was looking at that code I wouldn't have any idea what the difference is

DRK3: Or use different names or something

fqutishat (approved):

pkg/client/vcwallet/profile.go
DRK3: Talked to @sudeshrshetty about it - I misread the code - please disregard my original concern!

PR #2655, opened 2021-03-19
fix: fixes for didexchange interop with aca-py

Moopli:
accept old and new didex. message formats and mime types:
- application/ssi-agent-wire is accepted as an inbound mime type
- did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0 is accepted as
  an inbound PIURI for didexchange messages
- did doc can be parsed from "did_doc~attach" attachment in request
  messages, as per didexchange spec, in addition to being accepted
  inside "connection" field, as per connection-request spec

did doc format:
- Parse legacy peer did docs with publicKey members even if their
  context states they are using a newer format.
- Translate raw ed25519 public key values to did:key when reading
  recipient keys or routing keys


pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -44,6 +46,17 @@ const (
 	ResponseMsgType = PIURI + "/response"
 	// AckMsgType defines the did-exchange ack message type.
 	AckMsgType = PIURI + "/ack"
+	// OldPIURI is the old did-exchange protocol identifier URI.
+	OldPIURI = "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0"
```
troyronda, line 50: @swcurran @andrewwhitehead when does this old PIURI go away?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -431,6 +436,11 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
 		return nil, nil, fmt.Errorf("handle inbound request: %w", err)
 	}
 
+	fmt.Printf("###\ntheir Doc: %#v\n", requestDidDoc)
```
troyronda, line 439: Remove prints

pkg/didcomm/transport/http/inbound.go
```diff
@@ -21,6 +21,11 @@ import (
 
 var logger = log.New("aries-framework/http")
 
+const (
+	// acceptInboundContentType additional content type to be accepted for inbound messages.
+	acceptInboundContentType = "application/ssi-agent-wire"
```
troyronda, line 26: @swcurran @andrewwhitehead when does this old media type go away?

pkg/vdr/peer/vdr.go
```diff
@@ -50,5 +50,5 @@ func (v *VDR) Deactivate(did string, opts ...vdrapi.DIDMethodOption) error {
 
 // Accept did method.
 func (v *VDR) Accept(method string) bool {
-	return method == DIDMethod
+	return method == DIDMethod || method == "sov" // accept sov 'peer' did docs for interop
```
troyronda, line 53: @swcurran @andrewwhitehead @Moopli @llorllale I don’t understand this one. Why would a did:sov be interpreted as did:peer?

pkg/doc/did/doc.go
```diff
@@ -360,12 +360,61 @@ func ParseDocument(data []byte) (*Doc, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("populate proofs failed: %w", err)
+	}
+
+	doc.Proof = proofs
+
+	return doc, nil
+}
+
+func legacyParseSov(raw *rawDoc) (*Doc, error) {
```
troyronda, line 417: @swcurran @andrewwhitehead DID documents now use the verificationMethod property not publicKey. We are getting an invalid DID document.

pkg/didcomm/protocol/didexchange/service.go
```diff
@@ -246,7 +259,11 @@ func (s *Service) Accept(msgType string) bool {
 	return msgType == InvitationMsgType ||
 		msgType == RequestMsgType ||
 		msgType == ResponseMsgType ||
-		msgType == AckMsgType
+		msgType == AckMsgType ||
+		msgType == OldInvitationMsgType ||
```
troyronda, line 263: I'd prefer doing a string replace of `did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/` => `https://didcomm.org/` at a higher layer @Moopli @llorllale

pkg/didcomm/protocol/didexchange/service.go
troyronda: Not sure if this is due to the [msg transition](https://github.com/hyperledger/aries-rfcs/blob/233aeae8f730c8790533e91b52363ed0715ab225/features/0348-transition-msg-type-to-https/README.md), but DID Exchange was one of the "new" protocols. @swcurran @andrewwhitehead 

> Note: Any RFCs that already use the new "https" message type should continue to use the use new format in all cases—accepting and sending. New protocols defined in new and updated RFCs should use the new "https" format.

andrewwhitehead: At the moment ACA-Py has a startup option to use the new URI prefix. If the .NET framework now accepts that then we may be able to make it the default in the next release. Changing the prefix is automatic, so it doesn't have custom behaviour depending on the message type (ie. did-exchange).

pkg/didcomm/transport/http/inbound.go
andrewwhitehead: Again need to test with the .NET framework, last I checked it only accepts this one.

pkg/vdr/peer/vdr.go
andrewwhitehead: If our peer DID docs are using did:sov then that's likely an error, although standardizing on did:peer also doesn't seem ideal until that spec is finalized.

pkg/doc/did/doc.go
andrewwhitehead: Again need to check with the .NET framework but I don't think it will accept verificationMethod.

troyronda: @andrewwhitehead they will need to fix this. https://www.w3.org/TR/did-core/#verification-methods

pkg/vdr/peer/vdr.go
troyronda: @Moopli @andrewwhitehead we can't treat did:sov as did:peer - if we are being sent did:sov, we must resolve it using the did:sov driver.

If this is happening, AcaPy must fix it.

troyronda: @Moopli please revert this change.

pkg/didcomm/protocol/didexchange/service.go
troyronda: Looks like .NET has HTTPS handling [code](https://github.com/hyperledger/aries-framework-dotnet/blob/074dd0c14ee1bf50613523c55c3c2031a9046a21/src/Hyperledger.Aries/Agents/MessageTypesHttps.cs) for the HTTPs prefixes. Can you please switch the default @swcurran @andrewwhitehead @tmarkovski - we should be using the HTTPS prefix in the test harness.

pkg/didcomm/common/service/destination.go
```diff
@@ -23,6 +26,8 @@ type Destination struct {
 
 const (
 	didCommServiceType = "did-communication"
+	// legacyDIDCommServiceType is the non-spec service type used by legacy didcomm agent systems.
+	legacyDIDCommServiceType = "IndyAgent"
```
troyronda, line 30: @swcurran @andrewwhitehead When is the `IndyAgent` service type going away?

pkg/didcomm/transport/http/inbound.go
troyronda: https://github.com/hyperledger/aries-cloudagent-python/issues/1044

pkg/didcomm/protocol/didexchange/service.go
swcurran: @andrewwhitehead -- ACA-Py should be using the new prefix's on all new protocols anyway -- such as DID Exchange. 

We need to do a quick pass through the various mobile agents to ensure that all still work, but I think we are good with switching.

We really need to figure out how to do these community updates better -- how we can communicate when we can go from step to step reliably.  Basically -- that means we need participation from those involved -- software creators that build the frameworks and those that deploy them.  It's the latter that is the tricky part.

pkg/vdr/peer/vdr.go
swcurran: @andrewwhitehead -- let us know the state of this.  If this convention is in the wild, how do we eliminate it without breaking the interop we have today and add the interop we need tomorrow.

pkg/doc/did/doc.go
swcurran: It's about a community upgrade. When the first cut was implemented "publicKeys" was the attribute name in the then DID Core spec.

pkg/vdr/peer/vdr.go
andrewwhitehead: @swcurran  I'd start by just generating our current peer DID Doc implementation with the did:peer prefix to see how that works with interop (assuming we don't currently use did:peer, but I can't see any references to it in the code base). Maybe @sklump could have a look at that this week?

Moopli: @troyronda @andrewwhitehead I raised an issue over on aca-py for this https://github.com/hyperledger/aries-cloudagent-python/issues/1048

pkg/doc/did/doc.go
```diff
@@ -603,7 +650,7 @@ func getVerification(doc *Doc, rawVerification interface{},
 		return nil, errors.New("rawVerification is not map[string]interface{}")
 	}
 
-	if context == contextV011 {
+	if legacy || context == contextV011 {
```
troyronda, line 653: this is a bit unfortunate @fqutishat @Moopli

pkg/doc/did/doc.go
```diff
@@ -360,12 +360,61 @@ func ParseDocument(data []byte) (*Doc, error) {
 	} else if raw == nil {
 		return nil, errors.New("document payload is not provided")
 	}
-	// validate did document
-	err = validate(data, raw.schemaLoader())
+
+	switch {
+	case strings.HasPrefix(raw.ID, "did:sov:"):
```
troyronda, line 365: This might not be 100% correct. This is more a case of AcaPy currently sending some document in DID exchange as `did:sov` ...
@fqutishat @Moopli

fqutishat: yes

troyronda: @Moopli It might be safer to detect this case in DID Exchange code rather than here.

pkg/doc/did/doc.go
troyronda: Yup. Created https://github.com/hyperledger/aries-cloudagent-python/issues/1050

pkg/didcomm/common/service/did_comm_msg.go
```diff
@@ -25,6 +25,9 @@ const (
 	jsonThreadID       = "thid"
 	jsonParentThreadID = "pthid"
 	jsonMetadata       = "_internal_metadata"
+
+	oldPIURI = "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/"
+	newPIURI = "https://didcomm.org/"
```
troyronda, line 30: Let's not call this one "new".

pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -74,6 +74,11 @@ type Request struct {
 	Label      string            `json:"label,omitempty"`
 	Connection *Connection       `json:"connection,omitempty"`
 	Thread     *decorator.Thread `json:"~thread,omitempty"`
+	// DID the did of the requester. Optional, may be present within `connection` instead.
```
troyronda, line 77: @llorllale @Moopli - are these properties supposed to be optional or is this for interop?

pkg/doc/did/doc.go
```diff
@@ -360,6 +360,14 @@ func ParseDocument(data []byte) (*Doc, error) {
 	} else if raw == nil {
 		return nil, errors.New("document payload is not provided")
 	}
+
+	// Interop: handle legacy did docs that incorrectly indicate they use the new format
+	if requiresLegacyHandling(raw) {
+		ctx, _ := json.Marshal(raw.Context)
+		println("legacy:", string(ctx))
```
troyronda, line 367: @Moopli please remove prints.

pkg/framework/aries/framework.go
```diff
@@ -377,24 +379,44 @@ func createVDR(frameworkOpts *Aries) error {
...
-	for _, v := range frameworkOpts.vdr {
-		opts = append(opts, vdr.WithVDR(v))
-	}
 
 	p, err := peer.New(ctx.StorageProvider())
 	if err != nil {
 		return fmt.Errorf("create new vdr peer failed: %w", err)
 	}
 
+	// Interop: use a did:sov VDR that will prefer to store sov DID docs locally, by wrapping the did:peer VDR.
```
troyronda, line 388: @Moopli - no, you cannot override did:sov at this level. Needs to be within did exchange only.

pkg/vdr/sov/vdr.go
```diff
@@ -0,0 +1,155 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package sov
```
troyronda, line 7: @Moopli - no - this is not did:sov. It is confusing to have this package named sov in vdr.

Additionally, it doesn't really help. You need to fix the document in did exchange.

pkg/framework/aries/framework.go
troyronda: Let's remove this change from this PR. @Moopli

pkg/vdr/sov/vdr.go
troyronda: Let's remove this change from this PR @Moopli

pkg/doc/did/doc.go
```diff
@@ -404,6 +410,22 @@ func ParseDocument(data []byte) (*Doc, error) {
 	return doc, nil
 }
 
+func requiresLegacyHandling(raw *rawDoc) bool {
+	if len(raw.PublicKey) == 0 { // docs in v1 format don't have a top-level publicKey array
```
troyronda, line 414: || len(raw.VerificationMethod) > 0

Moopli: || or &&? if we get a doc that has _no keys_, should we do legacy tweaks?

pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -69,11 +69,18 @@ type Invitation struct {
...
-	ID         string            `json:"@id,omitempty"`
-	Label      string            `json:"label,omitempty"`
-	Connection *Connection       `json:"connection,omitempty"`
-	Thread     *decorator.Thread `json:"~thread,omitempty"`
+	Type   string            `json:"@type,omitempty"`
+	ID     string            `json:"@id,omitempty"`
+	Label  string            `json:"label,omitempty"`
+	Thread *decorator.Thread `json:"~thread,omitempty"`
+	// DID the did of the requester.
+	// Mandatory in did-exchange, but optional for backwards-compatibility with rfc 0160 connection protocol.
```
troyronda, line 77: @Moopli @llorllale do we have more description of why backwards-compatibility with RFC 0160 is needed?

llorllale: RFC0160 is not in the AIP 2.0 draft so far: https://github.com/hyperledger/aries-rfcs/pull/579/files?short_path=a9296f0#diff-a9296f04f44f2daa050b1b0724458240a85f029aa7b77ed83b17a5479ef97f85

We should probably confirm, but I believe there is a desire to deprecate RFC0160 going forward.

cc @swcurran @andrewwhitehead

pkg/didcomm/protocol/didexchange/models.go
llorllale: https://github.com/hyperledger/aries-rfcs/pull/608

pkg/didcomm/protocol/didexchange/service.go
llorllale: https://github.com/hyperledger/aries-rfcs/pull/579#issuecomment-804982644

pkg/didcomm/protocol/didexchange/models.go
Moopli: @troyronda I have to look through aca-py to see if any elements of 0160 are still used there - this, in particular, is a place where afgo was behind.

pkg/vdr/peer/vdr.go
llorllale: Relevant: 

* https://github.com/hyperledger/aries-rfcs/issues/611
* https://github.com/hyperledger/aries-rfcs/issues/609

troyronda (approved):

PR #2657, opened 2021-03-19
chore: delete outofband.Request

llorllale:
Syncronizing impl with RFC0434: the outofband.Request message is deleted as well as all relevant methods (CreateRequest, AcceptRequest, etc) from all exported artifacts (Go, JS/WASM, REST, mobile).


llorllale:
Tagging @Moopli and @haardikk21

troyronda (approved):

PR #2658, opened 2021-03-21
fix: Ensure store implementations disallow invalid tags

DRK3:
- Also updated dependency commit tags.
- edv module will be updated in another commit since it will depend on this commit.


sandrask (approved):

component/storage/leveldb/leveldb.go
```diff
@@ -72,6 +74,12 @@ func (p *Provider) OpenStore(name string) (storage.Store, error) {
 // SetStoreConfig isn't needed for LevelDB. For consistency with other store implementations, it saves
 // the store config for later retrieval.
 func (p *Provider) SetStoreConfig(name string, config storage.StoreConfiguration) error {
+	for _, tagName := range config.TagNames {
+		if strings.Contains(tagName, ":") {
```
troyronda, line 78: I think we should improve this solution in a followup - doing things that aren't in specs will cause later interop issues.

troyronda (approved):

PR #2659, opened 2021-03-21
fix: fixes for outofband interop with acapy

haardikk21:
Push 1:
- accept old and new outofband message format
- did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/out-of-band/1.0/invitation is accepted as an old invitation message type

Push 2:
- Currently `afgo` used to only look for a `service` field in the invitation message, but `acapy` sends `services` instead of `service`. Added that to the OOB Invitation struct to support parsing `services` and choose a target accordingly.
- Solves #2675 

Signed-off-by: Haardik Haardik haardik.haardik@securekey.com

### DIFF
--- a/pkg/didcomm/protocol/outofband/service.go
+++ b/pkg/didcomm/protocol/outofband/service.go
@@ -29,8 +29,12 @@ const (
 	Name = "out-of-band"
 	// PIURI is the Out-of-Band protocol's protocol instance URI.
 	PIURI = "https://didcomm.org/out-of-band/1.0"
+	// oldPIURI is the old OOB protocol's protocol instance URI.
+	oldPIURI = "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/out-of-band/1.0"
 	// InvitationMsgType is the '@type' for the invitation message.
 	InvitationMsgType = PIURI + "/invitation"
+	// OldInvitationMsgType is the `@type` for the old invitation message.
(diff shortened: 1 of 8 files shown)


pkg/didcomm/protocol/outofband/service.go
```diff
@@ -31,6 +31,8 @@ const (
 	RequestMsgType = "https://didcomm.org/oob-request/1.0/request"
 	// InvitationMsgType is the '@type' for the invitation message.
 	InvitationMsgType = "https://didcomm.org/out-of-band/1.0/invitation"
+	// OldInvitationMsgType is the '@type' for the old invitation message
+	OldInvitationMsgType = "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/out-of-band/1.0/invitation"
```
troyronda, line 35: https://github.com/hyperledger/aries-framework-go/pull/2655#discussion_r598115168

haardikk21: Seems like this was replaced in the DIDComm service on a higher level but that didn't seem to pass on for the OOB protocol

troyronda: ok

Moopli (approved):

troyronda (dismissed):

troyronda (approved):

troyronda (approved):

troyronda (approved):

PR #2660, opened 2021-03-21
test: add bdd test for vdr controller

fqutishat:


cmd/aries-js-worker/src/agent-rest-client.js
```diff
@@ -57,7 +57,7 @@ const pkgs = {
             method: "POST"
         },
         CreateDID: {
-            path: "/vdr/create",
+            path: "/vdr/did/create",
```
troyronda, line 60: I wonder if there is a way to combine the create and save endpoints.

troyronda (approved):

PR #2669, opened 2021-03-24
feat: VC wallet - Issue

### LINKED ISSUE #2668: VC wallet : Issue
Implementation for https://w3c-ccg.github.io/universal-wallet-interop-spec/#issue

sudeshrshetty:
- Implementation for universal wallet Issue interface.
- Support for Ed25519Signature2018, JSONWebSignature2020,
BbsBlsSignature2020
- Issue credential can happen only on unlocked wallet.
- Signer to use user profile based KMS.
- Fallback from VDR to profile based DID store while resolving DID.
- Validation on proof options (controller, verification method etc)
- Closes #2668

### DIFF
--- a/pkg/client/vcwallet/client.go
+++ b/pkg/client/vcwallet/client.go
@@ -8,17 +8,64 @@ package vcwallet
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/piprate/json-gold/ld"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	jld "github.com/hyperledger/aries-framework-go/pkg/doc/jsonld"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/presexch"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/jsonld"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/signer"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/suite"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/signature/suite/bbsblssignature2020"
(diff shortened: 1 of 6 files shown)


fqutishat (approved):

rolsonquadras (approved):

pkg/client/vcwallet/client.go
```diff
@@ -8,17 +8,64 @@ package vcwallet
...
+const (
+	// Ed25519Signature2018 ed25519 signature suite.
+	Ed25519Signature2018 = "Ed25519Signature2018"
+	// JSONWebSignature2020 json web signature suite.
+	JSONWebSignature2020 = "JsonWebSignature2020"
+	// BbsBlsSignature2020 BBS signature suite.
+	BbsBlsSignature2020 = "BbsBlsSignature2020"
+)
+
+// miscellaneous constants.
```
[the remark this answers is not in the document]
sudeshrshetty: I agree, I made adding context function generic.

pkg/client/vcwallet/client.go
```diff
@@ -260,9 +307,26 @@ func (c *Client) Query(query *QueryParams) ([]json.RawMessage, error) {
 //		- A verifiable credential with or without proof
 //		- Proof options
 //
-func (c *Client) Issue(credential json.RawMessage, options *ProofOptions) (json.RawMessage, error) {
-	// TODO to be added #2433
-	return nil, fmt.Errorf("to be implemented")
+func (c *Client) Issue(authToken string, credential json.RawMessage,
```
troyronda, line 310: I'm curious about the authToken as the first argument pattern. Couldn't Open return a Wallet object containing the authToken?

(and these functions would be on the Wallet object).

@fqutishat @sudeshrshetty @rolsonquadras

sudeshrshetty: Will be a nice option if we are only going to use SDK binding.
Since wallet server (REST command controller) is going to be stateless it will have to get new client instance each time, and opening wallet (instantiating kms, which is expensive indeed) each time we create new client instance is not going to be feasible here.

PR #2681, opened 2021-03-25
feat: vc wallet prove interface

### LINKED ISSUE #2680: vc wallet: prove
https://w3c-ccg.github.io/universal-wallet-interop-spec/#prove

sudeshrshetty:
- added prove interface based on universal wallet prove interface
(https://w3c-ccg.github.io/universal-wallet-interop-spec/#prove).
- support for both stored and raw credential arguments.
- Closes #2680

### DIFF
--- a/pkg/client/vcwallet/client.go
+++ b/pkg/client/vcwallet/client.go
@@ -225,12 +225,17 @@ func (c *Client) Issue(credential json.RawMessage,
 // Prove produces a Verifiable Presentation.
 //
 //	Args:
-//		- List of verifiable credentials IDs.
-//		- Proof options
+//		- list of interfaces (string of credential IDs which can be resolvable to stored credentials in wallet or
+//		raw credential).
+//		- proof options
 //
-func (c *Client) Prove(credentialIDs []string, options *wallet.ProofOptions) (json.RawMessage, error) {
-	// TODO to be added #2433
-	return nil, fmt.Errorf("to be implemented")
+func (c *Client) Prove(opts *wallet.ProofOptions, creds ...wallet.CredentialToProve) (*verifiable.Presentation, error) {
+	auth, err := c.auth()
+	if err != nil {
+		return nil, err
+	}
+
+	return c.wallet.Prove(auth, opts, creds...)
 }
 
 // Verify takes Takes a Verifiable Credential or Verifiable Presentation as input,.
--- a/pkg/wallet/wallet.go
+++ b/pkg/wallet/wallet.go
@@ -154,6 +154,31 @@ func WithUnlockExpiry(tokenExpiry time.Duration) UnlockOptions {
 	}
 }
 
+// proveOpts contains options for proving credentials.
+type proveOpts struct {
+	// credentials already saved in wallet.
+	storedCredentials []string
+	// credential being supplied to wallet to prove.
+	rawCredentials []json.RawMessage
+}
+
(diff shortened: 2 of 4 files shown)


rolsonquadras (approved):

pkg/client/vcwallet/client.go
```diff
@@ -225,12 +225,17 @@ func (c *Client) Issue(credential json.RawMessage,
...
-//		- List of verifiable credentials IDs.
-//		- Proof options
+//		- list of interfaces (string of credential IDs which can be resolvable to stored credentials in wallet or
+//		raw credential).
+//		- proof options
 //
-func (c *Client) Prove(credentialIDs []string, options *wallet.ProofOptions) (json.RawMessage, error) {
-	// TODO to be added #2433
-	return nil, fmt.Errorf("to be implemented")
+func (c *Client) Prove(credentials []interface{}, options *wallet.ProofOptions) (*verifiable.Presentation, error) {
```
troyronda, line 232: I'd rather not use an interface[]. Looking at the code, you have two different inputs to Prove: credential IDs or credentials.

troyronda: Also - It seems you switched from Prove being on stored credentials to being on anything passed-in. What was the reason?

troyronda: If you really need different inputs into the same function, you could also look at how NewPresentation was done:
https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/verifiable/presentation.go#L179-L221

sudeshrshetty: @troyronda things would have been simpler ([]string) if it is just for stored VCs. During selective disclosure, derived VCs aren't saved and we can not generate presentation unless we save it. But we cannot save them since it might replace the original VC (due to same `id` property)

sudeshrshetty: fixed, added variadic args for passing different kind of credential arguments.

pkg/wallet/wallet.go
```diff
@@ -154,6 +154,31 @@ func WithUnlockExpiry(tokenExpiry time.Duration) UnlockOptions {
...
+
+// WithStoredCredentials option for providing stored credential IDs for wallet to prove.
+func WithStoredCredentials(ids ...string) CredentialToProve {
+	return func(opts *proveOpts) {
+		opts.storedCredentials = ids
+	}
+}
+
+// WithRawCredentials option for providing raw credential for wallet to prove.
+func WithRawCredentials(raw ...json.RawMessage) CredentialToProve {
```
troyronda, line 176: Why raw credential instead of a Credential object?

troyronda, line 176: We might want to refactor this name.

pkg/wallet/wallet.go
```diff
@@ -406,6 +457,41 @@ func (c *Wallet) Verify(raw json.RawMessage) (bool, error) {
...
+		credential, err := verifiable.ParseCredential(raw, verifiable.WithDisabledProofCheck())
+		if err != nil {
+			return nil, err
+		}
+
+		response = append(response, credential)
+	}
+
+	for _, raw := range opts.rawCredentials {
+		credential, err := verifiable.ParseCredential(raw, verifiable.WithDisabledProofCheck())
```
troyronda, line 484: need comments on why proof checks are disabled.

troyronda (approved):

PR #2684, opened 2021-03-26
feat: EDV storage support for random document IDs

### LINKED ISSUE #2376: Revisit how we're generating EDV document IDs in EDV provider
The spec says that [Structured Documents](https://identity.foundation/confidential-storage/#structureddocument) and [Encrypted Documents](https://identity.foundation/confidential-storage/#encrypteddocument) must be Base58-encoded 128-bit random values. That was preventing us from adding multi-threaded value encryption with batching, so it was changed in https://github.com/hyperledger/aries-framework-go/pull/2359/files#diff-e94d69f254503724ec4dc06e46143db60be77f4f79812be13c26408d6eca6150R170 to instead be based off of a hash computed by an encrypted hash function (HMAC). The end result is an ID that's deterministic (much more efficient for document lookups), Base58-encoded (required by the server) but not a randomly generated ID. EDV servers can't tell if an ID is randomly generated, so the IDs will still be accepted by EDV servers but it's technically a bit out of spec.

### LINKED ISSUE #2493: New EDV REST provider implementation - avoid generating EDV document ID twice in Put calls with Batch endpoint extension disabled
When the batch endpoint extension is disabled, a call to Put() results in the EDV document ID being generated twice - once for the initial Get() to determine if the current operation is a create or update, and once just before putting the actual document in the server. It should be possible to refactor the code to avoid this.

DRK3:
EDV storage now uses random document IDs by default. The more performant deterministic ID generation also remains as an option.

closes #2376
closes #2493


component/storage/edv/encryptedformatter.go
```diff
@@ -118,52 +168,72 @@ func (e *EncryptedFormatter) getStructuredDocFromEncryptedDoc(
 	return structuredDoc, nil
 }
 
-// TODO (#2376) Revisit how we're generating EDV document IDs, since it's technically not 100% in line with the spec.
-//  (Spec requires randomly generated IDs)
 // Generates an encrypted document ID based off of key.
-func (e *EncryptedFormatter) formatKey(key string) (string, error) {
-	keyHash, err := e.macCrypto.ComputeMAC([]byte(key))
+func (e *EncryptedFormatter) generateDocumentIDDeterministically(prefix, key string) (string, error) {
```
aholovko, line 172: consider update naming to "generateDeterministicDocumentID"

component/storage/edv/encryptedformatter.go
```diff
@@ -118,52 +168,72 @@ func (e *EncryptedFormatter) getStructuredDocFromEncryptedDoc(
...
+	keyHash, err := e.macCrypto.ComputeMAC([]byte(prefix + key))
 	if err != nil {
 		return "", fmt.Errorf(`failed to compute MAC based on key "%s": %w`, key, err)
 	}
 
 	return base58.Encode(keyHash[0:16]), nil
 }
 
-func (e *EncryptedFormatter) formatTags(tagPrefix string, tags []spi.Tag) ([]spi.Tag, error) {
+func generateDocumentIDRandomly() (string, error) {
```
aholovko, line 185: consider update naming to "generateRandomDocumentID"

component/storage/edv/encryptedformatter.go
DRK3: I agree, that shorter and better. Done!

component/storage/edv/encryptedformatter.go
DRK3: I agree, that's shorter and better. Done!

fqutishat (approved):

aholovko (approved):

PR #2689, opened 2021-03-26
feat: verify presentation

### LINKED ISSUE #2686: VC wallet : Verify interface
https://w3c-ccg.github.io/universal-wallet-interop-spec/#verify

sudeshrshetty:
- Closes #2686

### DIFF
--- a/pkg/wallet/wallet.go
+++ b/pkg/wallet/wallet.go
@@ -306,7 +306,7 @@ func (c *Wallet) Issue(authToken string, credential json.RawMessage,
 //		- proof options
 //
 func (c *Wallet) Prove(authToken string, proofOptions *ProofOptions, credentials ...CredentialToPresent) (*verifiable.Presentation, error) { //nolint: lll
-	resolved, err := c.resolveCredentials(false, credentials...)
+	resolved, err := c.resolveCredentials(credentials...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve credentials from request: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


fqutishat (approved):

rolsonquadras (approved):

pkg/wallet/wallet.go
```diff
@@ -394,7 +389,7 @@ func (c *Wallet) resolveCredentials(verify bool, credentials ...CredentialToPres
 	for _, raw := range opts.rawCredentials {
 		// proof check is disabled while proving raw credentials,
 		// proof check has to be performed explicitly by using 'wallet.Verify()'.
-		credential, err := verifiable.ParseCredential(raw, credentialOpts...)
+		credential, err := verifiable.ParseCredential(raw, verifiable.WithDisabledProofCheck())
```
troyronda, line 392: these disabled proof checks probably still need a bit more explanation.

pkg/wallet/wallet.go
```diff
@@ -409,6 +404,33 @@ func (c *Wallet) resolveCredentials(verify bool, credentials ...CredentialToPres
 	return response, nil
 }
 
+func (c *Wallet) verifyCredential(credential json.RawMessage) (bool, error) {
+	// TODO resolve stored DIDs which can not be resolved by context?
```
troyronda, line 408: what's does this mean?

pkg/wallet/wallet.go
```diff
@@ -409,6 +404,33 @@ func (c *Wallet) resolveCredentials(verify bool, credentials ...CredentialToPres
...
+
+	_, err := verifiable.ParseCredential(credential, opts)
+	if err != nil {
+		return false, fmt.Errorf("credential verification failed: %w", err)
+	}
+
+	return true, nil
+}
+
+// TODO should we check proof of each credentials inside presentation? under discussion in universal wallet 2020.
```
troyronda, line 421: Why wouldn't the proofs be checked?

troyronda (approved):

pkg/wallet/wallet.go
sudeshrshetty: Yes, the point is should we check proof of each VCs inside presentation or we just have to check the holder signature in presentation. I am raising a question in universal wallet community since it is not clearly explained in spec. I will fix this in next PR once I get the response.

pkg/wallet/wallet.go
sudeshrshetty: We need custom VDRI for each wallet profile which also goes through stored DID in wallet content(cached DID Document). Part of my next PR. (ex: interop scenarios in wallet).

pkg/wallet/wallet.go
troyronda: We need to check both the VCs and the holder signature.

pkg/wallet/wallet.go
sudeshrshetty: fixed

pkg/wallet/wallet.go
sudeshrshetty: fixed

pkg/wallet/wallet.go
sudeshrshetty: resolved

PR #2701, opened 2021-03-30
feat: outofband.Invitation.Accept and media types

llorllale:
* add: `accept` property to the `outofband.Invitation` message as as per RFC0434.
* refactor: `transport.Envelope.CTY` -> `transport.Envelope.MediaType`
* refactor: packer constants for media types were replaced by new constants in the `transport` package
* delete: `Noop` packer that is not used anywhere


Baha-sk (approved):
it seems there's a unit-test error message update is needed.

otherwise, lgtm.

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -234,3 +232,13 @@ func (o *OutboundDispatcher) addTransportRouteOptions(req []byte, des *service.D
 
 	return req, nil
 }
+
+// TODO - inject MediaType selection strategy.
```
Baha-sk, line 236: I guess the packager should be deducting the mediaType based on its list of packers, this would be done in Packager.PackMessage() above, in this case, we should just return the full list of MediaTypes to the packager, what do you think @llorllale ?

rolsonquadras (approved):

pkg/didcomm/dispatcher/outbound.go
llorllale: @Baha-sk i wrote that TODO on the assumption that the user would declare their preferred media type when responding (`accept` is a list).

Personally i would leave the packager to choose the packer and the user to declare the media type they want to use.

fqutishat (approved):

pkg/didcomm/dispatcher/outbound.go
Baha-sk: correct, this is why it would make sense to return the full list of MediaTypes here, it can be done in a followup pr

PR #2705, opened 2021-03-31
feat: #2701 save inbound didexchange.Request envelope media type into connection record

llorllale:
closes #2701 

Note on implementation: 

This change could have been less disruptive if we simply handled this as a special case and add a new method to the didexchange protocol service. There is already a special case [here](https://github.com/hyperledger/aries-framework-go/blob/a4113ebda7e3e6a1918415960b09a309d86b19d3/pkg/framework/context/context.go#L181) for performance reasons; adding a new method and calling it will also make it a special case from a functional perspective. This did not sound quite right and instead I chose to refactor `HandleInbound` to accept a general purpose holder of the DIDComm context such that in the future additional cross-cutting properties can be added as needed without affecting function signatures.

### DIFF
--- a/pkg/didcomm/common/service/messenger.go
+++ b/pkg/didcomm/common/service/messenger.go
@@ -6,6 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 
 package service
 
+// DIDCommContextEnvelopeMediaTypeKey is DIDCommContext property key holding the DIDComm envelope's media type.
+const DIDCommContextEnvelopeMediaTypeKey = "DIDCommContextEnvelopeMediaType"
+
 // DIDCommMsg describes message interface.
 type DIDCommMsg interface {
 	ID() string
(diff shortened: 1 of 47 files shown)


pkg/didcomm/common/service/messenger.go
```diff
@@ -6,6 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 
 package service
 
+// DIDCommContextEnvelopeMediaTypeKey is DIDCommContext property key holding the DIDComm envelope's media type.
+const DIDCommContextEnvelopeMediaTypeKey = "DIDCommContextEnvelopeMediaType"
```
Baha-sk, line 10: will MediaType represent the envelope `typ`only or a combination of envelope `typ` + payload `cty`?

pkg/didcomm/common/service/messenger.go
```diff
@@ -55,7 +97,7 @@ type Messenger interface {
 type MessengerHandler interface {
 	Messenger
 	// HandleInbound handles all inbound messages
-	HandleInbound(msg DIDCommMsgMap, myDID, theirDID string) error
+	HandleInbound(msg DIDCommMsgMap, ctx DIDCommContext) error
```
soluchok, line 100: It would be better just to use build-in context with values.
```
HandleInbound(ctx, msg)
```

Baha-sk (approved):

fqutishat (approved):

pkg/didcomm/common/service/messenger.go
llorllale: @Baha-sk as per the RFCs, the **envelope's** media type determines both the `typ` and the `cty` values.

PR #2711, opened 2021-04-02
feat: implement new document loader for JSON-LD contexts

### LINKED ISSUE #2648: Implement document loader for JSON-LD contexts
Requirements:
* contexts should be stored in the persistent storage
* support for adding/removing contexts
* option for network being enabled or disabled (default)

Related issues:
* Update jsonld contexts that are pre-loaded by the framework (#2489)
* VC allowed/trusted JSON-LD contexts (#1325)
* DID allowed/trusted JSON-LD contexts (#1326)
* LD Signatures allowed/trusted JSON-LD contexts (#1341)

aholovko:
New document loader is an implementation of [ld.DocumentLoader](https://github.com/piprate/json-gold/blob/0b11d17575f79f1a359b0dca0ed32dc8ca86ddc9/ld/document_loader.go#L48) interface backed by storage.

There is no explicit caching support - whether caching is used or not is determined by the underlying storage (ie. use `cachedstore` from the framework to get caching capabilities).

Network support (fetching context from the remote URL if it's not found in the underlying storage) can be enabled using `WithRemoteDocumentLoader()` option upon loader creation. The [one from json-gold](https://github.com/piprate/json-gold/blob/0b11d17575f79f1a359b0dca0ed32dc8ca86ddc9/ld/document_loader.go#L54) can be used here.

`AddDocument` from the [old loader](https://github.com/hyperledger/aries-framework-go/blob/4f5a32ec42c7587198126398085153df5c600b7e/pkg/doc/jsonld/document_loader.go#L64) is not going to be supported as it encourages to add new documents at the later stages (and the idea is to have a loader with all needed documents already pre-populated during startup).

Default contexts are embedded into Go binary and can be preloaded using `WithContexts(jsonld.DefaultContexts...)` during loader creation.

Closes #2648


pkg/doc/jsonld/context/preload/trustbloc_credentials_v1.jsonld
```diff
@@ -0,0 +1,56 @@
+{
```
troyronda, line 1: This context shouldn't exist - use https://w3id.org/security/jws/v1

pkg/doc/jsonld/context/importer.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	return &Importer{store: store}, nil
+}
+
+func importContexts(store storage.Store) error {
+	var ops []storage.Operation
+
+	for u := range contexts {
+		c, err := readContextDocument(contexts[u].path)
```
troyronda, line 68: this seems to imply that the binary would need to be distributed with these files in a relative path?

pkg/doc/jsonld/context/importer.go
```diff
@@ -0,0 +1,144 @@
...
+	for i := range opts {
+		opts[i](importerOpts)
+	}
+
+	store, err := storageProvider.OpenStore(importerOpts.contextDBName)
+	if err != nil {
+		return nil, fmt.Errorf("open store for context DB: %w", err)
+	}
+
+	if importerOpts.preloadContexts {
```
troyronda, line 55: Suggest adding some comments that these preloaded contexts are embedded in the compilation.

pkg/doc/jsonld/context/importer.go
```diff
@@ -0,0 +1,144 @@
...
+
+		b, err := json.Marshal(remoteDocument)
+		if err != nil {
+			return fmt.Errorf("marshal remote document: %w", err)
+		}
+
+		ops = append(ops, storage.Operation{Key: u, Value: b})
+	}
+
+	if err := store.Batch(ops); err != nil {
```
troyronda, line 86: I assume if a context is already loaded into the DB then it is overwritten?

Note: We will have new contexts to load overtime so I think importer will logically be run on a fresh storage AND on version upgrades. We will need to add missing or updated contexts on these upgrades.

pkg/doc/jsonld/context/importer.go
```diff
@@ -0,0 +1,144 @@
...
+	"fmt"
+
+	"github.com/piprate/json-gold/ld"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/spi/storage"
+)
+
+const (
+	DefaultContextDBName = "ContextsDB"
```
troyronda, line 22: Not sure about the "DB" suffix - if we are worried about uniqueness, I suppose we could include jsonld (or similar) in the name.

pkg/doc/jsonld/context/contexts.go
```diff
@@ -0,0 +1,35 @@
...
+		path:        "preload/security_v2.jsonld",
+	},
+
+	"https://w3id.org/security/jws/v1": {
+		documentURL: "https://w3c-ccg.github.io/lds-jws2020/contexts/v1/",
+		path:        "preload/lds-jws2020_v1.jsonld",
+	},
+
+	"https://w3id.org/security/bbs/v1": {
+		documentURL: "https://w3c-ccg.github.io/ldp-bbs2020/contexts/v1/",
```
troyronda, line 32: use the canonical URL (not the redirected one): https://w3id.org/security/bbs/v1

pkg/doc/jsonld/context/contexts.go
```diff
@@ -0,0 +1,35 @@
...
+		path:        "preload/security_v1.jsonld",
+	},
+
+	"https://w3id.org/security/v2": {
+		documentURL: "https://w3c-ccg.github.io/security-vocab/contexts/security-v2.jsonld",
+		path:        "preload/security_v2.jsonld",
+	},
+
+	"https://w3id.org/security/jws/v1": {
+		documentURL: "https://w3c-ccg.github.io/lds-jws2020/contexts/v1/",
```
troyronda, line 27: use the canonical URL (not the redirected one): https://w3id.org/security/jws/v1

pkg/doc/jsonld/context/contexts.go
```diff
@@ -0,0 +1,35 @@
...
+
+// contexts contains a list of JSON-LD context documents that should be preloaded into DB.
+var contexts = map[string]contextDocument{
+	"https://w3id.org/security/v1": {
+		documentURL: "https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld",
+		path:        "preload/security_v1.jsonld",
+	},
+
+	"https://w3id.org/security/v2": {
+		documentURL: "https://w3c-ccg.github.io/security-vocab/contexts/security-v2.jsonld",
```
troyronda, line 22: use the canonical URL (not the redirected one): https://w3id.org/security/v2

pkg/doc/jsonld/context/contexts.go
```diff
@@ -0,0 +1,35 @@
...
+
+type contextDocument struct {
+	documentURL string
+	path        string
+}
+
+// contexts contains a list of JSON-LD context documents that should be preloaded into DB.
+var contexts = map[string]contextDocument{
+	"https://w3id.org/security/v1": {
+		documentURL: "https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld",
```
troyronda, line 17: use the canonical URL (not the redirected one): https://w3id.org/security/v1

pkg/doc/jsonld/context/importer.go
DRK3: Note that `Put` or `Batch` will overwrite data.

Baha-sk (approved):

troyronda:
We should note that the loaded documents can change during run-time. @aholovko @sudeshrshetty @rolsonquadras @fqutishat 

what should happen in the framework usage:
- base common documents are pre-loaded (as I see in this PR).
- Additional contexts are available from storage. These additional contexts are intended to be available across restarts.
- API provided to add/remove additional contexts into storage. There should also be bindings for our controllers (e.g., REST/JS/...).
- When an API is used to add/remove contexts in storage, these should be applied without restarting the framework.

aholovko:
> We should note that the loaded documents can change during run-time. @aholovko @sudeshrshetty @rolsonquadras @fqutishat @Baha-sk @llorllale
> 
> what should happen in the framework usage:
> 
> * base common documents are pre-loaded (as I see in this PR).
> * Additional contexts are available from storage. These additional contexts are intended to be available across restarts.
> * API provided to add/remove additional contexts into storage. There should also be bindings for our controllers (e.g., REST/JS/...).
> * When an API is used to add/remove contexts in storage, these should be applied without restarting the framework.

the missing part is API and controller for adding/removing contexts when framework is running - I am working on that

pkg/doc/jsonld/default_contexts.go
```diff
@@ -0,0 +1,39 @@
...
+// embedFS contains JSON-LD context documents from the "contexts" folder.
+//go:embed contexts/*.jsonld
+var embedFS embed.FS //nolint:gochecknoglobals
+
+// DefaultContexts contains a list of context documents embedded into Go binary. They can be preloaded into
+// the underlying storage using jsonld.WithContexts() option upon loader creation.
+// TODO: Update with more default contexts.
+var DefaultContexts = []ContextDocument{ //nolint:gochecknoglobals
+	{
+		URL:         "https://w3id.org/security/v1",
```
troyronda, line 20: Note: this is the important URL - this is the one that shows up in the documents.

pkg/doc/jsonld/contexts/bbs2020.jsonld
```diff
@@ -0,0 +1,93 @@
+{
```
troyronda, line 1: TODO: might want to more clearly indicate that these are embed data in the directory structure.

troyronda (approved):

PR #2720, opened 2021-04-09
feat:add ECDSA vdr keys in DIDExchange

Baha-sk:


pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -559,6 +555,99 @@ func (ctx *context) getDIDDocAndConnection(pubDID string, routerConnections []st
...
+
+		if !ok {
+			return "", errors.New("keyType is not kms.KeyType")
+		}
+
+		return keyType, nil
+	}
+
+	return "", errors.New("keyType option is needed for empty didDoc.VerificationMethod")
+} */
```
Baha-sk, line 603: a follow up pr will be created to add key vdr tests for keys other than Ed25519 and add default keys creation in Aries by specifying the key type where this code will be uncommented/used.

fqutishat (approved):

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -22,6 +22,12 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/vdr/fingerprint"
 )
 
+/* const (
+	legacyMediaType			 = "JWM/1.0"
+	didCommV1MediaType       = "application/didcomm-enc-env"
+	didCommV2MediaType       = "application/didcomm-encrypted+json"
+) */
+
```
DRK3, line 30: Did you intend to keep this commented out code here?

DRK3 (approved):
One comment to double check

pkg/didcomm/dispatcher/outbound.go
Baha-sk: yes. this is not ready to be used for now, it will be uncommented when encryption keys and packer selection based on MediaType is done, in a future pr.

PR #2726, opened 2021-04-12
Add resolving DID Document in VDR Registry when generating presentation by id.

msjeong482:
**Title:**
Add resolving DID Document in VDR Registry.

**Summary:**
When generating a presentation by ID, DID Documents are searched only in local storage, so even though DID Documents are in the VDR Registry, the process of storing DID Documents in local storage should have been additionally performed.
So, I added the process of finding DID Document in VDR Registry.


pkg/controller/command/verifiable/command.go
```diff
@@ -711,16 +711,24 @@ func (o *Command) GeneratePresentationByID(rw io.Writer, req io.Reader) command.
 		return command.NewValidationError(GeneratePresentationByIDErrorCode, fmt.Errorf("get vc by id : %w", err))
 	}
 
-	doc, err := o.didStore.GetDID(request.DID)
+	var didDoc *did.Doc
```
sudeshrshetty, line 714: I think it should be other way around. If someone caches DID then local storage should be looked first.

msjeong482: Thank you for sharing your good opinion. If the DID is cached, it has been modified to look in the local storage first.

I have an additional question.
When I look into the parts of DID search in many protocols, I look into the VDR Registry first, and if not, look into local storage. (I also referred to that part.)
Does this need to be modified?

sudeshrshetty (approved):

fqutishat (approved):

pkg/controller/command/verifiable/command.go
sudeshrshetty: @msjeong482 looks good to me now. Please rebase so that we can merge this PR.

fqutishat:
@msjeong482 please squash your commits to one commit

Issue #2495, opened 2021-02-03
didexchange updates


llorllale:
Need to implement outstanding updates to the did-exchange protocol. Mayor points:

* signing attachments with detached JWS
* new `complete` message

troyronda:
Related: https://github.com/hyperledger/aries-framework-go/issues/2415

llorllale:
@Moopli is working on this

Moopli:
- [ ] Emit didexchange-compliant messages
  - [ ] DID doc should go in `did_doc~attach`, not `connection.DIDDoc`
  - [ ] DID should go in `did`, not `connection.DID`

llorllale:
@Moopli - I added those tasks at the top of this issue

PR #2733, opened 2021-04-14
feat: didexchange interop with aca-py, with afgo as responder

Moopli:
Part of #2495:
- DID exchange `completed` state
- Attachment signatures: Sign/Verify methods for attachment decorators
  that use base64 payloads.
- Additional fixes that are locked behind a build flag.

The build flag `interop` enables "breaking" interop fixes - fixes
  for interop that break compliance with specs/RFCs, and are
  temporary workarounds to allow for interop testing before
  interop partners fully update their implementations.

The fixes behind the flag are:
- Accept did:sov did doc in didexchange request, and store it like
  a peer did
- Generate a did:sov did doc as a peer did for didexchange response
- DID Doc method to serialize into a non-spec-compliant 'hybrid' format
  when attaching it to a didexchange message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,12 @@ agent-rest:
 	@mkdir -p ./build/bin
 	@cd ${ARIES_AGENT_REST_PATH} && go build -o ../../build/bin/aries-agent-rest main.go
 
+.PHONY: agent-rest-acapy-interop
+agent-rest-acapy-interop:
+	@echo "Building aries-agent-rest for aca-py interop"
+	@mkdir -p ./build/bin
+	@cd ${ARIES_AGENT_REST_PATH} && go build -o ../../build/bin/aries-agent-rest -tags ACAPyInterop main.go
+
 .PHONY: agent-mobile
 agent-mobile:
 	@echo "Building aries-agent-mobile"
--- a/pkg/client/messaging/client_test.go
+++ b/pkg/client/messaging/client_test.go
@@ -606,8 +606,9 @@ func TestCommand_Reply(t *testing.T) {
 
 		go func() {
 			for {
-				if len(registrar.Services()) > 0 {
-					_, e := registrar.Services()[0].HandleInbound(
+				svcs := registrar.Services()
+				if len(svcs) > 0 {
+					_, e := svcs[0].HandleInbound(
 						replyMsg, service.NewDIDCommContext("sampleDID", "sampleTheirDID", nil))
 					require.NoError(t, e)
 				}
--- a/pkg/didcomm/protocol/decorator/decorator.go
+++ b/pkg/didcomm/protocol/decorator/decorator.go
@@ -11,7 +11,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/vdr/fingerprint"
 )
 
 const (
@@ -92,6 +98,8 @@ type AttachmentData struct {
 	// JSON is a directly embedded JSON data, when representing content inline instead of via links,
 	// and when the content is natively conveyable as JSON. Optional.
 	JSON interface{} `json:"json,omitempty"`
+	// JWS is a JSON web signature over the encoded data, in detached format.
+	JWS json.RawMessage `json:"jws,omitempty"`
 }
 
 // Fetch this attachment's contents.
@@ -118,7 +126,176 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
 
 	// TODO add support to fetch links
 
-	// TODO add support for jws signatures
-
 	return nil, errors.New("no contents in this attachment")
+}
+
+type rawSig struct {
+	Header    rawSigHeader `json:"header,omitempty"`
+	Signature string       `json:"signature,omitempty"`
+	Protected string       `json:"protected,omitempty"`
+}
+
+type rawSigHeader struct {
+	KID string `json:"kid,omitempty"`
+}
+
+type rawProtected struct {
+	JWK json.RawMessage `json:"jwk,omitempty"`
+	Alg string          `json:"alg,omitempty"`
+}
+
+// Sign signs the base64 payload of the AttachmentData, and adds the signature to the attachment.
+func (d *AttachmentData) Sign(c crypto.Crypto, kh, pub interface{}, pubBytes []byte) error { // nolint:funlen,gocyclo
+	didKey, _ := fingerprint.CreateDIDKey(pubBytes)
+
+	jwk, err := jose.JWKFromKey(pub)
+	if err != nil {
+		return fmt.Errorf("creating jwk from pub key: %w", err)
+	}
+
+	jwk.KeyID = didKey
+
+	jwkBytes, err := jwk.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("marshaling jwk: %w", err)
+	}
+
+	protected := rawProtected{
+		JWK: jwkBytes,
+	}
+
+	kty, err := jwk.KeyType()
+	if err != nil {
+		return fmt.Errorf("getting keytype: %w", err)
+	}
+
+	switch kty {
+	case kms.ED25519Type:
+		protected.Alg = "EdDSA"
+	case kms.ECDSAP256TypeIEEEP1363:
+		protected.Alg = "ES256"
+	case kms.ECDSAP384TypeIEEEP1363:
+		protected.Alg = "ES384"
+	case kms.ECDSAP521TypeIEEEP1363:
+		protected.Alg = "ES512"
+	default:
+		return fmt.Errorf("unsupported KeyType for attachment signing")
+	}
+
+	protectedBytes, err := json.Marshal(protected)
+	if err != nil {
+		return fmt.Errorf("marshaling protected header: %w", err)
+	}
+
+	protectedB64 := base64.RawURLEncoding.EncodeToString(protectedBytes)
+
+	var b64data string
+
+	// interop: the specific behaviour here isn't fully specified by the attachment decorator RFC (as of yet)
+	// see issue https://github.com/hyperledger/aries-cloudagent-python/issues/1108
+	if doACAPyInterop {
+		b64data = b64ToRawURL(d.Base64)
+	} else {
+		b64data = d.Base64
+	}
+
+	signedData := fmt.Sprintf("%s.%s", protectedB64, b64data)
+
+	sig, err := c.Sign([]byte(signedData), kh)
+	if err != nil {
+		return fmt.Errorf("signing data: %w", err)
+	}
+
+	jws := rawSig{
+		Header: rawSigHeader{
+			KID: didKey,
+		},
+		Protected: protectedB64,
+		Signature: base64.RawURLEncoding.EncodeToString(sig),
+	}
+
+	serializedJWS, err := json.Marshal(jws)
+	if err != nil {
+		return fmt.Errorf("marshaling jws: %w", err)
+	}
+
+	d.JWS = serializedJWS
+
+	return nil
+}
+
+func b64ToRawURL(s string) string {
+	return strings.ReplaceAll(strings.ReplaceAll(strings.Trim(s, "="), "+", "-"), "/", "_")
+}
+
+// Verify verify the signature on the attachment data.
+func (d *AttachmentData) Verify(c crypto.Crypto, keyManager kms.KeyManager) error { // nolint:funlen,gocyclo
+	if d.JWS == nil {
+		return fmt.Errorf("no signature to verify")
+	}
+
+	jws := rawSig{}
+
+	err := json.Unmarshal(d.JWS, &jws)
+	if err != nil {
+		return fmt.Errorf("parsing jws: %w", err)
+	}
+
+	sigKey, err := fingerprint.PubKeyFromDIDKey(jws.Header.KID)
+	if err != nil {
+		return fmt.Errorf("parsing did:key '%s': %w", jws.Header.KID, err)
+	}
+
+	protectedBytes, err := base64.RawURLEncoding.DecodeString(jws.Protected)
+	if err != nil {
+		return fmt.Errorf("decoding protected header: %w", err)
+	}
+
+	var protected rawProtected
+
+	err = json.Unmarshal(protectedBytes, &protected)
+	if err != nil {
+		return fmt.Errorf("parsing protected header: %w", err)
+	}
+
+	jwk := jose.JWK{}
+
+	err = jwk.UnmarshalJSON(protected.JWK)
+	if err != nil {
+		return fmt.Errorf("parsing jwk: %w", err)
+	}
+
+	keyType, err := jwk.KeyType()
+	if err != nil {
+		return fmt.Errorf("getting KeyType for jwk: %w", err)
+	}
+
+	kh, err := keyManager.PubKeyBytesToHandle(sigKey, keyType)
+	if err != nil {
+		return fmt.Errorf("creating key handle: %w", err)
+	}
+
+	sig, err := base64.RawURLEncoding.DecodeString(jws.Signature)
+	if err != nil {
+		return fmt.Errorf("decoding signature: %w", err)
+	}
+
+	var b64data string
+
+	// interop: the specific behaviour here isn't fully specified by the attachment decorator RFC (as of yet)
+	// see issue https://github.com/hyperledger/aries-cloudagent-python/issues/1108
+	if doACAPyInterop {
+		b64data = b64ToRawURL(d.Base64)
+	} else {
+		b64data = d.Base64
+	}
+
+	signedData := fmt.Sprintf("%s.%s", jws.Protected, b64data)
+
+	err = c.Verify(sig, []byte(signedData), kh)
+	if err != nil {
+		return fmt.Errorf("signature verification: %w", err)
+	}
+
+	return nil
 }
--- a/pkg/didcomm/protocol/didexchange/didex_default.go
+++ b/pkg/didcomm/protocol/didexchange/didex_default.go
@@ -1,0 +1,24 @@
+// +build !ACAPyInterop
+
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didexchange
+
+import (
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+const (
+	doACAPyInterop = false
+)
+
+// TODO interop: this is a stub method, that is substituted for special functionality
+//   when the ACAPyInterop flag is enabled.
+//   This can be removed when https://github.com/hyperledger/aries-cloudagent-python/issues/1048 is fixed.
+func convertPeerToSov(doc *did.Doc) (*did.Doc, error) {
+	return doc, nil
+}
--- a/pkg/didcomm/protocol/didexchange/didex_interop.go
+++ b/pkg/didcomm/protocol/didexchange/didex_interop.go
@@ -1,0 +1,54 @@
+// +build ACAPyInterop
+
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didexchange
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/btcsuite/btcutil/base58"
+
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+)
+
+const (
+	doACAPyInterop = true
+)
+
+// Interop: convert a peer did doc to a "sov-like" did doc, to accommodate current behaviour in aca-py,
+//          where sovrin dids are used as peer dids.
+// TODO interop: aca-py issue https://github.com/hyperledger/aries-cloudagent-python/issues/1048
+func convertPeerToSov(doc *did.Doc) (*did.Doc, error) {
+	if doc == nil {
+		return doc, nil
+	}
+
+	didParts := strings.Split(doc.ID, ":")
+	if len(didParts) != 3 {
+		return nil, fmt.Errorf("peer did not in 3 parts")
+	}
+
+	id := base58.Encode(base58.Decode(didParts[2])[:16])
+
+	newDID := fmt.Sprintf("did:sov:%s", id)
+
+	docBytes, err := doc.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+
+	docBytes = bytes.Replace(docBytes, []byte(doc.ID), []byte(newDID), -1)
+	err = doc.UnmarshalJSON(docBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return doc, nil
+}
--- a/pkg/didcomm/protocol/didexchange/models.go
+++ b/pkg/didcomm/protocol/didexchange/models.go
@@ -93,6 +93,12 @@ type Response struct {
 	ID                  string               `json:"@id,omitempty"`
 	ConnectionSignature *ConnectionSignature `json:"connection~sig,omitempty"`
 	Thread              *decorator.Thread    `json:"~thread,omitempty"`
+	// DID the did of the responder.
+	// Mandatory in did-exchange, but optional for backwards-compatibility with rfc 0160 connection protocol.
+	DID string `json:"did,omitempty"`
+	// DocAttach an attachment containing the did doc of the responder.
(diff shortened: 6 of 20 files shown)


pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +127,116 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+type rawProtected struct {
+	JWK json.RawMessage `json:"jwk,omitempty"`
+	Alg string          `json:"alg,omitempty"`
+}
+
+// Sign signs the base64 payload of the AttachmentData, and adds the signature to the attachment.
+func (d *AttachmentData) Sign(c crypto.Crypto, kh interface{}, pub []byte) error {
+	didKey, _ := fingerprint.CreateDIDKey(pub)
+
+	jwk, err := jose.JWKFromKey(ed25519.PublicKey(pub))
```
llorllale, line 152: Please don't hard code key types.

cc @Baha-sk

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -275,6 +275,28 @@ func (s *completed) ExecuteInbound(msg *stateMachineMsg, thid string, ctx *conte
...
+		action := func() error { return nil }
+
+		if msg.connRecord == nil {
+			return nil, &noOp{}, action, nil
+		}
+
+		connRec := *msg.connRecord
+
+		// Interop: aca-py did:sov DIDs are often sent without a did:sov: DID prefix, add the prefix here to handle it
+		if !strings.HasPrefix(connRec.TheirDID, "did:") {
```
llorllale, line 295: Shouldn't this be guarded by the new `interop` flag?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -441,6 +453,89 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
+		Thread: &decorator.Thread{
+			ID: request.ID,
+		},
+	}
+
+	if request.Thread != nil {
+		response.Thread.PID = request.Thread.PID
+	}
+
+	if !ctx.doInterop {
```
llorllale, line 471: I suggest renaming the flag to `doACAPyInterop` to make the intention clearer.

pkg/didcomm/protocol/decorator/decorator.go
Moopli: Right now, aca-py hardcodes to ed25519, but you're right, key type choice is made by the caller and `Sign()` should respect it.

pkg/doc/jose/jws.go
```diff
@@ -138,6 +138,40 @@ func (s JSONWebSignature) SerializeCompact(detached bool) (string, error) {
 		b64Signature), nil
 }
 
+/*
```
llorllale, line 141: looks like a leftover

pkg/didcomm/protocol/decorator/decorator.go
Baha-sk: if this is only for interop, then it's not a big deal as long as it will be removed when V1 interop is not needed, otherwise, this function has to be able to support signing with P-256 and P-384 keys as well

llorllale: `AttachmentData` is a standard DIDComm message structure not exclusive to interop.

troyronda: Agree with @llorllale - please don't hardcode key type.

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +129,158 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+	jwk := jose.JWK{}
+
+	err = jwk.UnmarshalJSON(protected.JWK)
+	if err != nil {
+		return fmt.Errorf("parsing jwk: %w", err)
+	}
+
+	var keyType kms.KeyType
+
+	switch jwk.Key.(type) {
```
Moopli, line 245: @Baha-sk is this a good way to select the kms key type from the jwk? Is there a way that you've already created to do that?

Baha-sk: this whole code is duplicated, why not use afgo's doc/jose package instead?

Moopli: I'm already using doc/jose, and it doesn't seem to support JSON JWS, only compact, so I'm using it just for the JWKs. There isn't any mechanism in doc/jose.JWK for determining the `kms.KeyType`, and I didn't find anything in the kms either.

Baha-sk: so figuring out the kms key type from a JWK doesn't exist directly.

Currently doc/jose/jwk.go has internal functions to check the key type like [this existing code](
https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/jose/jwk.go#L216-L260)

you can probably add a member function to JWK that returns the kms keyType based on the embedded key using these functions

pkg/doc/jose/jwk.go
```diff
@@ -213,6 +215,22 @@ func (j *JWK) MarshalJSON() ([]byte, error) {
 	return (&j.JSONWebKey).MarshalJSON()
 }
 
+// KeyType returns the kms KeyType of the JWK, or an error if the JWK is of an unrecognized type.
+func (j *JWK) KeyType() (kms.KeyType, error) {
```
Moopli, line 219: @Baha-sk here's the new method, taking the logic out of `decorator.AttachmentData.Verify()`

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -445,6 +458,89 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
+	if e != nil {
+		return nil, fmt.Errorf("getting sender verkey: %w", e)
+	}
+
+	pubKeyBytes, e := fingerprint.PubKeyFromDIDKey(invitationKey)
+	if e != nil {
+		return nil, fmt.Errorf("failed to extract pubKeyBytes from did:key [%s]: %w", invitationKey, e)
+	}
+
+	signingKID, e := localkms.CreateKID(pubKeyBytes, kms.ED25519Type)
```
Baha-sk, line 522: as discussed.. `kms.ED25519Type` will be changed to `ctx.KeyType` after I push my PR to make this key dynamic

Baha-sk: you can probably add a TODO here for this purpose

pkg/doc/jose/jwk.go
Baha-sk: it can be improved to check if jwk.Key is ECDSA keys .. it can be IEEE1363Type keys for ecdsa keys

Baha-sk (approved):

pkg/didcomm/protocol/decorator/decorator.go
Moopli: This is done

pkg/didcomm/protocol/didexchange/states.go
Moopli: Done

pkg/doc/jose/jws.go
Moopli: removed

Makefile
```diff
@@ -124,6 +124,12 @@ agent-rest:
 	@mkdir -p ./build/bin
 	@cd ${ARIES_AGENT_REST_PATH} && go build -o ../../build/bin/aries-agent-rest main.go
 
+.PHONY: agent-rest-interop
```
troyronda, line 127: we should make these targets specific to acapy-interop.

Moopli: @troyronda could you clarify what you mean?
rename the target, or remove the target and have the flagged build only built in interop tests?

troyronda: similar to @llorllale comment earlier, interop is too generic.

pkg/didcomm/protocol/didexchange/didex_default.go
```diff
@@ -0,0 +1,14 @@
+// +build !interop
```
troyronda, line 1: build tag should be more specifically named.

Moopli: Done

pkg/didcomm/protocol/didexchange/didex_interop.go
```diff
@@ -0,0 +1,15 @@
...
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didexchange
+
+const (
+	// Interop: create "sov" docs when creating peer DIDs
```
troyronda, line 12: @Moopli Have you verified that AcaPy hasn't already fixed this?

pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -68,6 +68,9 @@ type Invitation struct {
 	Thread *decorator.Thread `json:"~thread,omitempty"`
 }
 
+// todo: emit didex request/response messages with did_doc~attach and did, not connection
```
troyronda, line 71: todo

pkg/client/messaging/client_test.go
```diff
@@ -606,8 +606,9 @@ func TestCommand_Reply(t *testing.T) {
 
 		go func() {
 			for {
-				if len(registrar.Services()) > 0 {
-					_, e := registrar.Services()[0].HandleInbound(
+				svcs := registrar.Services()
```
Moopli, line 609: Note: this is a fix for an intermittent unit test failure, where the `if` block was entered, but the services list was somehow empty within (I presume it's a concurrency issue).

pkg/didcomm/protocol/didexchange/models.go
```diff
@@ -100,10 +112,20 @@ type ConnectionSignature struct {
 	Signature  string `json:"signature,omitempty"`
 	SignedData string `json:"sig_data,omitempty"`
 	SignVerKey string `json:"signers,omitempty"`
+	// InteropKey Interop: signer key compatible with aca-py.
+	InteropKey string `json:"signer,omitempty"`
```
troyronda, line 116: What's InteropKey?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -275,6 +276,28 @@ func (s *completed) ExecuteInbound(msg *stateMachineMsg, thid string, ctx *conte
...
+
+		action := func() error { return nil }
+
+		if msg.connRecord == nil {
+			return nil, &noOp{}, action, nil
+		}
+
+		connRec := *msg.connRecord
+
+		// Interop: aca-py did:sov DIDs are often sent without a did:sov: DID prefix, add the prefix here to handle it
```
troyronda, line 295: @Moopli Have you verified that AcaPy hasn't already fixed this? Is there an open issue?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -445,6 +456,90 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
+	if err != nil {
+		return nil, fmt.Errorf("connection signature: %w", err)
+	}
+
+	response.ConnectionSignature = encodedConnectionSignature
+
+	return response, nil
+}
+
+func (ctx *context) prepareResponseInterop(request *Request, response *Response,
```
troyronda, line 489: Missing comment on what is the AcaPy interop difference and link to open issue in AcaPy.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -445,6 +456,90 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
+	if e != nil {
+		return nil, fmt.Errorf("failed to get key handle: %w", e)
+	}
+
+	e = docAttach.Data.Sign(ctx.crypto, kh, ed25519.PublicKey(pubKeyBytes), pubKeyBytes)
+	if e != nil {
+		return nil, fmt.Errorf("signing did_doc~attach: %w", e)
+	}
+
+	// Interop: aca-py expects naked DID method-specific identifier for sov DIDs
```
troyronda, line 536: please link to open issues in acapy.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -669,8 +764,14 @@ func (ctx *context) resolveDidDocFromConnection(conn *Connection) (*did.Doc, err
 		return nil, fmt.Errorf("resolveDidDocFromConnection: failed to parse DID [%s]: %w", didDoc.ID, err)
 	}
 
+	// Interop: sov -> peer
```
troyronda, line 767: please link to open issues in acapy.

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -720,6 +821,18 @@ func (ctx *context) prepareConnectionSignature(connection *Connection,
 		return nil, fmt.Errorf("sign response message: %w", err)
 	}
 
+	if ctx.doACAPyInterop {
+		// Interop: aca-py expects raw base58 Ed25519 key instead of did:key
```
troyronda, line 825: please link to open issues in acapy.

pkg/doc/did/serialize_interop.go
```diff
@@ -0,0 +1,215 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/btcsuite/btcutil/base58"
+)
+
+/*
+This file contains interop fixes for DID doc serialization, that break compliance with the DID spec.
```
troyronda, line 21: please link to open issues in acapy.

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +126,140 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+	jwk.KeyID = didKey
+
+	jwkBytes, err := jwk.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("marshaling jwk: %w", err)
+	}
+
+	protected := rawProtected{
+		JWK: jwkBytes,
+		Alg: "EdDSA",
```
llorllale, line 166: shouldn't hard code keytypes/algos

pkg/framework/aries/build_interop.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	var opts []vdr.Option
+
+	p, err := peer.New(ctx.StorageProvider())
+	if err != nil {
+		return fmt.Errorf("create new vdr peer failed: %w", err)
+	}
+
+	// Interop: use a did:sov VDR that wraps the did:peer VDR and httpbinding VDRs,
```
troyronda, line 46: Please check that this hasn't already been fixed. I'd rather not have this.

pkg/vdr/sov/vdr.go
```diff
@@ -0,0 +1,188 @@
+// +build ACAPyInterop
```
troyronda, line 1: this package is not sov - this would be confusing.

pkg/didcomm/protocol/didexchange/models.go
Moopli: ACA-py _was_ expecting the connection signature's `signer` field to contain the base58-encoded sender verification key.

Just ran a test to verify if it's still necessary (commented this out and ran the interop test) and it looks like this issue is no longer present, so I'll remove the code.

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +126,140 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+	}
+
+	protectedBytes, err := json.Marshal(protected)
+	if err != nil {
+		return fmt.Errorf("marshaling protected header: %w", err)
+	}
+
+	protectedB64 := base64.RawURLEncoding.EncodeToString(protectedBytes)
+
+	b64data := strings.Trim(d.Base64, "=")
```
Baha-sk, line 176: Manipulating the encoding manually is not recommended.

Prefer to update the code to make the value of `d.Base64` the result of `RawURLEncoding` instead of `StdURLEncoding` and remove the `Trim()` call here.

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +126,140 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+	if err != nil {
+		return fmt.Errorf("creating key handle: %w", err)
+	}
+
+	sig, err := base64.RawURLEncoding.DecodeString(jws.Signature)
+	if err != nil {
+		return fmt.Errorf("decoding signature: %w", err)
+	}
+
+	b64data := strings.Trim(d.Base64, "=")
```
Baha-sk, line 255: same as above comment

pkg/didcomm/protocol/decorator/decorator.go
Moopli: It looks like that would break other protocols that currently use attachments, since they interop using StdEncoding, and [aries-rfcs/#558](https://github.com/hyperledger/aries-rfcs/pull/558) is not yet merged - moreover, that PR states that we should accept _either_ URLEncoding or RawURLEncoding for the data - therefore for signing we'd need to keep this trim, since ACA-Py trims the padding from the attachment base64 data, for a consistent signed payload even if the encoding is different across different uses.

Actually I just noticed, they don't just convert to Raw, they also convert to URL encoding - that is, no matter which base64 encoding the data uses (since it is, as of yet, allowed to vary based on use), the signature will be over the RawURL encoding of the data. I'll amend this code to make sure it does the same.

pkg/vdr/sov/vdr.go
Moopli: Addressed this by removing the "fake sov" interop vdr and refactoring the conversion logic into didexchange itself.

pkg/framework/aries/build_interop.go
Moopli: did some refactoring and got rid of it :+1:

pkg/didcomm/protocol/decorator/decorator.go
Moopli: done, now it switches on the keytype

pkg/doc/did/serialize_interop.go
Moopli: done

pkg/didcomm/protocol/didexchange/states.go
Moopli: done

pkg/didcomm/protocol/didexchange/states.go
Moopli: removed this path, as refactoring had made it dead code (and it was unnecessary)

pkg/didcomm/protocol/didexchange/states.go
Moopli: done

pkg/didcomm/protocol/didexchange/models.go
Moopli: removed this

pkg/didcomm/protocol/didexchange/states.go
Moopli: removed this specific code as I'm not sure this bit will actually be necessary, but yes, the sov stuff is still an issue that needs to be handled elsewhere.

troyronda: @Moopli I had thought that AcaPy might have introduced their own build flag to not emit sov. Have you double checked?

Moopli: Well, they don't have a startup flag for it, and I can't find anything in their didexchange manager or did doc implementation that would suggest the existence of a toggle, so I suspect not. I'll do some more digging if you want me to really really confirm though ;)

Moopli: @troyronda looks like we still need to accommodate sov "peer" for now

pkg/didcomm/protocol/didexchange/models.go
Moopli: removed

pkg/didcomm/protocol/didexchange/didex_interop.go
Moopli: verified, they don't

pkg/didcomm/protocol/didexchange/states.go
Moopli: renamed method to be more clear, and added an explanatory comment where it's invoked - this isn't an interop fix, it's an AIP2.0 fix, and making this behaviour default (instead of limited to interop testing with aca-py) will be my next PR.

pkg/didcomm/protocol/decorator/decorator.go
Baha-sk: then the special trimming logic should be executed for the interop case only, default case should be just RawURLEncoded without trimming

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -391,63 +409,152 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
-	connRec.MyDID = connection.DID
-	connRec.TheirLabel = request.Label
+	if request.Thread != nil {
+		response.Thread.PID = request.Thread.PID
+	}
 
-	destination, err := service.CreateDestination(requestDidDoc)
+	// TODO followup: bring this out from under doACAPYInterop, and instead under an AIP 2.0 flag
+	//  part of issue https://github.com/hyperledger/aries-framework-go/issues/2495
+	if ctx.doACAPyInterop {
```
troyronda, line 484: what's the acapy issue?

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -391,63 +409,152 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
+			return nil, fmt.Errorf("verifying signature on doc~attach: %w", err)
+		}
 	}
 
-	senderVerKey, err := recipientKey(responseDidDoc)
+	var docBytes []byte
+
+	var err error
+
+	if ctx.doACAPyInterop {
```
troyronda, line 512: what's the acapy issue?

troyronda: https://github.com/hyperledger/aries-framework-go/pull/2733/files#diff-23f6a42334371adf3244d1e0517e14388a004b149c9dc05401b3c56abaea9368R156-R160

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -391,63 +409,152 @@ func (ctx *context) handleInboundRequest(request *Request, options *options,
...
 
-	connRec.TheirDID = reqConn.DID
-	connRec.MyDID = connection.DID
-	connRec.TheirLabel = request.Label
+	if request.Thread != nil {
+		response.Thread.PID = request.Thread.PID
+	}
 
-	destination, err := service.CreateDestination(requestDidDoc)
+	// TODO followup: bring this out from under doACAPYInterop, and instead under an AIP 2.0 flag
```
troyronda, line 482: we won't have a flag for AIP 2 - we only care about AIP2+.

pkg/client/messaging/client_test.go
troyronda: This doesn't really make sense to me.

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +126,171 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+	protectedBytes, err := json.Marshal(protected)
+	if err != nil {
+		return fmt.Errorf("marshaling protected header: %w", err)
+	}
+
+	protectedB64 := base64.RawURLEncoding.EncodeToString(protectedBytes)
+
+	var b64data string
+
+	if doACAPyInterop {
```
troyronda, line 194: aca-py issue?

pkg/didcomm/protocol/decorator/decorator.go
```diff
@@ -118,7 +126,171 @@ func (d *AttachmentData) Fetch() ([]byte, error) {
...
+	}
+
+	sig, err := base64.RawURLEncoding.DecodeString(jws.Signature)
+	if err != nil {
+		return fmt.Errorf("decoding signature: %w", err)
+	}
+
+	var b64data string
+
+	if doACAPyInterop {
```
troyronda, line 283: aca-py issue?

Baha-sk (approved):

pkg/didcomm/protocol/didexchange/states.go
Moopli: fixed comment

pkg/didcomm/protocol/decorator/decorator.go
Moopli: added to comment right above this line

pkg/didcomm/protocol/decorator/decorator.go
Moopli: added to comment right above this line

llorllale (approved):

PR #2740, opened 2021-04-16
fix: media types not fetched from connection when initiating protocol

llorllale:
This PR makes the outbound dispatcher fetch the receiving agent's supported media types from the existing connection record.

Other changes:

* Renamed "MediaTypes" -> "MediaTypeProfiles"
* the framework was falling back to the envelope media type in the inbound didexchange `request` message if the did doc embedded in the message did not contain an `accept` property. This behaviour has been removed. The media type profile is solely determined by either a) the OOB invitation (has precedence), or b) the `service` block's `accept` property

### DIFF
--- a/pkg/didcomm/transport/media_type.go
+++ b/pkg/didcomm/transport/media_type.go
@@ -6,9 +6,10 @@ SPDX-License-Identifier: Apache-2.0
 
 package transport
 
-import "fmt"
-
 const (
+	// MediaTypeRFC0019EncryptedEnvelope is the original media type for DIDComm V1 encrypted envelopes as per
+	// Aries RFC 0019.
+	MediaTypeRFC0019EncryptedEnvelope = "JWM/1.0"
 	// MediaTypeV1EncryptedEnvelope is the media type for DIDComm V1 encrypted envelopes as per Aries RFC 0044.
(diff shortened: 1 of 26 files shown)


Baha-sk (approved):

troyronda (approved):

pkg/didcomm/transport/media_type.go
```diff
@@ -6,9 +6,10 @@ SPDX-License-Identifier: Apache-2.0
 
 package transport
 
-import "fmt"
-
 const (
+	// MediaTypeRFC0019EncryptedEnvelope is the original media type for DIDComm V1 encrypted envelopes as per
+	// Aries RFC 0019.
+	MediaTypeRFC0019EncryptedEnvelope = "JWM/1.0"
```
Baha-sk, line 12: I believe we also need to add media [profiles](https://github.com/hyperledger/aries-rfcs/tree/master/features/0044-didcomm-file-and-mime-types#defined-profiles) as well

llorllale: I've added them in the `outofband` package.

Baha-sk: I noticed @llorllale updated the pr specifically for media profiles. It looks good now.

Baha-sk (approved):

Issue #2682, opened 2021-03-25
add KeyAgreementKey in DID doc to store JWE encryption keys


Baha-sk:
And add dynamic authentication and KeyAgreement kms.KeyType in the framework and DIDExchange client.

Baha-sk:
this is supposed to be closed by #2743

PR #2743, opened 2021-04-20
feat: add KeyAgreement to DIDDoc and variable key type

### LINKED ISSUE #1993: Generate DID docs with Encryption key as KeyAgreement

Baha-sk:
This change adds the following features:
    1. A new encryption key is added into the DID doc in DIDcomm, stored as a KeyAgreement verification.
    2. Includes variable key types for authentication (signing) and encryption keys in the framework and the DIDExchange service.
    3. Introduces a new JWK helper function (PubKeyBytesToJWK) to create a JWK from raw key bytes and a kms key type.
    4. Finally it updates the DID doc context to the official V1 schema url "https://www.w3.org/ns/did/v1".

closes #1993, #2682

### DIFF
--- a/pkg/crypto/eckey.go
+++ b/pkg/crypto/eckey.go
@@ -1,0 +1,41 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"fmt"
+	"math/big"
+)
+
+// ToECKey converts key to an ecdsa public key. It returns an error if the curve is invalid.
+func ToECKey(key *PublicKey) (*ecdsa.PublicKey, error) {
+	crv, err := toCurve(key.Curve)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ecdsa.PublicKey{
+		Curve: crv,
+		X:     new(big.Int).SetBytes(key.X),
+		Y:     new(big.Int).SetBytes(key.Y),
+	}, nil
+}
+
+func toCurve(crv string) (elliptic.Curve, error) {
+	switch crv {
+	case "P-256", "NIST_P256":
+		return elliptic.P256(), nil
+	case "P-384", "NIST_P384":
+		return elliptic.P384(), nil
+	case "P-521", "NIST_P521":
+		return elliptic.P521(), nil
+	}
+
+	return nil, fmt.Errorf("invalid curve '%s'", crv)
+}
--- a/pkg/didcomm/protocol/didexchange/keys.go
+++ b/pkg/didcomm/protocol/didexchange/keys.go
@@ -1,0 +1,155 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package didexchange
+
+import (
+	"encoding/json"
+	"fmt"
+
+	gojose "github.com/square/go-jose/v3"
+
+	"github.com/hyperledger/aries-framework-go/pkg/crypto"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/jose"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+)
+
+func createNewKeyAndVM(didDoc *did.Doc, keyType, keyAgreementType kms.KeyType, keyManager kms.KeyManager) error {
+	vm, err := createSigningVM(keyManager, getVerMethodType(keyType), keyType)
+	if err != nil {
+		return err
+	}
+
+	kaVM, err := createEncryptionVM(keyManager, getVerMethodType(keyAgreementType), keyAgreementType)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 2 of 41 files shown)


aholovko (approved):

pkg/crypto/eckey.go
```diff
@@ -0,0 +1,41 @@
...
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"fmt"
+	"math/big"
+)
+
+// ToECKey converts key to an ecdsa public key. It panics if the curve is invalid.
```
DRK3, line 16: Does this function really panic if the curve is invalid? It looks to me like it would only panic if `key` is `nil`.

pkg/didcomm/protocol/didexchange/keys.go
```diff
@@ -0,0 +1,143 @@
...
+	didDoc.KeyAgreement = append(didDoc.KeyAgreement, *did.NewReferencedVerification(kaVM, did.KeyAgreement))
+
+	return nil
+}
+
+func createSigningVM(km kms.KeyManager, vmType string, keyType kms.KeyType) (*did.VerificationMethod, error) {
+	kid, pubKeyBytes, err := km.CreateAndExportPubKeyBytes(keyType)
+	if err != nil {
+		return nil, err
+	}
```
DRK3, line 44: Might be a good idea to wrap these errors

pkg/didcomm/protocol/didexchange/keys_test.go
```diff
@@ -0,0 +1,87 @@
...
+		require.Equal(t, x25519KeyAgreementKey2019, didDoc.KeyAgreement[0].VerificationMethod.Type)
+	})
+
+	t.Run("createNewKeyAndVM invalid keyType export signing key", func(t *testing.T) {
+		didDoc := &did.Doc{}
+
+		err := createNewKeyAndVM(didDoc, kms.HMACSHA256Tag256Type, kms.X25519ECDHKWType, k)
+		require.EqualError(t, err, "createAndExportPubKeyBytes: failed to export new public key bytes: "+
+			"exportPubKeyBytes: failed to export marshalled key: exportPubKeyBytes: failed to get public keyset "+
+			"handle: keyset.Handle: keyset.Handle: keyset contains a non-private key")
```
DRK3, line 37: Looks like you have some repetition (`keyset.Handle: keyset.Handle:`) in your full error message. If you rework the error messages at each level where they get wrapped you should be able to avoid this

pkg/doc/did/doc_test.go
```diff
@@ -445,10 +445,8 @@ func TestPopulateKeyAgreements(t *testing.T) {
 		require.NoError(t, err)
 
 		_, err = ParseDocument(bytes)
-		require.Error(t, err)
-
-		expected := fmt.Sprintf("key %s not exist in did doc verification method", missingPubKeyID)
-		require.Contains(t, err.Error(), expected)
+		require.EqualError(t, err, fmt.Sprintf("populate key agreements failed: key %s not exist in did doc"+
```
DRK3, line 448: Looks like this isn't something you changed in this PR, but figured it's worth mentioning anyway: should be `does not exist` instead of `not exist`

fqutishat (approved):

DRK3 (approved):
Made minor comments

pkg/crypto/eckey.go
troyronda: yeh we can't be panic ing!

pkg/didcomm/protocol/didexchange/keys_test.go
Baha-sk: anything after `exportPubKeyBytes: failed to get public keyset handle: `  is inside Tink unfortunately

pkg/doc/did/doc_test.go
Baha-sk: will update :)

pkg/crypto/eckey.go
Baha-sk: it's an outdated comment, it has been updated.

pkg/didcomm/protocol/didexchange/keys.go
Baha-sk: done

pkg/doc/did/doc_test.go
Baha-sk: updated

Moopli (approved):

PR #2785, opened 2021-05-09
add node.js modules to .gitignore

iRaymond0724:


iRaymond0724:
Hello, I am following this document 
( https://github.com/hyperledger/aries-framework-go/blob/main/docs/test/build.md#Prerequisites-(for-running-tests-and-demos) ) 
Build Aries, node.js will produce something that can be added and ignored

.gitignore
```diff
@@ -42,3 +42,7 @@ profile.out
 
 # mobile related files
 cmd/aries-agent-mobile/build/
+
+# node.js modules
+node_modules/
+package-lock.json
```
troyronda, line 48: I think the package-lock.json should not be ignored for folders that become npm modules.
Particularly https://github.com/hyperledger/aries-framework-go/tree/main/cmd/aries-js-worker

I've been somewhat on the fence about test directories containing package-lock.json.

.gitignore
```diff
@@ -42,3 +42,7 @@ profile.out
 
 # mobile related files
 cmd/aries-agent-mobile/build/
+
+# node.js modules
+node_modules/
```
troyronda, line 47: I think the directories containing package.json already have a .gitignore with node_modules. Although it doesn't hurt to also make it a global ignore.

.gitignore
iRaymond0724: Thank you for your reply and remind. ^^
I did not notice that , thank you.

.gitignore
iRaymond0724: Yes , sir !
Because after I compiled, node_modules appeared in the root directory of the project, so I wanted to say whether it is possible to add node_modules to the global .gitignore . Obviously my thinking is too simple, thank you very much for your reply ! ^^

PR #2798, opened 2021-05-18
test: RFC0593: Go SDK BDD tests

llorllale:

### DIFF
--- a/pkg/didcomm/protocol/issuecredential/states.go
+++ b/pkg/didcomm/protocol/issuecredential/states.go
@@ -331,12 +331,19 @@ func (s *offerReceived) ExecuteInbound(md *MetaData) (state, stateAction, error)
 		return nil, nil, fmt.Errorf("decode: %w", err)
 	}
 
+	response := &RequestCredential{
+		Type:           RequestCredentialMsgType,
+		RequestsAttach: offer.OffersAttach,
+	}
+
+	if md.RequestCredential() != nil {
+		response = md.RequestCredential()
+		response.Type = RequestCredentialMsgType
+	}
+
 	// creates the state's action
 	action := func(messenger service.Messenger) error {
-		return messenger.ReplyToMsg(md.Msg, service.NewDIDCommMsgMap(RequestCredential{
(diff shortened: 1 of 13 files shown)


pkg/didcomm/protocol/issuecredential/states.go
```diff
@@ -331,12 +331,19 @@ func (s *offerReceived) ExecuteInbound(md *MetaData) (state, stateAction, error)
 		return nil, nil, fmt.Errorf("decode: %w", err)
 	}
 
+	response := &RequestCredential{
+		Type:           RequestCredentialMsgType,
+		RequestsAttach: offer.OffersAttach,
+	}
+
+	if md.RequestCredential() != nil {
```
llorllale, line 339: quick workaround - @soluchok should probably take a look at this

pkg/framework/aries/default.go
```diff
@@ -226,7 +226,10 @@ func assignVerifiableStoreIfNeeded(aries *Aries, storeProvider storage.Provider)
 		return nil
 	}
 
-	provider, err := context.New(context.WithStorageProvider(storeProvider))
+	provider, err := context.New(
+		context.WithStorageProvider(storeProvider),
+		context.WithJSONLDDocumentLoader(aries.jsonldDocumentLoader),
```
llorllale, line 231: uncovered this bug

pkg/didcomm/protocol/issuecredential/states.go
soluchok: @llorllale it seems to be fine.
This logic was added by @pfeairheller. We missed this part. Would be nice to cover it with tests (TODO is fine).
https://github.com/hyperledger/aries-framework-go/commit/d4d3721fdf9ca694188e3f15b00204e4fffdbb1e#diff-86393b3003ab06a03e9ec28f2f603b63688888c366300c4dcf28b557d39fe9b1

fqutishat (approved):

Baha-sk (approved):

PR #2811, opened 2021-05-31
feat: add CBC+HMAC Tink Key

### LINKED ISSUE #2810: Support AES_CBC+HMAC Tink Key
In order to prepare for [ECDH-1PU draft4](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04). This new Tink key and the ECDH primitive using the new related AEAD primitive is required.

Baha-sk:
This change adds a new Tink AEAD key using CBC+HMAC algorithm as Tink AEAD primitive.
It also updates the ECDH Tink primitive using the newly created primitive.

closes #2810


rolsonquadras (approved):

troyronda (approved):

pkg/crypto/tinkcrypto/primitive/aead/subtle/aes_cbc.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if len(plaintext) == 0 {
+		return plaintext, nil
+	}
+
+	// unpad plaintext if not block size.
+	last := plaintext[len(plaintext)-1]
+	count := int(last)
+
+	if count == 0 || count > blockSize || count > len(plaintext) {
```
thaidn, line 99: This looks wrong to me, because it'd reject any 16-byte plaintext that ends with a byte that is 0 or 17 or higher.

You don't want to make padding optional. Even if the plaintext length is already a multiple of the block size, add a dummy block consisting of 16 bytes of value 15.

This also helps make sure the ciphertext of the empty plaintext is indistinguishable.

pkg/crypto/tinkcrypto/primitive/aead/subtle/aes_cbc.go
```diff
@@ -0,0 +1,154 @@
...
+// Pad text to blockSize.
+func Pad(text []byte, originalTextSize, blockSize int) []byte {
+	// pad to block size if needed.
+	missing := blockSize - (originalTextSize % blockSize)
+	if missing > 0 && originalTextSize > 0 {
+		text = append(text, bytes.Repeat([]byte{byte(missing)}, missing)...)
+	}
+
+	return text
+}
```
thaidn, line 154: Don't make padding optional because otherwise the ciphertext of the empty plaintext would be distinguishable. This violates semantic security.

Add a block of padding when the plaintext size is a multiple of the blocksize.

thaidn: Hmm I'm wrong because IIRC semantic security ignores message length. FWIW, CTF also leaks when the empty plaintext is encrypted.

I still think you should add a padding block. Page 17 of https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf also recommends that.

pkg/crypto/tinkcrypto/primitive/aead/aead_factory.go
```diff
@@ -0,0 +1,117 @@
+/*
```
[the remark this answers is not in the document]
Baha-sk: makes sense, I'll remove it and use Tink's aead_factory.go instead.

pkg/crypto/tinkcrypto/primitive/aead/aes_cbc_hmac_aead_key_manager.go
```diff
@@ -0,0 +1,212 @@
...
+	key := new(aeadpb.AesCbcHmacAeadKey)
+	if err := proto.Unmarshal(serializedKey, key); err != nil {
+		return nil, errInvalidAESCBCHMACAEADKey
+	}
+
+	if err := km.validateKey(key); err != nil {
+		return nil, fmt.Errorf("aes_cbc_hmac_aead_key_manager: %w", err)
+	}
+
+	ctr, err := subtle.NewAESCBC(key.AesCbcKey.KeyValue)
```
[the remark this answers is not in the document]
Baha-sk: yes, I've corrected this in a subsequent [pr](https://github.com/hyperledger/aries-framework-go/pull/2814/files#diff-51e77f5c903352f7a32add583ae650ca19f9f9eee8fc936995ccad034e66a8cfR78)

pkg/crypto/tinkcrypto/primitive/aead/subtle/aes_cbc.go
```diff
@@ -0,0 +1,154 @@
...
+
+// newCipher creates a new AES-CBC cipher using the given key, IV and the crypto library.
+func newCipher(key, iv []byte, decrypt bool) (cipher.BlockMode, error) {
+	block, err := aes.NewCipher(key)
+	if err != nil {
+		return nil, fmt.Errorf("aes_cbc: failed to create block cipher, error: %w", err)
+	}
+
+	// If the IV is less than BlockSize bytes we need to pad it with zeros otherwise NewCBCEncrypter will panic.
+	if len(iv) < aes.BlockSize {
```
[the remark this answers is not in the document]
Baha-sk: makes sense since  we [generate](https://github.com/hyperledger/aries-framework-go/pull/2811/files/9d05bdf574fdad1d59b54e5343e266b5f4ed7ae6#diff-108bd5df34a666a2a61133481aa53ee51e671debf7037f79de5a048f4e74bcb4R47) a new iv and pass it to newCipher(), iv will never be < BlockSize

I'll remove this padding logic and return an error instead

pkg/crypto/tinkcrypto/primitive/aead/subtle/aes_cbc.go
Baha-sk: I see, thank you for this comment, I'll force add an extra padding block as per Page 17 of this NIST doc, quoted below:

> One way to ensure that the receiver does not mistakenly remove bits from an unpadded message is to require the sender to pad every message, including messages in which the final block (segment) is already complete. For such messages, an entire block (segment) of padding is appended.

pkg/crypto/tinkcrypto/primitive/aead/subtle/aes_cbc.go
Baha-sk: thank you @thaidn, it makes sense, but why this last block must contain bytes of value 15? 

What if the plaintext needs padding, shouldn't this extra block contain bytes of the padding value which is the padding size, eg if plaintext is missing 6 bytes to the block, then pad 6 bytes of value 6, and add a new block of 16 bytes of value 6, not 15? Or are you saying padding should always be hard coded with value 15 for both the dummy block and the padding bytes (for plaintext length not multiple of the block size)?

thaidn: 15 is a typo, it should be 16.

You want to implement: https://en.wikipedia.org/wiki/Padding_(cryptography)#PKCS#5_and_PKCS#7.

The value of each added byte is the number of bytes that are added, i.e. N bytes, each of value N are added. Let's say the plaintext size is 5. You would need to add 9 bytes, each of value 9. If the plaintext size is 16, you'd need to add a whole block of padding, meaning 16 bytes, each of value 16.

Baha-sk: So adding a full dummy block is only added when plaintext's size is a multiple of block size. This doesn't match the NIST examples though, since plaintext size is a multiple of block size in these examples. See CBC test vectors on page 27 of the [NIST document 800-3a](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf), plaintext are not padded with a dummy buffer.

what to do in this case? 

In the meantime I will push a pr that adds a dummy block for plaintext messages with size being a multiple of block size.

PR #2826, opened 2021-06-03
chore: Update EDV storage provider dependencies, remove replace

DRK3:
- Updated EDV storage provider dependencies, allowing additional unit tests to be run now, increasing code coverage
- Removed replace statement that was accidentally left in


DRK3:
The other go modules had to be updated to deal with the "implicitly required module" errors

troyronda (approved):

PR #2828, opened 2021-06-04
feat: support published sov DIDs as returned by uniresolver

Moopli:

### DIFF
--- a/pkg/didcomm/protocol/didexchange/states_test.go
+++ b/pkg/didcomm/protocol/didexchange/states_test.go
@@ -1740,6 +1740,84 @@ func TestGetDIDDocAndConnection(t *testing.T) {
 	})
 }
 
+const sovDoc = `{
+  "@context": "https://www.w3.org/2019/did/v1",
+  "id": "did:sov:17hRTxZFuRqqwFPxXnnuLj",
+  "service": [
+    {
+      "type": "endpoint",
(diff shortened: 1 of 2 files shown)


fqutishat (approved):

pkg/didcomm/protocol/didexchange/states_test.go
```diff
@@ -1776,6 +1854,26 @@ func TestGetVerKey(t *testing.T) {
...
+			ResolveValue: publicDID,
+		}
+		ctx.doACAPyInterop = true
+
+		err = ctx.connectionRecorder.SaveInvitation(invitation.ThreadID, invitation)
+		require.NoError(t, err)
+
+		result, err := ctx.getVerKey(invitation.ThreadID)
+		require.NoError(t, err)
+		require.Equal(t, publicDID.Service[0].RecipientKeys[0], result)
```
Baha-sk, line 1872: This Equal() call means both expected and actual results are the same object reference. Usually it's better to use EqualValues if you're looking for content equality.

Baha-sk (approved):
lgtm

pkg/didcomm/protocol/didexchange/states_test.go
Moopli: Those are strings, `Equal()` is fine here :)

fqutishat (approved):

PR #2841, opened 2021-06-22
feat: finalize 1PU draft4 implementation

### LINKED ISSUE #2816: ECDH-1PU draft4 upgrade
Upgrade Authcrypt packer with the new [draft4 implementation](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04) of ECDH-1PU.

This means:
- [x]  Use the new AES-CBC-HMAC as the content encryption AEAD cipher #2832.
- Move up the content encryption **before** the recipient's key wrapping.
- Extract the auth Tag from the above encryption and use it in key wrapping as described in the new version of the draft.
- Update Compact and Flattened JWE format since recipients headers are merged in the protected headers.
- Ensure JWE compatibility with the new example in draft4's [Appendix-B](https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#appendix-B).

Baha-sk:
This change concludes 1PU Draft4 implementation by:
- using AES-CBC+HMAC-SHA content encryption
- make content encryption execute before key wrapping
- pass encrytpion tag to key wrapping
- make epk, apu, apv and alg recipient headers common in ProtectHeaders
- ensures JWE/KW compatibility with ecdh-1pu draft4 example B test vector for both content encryption and full JWE decryption

closes #2816


pkg/doc/jose/encrypter.go
```diff
@@ -603,3 +888,24 @@ func computeAuthData(protectedHeaders map[string]interface{}, aad []byte) ([]byt
 
 	return output, nil
 }
+
+func jwkMarshalEPK(protectedHeadersJSON map[string]json.RawMessage) error {
```
Baha-sk, line 892: This function is crucial in building the EPK in the ProtectedHeaders when decryption (the same way as encryption)

fqutishat (approved):

PR #2842, opened 2021-06-22
feat: JSON-LD Context API

aholovko:
This PR mainly adds support for remote JSON-LD context providers (represented by endpoints on remote servers where JSON-LD contexts are hosted).

One example with static server setup can be found in BDD tests (`file-server.example.com` in `test/bdd/fixtures/agent-rest/docker-compose.yml`).

Remote provider(s) can be added to the agent instance in the following ways:

* statically with a custom document loader and `WithRemoteProvider`
* via configuration option: `ARIESD_CONTEXT_PROVIDER_URL` for rest agent, `context-provider-url` for agent js
* dynamically using REST API (controller API, client, js)

Data layer for contexts and providers was encapsulated in 2 new repositories:
* ContextStore (`pkg/store/ld/context_store.go`)
* RemoteProviderStore (`pkg/store/ld/remote_provider_store.go`)

The other big chunks are changes in package organizations (document loader now lives in `ld` package, context document definition in a separate `ldcontext` package, etc.).


fqutishat (approved):

cmd/aries-js-worker/src/agent-rest-client.js
```diff
@@ -457,9 +457,31 @@ const pkgs = {
             method: "POST",
         },
     },
-    context: {
-        Add: {
-            path: "/context/add",
+    ld: {
+        AddContexts: {
+            path: "/ld/context/add",
```
rolsonquadras, line 462: imo we don't need add suffix - usually HTTP POST verb implicitly denotes that.

cmd/aries-js-worker/src/agent-rest-client.js
```diff
@@ -457,9 +457,31 @@ const pkgs = {
...
+            method: "POST",
+        },
+        AddRemoteProvider: {
+            path: "/ld/remote-provider/add",
```
rolsonquadras, line 466: same as above comment (HTTP POST vs add suffix)

cmd/aries-js-worker/src/agent-rest-client.js
```diff
@@ -457,9 +457,31 @@ const pkgs = {
...
+            method: "POST"
+        },
+        RefreshRemoteProvider: {
+            path: "/ld/remote-provider/{id}/refresh",
+            method: "POST",
+            pathParam: "id"
+        },
+        RemoveRemoteProvider: {
+            path: "/ld/remote-provider/{id}/remove",
+            method: "POST",
```
rolsonquadras, line 476: We can remove `/remove` suffix and change the HTTP method to DELETE.

rolsonquadras (commented):
some minor comments

pkg/client/ld/client.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Client represents a REST client for JSON-LD API.
+type Client struct {
+	endpoint   string
+	httpClient HTTPClient
+}
+
+// NewClient returns a new JSON-LD REST client.
+func NewClient(endpoint string, opts ...Option) *Client {
```
sudeshrshetty, line 40: just curious: which `endpoint` is this? remote provider?

aholovko: this is agent controller endpoint, the one for remote provider is passed to `AddRemoteProvider` method as `remoteProviderEndpoint` argument

sudeshrshetty: but `pkg/client` is typically SDK clients not the client's for agent's rest controller.

It should be other way around where command controllers use clients for operations.

cmd/aries-js-worker/src/agent-rest-client.js
rolsonquadras: Not directly related to this PR, but we need to remove /add suffix [here](https://github.com/hyperledger/aries-framework-go/blob/eec182deab9a345ad2c66dc58d169233edbcdaaa/pkg/controller/rest/jsonld/context/operation.go#L22) as well - we can create a new issue/PR.

pkg/client/ld/client.go
aholovko: Thanks Sudesh for catching that 👍. I see that this is not in line with what other clients do here, so, for consistency, it would be better to remove that functionality. If SDK client for JSON-LD stuff is something that we really need, we can implement it in a separate task.

sudeshrshetty: Yes, because some AFG clients don't need controllers, they can work with pure SDK for all the features.

You already have APIs just call them directly here instead of going through controller.

aholovko: The problem here is that I don't have dedicated service for JSON-LD functionality (e.g. as in case of `wallet`). Command handlers act here as a "domain layer". Probably I will need to create a separate package on a root level (`pkg\ld`), put core logic there and then use it from command handlers and SDK client. I don't think that calling command handlers from the SDK client would be a good idea. What do you think on that? cc @sudeshrshetty

sudeshrshetty: better moving core logic to separate package, we did something similar for `pkg/wallet`,  and we have `vcwallet` package for SDK, command controller built on top of `pkg/wallet`

sudeshrshetty (approved):

rolsonquadras (approved):

PR #2847, opened 2021-06-24
did:key updates for JsonWebKey2020

davidastark:
All the changes applies to the did:key method.

Updated the serialization of the EC keys (P-256, P-384, P-521) to use the compressed format as specified in the multicodec specs.
When creating a did:key with JsonWebKey2020 the caller must pass in a json web key.
When reading a did:key with JsonWebKey2020 then return a did doc with a json web key
When reading a did:key enforce that the did method is "key"


fqutishat (approved):

pkg/vdr/key/creator.go
```diff
@@ -111,13 +126,21 @@ func getKeyCode(keyType kms.KeyType, verificationMethod *did.VerificationMethod)
...
-		64:  fingerprint.P256PubKeyMultiCodec,
-		96:  fingerprint.P384PubKeyMultiCodec,
-		132: fingerprint.P521PubKeyMultiCodec,
+	// ec keys must pass in a jwk
+	if method.JSONWebKey() == nil {
+		return 0, fmt.Errorf("jsonWebKey is required for EC key")
 	}
 
-	return ecdsaCodesByKeyLen[len(method.Value)], nil
+	switch method.JSONWebKey().Crv {
```
Baha-sk, line 134: The point of the map is to avoid switch/case.. if you want to use the curve type instead of key size, then replace the map keys to be curves instead of key size. This way you can avoid doing switch/case and use the map directly.

pkg/vdr/key/resolver.go
```diff
@@ -63,6 +70,50 @@ func createBase58DIDDoc(kid, keyType string, pubKeyBytes []byte) (*did.Doc, erro
...
 }
 
+func createJSONWebKey2020DIDDoc(kid string, code uint64, pubKeyBytes []byte) (*did.Doc, error) {
+	didKey := fmt.Sprintf("did:key:%s", kid)
+
+	keyID := fmt.Sprintf("%s#%s", didKey, kid)
+
+	var curve elliptic.Curve
+
+	switch code {
```
Baha-sk, line 80: same comment as above, using a map is faster than switch/case

Baha-sk (approved):
LGTM, I've added a couple of minor comments.

pkg/vdr/key/creator.go
davidastark: I'm not sure what the preference is of avoiding a switch statement, but if I revert to using a map then an additional if statement will be required in case the curve is not recognized at all. Note that this was an issue with the prior implementation as well in that when the key length did not match any of the values the function was just returning 0 and no error.

If you still prefer a map and an if statement then I can change it to that.

Baha-sk: yes, with a map, we'd still want to check if the value is found, otherwise return an error, it was probably missed before.

you can keep switch/case then for now, no need for additional changes.

PR #2851, opened 2021-06-26
feat: enable ecdh-1pu draft4 JWE decryption test

### LINKED ISSUE #2850: enable 1pu Appendix B decryption test
Require fixing the example to be compatible with ecdh-1pu draft4

Baha-sk:
This change enables the commented out test for full compatibility with ECDH-1PU draft4 Appendix B example.

It also includes the following changes:
* fixes CBC+HMAC keys extracted from the ECDH cek. The keys were inversed.
* moved recipients' kids list from AAD to APV (recipients) protectedHeader
* use originalProtectedHeaders if populated instead of protectedHeader when building authData. To avoid marshal/unmarshal calls from changing the json order.
* ECDH-1PU messages with multi-recipients can now read the sender kid (skid) from APU if skid wass empty since it's an optional header.

closes #2850

### DIFF
--- a/pkg/doc/jose/encrypter_decrypter_test.go
+++ b/pkg/doc/jose/encrypter_decrypter_test.go
@@ -46,370 +46,429 @@ const (
 	DIDCommContentEncodingType = "application/didcomm-plain+json"
 )
 
+//nolint:gocognit
 func TestJWEEncryptRoundTrip(t *testing.T) {
 	_, err := ariesjose.NewJWEEncrypt("", "", "", "", nil, nil, nil)
 	require.EqualError(t, err, "empty recipientsPubKeys list",
 		"NewJWEEncrypt should fail with empty recipientPubKeys")
 
+	singleRecipientNISTPKWError := "jwedecrypt: failed to unwrap cek: [unwrapKey: deriveKEKAndUnwrap:" +
+		" failed to AES unwrap key: square/go-jose: key wrap input must be 8 byte blocks]"
+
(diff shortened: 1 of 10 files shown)


pkg/doc/jose/encrypter_decrypter_test.go
```diff
@@ -556,17 +557,31 @@ func TestJWEEncryptRoundTrip(t *testing.T) {
...
 						},
-					},
+					})
 				}
 
 				_, err = jweDecrypter.Decrypt(badJWE)
-				require.EqualError(t, err, "jwedecrypt: failed to build recipients WK: unable to read "+
-					"JWK: invalid character 's' looking for beginning of value")
+				if tc.nbRec == 1 {
+					errStr := "jwedecrypt: failed to unwrap cek: [unwrapKey: deriveKEKAndUnwrap:" +
```
Moopli, line 573: couldn't errStr be a field in the testcase, instead of being constructed through some logic based on the testcase parameters?

Moopli (approved):

pkg/doc/jose/encrypter_decrypter_test.go
Baha-sk: although this is local error for single recipient messages, it's a sub-test, but I guess I can add test case error string for this specific case..

Baha-sk: updated

troyronda (approved):

PR #2856, opened 2021-07-06
feat: support packer selection based on destination MediaType

### LINKED ISSUE #272: Encrypted interface - add key type argument

### LINKED ISSUE #749: Support different key schemes for each recipient

### LINKED ISSUE #1112: Configurable packing algorithm(auth/anon crypt) for Forward(router) message
Forward(router) message packing algorithm(auth/anon crypt) should be configurable.

https://github.com/hyperledger/aries-framework-go/blob/ee769a304613c3dc3fb44edad780d674921d567a/pkg/didcomm/dispatcher/outbound.go#L170-L172

### LINKED ISSUE #2858: DIDComm Outbound - remove mandatory connection requirement
The SendToDID function in outbound dispatcher expects a connection to exists between sender and receiver DID (passed as part of function argument). This was added as part of [media type support PR](https://github.com/hyperledger/aries-framework-go/pull/2740/files#diff-ff1356a25eb14f75939a1f9bab5ceee8e783a09fa6ea8500276693d04e8ad5c9R75-R84). We need to provide support to send DIDComm messages without connections (as it was before the PR).

https://github.com/hyperledger/aries-framework-go/blob/e07a5c4905f6371201e94be1ec27121fd950e134/pkg/didcomm/dispatcher/outbound.go#L74-L83

Baha-sk:
This change sets the media type priority based on the destination according to Aries RFCs 19, 44 and 587.
It sets the JWE message's CTY header based on the received media type.
It also updates the default authcrypt packing sender key type to be AES256CBCHS512.

closes #1112 (packer is selected based on mediaTypeProfile)
closes #272 (key type is managed by the kms)
closes #749 (diffrent keys are supported by the packers, except LegacyPacker that supports ED25519 keys only)
closes #2858 (SendToDID now can be used without a DID connection, as it was before)

### DIFF
--- a/pkg/didcomm/dispatcher/outbound.go
+++ b/pkg/didcomm/dispatcher/outbound.go
@@ -8,11 +8,13 @@ package dispatcher
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/google/uuid"
 
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/model"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
@@ -33,6 +35,7 @@ type provider interface {
 	KMS() kms.KeyManager
 	ProtocolStateStorageProvider() storage.Provider
 	StorageProvider() storage.Provider
+	MediaTypeProfiles() []string
 }
 
 type connectionLookup interface {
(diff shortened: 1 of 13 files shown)


rolsonquadras (approved):

pkg/didcomm/dispatcher/outbound.go
```diff
@@ -266,11 +319,25 @@ func (o *OutboundDispatcher) addTransportRouteOptions(req []byte, des *service.D
...
+			case transport.MediaTypeV1EncryptedEnvelope, transport.MediaTypeV2EncryptedEnvelopeV1PlaintextPayload:
+				v1MT = mtp
+			case transport.MediaTypeV2EncryptedEnvelope, transport.MediaTypeV2PlaintextPayload:
+				// V2 is highest priority, if found use it directly.
+				return mtp
+			}
+		}
 	}
 
 	return mt
```
Moopli, line 343: `mt` is never set to anything, do you mean to return `v1MT` instead?

Baha-sk: indeed, `v1MT` is unnecessary, I've updated to the code to use `mt` instead.

Moopli (approved):

troyronda (approved):

Issue #2815, opened 2021-06-02
Don't import edv package directly in wallet package


DRK3:
Currently the `wallet` package imports the `edv` storage provider, which is in another module. This creates an undesirable dependency on the `edv` module from the main aries go module. We should rework the `wallet` package to not be hard-coded to `edv`, and instead allow the encrypted storage solution to be injected in. This may require creating a new interface in `spi`.

sudeshrshetty:
wallet should have a internal default EDV implementation where wallet user can just pass an EDV URL & keys and expect wallet to use that EDV for storing contents.

DRK3:
Another solution: move the wallet package to a new module

PR #2863, opened 2021-07-08
chore: Update storage module versions

DRK3:


DRK3:
Note that once #2815 is done, this module doesn't need to be imported from the main Aries module anymore.

fqutishat (approved):

sudeshrshetty (approved):

PR #2864, opened 2021-07-08
fix: resolve local did key references when handling request via implicit invitation

Moopli:

### DIFF
--- a/pkg/didcomm/protocol/didexchange/states.go
+++ b/pkg/didcomm/protocol/didexchange/states.go
@@ -503,9 +503,9 @@ func (ctx *context) didDocAttachment(doc *did.Doc, myVerKey string) (*decorator.
 	// Interop: signing did_doc~attach has been removed from the spec, but aca-py still verifies signatures
 	// TODO make aca-py issue
 	if ctx.doACAPyInterop {
-		pubKeyBytes, err := fingerprint.PubKeyFromDIDKey(myVerKey)
+		pubKeyBytes, err := ctx.resolvePublicKey(myVerKey)
 		if err != nil {
-			return nil, fmt.Errorf("failed to extract pubKeyBytes from did:key [%s]: %w", myVerKey, err)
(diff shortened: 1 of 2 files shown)


pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -528,6 +533,33 @@ func (ctx *context) didDocAttachment(doc *did.Doc, myVerKey string) (*decorator.
...
+
+		vm, ok := did.LookupPublicKey(kid, pubDoc.DIDDocument)
+		if !ok {
+			return nil, fmt.Errorf("failed to lookup public key for ID %s", kid)
+		}
+
+		return vm.Value, nil
+	}
+
+	return nil, nil
```
Baha-sk, line 560: return error here

Baha-sk (approved):

rolsonquadras (approved):

troyronda (approved):

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -528,6 +528,33 @@ func (ctx *context) didDocAttachment(doc *did.Doc, myVerKey string) (*decorator.
 	return docAttach, nil
 }
 
+func (ctx *context) resolvePublicKey(kid string) ([]byte, error) {
+	if strings.HasPrefix(kid, "did:key:") {
```
rolsonquadras, line 532: Is this logic specific to Interop as this is called inside if `ctx.doACAPyInterop` is true ?

sudeshrshetty (approved):

pkg/didcomm/protocol/didexchange/states.go
Moopli: It's not technically interop-only, but it's only needed in a specific case during aca-py interop.

Lifting it out requires some refactoring to change how we handle our key references across more of didexchange, and that wasn't necessary here so I avoided it.

PR #2894, opened 2021-07-27
fix: presexch shouldn't use only the final schema match when checking for schema match

Moopli:
Also include context needed for test harness tests.

### DIFF
--- a/pkg/doc/presexch/definition.go
+++ b/pkg/doc/presexch/definition.go
@@ -883,10 +883,12 @@ func filterSchema(schemas []*Schema, credentials []*verifiable.Credential) []*ve
 		var applicable bool
 
 		for _, schema := range schemas {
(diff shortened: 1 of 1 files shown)


fqutishat (approved):

pkg/doc/jsonld/contexts.go
```diff
@@ -122,4 +124,9 @@ var embedContexts = []ContextDocument{ //nolint:gochecknoglobals
 		DocumentURL: "https://ns.did.ai/suites/secp256k1-2019/v1/",
 		Content:     secp256k12019,
 	},
+	{
+		URL:         "https://w3id.org/citizenship/v1",
```
troyronda, line 128: This context cannot be embedded into the core framework loader - needs to be loaded by an application or test scenario.

troyronda (dismissed):
You need to externalize contexts out of the core framework package.

PR #2903, opened 2021-08-03
fix: presentation schema filtering respects schema IDs in contexts

### LINKED ISSUE #2756: presentation-exchange: improve `schema.uri` matching
PR #2755 fixes `schema.uri` matching for presentation definitions by requiring them to be the full URI to the type's jsonld type.

One shortcut made is the assumption that the `#` character is always the separator between the base context IRI and the type's specific name identifier. This works in many cases but not all (eg. `https://schema.org/Person`). This means that the verifier must populate `schema.uri` not necessarily with the type's full IRI, but with this artificial construct joining the context IRI and the type's name with `#`.

The algorithm should be generalized and made to work with the type's actual IRI always.

Moopli:
Fixes #2756


pkg/wallet/query.go
```diff
@@ -98,7 +98,7 @@ func (q *Query) PerformQuery(credentials map[string]json.RawMessage) ([]*verifia
 
 		credentials, err := q.getCredentials(qType, vcs, param.Query...)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("getCredentials: %w", err)
```
sudeshrshetty, line 101: what is the reason for prefixing function names here?

sudeshrshetty (approved):

pkg/wallet/query.go
Moopli: Added them for more verbose logging during debugging - I can remove these, they're not really necessary.

troyronda: you shouldn't put function names in the message.

troyronda (approved):

PR #2909, opened 2021-08-05
fix: Fixes DIDs resolution when inbound message received.

soluchok:

### DIFF
--- a/pkg/framework/context/context.go
+++ b/pkg/framework/context/context.go
@@ -9,8 +9,10 @@ package context
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/btcsuite/btcutil/base58"
+	"github.com/cenkalti/backoff/v4"
 	"github.com/piprate/json-gold/ld"
 
 	"github.com/hyperledger/aries-framework-go/pkg/crypto"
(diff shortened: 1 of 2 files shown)


fqutishat (approved):

pkg/framework/context/context.go
```diff
@@ -224,19 +233,29 @@ func (p *Provider) InboundMessageHandler() transport.InboundMessageHandler {
...
-	}
+	var (
+		myDID    string
+		theirDID string
+		err      error
+	)
+
+	return myDID, theirDID, backoff.Retry(func() error {
+		var notFound bool
+		myDID, err = p.didConnectionStore.GetDID(base58.Encode(envelope.ToKey))
```
rolsonquadras, line 244: @soluchok can you please add a comment/document on why we need the retry ? The function fetches the mapping between the key and DID from the store (this is not actual DID Resolution).

soluchok: When a connection exists for `envelope.ToKey` then it should exist for `envelope.FromKey`.
The bug I found was bc we had a connection for `envelope.ToKey` but we didn't for   `envelope.FromKey`.
Now I added `notFound` flag which allows us to determine whether the connection for `envelope.FromKey` should be present. If it should be present but the actual result is not found we will do a retry logic.

PR #2920, opened 2021-08-09
fix: schema uri matching handles IRIs expanded in nested contexts

Moopli:

### DIFF
--- a/pkg/doc/presexch/package_test.go
+++ b/pkg/doc/presexch/package_test.go
@@ -7,17 +7,32 @@ SPDX-License-Identifier: Apache-2.0
 package presexch
 
 import (
+	_ "embed"
 	"fmt"
 	"testing"
 
 	"github.com/piprate/json-gold/ld"
 	"github.com/stretchr/testify/require"
+
+	jld "github.com/hyperledger/aries-framework-go/pkg/doc/ld"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/ldcontext"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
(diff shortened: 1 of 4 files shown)


fqutishat (approved):

pkg/doc/presexch/package_test.go
```diff
@@ -53,3 +59,128 @@ func TestGetContext(t *testing.T) {
...
+			Context: []string{
+				verifiable.ContextURI,
+				"https://trustbloc.github.io/context/vc/examples/mdl-v1.jsonld",
+			},
+			Types: []string{verifiable.VCType, "mDL"},
+		},
+	}
+
+	t.Run("success", func(t *testing.T) {
+		mDLjson := `{
```
troyronda, line 79: test files should be created as a file in a testdata subfolder and then accessed using embed.

example:
https://github.com/hyperledger/aries-framework-go/blob/af30e870c3c085ed1debfe304de293ba2b995516/pkg/doc/signature/jsonld/processor_test.go#L509

pkg/doc/presexch/package_test.go
```diff
@@ -53,3 +59,128 @@ func TestGetContext(t *testing.T) {
...
+		}))
+
+		require.NoError(t, err)
+
+		matched := filterSchema(schemas, creds, docLoader)
+		require.Len(t, matched, 1)
+	})
+
+	t.Run("failure: fail to parse child ctx", func(t *testing.T) {
+		badctx := `{
```
troyronda, line 145: test files should be created as a file in a testdata subfolder and then accessed using embed.

pkg/doc/presexch/package_test.go
Moopli: fixed

PR #2937, opened 2021-08-17
refactor: use did:key in packagers (for didcomm v1)

### LINKED ISSUE #1207: fix hardcode base58

### LINKED ISSUE #1604: Inlined keys for didcomm
Service entries of type `did-communication` in did documents must be `did:key`s as per [didcomm conventions](https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions)

### LINKED ISSUE #1659: Update PublicKey parsing routine to support did:key method
Currently, Aries Framework supports the following encodings for public key: publicKeyBase58, publicKeyHex, publicKeyPem, and publicKeyJwk. However public key can be represented with did:key method as well.
```json
{
    "@type": "somemessagetype",
    "~service": {
        "recipientKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
    }
}
```

In the example above, did:key method can be used in "recipientKeys" instead of the "naked" key:

```json
{
    "@type": "somemessagetype",
    "~service": {
        "recipientKeys": ["did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"],
    }
}
```

For more details please check https://github.com/hyperledger/aries-rfcs/tree/master/features/0360-use-did-key.

### LINKED ISSUE #1847: Preload framework defaults with key vdri
As a framework user, I would like to have the vdri registry to be preloaded with did:key vdri, so the default Aries instance has did:key resolving capabilities.

Baha-sk:
This change removes Base58 key encoding for recipients and sender for did:key (execept for legacy packer which will remain using bsae58 encoding).
It also includes vdr peer did creation with all VM and KeyAgreement keys, not only the first key.
This change adds key resolvers for the packers, one for did:key resolution and another one used for the 1PU test vector verification purposes using a thirdparty store resolution.
Finally, this change also moves did:key fingerprint code referencing did.Doc into a separate subpackage (pkg/vdr/fingerprint/didfp) to avoid cyclic dependency with jwk fingerprint.

closes #1604
closes #1207
closes #1659
closes #1847

### DIFF
--- a/pkg/didcomm/common/service/destination_default.go
+++ b/pkg/didcomm/common/service/destination_default.go
@@ -10,6 +10,7 @@ package service
 
 import (
 	"fmt"
+	"strings"
 
 	diddoc "github.com/hyperledger/aries-framework-go/pkg/doc/did"
 )
@@ -30,8 +31,12 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
 		return nil, fmt.Errorf("create destination: no recipient keys on didcomm service block in diddoc: %+v", didDoc)
 	}
 
-	// TODO ensure recipient keys are did:key's
-	//  https://github.com/hyperledger/aries-framework-go/issues/1604
+	for i, k := range didCommService.RecipientKeys {
+		if !strings.HasPrefix(k, "did:") {
+			return nil, fmt.Errorf("create destination: recipient key %d:[%v] of didComm '%s' not a did:key", i+1,
+				k, didCommService.ID)
+		}
+	}
 
 	return &Destination{
 		RecipientKeys:     didCommService.RecipientKeys,
--- a/pkg/didcomm/protocol/mediator/service.go
+++ b/pkg/didcomm/protocol/mediator/service.go
@@ -361,6 +361,8 @@ func (s *Service) handleInboundRequest(c *callback) error {
 		c.options,
 		s.endpoint,
 		func() (string, error) {
+			// TODO why use hard coded key type and key creation here? pass ctx.Keytype to s *service and use it instead
+			// TODO of hard coded ED25519 type if new key is still required.
 			_, pubKeyBytes, er := s.kms.CreateAndExportPubKeyBytes(kms.ED25519Type)
 			if er != nil {
 				return "", fmt.Errorf("outboundGrant from handleInboundRequest: kms failed to create and "+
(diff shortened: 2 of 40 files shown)


fqutishat (approved):

pkg/didcomm/protocol/mediator/service.go
```diff
@@ -748,6 +750,9 @@ func (s *Service) setKeyUpdateResponseCh(msgID string, keyUpdateCh chan *Keylist
 }
 
 func (s *Service) ensureConnectionExists(connID string) error {
+	//nolint:forbidigo
+	fmt.Println(fmt.Sprintf("~~~~ get router connID in routeStore: "+routeConnIDDataKey, connID))
```
rolsonquadras, line 754: Can we remove this fmt statement ? If needed, we can use logger instead.

rolsonquadras (approved):

pkg/didcomm/common/service/destination_default.go
```diff
@@ -30,8 +31,12 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
 		return nil, fmt.Errorf("create destination: no recipient keys on didcomm service block in diddoc: %+v", didDoc)
 	}
 
-	// TODO ensure recipient keys are did:key's
-	//  https://github.com/hyperledger/aries-framework-go/issues/1604
+	for i, k := range didCommService.RecipientKeys {
+		if !strings.HasPrefix(k, "did:") {
```
Moopli, line 35: This is testing against prefix `did:`, so I guess we're still accepting did-url key references, right?

Moopli (approved):
Noticed a few small places where we could refactor to clean up a bit, but those can come later.
LGTM :+1:

pkg/didcomm/common/service/destination_default.go
Baha-sk: This is a check to ensure RecipientKeys are always didkeys

The didCommService is already updated to have RecipientKeys as did keys. This is only a validation as per the removed TODO comment above.

pkg/didcomm/protocol/mediator/service.go
Baha-sk: removed

pkg/didcomm/common/service/destination_default.go
Baha-sk: > Noticed a few small places where we could refactor to clean up a bit, but those can come later.
> LGTM 👍

Indeed there's lot's of room for improvements.

Baha-sk:
> Noticed a few small places where we could refactor to clean up a bit, but those can come later.
> LGTM 👍

Indeed there's lot's of room for improvements.

aholovko (approved):

PR #2964, opened 2021-09-01
feat: mobile bindings for vcwallet

skflorin:
Signed-off-by: Florin Birsan <71101813+skflorin@users.noreply.github.com>

Added VCWalletController to be used in native code


sudeshrshetty (approved):

fqutishat (approved):

aholovko:
@skflorin I would put `contexts` under `testdata` folder and move test document loader to the place where it is used (`vcwallet_test.go`) to avoid confusion that the loader from `document_loader.go` file is not a production code and is needed only for tests. But this is just a matter of style, I am ok with what we have now as well.

aholovko (approved):

skflorin:
> @skflorin I would put `contexts` under `testdata` folder and move test document loader to the place where it is used (`vcwallet_test.go`) to avoid confusion that the loader from `document_loader.go` file is not a production code and is needed only for tests. But this is just a matter of style, I am ok with what we have now as well.

@aholovko I will make the change as suggested by you. This was my first piece of Go code and it was done on agent-sdk first before the document loader was added. I will clean it up

PR #2992, opened 2021-09-21
fix: add interop support to new didexchange key handling

Moopli:

### DIFF
--- a/pkg/didcomm/packager/packager.go
+++ b/pkg/didcomm/packager/packager.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/packer"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/packer/authcrypt"
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/jose/jwk/jwksupport"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/util/jwkkid"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/util/kmsdidkey"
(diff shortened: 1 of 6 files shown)


fqutishat (approved):

pkg/didcomm/packager/packager.go
```diff
@@ -117,30 +118,40 @@ func (bp *Packager) PackMessage(messageEnvelope *transport.Envelope) ([]byte, er
...
+		case strings.HasPrefix(receiverKeyID, "did:key"):
+			marshalledKey, err := addDIDKeyToRecipients(i, receiverKeyID, isLegacy)
 			if err != nil {
 				return nil, nil, err
 			}
 
 			recipients = append(recipients, marshalledKey)
-		case strings.Index(receiverKey, "#") > 0:
-			marshalledKey, err := bp.resolveKeyAgreementFromDIDDoc(receiverKey)
+		case strings.Index(receiverKeyID, "#") > 0:
```
Baha-sk, line 136: this condition is not meant for legacy packing

pkg/didcomm/packager/packager.go
```diff
@@ -161,26 +172,28 @@ func (bp *Packager) prepareSenderAndRecipientKeys(cty string, envelope *transpor
 		}
 	//nolint:gocritic // need to check with strings not bytes
 	case strings.Index(string(envelope.FromKey), "#") > 0:
```
Baha-sk, line 174: this condition is not meant for legacy packing

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -687,7 +687,7 @@ func (ctx *context) getMyDIDDoc(pubDID string, routerConnections []string, servi
 
 	if newService {
 		switch newDID.Service[0].Type {
-		case didCommServiceType:
+		case didCommServiceType, "IndyAgent":
```
Baha-sk, line 690: is this a new DIDComm service type? There's many places where this needs to be updated and must follow logic of mediaType profile for DIDComm V1 + Interop Flag. It means the Interop flag must become part of AFGO's context.

I'm currently updating bdd tests with DID docs missing the MediaTypeProfiles() and KeyAgreement() contexts, the new Interop flag must be added on top of these changes as well.

It's not enough to change this here. There's at least oob invitations as well.

pkg/didcomm/packager/packager.go
```diff
@@ -361,58 +374,77 @@ func (bp *Packager) getCTYAndPacker(envelope *transport.Envelope) (string, packe
...
+		// the kms kid value during unpack.
+		recKey.KID = keyAgrID
+	case x25519KeyAgreementKey2019:
+		recKey = &crypto.PublicKey{
+			KID:   keyAgrID,
+			X:     vm.Value,
+			Curve: "X25519",
+			Type:  "OKP",
+		}
+	case "Ed25519VerificationKey2018":
```
Baha-sk, line 437: This condition is for legacy packer only which is based on recipientKey as base58 encoded keys coming from destination.RecipientKeys, the packager must not get DID doc's VMs for legacy packing. Let's not propagate legacy stuff further in the framework.

Baha-sk (changes requested):

pkg/didcomm/packager/packager.go
Moopli: and yet, some did docs in legacy interop scenarios _do_ use DID url key references - and the new handling, that assumes this only happens in non-legacy scenarios, broke interop tests that previously worked.

pkg/didcomm/protocol/didexchange/states.go
Moopli: > is this a new DIDComm service type?
It's an old service type, from AIP1, that we only need to recognize in limited cases, with the intent that it goes away once ACA-Py is sure to no longer emit it. Adding support everywhere would be counterproductive.

Baha-sk (approved):

PR #2998, opened 2021-09-23
refactor: use Mediatype Profile instead of MediaType as default

Baha-sk:


fqutishat (approved):

pkg/framework/aries/default.go
```diff
@@ -229,10 +229,11 @@ func setAdditionalDefaultOpts(frameworkOpts *Aries) error {
 	}
 
 	if frameworkOpts.mediaTypeProfiles == nil {
-		// for now only set legacy media type profile to match default key type and primary packer above.
+		// For now only set legacy media type profile to match default key type and primary packer above.
+		// Using media type profile, not just a media type, in order to align with OOB invitations' Accept header.
 		// TODO once keyAgreement is added in the packers, this can be switched to DIDcomm V2 media type as well as
 		// 		switching legacyPacker with authcrtypt as primary packer and using an ECDH-1PU key as default key above.
-		frameworkOpts.mediaTypeProfiles = []string{transport.MediaTypeRFC0019EncryptedEnvelope}
+		frameworkOpts.mediaTypeProfiles = []string{transport.MediaTypeProfileDIDCommAIP1}
```
troyronda, line 236: Our default is actually AIP2 - why make it AIP1?

troyronda (changes requested):
AIP2

pkg/framework/aries/default.go
Baha-sk: Shouldn't we keep AIP1 till it's retired to avoid compatibility issues? putting AIP2 here will force people to manually add: `WithMediaTypeProfiles([]string{MediaTypeProfileDIDCommAIP1})` option in their existing clients to maintain existing behaviour.

Or starting AFGO 0.1.8, default should be DIDCommV2 (`MediaTypeDIDCommV2Profile = "didcomm/v2"`) ?

as for `AIP2`, there are 2 flavours:
`MediaTypeAIP2RFC0019Profile = "didcomm/aip2;env=rfc19"`
and 
`MediaTypeAIP2RFC0587Profile = "didcomm/aip2;env=rfc587"`

did you mean the second?

Baha-sk: I'll update to `MediaTypeAIP2RFC0587Profile = "didcomm/aip2;env=rfc587"` when aries-js-worker clients are fully updated to support MediaTypeProfiles option,

for now, i'll set the default to `MediaTypeAIP2RFC0019Profile = "didcomm/aip2;env=rfc19"`

Baha-sk:
updated packager to support unpacking base64 encoded and double quoted envelopes to avoid duplicate unpacking in the ws pool listener/http inbound handler

PR #3005, opened 2021-09-29
fix: include status in issue-cred ack

Moopli:


Moopli:
This issue just cropped up in testing today, seems to have been a small change in aca-py to expect this ack to have a status as well.

fqutishat (approved):

PR #3032, opened 2021-11-01
fix: now decrypt work properly for rotated keys

vkubiv:

**Title:**
Fix tinkcrypto.Decrypt to work properly for rotated keys.

**Summary:**

Iterate through all key prefixes in the keyset to try to decrypt data.


Baha-sk (approved):
as discussed, spot on and it looks good @vkubiv 

I've added a small comment detail in the unit-test, but i'm approving since the code is correct as suggested.

fqutishat (approved):

PR #3035, opened 2021-11-03
feat: two-stage protocol service initialization & inbound dispatcher package

Moopli:

### DIFF
--- a/pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
+++ b/pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
@@ -1,0 +1,267 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package inbound
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/cenkalti/backoff/v4"
+
+	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/crypto"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/dispatcher"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api"
+	"github.com/hyperledger/aries-framework-go/pkg/store/did"
+	"github.com/hyperledger/aries-framework-go/pkg/vdr/fingerprint"
+)
+
+var logger = log.New("dispatcher/inbound")
+
+const (
+	kaIdentifier = "#"
+)
+
+// MessageHandler handles inbound envelopes, processing then dispatching to a protocol service based on the
+// message type.
+type MessageHandler struct {
+	didConnectionStore     did.ConnectionStore
+	msgSvcProvider         api.MessageServiceProvider
+	services               []dispatcher.ProtocolService
+	getDIDsBackOffDuration time.Duration
+	getDIDsMaxRetries      uint64
+	messenger              service.InboundMessenger
+	initialized            bool
+}
+
+type provider interface {
+	DIDConnectionStore() did.ConnectionStore
+	MessageServiceProvider() api.MessageServiceProvider
+	AllServices() []dispatcher.ProtocolService
+	GetDIDsBackOffDuration() time.Duration
+	GetDIDsMaxRetries() uint64
+	InboundMessenger() service.InboundMessenger
+}
+
+// NewInboundMessageHandler creates an inbound message handler, that processes inbound message Envelopes,
+// and dispatches them to the appropriate ProtocolService.
+func NewInboundMessageHandler(p provider) *MessageHandler {
+	h := MessageHandler{}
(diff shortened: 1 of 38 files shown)


Moopli:
Marking as ready, only needs test coverage.

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	isV2, err := service.IsDIDCommV2(&msg)
+	if err != nil {
+		return err
+	}
+
+	// if msg is a didcomm v2 message, do additional handling
+	if isV2 {
+		logger.Debugf("didcomm v2 message handling before service...")
```
Baha-sk, line 101: add a TODO here?

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
```diff
@@ -0,0 +1,238 @@
...
+
+			return err
+		}
+	}
+
+	// in case of no services are registered for given message type,
+	// find generic inbound services registered for given message header
+	for _, svc := range handler.msgSvcProvider.Services() {
+		h := struct {
+			Purpose []string `json:"~purpose"`
```
Baha-sk, line 129: is this a V1 format?

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
```diff
@@ -0,0 +1,238 @@
...
+
+			return handler.tryToHandle(svc, msg, service.NewDIDCommContext(myDID, theirDID, nil))
+		}
+	}
+
+	return fmt.Errorf("no message handlers found for the message type: %s", msg.Type())
+}
+
+//nolint:nestif,gocognit,funlen,gocyclo
+func (handler *MessageHandler) getDIDs(envelope *transport.Envelope) (string, string, error) {
```
Baha-sk, line 151: just to confirm, this was extract as-is from the original inboundhandler, correct?

pkg/didcomm/protocol/mediator/service.go
```diff
@@ -149,55 +149,77 @@ type Service struct {
...
 	mp, err := prov.Service(messagepickup.MessagePickup)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	messagePickupSvc, ok := mp.(messagepickup.ProtocolService)
 	if !ok {
-		return nil, errors.New("cast service to message pickup service failed")
+		return errors.New("cast service to message pickup service failed")
```
Baha-sk, line 201: why not add Type check in the error here as  was done for the providers above?

pkg/didcomm/protocol/outofband/service.go
```diff
@@ -156,59 +157,80 @@ type Provider interface {
...
 	svc, err := p.Service(didexchange.DIDExchange)
 	if err != nil {
-		return nil, fmt.Errorf("failed to initialize outofband service : %w", err)
+		return fmt.Errorf("failed to initialize outofband service : %w", err)
 	}
 
 	didSvc, ok := svc.(didExchSvc)
 	if !ok {
-		return nil, errors.New("failed to cast the didexchange service to satisfy our dependency")
+		return errors.New("failed to cast the didexchange service to satisfy our dependency")
```
Baha-sk, line 188: same, simply say something like: oob depends on service didExchangeService, got (%T) instead

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
Moopli: Almost completely as-is, I'll note the only change

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
```diff
@@ -0,0 +1,238 @@
...
+
+	return myDID, theirDID, backoff.Retry(func() error {
+		var notFound bool
+
+		// nolint: gocritic
+		if id := strings.Index(string(envelope.ToKey), kaIdentifier); id > 0 &&
+			strings.Index(string(envelope.ToKey), "\"kid\":\"did:") > 0 {
+			myDID, err = pubKeyToDID(envelope.ToKey)
+			if err != nil {
+				return backoff.Permanent(fmt.Errorf("getMyDID: envelope.ToKey as myDID: %w", err))
```
Moopli, line 166: @Baha-sk the only change is that these errors are now wrapped in `backoff.Permanent()`. If pubKeyToDID fails, then it's never going to be able to succeed in a retry, so wrapping it in `backoff.Permanent` keeps it from retrying a case that can only fail over and over.

Baha-sk (approved):

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
Moopli: Yes - I'm not sure what will happen with the message service and purpose decorator when/if it gets ported to didcomm/v2, but this part is v1 only.

I can put it inside an `if`.

Baha-sk: yes, it would be better

fqutishat (approved):

pkg/didcomm/protocol/mediator/service.go
Moopli: minor issue, baha agreed there's no need

pkg/didcomm/protocol/outofband/service.go
Moopli: minor issue, baha agreed there's no need

pkg/didcomm/dispatcher/inbound/inbound_message_handler.go
Moopli: done

PR #3038, opened 2021-11-04
feat: OOB V2

Baha-sk:
This change introduces OOB V2 service and client first draft.
It also injects the Accept oob property in the BDD tests.

### DIFF
--- a/pkg/client/outofbandv2/client.go
+++ b/pkg/client/outofbandv2/client.go
@@ -1,0 +1,174 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package outofbandv2
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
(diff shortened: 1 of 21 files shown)


fqutishat (approved):

pkg/client/outofbandv2/client.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	result := make([]oobv2.Action, len(actions))
+	copy(result, actions)
+
+	return result, nil
+}
+
+// ActionContinue allows continuing with the protocol after an action event was triggered.
+func (c *Client) ActionContinue(piID string, opts oobv2.Options) error {
```
rolsonquadras, line 153: When does a client need to invoke this function ? Do they have to call `AcceptInvitation` first and then invoke this ?

Baha-sk: so ActionContinue is supposed to trigger the other services, would it be the case? If it's not needed, i can definitely remove it. We just need to figure out where/how to trigger the other protocol services via oob.

Moopli (approved):
Selecting and triggering the followup protocol (ie, whatever the OOB message wraps) is in a followup PR, so this looks good to me :+1:

pkg/client/outofbandv2/client.go
Baha-sk: removed ActionContinue(), Actions() and ActionStop()

PR #3044, opened 2021-11-09
feat: Credential Manifest

DRK3:

### DIFF
--- a/pkg/doc/credentialmanifest/credentialmanifest.go
+++ b/pkg/doc/credentialmanifest/credentialmanifest.go
@@ -1,0 +1,427 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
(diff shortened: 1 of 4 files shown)


pkg/doc/credentialmanifest/credentialmanifest.go
```diff
@@ -0,0 +1,422 @@
...
+		resolveDynamicDisplayMappingObjects(outputDescriptor.Display.Properties, vc)
+	if err != nil {
+		return ResolvedDataDisplayDescriptor{}, err
+	}
+
+	return resolvedDataDisplayDescriptor, nil
+}
+
+func resolveStaticDisplayMappingObjects(outputDescriptor *OutputDescriptor,
+	vc map[string]interface{}) (string, string, string, error) {
```
talwinder50, line 191: Can a return here be the struct rather than string, string , string ??  like ResolvedDataDisplayDescriptor

fqutishat (approved):

PR #3049, opened 2021-11-12
fix(vcwallet): convert key to kid before issuing vc

### LINKED ISSUE #3029: VC wallet: Unable to issue credential
* Are you filing a bug report?

**What I'm trying to do**

I'm trying to issue a credential via the `/vcwallet/issue` endpoint with a key that I created beforehand via `/vcwallet/create-key-pair`.

**Expected result**

Credential with an added proof section.

**Actual result**

Error:

```json
{
  "code": 12010,
  "message": "failed to issue credential: getKeySet: failed to read json keyset from reader: cannot read data for keysetID z6MksNRMoWkz5UdGgSkpbMS83FvRkKzNN5XAbbF2gLgWCmCe: failed to get DB entry: data not found"
}
```

**Sample code / test case**

1. I created a key pair via the `/vcwallet/create-key-pair` endpoint

```json
{
  "keyID": "zteLhw3qQMOozGq8ls60AKxBxWBUNl7T1k-ypvWjEIc",
  "publicKey": "v-qOrzVcRfrqXrgy8-JvajJfZotJcsuV1zA54fX7z_s"
}
```

2. I computed the DID for the public key

```
did:key:z6MksNRMoWkz5UdGgSkpbMS83FvRkKzNN5XAbbF2gLgWCmCe
```

3. I sent an unsigned VC to the `/vcwallet/issue` endpoint

```json
{
  "auth": "{{TOKEN}}",
  "credential": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1"
    ],
    "id": "http://example.edu/credentials/58473",
    "issuer": "did:key:z6MksNRMoWkz5UdGgSkpbMS83FvRkKzNN5XAbbF2gLgWCmCe",
    "issuanceDate": "2021-10-25T16:49:57Z",
    "type": [
      "VerifiableCredential"
    ],
    "credentialSubject": {
      "id": "did:example:ebfeb1f712ebc6f1c276e12ec21"
    }
  },
  "proofOptions": {
    "challenge": "test-challenge",
    "controller": "did:key:z6MksNRMoWkz5UdGgSkpbMS83FvRkKzNN5XAbbF2gLgWCmCe",
    "created": "2021-10-20T06:28:12.840Z",
    "domain": "identinet.io",
    "proofRepresentation": 0,
    "proofType": "Ed25519Signature2018",
    "verificationMethod": "did:key:z6MksNRMoWkz5UdGgSkpbMS83FvRkKzNN5XAbbF2gLgWCmCe#z6MksNRMoWkz5UdGgSkpbMS83FvRkKzNN5XAbbF2gLgWCmCe"
  },
  "userID": "user1"
}
```

It looks like the verification method isn't properly turned into a keyID that can key found by the key manager. Instead the verification method is directly fed to `keyManager.Get()`:
https://github.com/hyperledger/aries-framework-go/blob/main/pkg/wallet/kmsclient.go#L271

Also the test cases only test the error cases but not the happy case:
https://github.com/hyperledger/aries-framework-go/blob/main/pkg/wallet/kmsclient_test.go#L587

Furthermore, I'm wondering whether splitting the verification method is enough to find the right key. In my understanding the steps to find the right key are:
1. resolve the DID document referenced by `verificationMethod` (`split('#')[0]`) - this step happens
2. find the public key with id `split('#')[1]` - this step seems to be missing
3. decode the public key and pass it to the key manager - this step seems to be missing
   - instead  `split('#')[1]` is used as public key or key id

jceb:
**Title:**

Fix VC wallet: Unable to issue credential

**Description:**

Fixes #3029

**Summary:**

Converts the verification method / public key to the key ID that's required by keyManager to retrieve the private key for signing. The conversion steps were taken from the standard VC issuing endpoint.

### DIFF
--- a/pkg/controller/command/verifiable/command.go
+++ b/pkg/controller/command/verifiable/command.go
@@ -142,8 +142,14 @@ const (
 	// P521KeyCurve EC P-521 curve.
 	P521KeyCurve = "P-521"
 
-	// Ed25519VerificationKey ED25519 verification key type.
-	Ed25519VerificationKey = "Ed25519VerificationKey"
+	// Ed25519VerificationKeyPrefix ED25519 verification key type.
+	Ed25519VerificationKeyPrefix = "Ed25519VerificationKey"
+
+	// Ed25519VerificationKey2018 ED25519 2018 verification key type.
+	Ed25519VerificationKey2018 = "Ed25519VerificationKey2018"
+
+	// Bls12381G2Key2020 BLS12-381 signature key type.
+	Bls12381G2Key2020 = "Bls12381G2Key2020"
 
 	// JSONWebKey2020 verification key type.
 	JSONWebKey2020 = "JsonWebKey2020"
@@ -1082,9 +1088,11 @@ func buildKIDOption(opts *ProofOptions, vms []did.VerificationMethod) error {
 				kt := kms.ED25519Type
 
 				switch vm.Type {
-				case Ed25519VerificationKey:
+				case Ed25519VerificationKey2018:
+				case Bls12381G2Key2020:
+					kt = kms.BLS12381G2Type
 				case JSONWebKey2020:
-					kt = kmsKeyTypeByJWKCurve(vm.JSONWebKey().Crv)
+					kt = KmsKeyTypeByJWKCurve(vm.JSONWebKey().Crv)
 				}
 
 				kid, err := localkms.CreateKID(vm.Value, kt)
@@ -1100,7 +1108,8 @@ func buildKIDOption(opts *ProofOptions, vms []did.VerificationMethod) error {
 	return nil
 }
 
-func kmsKeyTypeByJWKCurve(crv string) kms.KeyType {
+// KmsKeyTypeByJWKCurve returns key type for the JWK curve's name.
+func KmsKeyTypeByJWKCurve(crv string) kms.KeyType {
 	kt := kms.ED25519Type
 
 	switch crv {
@@ -1123,7 +1132,7 @@ func getDefaultVerificationMethod(didDoc *did.Doc) (string, error) {
 		var publicKeyID string
 
 		for _, k := range didDoc.VerificationMethod {
-			if strings.HasPrefix(k.Type, Ed25519VerificationKey) {
+			if strings.HasPrefix(k.Type, Ed25519VerificationKeyPrefix) {
 				publicKeyID = k.ID
 
 				break
--- a/pkg/kms/localkms/privkey_import.go
+++ b/pkg/kms/localkms/privkey_import.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"crypto/ecdsa"
 	"crypto/ed25519"
+	"crypto/elliptic"
 	"fmt"
 
 	"github.com/golang/protobuf/proto"
@@ -23,6 +24,7 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/composite/ecdh"
 	bbspb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/bbs_go_proto"
 	ecdhpb "github.com/hyperledger/aries-framework-go/pkg/crypto/tinkcrypto/primitive/proto/ecdh_aead_go_proto"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/util/jwkkid"
 	"github.com/hyperledger/aries-framework-go/pkg/kms"
 )
 
@@ -34,8 +36,10 @@ const (
 	nistpECDHKWPrivateKeyTypeURL = "type.hyperledger.org/hyperledger.aries.crypto.tink.NistPEcdhKwPrivateKey"
 )
 
+//nolint:funlen,gocyclo
 func (l *LocalKMS) importECDSAKey(privKey *ecdsa.PrivateKey, kt kms.KeyType,
-	opts ...kms.PrivateKeyOpts) (string, *keyset.Handle, error) {
+	opts ...kms.PrivateKeyOpts,
+) (string, *keyset.Handle, error) {
 	var params *ecdsapb.EcdsaParams
 
 	err := validECPrivateKey(privKey)
@@ -86,6 +90,34 @@ func (l *LocalKMS) importECDSAKey(privKey *ecdsa.PrivateKey, kt kms.KeyType,
 		return "", nil, fmt.Errorf("import private EC key failed: invalid ECDSA key type")
 	}
 
+	if len(opts) == 0 {
+		var kid string
+
+		// since the public key exists already in decoded form, use the
+		// IEEEP1363 encoding for marhalling and computing the key id as it is
+		// easier to use
+		tmpKt := kms.ECDSAP256TypeIEEEP1363
+
+		switch kt {
+		case kms.ECDSAP256TypeIEEEP1363, kms.ECDSAP256TypeDER:
+			tmpKt = kms.ECDSAP256TypeIEEEP1363
+		case kms.ECDSAP384TypeIEEEP1363, kms.ECDSAP384TypeDER:
+			tmpKt = kms.ECDSAP384TypeIEEEP1363
+		case kms.ECDSAP521TypeIEEEP1363, kms.ECDSAP521TypeDER:
+			tmpKt = kms.ECDSAP521TypeIEEEP1363
+		}
+
+		c := jwkkid.GetCurveByKMSKeyType(tmpKt)
+		pubKeyBytes := elliptic.Marshal(c, privKey.PublicKey.X, privKey.PublicKey.Y)
+
+		kid, err = CreateKID(pubKeyBytes, tmpKt)
+		if err != nil {
+			return "", nil, fmt.Errorf("import private EC  key failed: %w", err)
+		}
+
+		opts = append(opts, kms.WithKeyID(kid))
+	}
+
 	mKeyValue, err := getMarshalledECDSAPrivateKey(privKey, params)
 	if err != nil {
 		return "", nil, fmt.Errorf("import private EC key failed: %w", err)
@@ -97,7 +129,8 @@ func (l *LocalKMS) importECDSAKey(privKey *ecdsa.PrivateKey, kt kms.KeyType,
 }
 
 func (l *LocalKMS) buildAndImportECDSAPrivateKeyAsECDHKW(privKey *ecdsa.PrivateKey, kt kms.KeyType,
-	opts ...kms.PrivateKeyOpts) (string, *keyset.Handle, error) {
+	opts ...kms.PrivateKeyOpts,
+) (string, *keyset.Handle, error) {
 	var keyTemplate *tinkpb.KeyTemplate
 
 	switch kt {
@@ -157,7 +190,8 @@ func getMarshalledECDSAPrivateKey(privKey *ecdsa.PrivateKey, params *ecdsapb.Ecd
 }
 
 func (l *LocalKMS) importEd25519Key(privKey ed25519.PrivateKey, kt kms.KeyType,
-	opts ...kms.PrivateKeyOpts) (string, *keyset.Handle, error) {
+	opts ...kms.PrivateKeyOpts,
+) (string, *keyset.Handle, error) {
 	if privKey == nil {
 		return "", nil, fmt.Errorf("import private ED25519 key failed: private key is nil")
 	}
@@ -171,6 +205,17 @@ func (l *LocalKMS) importEd25519Key(privKey ed25519.PrivateKey, kt kms.KeyType,
 		return "", nil, fmt.Errorf("import private ED25519 key failed: %w", err)
 	}
 
+	if len(opts) == 0 {
+		var kid string
+
+		kid, err = CreateKID(privKeyProto.PublicKey.KeyValue, kt)
+		if err != nil {
+			return "", nil, fmt.Errorf("import private ED25519 key failed: %w", err)
+		}
+
+		opts = append(opts, kms.WithKeyID(kid))
+	}
+
 	mKeyValue, err := proto.Marshal(privKeyProto)
 	if err != nil {
 		return "", nil, fmt.Errorf("import private ED25519 key failed: %w", err)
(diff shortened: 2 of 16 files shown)


pkg/controller/command/verifiable/command.go
```diff
@@ -1100,7 +1100,7 @@ func buildKIDOption(opts *ProofOptions, vms []did.VerificationMethod) error {
 	return nil
 }
 
-func kmsKeyTypeByJWKCurve(crv string) kms.KeyType {
+func KmsKeyTypeByJWKCurve(crv string) kms.KeyType {
```
Baha-sk, line 1103: please add a function comment since this function is now public

baha-ai (approved):
thank you your contribution @jceb 

The change looks good, i left a tiny comment to help make the linter check pass

jceb:
@Baha-sk 
Unfortunately, the PR isn't ready yet. A bunch of tests are failing due to the changes that I made. -> This brings me to a larger question:

I noticed that the ImportPrivateKey function accepts a custom key id which serves as the handle to the key store for future operations like issuing a credential. At least for the REST interface it's currently not possible to specify a custom key id that produces a signed VC. Instead, the key id is deterministically derived from the public key that's referenced in the DID document. As a developer I don't have to do anything to manage the key id. This behavior makes a lot of sense to me, because operations like issuing a VC require an identifier - if I can pass in the identifier/controller and the verification method that points to a public key in the DID document I see lots of consistency and easy to follow references.

With custom key ids, I have a much harder time grasping what's going on because the custom key ids are not part of any DID document. However, the current tests are using custom key ids a lot to store the same key multiple times in the database.

Do you see any use cases that require custom key ids apart from storing symmetric keys? As a user of aries-framework-go, I'd prefer to not deal with custom key ids at all because it requires additional work and it might even cause problems by signing a VC with a key that's not associated with the DID that I'm using.

Baha-sk:
> @Baha-sk Unfortunately, the PR isn't ready yet. A bunch of tests are failing due to the changes that I made. -> This brings me to a larger question:
> 
> I noticed that the ImportPrivateKey function accepts a custom key id which serves as the handle to the key store for future operations like issuing a credential. At least for the REST interface it's currently not possible to specify a custom key id that produces a signed VC. Instead, the key id is deterministically derived from the public key that's referenced in the DID document. As a developer I don't have to do anything to manage the key id. This behavior makes a lot of sense to me, because operations like issuing a VC require an identifier - if I can pass in the identifier/controller and the verification method that points to a public key in the DID document I see lots of consistency and easy to follow references.
> 
> With custom key ids, I have a much harder time grasping what's going on because the custom key ids are not part of any DID document. However, the current tests are using custom key ids a lot to store the same key multiple times in the database.
> 
> Do you see any use cases that require custom key ids apart from storing symmetric keys? As a user of aries-framework-go, I'd prefer to not deal with custom key ids at all because it requires additional work and it might even cause problems by signing a VC with a key that's not associated with the DID that I'm using.

@jceb setting a custom KID in the ImportPrivateKey is optional, it was added for users who want to preset it to whatever KID they have externally (but doesn't guarantee the key with custom KID will work in all cases in the framework). The KMS does not require it as it uses the JWK thumbprint of a key as the KID. 

In other words, you don't have to set a custom KID when calling ImportPrivateKey(), simply use the generated KID returned by the function to keep track of the key in AFGO's KMS.

Baha-sk:
I see now, the unit-tests are failing because the KID is not found:
```
--- FAIL: TestWallet_Issue (0.54s)
    --- FAIL: TestWallet_Issue/Test_VC_wallet_issue_using_controller_-_success (0.04s)
        wallet_test.go:1271: 
            	Error Trace:	wallet_test.go:1271
            	Error:      	Received unexpected error:
            	            	failed to issue credential: getKeySet: failed to read json keyset from reader: cannot read data for keysetID dfiDkP85Xq5Cald-Q7nT4515MNAcguRJzJpD4CsepAg: data not found
```

I would remove `, kms.WithKeyID(kid)` at line 1265 of wallet_test.go: https://github.com/hyperledger/aries-framework-go/blob/main/pkg/wallet/wallet_test.go#L1265

and all other tests explicitly calling `WithKeyID(kid)`

pkg/wallet/wallet.go
```diff
@@ -37,8 +36,10 @@ import (
 	"github.com/hyperledger/aries-framework-go/pkg/doc/verifiable"
 	"github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
 	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/kms/localkms"
 	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
 	"github.com/hyperledger/aries-framework-go/spi/storage"
+	"github.com/piprate/json-gold/ld"
```
Baha-sk, line 42: please revert this last import line and keep it under `"github.com/google/uuid"` as it was prior to this pr, this is not an AFGO import, so it must be part of third party dependencies block above AFGO's dependencies.

jceb:
Unfortunately, the kid produced by ImportPrivateKey and CreateKeyPair are different. I'll update the PR to fix that issue. I'll also implement the use case that allows for a custom kid while unifying the kids returned by the create and import methods.

Baha-sk:
> CreateKeyPair

as per the wallet's CreateKeyPair() function, it uses the KMS to create the key and KID, so it must match AFGO's KMS (unless the unit-tests use a mock KMS that returns a fixed key/KID):
https://github.com/hyperledger/aries-framework-go/blob/main/pkg/wallet/wallet.go#L608

jceb:
@Baha-sk the pull request turned out to be bigger than expected. Please have a look whether the changes are okay.

I'll continue to make the remainder of the tests pass.

pkg/wallet/wallet.go
Baha-sk: this is still not reverted, please revert this line with the removed line 18 above

pkg/kms/localkms/privkey_import.go
```diff
@@ -80,6 +82,34 @@ func (l *LocalKMS) importECDSAKey(privKey *ecdsa.PrivateKey, kt kms.KeyType,
...
+		pubKeyBytes := elliptic.Marshal(c, privKey.PublicKey.X, privKey.PublicKey.Y)
+
+		kid, err = CreateKID(pubKeyBytes, tmpKt)
+		if err != nil {
+			return "", nil, fmt.Errorf("import private EC  key failed: %w", err)
+		}
+
+		opts = append(opts, kms.WithKeyID(kid))
+	}
+
```
Baha-sk, line 112: this logic is unnecessary since if no KID option is being passed, the function will call `CreateKID()` implicitly.

Baha-sk: I take away the above statement, CreateKID() was built after ImportPrivKey(), so it's never called, a random KID is generated instead.

This logic is therefore valid for ECDSA keys here and similar CreateKID() calls you added for Ed25519 and BBS keys.

jceb:
@Baha-sk one more thing - the add content (https://github.com/identinet/aries-framework-go/blob/fix-vcwallet-correct-issue-vc/pkg/wallet/wallet.go#L407) doesn't return a value yet. So if a key gets added, the developer/user has to manually compute the key id from the public key to find out where the key's stored. This is inconvenient and I'd prefer as a developer to just receive back the id of the key in particular and everything that I add in general, to have an easy way of tracking what I added to my wallet.

The non key material can easily be queried but stored keys can't be queried. So for an application to keep track of the keys that the user added, it's important to receive back the id.

What do you think, shall we make this change to the `Add` function signature in this PR?

Baha-sk:
> @Baha-sk one more thing - add content (https://github.com/identinet/aries-framework-go/blob/fix-vcwallet-correct-issue-vc/pkg/wallet/wallet.go#L407) doesn't return the content id yet. So if a key gets added, the developer/user has to manually compute the key id from the public key to find out where the key's stored. This is inconvenient and I'd prefer as a developer to just receive back the id of the key in particular and everything that I add in general, to have an easy way of tracking what I added to my wallet.
> 
> Non-key material can easily be queried but stored keys can't be queried. So for an application to keep track of the keys that the user added, it's important to receive back the id.
> 
> What do you think, shall we make this change to the `Add` function signature in this PR as well?

cc @sudeshrshetty, the wallet implementer. I believe Add() should return the `kid` as string as well. I don't see the Universal Wallet API imposing specific returns for the [Add function](https://w3c-ccg.github.io/universal-wallet-interop-spec/#add) so it should be able to return `kid`. @sudeshrshetty thoughts?

Baha-sk:
@jceb the pr looks good to me, regardless of the above discussion about returning the `kid` when adding a key to the wallet.

cc @troyronda @fqutishat @sudeshrshetty

pkg/wallet/wallet.go
```diff
@@ -986,14 +988,29 @@ func (c *Wallet) validateProofOption(authToken string, opts *ProofOptions, metho
...
 		if opts.VerificationMethod == vm.VerificationMethod.ID {
+			switch vm.VerificationMethod.Type {
+			case verifiablec.Ed25519VerificationKey2018:
+			case verifiablec.Bls12381G2Key2020:
+				kt = kms.BLS12381G2Type
+			case verifiablec.JSONWebKey2020:
+				kt = verifiablec.KmsKeyTypeByJWKCurve(vm.VerificationMethod.JSONWebKey().Crv)
+			}
+
+			kid, err := localkms.CreateKID(vm.VerificationMethod.Value, kt)
```
sudeshrshetty, line 1007: KID shouldn't be computed here in wallet and shouldn't replace the option KID. If In case client wants to use custom KID then it should be computed outside and passed as an option.

Main reason for this is that, different DID method implementations use their own way of preparing key IDs. 
Creating and using  our own key IDs will break the interop in this wallet. I guess yet there is no standard for creating key IDs in DIDs .

Imagine a case where someone configured VCWallet with their own VDR and remote KMS (webkms), then this change will force remote kms implementation to use this `localkms.CreateKID(...)` logic to create their key IDs.

You can refer [this test in trustbloc](https://github.com/trustbloc/agent-sdk/blob/main/cmd/wallet-js-sdk/test/specs/credential/credential-manager.spec.js#L114-L145) where `wallet.CreateKeyPair()` is used to create keys for a DID to issue/verify credentials.

We are yet to add bddtests for vcwallet in aries. I would love to have chat if you want to discuss more about this.

pkg/wallet/models.go
```diff
@@ -30,6 +30,7 @@ type QueryParams struct {
 // To be used as options for issue/prove wallet features.
 //
 type ProofOptions struct {
+	KID string `json:"kid,omitempty"`
```
sudeshrshetty, line 33: I am ok to pass this as an option for generating proofs, but it shouldn't get replaced internally by wallet. That will break the interoperability.

sudeshrshetty (changes requested):

sudeshrshetty:
@jceb @Baha-sk would love to get on a call and solve this issue quickly.

jceb:
Let me know when it suits you. I'm in the CET timezone. You can send me an invite to jan-christoph.ebersbach@identinet.io

sudeshrshetty:
> Let me know when it suits you. I'm in the CET timezone. You can send me an invite to [jan-christoph.ebersbach@identinet.io](mailto:jan-christoph.ebersbach@identinet.io)

thanks for the time @Baha-sk @jceb 

here are the outcomes of our discussion:

- in case of issue if `KID` is provided by client then key will be directly fetched using that key ID and verification method splitting logic will be skipped 
- verification method skipping logic will have fallback in case of KEY NOT FOUND and wallet will try to fetch key using fingerprint as you mentioned
-`vcwallet.Add(key)` to accept custom key ID or compute one if missing (need update in universal wallet spec)

jceb:
@sudeshrshetty @Baha-sk the PR is complete from my side. Thank you for your help!

jceb:
Hi @sudeshrshetty 

> here are the outcomes of our discussion:
> 
>     * in case of signing, if `KID` is provided by client then key will be fetched from KMS using that key ID and verification method splitting logic will be skipped .

👍 has been implemented.

>     * verification method splitting logic will have fallback in case of KEY NOT FOUND and wallet will try to fetch key again using fingerprint as you mentioned.

I'm not sure about the refetch logic. At the moment the key ID that's provided or if none is provided, it's computed, is used.

>     * `vcwallet.Add(key)` to accept custom key ID or compute one if missing (need update in universal wallet spec).

From analyzing the code I learned that the Universal Wallet spec lets us add a key. In the content of the key there's an ID field. With this in place, I don't see the need to extend the specification because we can use that field. The current implementation either stores the key under the provided ID or, if ID is an empty string, the key ID is computed.

Is there anything else that you'd like me to change in the PR?

jceb:
@sudeshrshetty @Baha-sk happy new year to you! Is there anything else I can do to get the PR merged?

jceb:
@Baha-sk @sudeshrshetty I updated the patch to the version in main and extracted #3222 as it's not directly related. How could we get the patch in shape to be merged into main?

fastman61:
@jceb, @Baha-sk , @sudeshrshetty

How can I solve this issue? Will the PR be merged? Is there a plan?

When I merged this PR with solving merge conflicts to my local main branch, It gives the error below. What can I do to solve this issue using BbsBlsSignature2020 signature?
```
{
    "code": 12010,
    "message": "failed to issue credential: getKeySet: failed to read json keyset from reader: keyset.Handle: decryption failed: cipher: message authentication failed"
}
``` 
```
/vcwallet/issue
{
    "auth": "{{wallet_token}}",
    "credential": {
        "@context": [
            "https://w3id.org/citizenship/v1"
        ],
        "type": [
            "VerifiableCredential",
            "PermanentResident"
        ],
        "id": "https://credential.example.com/residents/1234567890",
        "issuer": "did:key:zUC74hLbUrGUFKyVb3ehRU3rvpeb7HVYrCEQsbki5B18r5K4X2iKxBgygjbLstfLa7ZNLoQrsWxUcNJpjSBD5fDbXgyrm9opUPieU6dP98HGg1uKhq4yUNCq8gp4a363BzVUEcx",
        "issuanceDate": "2020-01-01T12:00:00Z",
        "credentialSubject": {
            "type": [
                "PermanentResident"
            ],
            "id": "did:peer:1zQmSiUedvbpDzi245g1wWXPz97kaBbRS6BBALW9DnP5Nk7Z",
            "givenName": "ALICE",
            "familyName": "SMITH",
            "gender": "Female",
            "birthCountry": "Bahamas",
            "birthDate": "1958-07-17"
        }
    },
    "proofOptions": {
        "controller": "did:key:zUC74hLbUrGUFKyVb3ehRU3rvpeb7HVYrCEQsbki5B18r5K4X2iKxBgygjbLstfLa7ZNLoQrsWxUcNJpjSBD5fDbXgyrm9opUPieU6dP98HGg1uKhq4yUNCq8gp4a363BzVUEcx",
        "proofRepresentation": 0,
        "proofType": "BbsBlsSignature2020"
    },
    "userID": "{{wallet_user_id}}"
}
```

jceb:
Not sure, I haven't tried it but BLS signatures yet. Not sure either why this PR isn't going forward.

PR #3057, opened 2021-11-15
fix: allow payloads of any size in hkdf and pbkdf2 encrypt

aholovko:
Signed-off-by: Andrii Holovko <andriy.holovko@gmail.com>


Baha-sk (approved):
The plaintext fed to Encrypt() does not need to be limited in size. Looks good.

fqutishat (approved):

PR #3064, opened 2021-11-19
feat: didcomm v2 DID rotation

Moopli:

### DIFF
--- a/test/bdd/pkg/connection/connection_sdk_steps.go
+++ b/test/bdd/pkg/connection/connection_sdk_steps.go
@@ -1,0 +1,145 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package connection
+
+import (
+	"fmt"
+
+	"github.com/cucumber/godog"
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/client/connection"
+	"github.com/hyperledger/aries-framework-go/pkg/client/didexchange"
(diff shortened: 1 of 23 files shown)


fqutishat (approved):

test/bdd/pkg/connection/connection_sdk_steps.go
```diff
@@ -0,0 +1,145 @@
...
+
+	myDoc := s.bddContext.PublicDIDDocs[agentID]
+	theirDoc := s.bddContext.PublicDIDDocs[otherAgent]
+
+	didExClient, err := didexchange.New(agentCtx)
+	if err != nil {
+		return err
+	}
+
+	conns, err := didExClient.QueryConnections(&didexchange.QueryConnectionsParams{
```
Baha-sk, line 58: didcommv2 does not require connections, in fact oobv2 is connectionless but target service calls can then create a connection.

maybe as an improvement: add a case of did rotation without a connection? to be discussed if it makes sense..

Baha-sk (approved):
LGTM,

I've added a comment that needs clarification around DID rotations and connections

test/bdd/pkg/connection/connection_sdk_steps.go
Moopli: Not sure this works, we still create and save a connection record when we do a didcomm v2 communication, and various places in the code expect a connection record to exist. If we get rid of the connection record, then did rotation will still need some persistence, since it has to add the rotation JWS to outbound messages until it receives a message sent to the new DID (which acknowledges the rotation), and that requires persistent state.

PR #3084, opened 2021-12-03
feat: Added option for optimizing storage batch put operations

DRK3:
- Added the ability to set an option on a put operation that indicates that the key being used does not already exist in the database. For some storage providers, knowing this beforehand allows them to operate faster.
- Also updated some other documentation in the storage interface to make it explicit that data is normally overwritten if an existing key is being used.

### DIFF
--- a/spi/storage/storage.go
+++ b/spi/storage/storage.go
@@ -116,11 +116,21 @@ type Tag struct {
 	Value string `json:"value,omitempty"`
 }
 
+// PutOptions represents options for a Put Operation.
+type PutOptions struct {
+	// This is an optimization for a Put Operation. Some storage providers may be able to store data faster if they
+	// know beforehand that this key does not currently exist in the database. Unexpected behaviour may occur if
+	// this is set to true and the key already exists. See the documentation for the specific storage provider to
+	// see if and how this option is used.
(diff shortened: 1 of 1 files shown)


spi/storage/storage.go
```diff
@@ -116,11 +116,21 @@ type Tag struct {
...
+
 // Operation represents an operation to be performed in the Batch method.
 type Operation struct {
-	Key   string `json:"key,omitempty"`
-	Value []byte `json:"value,omitempty"` // A nil value will result in a delete operation.
-	Tags  []Tag  `json:"tags,omitempty"`  // Optional.
+	Key        string     `json:"key,omitempty"`
+	Value      []byte     `json:"value,omitempty"`      // A nil value will result in a delete operation.
+	Tags       []Tag      `json:"tags,omitempty"`       // Optional.
+	PutOptions PutOptions `json:"putOptions,omitempty"` // Optional. Only used for Put Operations.
```
bstasyszyn, line 133: I wonder if it would be better to use a pointer here?

DRK3: Yes - done!

fqutishat (approved):

bstasyszyn (approved):

Issue #3083, opened 2021-12-03
Cannot register a new message service on mobile


adn-misa:
**What I'm trying to do**

I'm trying to register a new message service on mobile.

**Expected result**

It should successfully register the new message service.
The same call over REST is working correctly.

**Actual result**

The library crashes due to a null pointer reference after executing the following code:

```kotlin
val messagingController = ariesAgent.messagingController
val res = messagingController.registerService(RequestEnvelope("{\"name\":\"basic-message\", \"purpose\":[\"chat\"], \"type\":\"https://didcomm.org/basicmessage/2.0\"}".toByteArray(Charsets.UTF_8)))
```

Stacktrace:

```
E/Go: panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x7310a15e20]
    goroutine 17 [running, locked to thread]:
    github.com/hyperledger/aries-framework-go/pkg/client/messaging.(*Client).RegisterService(0x4000034180, 0x4000504130, 0xd, 0x40000bc120, 0x2c, 0x40003ee140, 0x1, 0x4, 0x4000422b50, 0x400037ac00)
    	aries-framework-go/pkg/client/messaging/client.go:143 +0xd0
    github.com/hyperledger/aries-framework-go/pkg/controller/command/messaging.(*Command).registerMessageService(0x4000010030, 0x40003ee100, 0x40003ee100, 0x0)
E/Go: 	aries-framework-go/pkg/controller/command/messaging/command.go:283 +0x290
    github.com/hyperledger/aries-framework-go/pkg/controller/command/messaging.(*Command).RegisterService(0x4000010030, 0x7310d1e308, 0x40003ec120, 0x7310d1e328, 0x40003ec150, 0x7310c7c4a0, 0x400029c001)
    	aries-framework-go/pkg/controller/command/messaging/command.go:125 +0x2ac
    github.com/hyperledger/aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/command.exec(0x40000322d0, 0x7310c7f920, 0x40003ee0c0, 0x40003ee0c0, 0x40002983d0, 0x0, 0x0)
    	aries-framework-go/cmd/aries-agent-mobile/pkg/wrappers/command/command.go:27 +0x188
[...]

PR #3086, opened 2021-12-06
fix: Add the missing message handler for the messaging service on mobile

adn-misa:
**Title:**
Mobile implementations could not register a new message service.

**Description:**
Related issue: #3083.

Fixed the issue by creating the `Registrar` on a higher level and passed it as part of the mobile wrapped options to the required functions.

**Summary:**
There was a missing message handler on mobile that did cause a null dereferencing error when trying to register a new message service


fqutishat (approved):

sudeshrshetty (commented):
Looks good to me but please squash all your commits before merging it

PR #3102, opened 2021-12-10
feat: Method to add Credential Fulfillment to Verifiable Presentation

DRK3:


pkg/doc/cm/credentialfulfillment.go
```diff
@@ -103,6 +104,41 @@ func (cf *CredentialFulfillment) validate() error {
...
 
+// AddCredentialFulfillmentToPresentation takes the given presentation and adds Credential Fulfillment data based on
+// credentialManifest. This method should only be called on a presentation without Credential Fulfillment data.
+// Note the following assumptions/limitations of this method:
+// 1. The format of all credentials is assumed to be ldp_vc.
+// 2. The location of the Verifiable Credentials is assumed to be an array at the root under a field called
+//    "verifiableCredential".
+// 3. The Verifiable Credentials in the presentation is assumed to be in the same order as the Output Descriptors in
+//    the Credential Manifest.
+func AddCredentialFulfillmentToPresentation(presentation *verifiable.Presentation,
```
sudeshrshetty, line 115: I would suggest renaming it to `PresentCredentialFulfillment` to follow naming convention of other packages doing similar feature.

If Input presentation is nil then build one instead of panicing (you can do it as followup PR if you like)

DRK3: Done! Made other changes per our earlier conversation too

sudeshrshetty (approved):

PR #3104, opened 2021-12-13
chore: upgrade to Go 1.17 and fix MacOS Monterey Go Bug

Baha-sk:


cmd/aries-agent-mobile/go.mod
```diff
@@ -17,9 +17,56 @@ require (
...
+	golang.org/x/crypto v0.0.0-20210616213533-5ff15b29337e // indirect
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
+	google.golang.org/protobuf v1.27.1 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+)
+
 replace (
 	github.com/hyperledger/aries-framework-go => ../../
-	github.com/hyperledger/aries-framework-go/component/storage/edv => ../../component/storage/edv // TODO (#2815) remove this once the wallet package doesn't import edv
+	// github.com/hyperledger/aries-framework-go/component/storage/edv => ../../component/storage/edv // TODO (#2815) remove this once the wallet package doesn't import edv
```
Baha-sk, line 69: @DRK3 this was causing compilation errors in Go 1.17 (was complaining about aries-agent-mobile not using the correct edv revision). Tests seem to pass, not sure the impact on dependant projects though.

cmd/aries-agent-rest/go.mod
```diff
@@ -4,25 +4,73 @@
...
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
+	google.golang.org/protobuf v1.27.1 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	nhooyr.io/websocket v1.8.3 // indirect
+)
+
 replace (
 	github.com/hyperledger/aries-framework-go => ../..
-	github.com/hyperledger/aries-framework-go/component/storage/edv => ../../component/storage/edv // TODO (#2815) remove this once the wallet package doesn't import edv
+	//	github.com/hyperledger/aries-framework-go/component/storage/edv => ../../component/storage/edv // TODO (#2815) remove this once the wallet package doesn't import edv
```
Baha-sk, line 73: @DRK3 the same here

cmd/aries-agent-mobile/go.mod
DRK3: I think we'll still need this replace. I think you can resolve those errors about the `edv` revision by running `go mod tidy` in the various go modules - give that a try and let me know if it works

Baha-sk: I did, Go 1.17 adds those "indirect" entries in a new section in go.mod and doesn't seem to want to resolve edv dep instead of doing those replaces. Not sure why.

DRK3: I'm having trouble remembering exactly why we had to use the replace... had something to do with the circular dependency and trying to update one of the modules.
@Baha-sk Do you know if this issue happens with other `replace` statements as well? I wonder if we'll run into the issue elsewhere as well...


(Really what we should do is resolve #2815... perhaps @sudeshrshetty or I could work on it if this ends up being a blocker)

DRK3 (approved):

fqutishat (approved):

PR #3105, opened 2021-12-13
refactor: update goal_code invalid name for OOB V2

Baha-sk:

### DIFF
--- a/pkg/wallet/invitation.go
+++ b/pkg/wallet/invitation.go
@@ -38,7 +38,7 @@ type GenericInvitation struct {
 	Type      string                        `json:"type"`
 	Label     string                        `json:"label,omitempty"`
 	Goal      string                        `json:"goal,omitempty"`
-	GoalCode  string                        `json:"goal_code,omitempty"`
+	GoalCode  string                        `json:"goal-code,omitempty"`
 	Services  []interface{}                 `json:"services"`
 	Accept    []string                      `json:"accept,omitempty"`
 	Protocols []string                      `json:"handshake_protocols,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/wallet/invitation.go
```diff
@@ -38,7 +38,7 @@ type GenericInvitation struct {
 	Type      string                        `json:"type"`
 	Label     string                        `json:"label,omitempty"`
 	Goal      string                        `json:"goal,omitempty"`
-	GoalCode  string                        `json:"goal_code,omitempty"`
+	GoalCode  string                        `json:"goal-code,omitempty"`
```
sudeshrshetty, line 41: This one will break current DIDComm V1 support for wallet

sudeshrshetty (changes requested):

pkg/wallet/invitation.go
Baha-sk: Then we need 2 distinct models of `GenericInvitation` and `rawInvitation` and separate V1 from V2

Baha-sk: actually, this doesn't break V1 support, notice ID and Type above are V2 as well. The logic of flipping between V1 and V2 is used elsewhere. No additional change is needed here.

sudeshrshetty: but it is used as request model in command controller here. It comes down to problem we discussed yesterday where command controller API doc shows different model than the one command is expecting.

https://github.com/hyperledger/aries-framework-go/blob/main/pkg/controller/command/vcwallet/models.go#L351

Created [issue](https://github.com/hyperledger/aries-framework-go/issues/3106) for that, and this one can be merged, but need to keep an eye.

sudeshrshetty (approved):

PR #3108, opened 2021-12-15
refactor: move oobv2 'From' field up from attachment

Baha-sk:
As per the DIDComm V2 spec, the From field should be in the oobv2 struct directly, not in the attachment.

### DIFF
--- a/pkg/didcomm/protocol/outofbandv2/service.go
+++ b/pkg/didcomm/protocol/outofbandv2/service.go
@@ -194,12 +194,16 @@ func (s *Service) HandleOutbound(_ service.DIDCommMsg, _, _ string) (string, err
 
 // AcceptInvitation from another agent.
 //nolint:funlen,gocyclo
-func (s *Service) AcceptInvitation(i *Invitation) (string, error) { // nolint: gocognit
+func (s *Service) AcceptInvitation(i *Invitation) (string, error) {
 	msg := service.NewDIDCommMsgMap(i)
 
 	err := validateInvitationAcceptance(msg, s.myMediaTypeProfiles)
 	if err != nil {
 		return "", fmt.Errorf("oob/2.0 unable to accept invitation: %w", err)
+	}
+
+	if i.From == "" {
+		return "", fmt.Errorf("oob/2.0 does not have from field")
 	}
 
 	clbk := &callback{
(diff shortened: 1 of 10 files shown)


pkg/didcomm/protocol/outofbandv2/service.go
```diff
@@ -242,76 +246,46 @@ func (s *Service) AcceptInvitation(i *Invitation) (string, error) { // nolint: g
...
+		return "", fmt.Errorf("oob/2.0 failed to create destination: %w", err)
+	}
 
-		if err := s.connectionRecorder.SaveConnectionRecord(connRecord); err != nil {
-			return "", err
-		}
+	connRecord := &connection.Record{
+		ConnectionID:      uuid.New().String(),
+		ParentThreadID:    i.ID,
+		State:             "null",
```
Moopli, line 267: I was wondering, why do you set state "null"? This confuses the `connection.Recorder`, which only saves the record with certain metadata (like, with all the tags needed for lookups elsewhere in the code) when the state is 'completed'.

Baha-sk: updated to `connection.StateNameCompleted`

Moopli: (resolved now)

Moopli (approved):
lgtm :+1:

pkg/didcomm/protocol/outofbandv2/service.go
```diff
@@ -242,76 +246,46 @@ func (s *Service) AcceptInvitation(i *Invitation) (string, error) { // nolint: g
...
+	connRecord := &connection.Record{
+		ConnectionID:      uuid.New().String(),
+		ParentThreadID:    i.ID,
+		State:             connection.StateNameCompleted,
+		InvitationID:      i.ID,
+		ServiceEndPoint:   destination.ServiceEndpoint,
+		RecipientKeys:     destination.RecipientKeys,
+		RoutingKeys:       destination.RoutingKeys,
+		TheirLabel:        i.Label,
+		TheirDID:          i.From,
```
Moopli, line 273: ah wait, don't we also need to set MyDID in the record?

Baha-sk: technically it's the same as `InvitationDID` further below... but i've added MyDID for the sake of completeness.

Moopli: :+1: yeah it's a bit of cruft from the record being used in different ways in didexchange vs elsewhere
but `MyDID` is needed in a lot of places, while `InvitationDID` is mostly ignored.

Moopli (approved):

fqutishat (approved):

PR #3138, opened 2022-01-24
test: Enhancements to WACI issuance BDD tests

DRK3:
- Attachments are now generated on the fly using pre-defined Credential Manifest, Credential Application, and Credential Fulfillment objects.
- Added constants to Credential Manifest package for attachment formats and contexts.


test/bdd/pkg/waci/waci_issuance_sdk_steps.go
```diff
@@ -70,6 +81,7 @@ func (i *IssuanceSDKSteps) SetContext(ctx *context.BDDContext) {
 }
 
 // RegisterSteps registers the BDD test steps on the suite.
+// Note that VC proofs are not checked in this test suite.
```
DRK3, line 84: Should I also check VC proofs? (if so, I can do that in my next PR) @sudeshrshetty @rolsonquadras

test/bdd/pkg/waci/waci_issuance_sdk_steps.go
```diff
@@ -699,6 +674,275 @@ type prop interface {
...
+		ID:        uuid.New().String(),
+		MediaType: "application/json",
+		Format:    cm.CredentialFulfillmentAttachmentFormat,
+		Data:      decorator.AttachmentData{JSON: attachmentData},
+	}
+
+	return &credentialFulfillmentAttachment, nil
+}
+
+func generateCredentialFulfillmentProof() map[string]string {
```
DRK3, line 922: Not sure if this is the best way to do this... let me know if there's something else I should do instead. Also let me know if `did:orb` is ok here.

test/bdd/pkg/waci/credential_application_drivers_license.json
```diff
@@ -0,0 +1,12 @@
+{
```
Moopli, line 1: is it possible to move these files (and the others in the `waci` folder) to a `testdata` subdirectory like elsewhere in `test/bdd/pkg` ?

DRK3: Done!

Moopli (approved):

test/bdd/pkg/waci/waci_issuance_sdk_steps.go
sudeshrshetty: that's ok, but nice to have feature. If it can be quick then do it.

sudeshrshetty (approved):

PR #3152, opened 2022-02-03
feat: support for websocket read limit

### LINKED ISSUE #3147: WebSocket - support to pass read data size limit as framework opts
AFG uses [`nhooyr.io/websocket`](https://pkg.go.dev/nhooyr.io/websocket) library to support WebSocket flows. The default data size for reading is set to 32KB. If the data size is more than this, the socket gets closed and AFG throws an error. 

The [`SetReadLimit`](https://pkg.go.dev/nhooyr.io/websocket#Conn.SetReadLimit) function in the library provides support to override this default limit. AFG needs to expose this option to the client when configuring the framework ([aries.New(opts)](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/framework/aries/framework.go#L92)).

aholovko:
Closes #3147

Signed-off-by: Andrii Holovko <andriy.holovko@gmail.com>

### DIFF
--- a/pkg/framework/aries/defaults/defaults.go
+++ b/pkg/framework/aries/defaults/defaults.go
@@ -26,10 +26,16 @@ func WithInboundHTTPAddr(internalAddr, externalAddr, certFile, keyFile string) a
 	}
 }
 
-// WithInboundWSAddr return new default ws inbound transport.
-func WithInboundWSAddr(internalAddr, externalAddr, certFile, keyFile string) aries.Option {
+// WithInboundWSAddr return new default ws inbound transport. If readLimit is 0, the default value of 32kB is set.
+func WithInboundWSAddr(internalAddr, externalAddr, certFile, keyFile string, readLimit int64) aries.Option {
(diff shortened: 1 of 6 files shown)


pkg/framework/aries/defaults/defaults.go
```diff
@@ -27,9 +27,15 @@ func WithInboundHTTPAddr(internalAddr, externalAddr, certFile, keyFile string) a
 }
 
 // WithInboundWSAddr return new default ws inbound transport.
-func WithInboundWSAddr(internalAddr, externalAddr, certFile, keyFile string) aries.Option {
+func WithInboundWSAddr(internalAddr, externalAddr, certFile, keyFile string, readLimit int64) aries.Option {
```
rolsonquadras, line 30: it would be better to add a comment here about the default value when readLimit is 0. Or may be we can add another option function with readLimit.

sudeshrshetty (approved):

rolsonquadras (approved):

PR #3183, opened 2022-02-28
Size computation for allocation may overflow

bhaskarvilles:
Performing calculations involving the size of potentially large strings or slices can result in an overflow (for signed integer types) or a wraparound (for unsigned types). An overflow causes the result of the calculation to become negative, while a wraparound results in a small (positive) number.

This can cause further issues. If, for example, the result is then used in an allocation, it will cause a runtime panic if it is negative, and allocate an unexpectedly small buffer otherwise.

Recommendation:
Always guard against overflow in arithmetic operations involving potentially large numbers by doing one of the following:

Validate the size of the data from which the numbers are computed.
Define a guard on the arithmetic expression, so that the operation is performed only if the result can be known to be less than, or equal to, the maximum value for the type.
Use a wider type (such as uint64 instead of int), so that larger input values do not cause overflow.

### DIFF
--- a/pkg/vdr/fingerprint/fingerprint.go
+++ b/pkg/vdr/fingerprint/fingerprint.go
@@ -5,14 +5,17 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 package fingerprint
+package main
 
 import (
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/elliptic"
 	"encoding/binary"
+	"encoding/json"
 	"errors"
 	"fmt"
+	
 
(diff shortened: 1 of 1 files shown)


sudeshrshetty:
@bhaskarvilles please suggest/replace the existing code which you think is having problem and also please add unit tests for the better coverage of new function you added.

Also, please refer `required` [guidelines](https://github.com/hyperledger/aries-framework-go/blob/cea4427c93a6c0bed602c253f5e577448c4f88df/.github/CONTRIBUTING.md#pull-request) for creating a pull request.

Issue #3175, opened 2022-02-22
EDV/KMS improve perf by combining KeyWrap and HMAC rest calls to the KMs server


Baha-sk:
Currently, when formatting a new EDV doc, there are at least 3 calls to KMS for several operations (and 2 extra calls for each tag created). These calls are made mainly for HMAC to generate Doc/Tag IDs and KeyWrapping for encrypting the doc. They are all separate RESTful calls to the KMS server. 

In the case of multiple doc creations, these calls are multiplied for each doc created. 

To improve user experience, combine all these RESTfull calls into 1 and have the KMS server offer a corresponding endpoint returning multiple crypto results in the response.

Baha-sk:
closing this issue as AFGO changes to combin crypto operations between KMS/Crypto/JWEEncrypter is not a valid solution. These operations should not be combined.

PR #3189, opened 2022-03-02
feat: add BatchCrypto support in EDV

Baha-sk:
This change includes EDV batch Crypto computation to help support combining all MAC/Encryption/KW operations in one 1 call

Part of #3175


component/storage/edv/encryptedformatter.go
```diff
@@ -126,18 +142,46 @@ func (e *EncryptedFormatter) format(keyAndTagPrefix, key string, value []byte, t
 
 	var formattedValue []byte
 
-	if documentID != "" {
```
DRK3, line 129: Looks like you removed the `if documentID != ""` check - was that intentional? I can't remember why I put that there in the first place, so if that was an intentional change we should discuss briefly just to make sure it doesn't break something

Baha-sk:
```
	if e.useDeterministicDocumentIDs {
		documentID, err = e.generateDeterministicDocumentID(keyAndTagPrefix, key)
		if err != nil {
			return "", nil, nil, fmt.Errorf("failed to format key into an encrypted document ID: %w", err)
		}
	} else {
		documentID, err = generateRandomDocumentID()
		if err != nil {
			return "", nil, nil, fmt.Errorf("failed to generate EDV-compatible ID: %w", err)
		}
	}
```

above will ensure `documentID` is not empty, unless one of those calls return an empty value, which is not the case.

DRK3: Makes sense!

DRK3 (approved):

component/storage/edv/encryptedformatter.go
Baha-sk: reverted this change to minimize impact of this pr

Baha-sk:
> I see [scripts/start_edv_test_docker_images.sh](https://github.com/hyperledger/aries-framework-go/pull/3189/files#diff-e924216c475870598d056b7998f70d374bce2323a280fe61813390a55f9b9f80) in the changelist but no actual line changes in the diff - was something actually changed?

actually, it's just security access to this file to be executable, i just made it accessible for local execution, current mode was read/write only, now it's also executable. Notice the `chomd` mode udpate: `100644 → 100755`

component/storage/edv/encryptedformatter.go
DRK3: Actually on closer look, `generateDeterministicDocumentID` can return a blank ID, so yeah reverting is the right thing to do

fqutishat (approved):

PR #3224, opened 2022-04-20
Draft: Port Aries Agent to AWS Lambda

borancar:
**Title:**
Port Aries Agent to AWS Lambda

**Description:**
Port the Agent to run on AWS Lambda, using APIGW and DynamoDB as the supporting infrastructure for transport and storage, respectively.

**Summary:**
The PR includes both a Go application and the Serverless + Serverless Offline framework in order to deploy and run locally for debug purposes. Yarn is used to track the Serverless part of the stack.
The Go application is the agent itself, build to run on AWS Lambda. The main communication with Lambda is handled via github.com/aws/aws-lambda-go/lambda.It uses DynamoDB as both persistent and temporary storage since Lambda runs are ephemeral. The DIDComm transports are handled using APIGW, with WebSockets requiring a bit more dedicated steps. One issue we've identified here is the main app finishing sooner and Lambda terminating the app before auto-approve has had a chance to respond (this only seems to affect the deployed app and does not affect the locally running Serverles Offline app). The temporary hacks we've put in place add a 5 second delay and that seems to work most of the time. This is only a draft for now, as we'd like help on clearing that (perhaps by using WaitGroups or another synchronization mechanism?).


cmd/aries-agent-lambda/aries.go
```diff
@@ -0,0 +1,387 @@
...
+	autoAccept                                     bool
+	msgHandler                                     command.MessageHandler
+	dbParam                                        *dbParam
+	autoExecuteRFC0593                             bool
+}
+
+const (
+	httpProtocol      = "http"
+	websocketProtocol = "ws"
+	defaultWebhookUrl = "https://sly.slygo.scaffoldly.dev/aries-webhooks/api/v1/webhooks"
```
borancar, line 59: We've noticed we needed a webhook, otherwise we wouldn't get didcomm working. Can you point us to an easy way of avoiding the need for a webhook?

cmd/aries-agent-lambda/aries.go
```diff
@@ -0,0 +1,387 @@
...
+	internalHttpHost := httpHost
+	// internalWsHost := wsHost
+	basePath := os.Getenv("SERVICE_SLUG")
+	webhookPath := fmt.Sprintf("/%s/%s", basePath, "webhook")
+	if stage == "local" {
+		httpProto = "http"
+		wsProto = "ws"
+		httpHost = "localhost:3000"
+		wsHost = "localhost:3001"
+		internalHttpHost = "host.docker.internal:3000"
```
borancar, line 101: We needed this hack at the time as Linux uses a proper bridge vs other platforms proxying on localhost. Possibly `serverless-offline` has mitigated the issue in the meantime.

cmd/aries-agent-lambda/aries.go
```diff
@@ -0,0 +1,387 @@
...
+		// server:               server,
+		host: httpHost,
+		// token:                token,
+		inboundHostInternals: []string{inboundWsHost},
+		inboundHostExternals: []string{inboundHostWsExternalUrl},
+		dbParam:              dbParam,
+		// defaultLabel:         defaultLabel,
+		webhookURLs: webhookUrls,
+		// httpResolvers:        httpResolvers,
+		outboundTransports: []string{"ws"},
```
borancar, line 126: We've noticed invitations returning the http path, if we had `[]string{"ws", "http"}` here.

cmd/aries-agent-lambda/aries.go
```diff
@@ -0,0 +1,387 @@
...
+	return router, nil
+}
+
+func createAriesAgent(parameters *agentParameters) (*ariescontext.Provider, error) {
+	var opts []aries.Option
+
+	storePro := dynamodb.NewProvider() // TODO Other Storage Providers
+
+	opts = append(opts, aries.WithStoreProvider(storePro))
+	opts = append(opts, aries.WithProtocolStateStoreProvider(storePro))
```
borancar, line 176: We're thinking of pushing this to Redis, or other type of temporary storage. For now, DynamoDB does make it work nicely.

cmd/aries-agent-lambda/gorillamux.go
```diff
@@ -0,0 +1,129 @@
+/*
```
borancar, line 1: We started with https://github.com/awslabs/aws-lambda-go-api-proxy, but we think we can cut out gorillamux in favor of the standard library, to make it simpler.

cmd/aries-agent-lambda/storage/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,562 @@
...
+
+	// TODO
+
+	fmt.Printf("Querying %s\n", expression)
+
+	if expression == "" {
+		return nil, errInvalidQueryExpressionFormat
+	}
+
+	expr, err := dbexpr.NewBuilder().WithFilter(dbexpr.Name("Tags").Contains(expression)).Build()
```
borancar, line 387: Perhaps imperfect (would need tweaks not to get substrings, like `id:bet` maching `id:better`, but it was performing nicely in our tests).

cmd/aries-agent-lambda/transport/apigw_ws/inbound.go
```diff
@@ -0,0 +1,175 @@
...
+			getConnPool().addKey(unpackMsg, trans, connectionId, respChannel)
+
+			err = prov.InboundMessageHandler()(unpackMsg)
+			if err != nil {
+				log.Fatalf("incoming msg processing failed: %v", err)
+				return
+			}
+
+			select {
+			//case response := <-respChannel:
```
borancar, line 126: Our original thinking here was to try and deliver the autoapprove reply via channels (wait on the reply or timeout after 5 seconds). This didn't quite work as we expected, so we just reverted to having a timeout by default.

cmd/aries-agent-lambda/transport/apigw_ws/inbound.go
```diff
@@ -0,0 +1,175 @@
...
+			//	w.WriteHeader(http.StatusOK)
+			//	log.Printf(">>>>>>>>> Got response\n")
+			//	w.Write(response)
+			case <-time.After(5 * time.Second):
```
borancar, line 130: We needed this, otherwise auto-approve would never send the reply back (Lambda would terminate the app once the main function finished, and waiting on all goroutines wasn't feasible as some weren't meant to terminate).

cmd/aries-agent-lambda/transport/apigw_ws/outbound.go
```diff
@@ -0,0 +1,153 @@
...
+
+	conn, channel, cleanup, err := cs.getConnection(destination)
+	defer cleanup()
+
+	if err != nil {
+		return "", fmt.Errorf("get websocket connection : %w\n", err)
+	}
+
+	if channel != nil {
+		//fmt.Printf(">>>>>>> Using Go channels\n")
```
borancar, line 85: Remnant from attempting to reply via Go channels.

cmd/aries-agent-lambda/serverless.yml
```diff
@@ -0,0 +1,389 @@
...
+
+resources:
+  Resources:
+    Secret:
+      Type: AWS::SecretsManager::Secret
+      Properties:
+        Name: lambda/${opt:stage, "local"}/${self:service}
+        SecretString: '{}'
+
+    CoordinateMediation:
```
borancar, line 75: These create the needed DyanamoDB tables. Rather than allowing the app higher privileges, we thought this as the better solution. It does mean that adding new tables has to be accompanied with changes here.

cmd/aries-agent-lambda/serverless.yml
```diff
@@ -0,0 +1,389 @@
...
+    timeout: 30
+    events:
+      - http:
+          path: /
+          method: any
+      - http:
+          path: /
+          method: options
+      - http:
+          path: /{proxy+}
```
borancar, line 54: These match arbitrary paths and forward them to the lambda handler.

cmd/aries-agent-lambda/aries.go
troyronda: @rolsonquadras @fqutishat @sudeshrshetty

cmd/aries-agent-lambda/aries.go
troyronda: @rolsonquadras @sudeshrshetty @fqutishat

cmd/aries-agent-lambda/aries.go
sudeshrshetty: since you are using aries REST controller, you need webhook to get notification of incoming protocol messages during various protocol execution.



If you don't want to host external webhooks then you can use `controller.WithNotifier()` instead of `controller.WithWebhookURLs()` to provide your own custom notifier.

sudeshrshetty: since you are using aries REST controller, you need webhook to get notification of incoming protocol messages during various protocol execution.



If you don't want to host external webhooks then you can use `controller.WithNotifier()` instead of `controller.WithWebhookURLs()` to provide your own custom notifier.

cmd/aries-agent-lambda/aries.go
sudeshrshetty: @borancar Are you talking about paths in service endpoints of invitations?

cmd/aries-agent-lambda/gorillamux.go
sudeshrshetty: As @rolsonquadras mentioned, once you move this to [AFG-Ext](https://github.com/hyperledger/aries-framework-go-ext) repo as an independent module, then you don't have to worry about using gorillamux.

cmd/aries-agent-lambda/transport/apigw_ws/inbound.go
sudeshrshetty: How are you writing response here?

borancar: We just let the `transport/apigw_ws/outbound.go` handle the response writing (which is basically the same behavior as default WebSockets have). So technically we'd just need a way of getting notified that:
1. Yes, there's a reply to be sent now as part of this run, and
2. When that reply has actually been sent, so we can resume the flow here.

cnuss:
@rolsonquadras @borancar I created a new PR on AFT (#3229) which externalizes some of the functionality of the REST client so we can use it in AFG-Ext

PR #3237, opened 2022-04-29
feat: Use new EDV add index endpoint when setting store config

### LINKED ISSUE #2492: New EDV REST provider implementation - store store config somewhere in the vault
EDV REST doesn't have any actual use of the store configuration currently. For consistency with other store implementations, though, it stores the store configuration in memory, but to truly be consistent it should be stored in persistent storage in the EDV server.

DRK3:
closes #2492

Also updated TrustBloc EDV image used for tests and removed some out-of-date comments.

### DIFF
--- a/component/storage/edv/restclient.go
+++ b/component/storage/edv/restclient.go
@@ -230,6 +230,32 @@ func (c *restClient) deleteDocument(vaultID, docID string) error {
 	return fmt.Errorf(failResponseFromEDVServer, statusCode, respBytes)
 }
 
+func (c *restClient) addIndex(vaultID string, attributeNames []string) error {
+	addIndexOperation := indexOperation{
+		Operation:      "add",
+		AttributeNames: attributeNames,
+	}
+
+	jsonToSend, err := json.Marshal(addIndexOperation)
+	if err != nil {
+		return fmt.Errorf("failed to marshal index operation: %w", err)
+	}
+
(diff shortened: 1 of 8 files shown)


component/storage/edv/restclient.go
```diff
@@ -230,6 +230,32 @@ func (c *restClient) deleteDocument(vaultID, docID string) error {
...
+	jsonToSend, err := json.Marshal(addIndexOperation)
+	if err != nil {
+		return fmt.Errorf("failed to marshal index operation: %w", err)
+	}
+
+	endpoint := fmt.Sprintf("%s/%s/index", c.edvServerURL, url.PathEscape(vaultID))
+
+	statusCode, _, respBytes, err := c.sendHTTPRequest(http.MethodPost, endpoint, jsonToSend, c.headersFunc)
+	if err != nil {
+		return err
```
bstasyszyn, line 248: It would be good to wrap the error with additional info, e.g.

`return fmt.Errorf("send HTTP request: %w", err)`

Sometimes when you look at an error in the log, you don't know which path the error took. :)

bstasyszyn (approved):

component/storage/edv/restclient.go
DRK3: Done!

fqutishat (approved):

PR #3239, opened 2022-05-09
refactor: set DIDcomm V1 VS V2 Endpoint support

### LINKED ISSUE #3219: Update DIDComm V2 Endpoint to include a URI, routingKeys and Accept fields
spec udpate ref: https://github.com/decentralized-identity/didcomm-messaging/pull/352

Baha-sk:
This change sets back the original DIDComm V1's Endpoint logic that was supperceded with the Endpoint struct update for DIDComm V2.
With this change, both ServiceEndpoint structures for V1 and V2 are not supported.

closes #3219


pkg/didcomm/dispatcher/outbound/outbound.go
```diff
@@ -158,16 +160,42 @@ func (o *Dispatcher) SendToDID(msg interface{}, myDID, theirDID string) error {
...
+			dest.MediaTypeProfiles = make([]string, len(connRec.MediaTypeProfiles))
+			copy(dest.MediaTypeProfiles, connRec.MediaTypeProfiles)
+		}
+	case transport.MediaTypeAIP2RFC0587Profile, transport.MediaTypeDIDCommV2Profile:
+		var (
+			url                 string
+			accept, routingKeys []string
+		)
+
+		url, err = connRec.ServiceEndPoint.URI()
```
fqutishat, line 179: You trying to use ServiceEndPoint from connRec but the problem is the ServiceEndPoint fields are not exported so when we store in connRec in db as JSON the values of ServiceEndPoint are gone

pkg/didcomm/protocol/didexchange/states.go
```diff
@@ -1006,14 +1085,21 @@ func (ctx *context) getServiceBlock(i *OOBInvitation) (*did.Service, error) {
 
 		err = decoder.Decode(svc)
 		if err != nil {
-			return nil, fmt.Errorf("failed to decode service block : %w", err)
+			// for DIDCommV2, decoder.Decode(svc) doesn't support serviceEndpoint as []interface{} representing an array
+			// for model.Endpoint. Manually build the endpoint here in this case.
+			if strings.Contains(err.Error(), "'serviceEndpoint' expected a map, got 'slice'") {
+				extractDIDCommV2EndpointIntoService(svc, s)
```
fqutishat, line 1091: you are passing s by value you need to pass it by reference because you are changing the s struct inside the method you passing

fqutishat (dismissed):

pkg/didcomm/dispatcher/outbound/outbound.go
Baha-sk: dest.ServiceEndpoint don't need to be updated as it's build from the DIDdocument. I reverted this change

pkg/didcomm/protocol/didexchange/states.go
Baha-sk: service is now passed as a reference to allow updating its `ServiceEndpoint` member field.

fqutishat (approved):

PR #3241, opened 2022-05-12
feat: add an option to disable required properties for CredentialParse.

vkubiv:

**Title:**
Add an option for CredentialParse to use customized vc default schema.

**Description:**
https://github.com/trustbloc/vcs/issues/725

**Summary:**

Now CredentialParse can use a customized vc default schema. 
CreateDefaultSchema function added to simplify the creation of custom default schema.

### DIFF
--- a/pkg/doc/verifiable/credential.go
+++ b/pkg/doc/verifiable/credential.go
@@ -26,14 +26,18 @@ import (
 
 var logger = log.New("aries-framework/doc/verifiable")
 
-// DefaultSchema describes default schema.
-const DefaultSchema = `{
+const (
+	schemaPropertyType              = "type"
+	schemaPropertyCredentialSubject = "credentialSubject"
+	schemaPropertyIssuer            = "issuer"
+	schemaPropertyIssuanceDate      = "issuanceDate"
+)
+
+// DefaultSchemaTemplate describes default schema.
+const DefaultSchemaTemplate = `{
   "required": [
-    "@context",
-    "type",
-    "credentialSubject",
-    "issuer",
-    "issuanceDate"
+    "@context"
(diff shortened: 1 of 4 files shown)


pkg/doc/verifiable/credential.go
```diff
@@ -1197,11 +1209,67 @@ func getSchemaLoader(schemas []TypedID, opts *credentialOpts) (gojsonschema.JSON
...
+
+// WithDisableRequiredField disabled check of required field in default schema.
+func WithDisableRequiredField(fieldName string) DefaultSchemaOpt {
+	return func(opts *defaultSchemaOpts) {
+		opts.disabledChecks = append(opts.disabledChecks, schemaPropertyIssuanceDate)
+	}
+}
+
+// CreateDefaultSchema creates default schema with the option to disable the check of specific properties.
+func CreateDefaultSchema(opts ...DefaultSchemaOpt) string {
```
rolsonquadras, line 1230: I am bit unsure about Default name here - default is already configured in the code base, this option is used by clients to override the those options. I feel, following names would be more concise
- CreateDefaultSchema - > JSONSchemaLoader : similar to JSON-LD schema loader https://github.com/hyperledger/aries-framework-go/blob/aa5f611f87a0b3f6c335ee8d6f2e699ac6509428/pkg/doc/verifiable/credential.go#L301

- DefaultSchemaOpt -> SchemaOpt : since we are overriding schema

https://github.com/hyperledger/aries-framework-go/blob/aa5f611f87a0b3f6c335ee8d6f2e699ac6509428/pkg/doc/verifiable/credential.go#L301

fqutishat (approved):

PR #3242, opened 2022-05-16
feat: Add multi-tag support to mem DB

bstasyszyn:
The mem DB now supports queries with multiple tags using the '&&' operator.

Signed-off-by: Bob Stasyszyn <Bob.Stasyszyn@securekey.com>

### DIFF
--- a/component/storageutil/mem/mem_test.go
+++ b/component/storageutil/mem/mem_test.go
@@ -67,3 +67,139 @@ func TestProvider_Ping(t *testing.T) {
 	err := provider.Ping()
 	require.NoError(t, err)
 }
+
+func TestMemStore_Query(t *testing.T) {
+	provider := mem.NewProvider()
+
+	store, err := provider.OpenStore("TestStore")
+	require.NoError(t, err)
+
+	const (
+		tagName1 = "TagName1"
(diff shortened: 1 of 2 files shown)


DRK3 (approved):

fqutishat (approved):

component/storageutil/mem/mem_test.go
```diff
@@ -67,3 +67,139 @@ func TestProvider_Ping(t *testing.T) {
...
+		require.NoError(t, err)
+		require.True(t, containsKey(key, key1, key4))
+
+		ok, err = iterator.Next()
+		require.NoError(t, err)
+		require.False(t, ok)
+	})
+
+	t.Run("Invalid value for tag -> no results", func(t *testing.T) {
+		iterator, err := store.Query("TagName1:TagValueX")
```
DRK3, line 188: I think this is a valid tag value, it's just that there happens to be no results. When I think of an "invalid" tag value, I think of one that contains a `:` character, for instance.

bstasyszyn: I guess I shouldn't have said 'invalid'. Instead, non-matching.

Issue #3260, opened 2022-06-10
VCWallet - Data Model Validation


sudeshrshetty:
Data models to be validated against their JSON LD context before being saved into the wallet content store. 

- Use JSON LD compact mechanism used for validating verifiable credentials. [Reference code.](https://github.com/hyperledger/aries-framework-go/blob/c06a49a517fc6609d538b0ae6fbd42716c586be3/pkg/doc/verifiable/jsonld.go#L27)
- Aries Data models like DID Documents are already validated by downstream APIs. For better performance avoid validating those data models.
  - As per current set of data models these require JSON LD context validation.
  - Collection
  - Connection
  - Key
  - Credential
  - Metadata
- Make JSON LD data model validation feature configurable so that a wallet implementation can choose to disable it over performance concerns or if they have pre-defined/validated data models.

PR #3261, opened 2022-06-13
feat: validate data model of objects saved in a wallet.

vkubiv:

**Title:**
VC Wallet Data Model validation

**Description:**
https://github.com/hyperledger/aries-framework-go/issues/3260

**Summary:**
Data models are validated against their JSON LD context before being saved into the wallet content store.


pkg/wallet/contents.go
```diff
@@ -178,13 +187,25 @@ func (cs *contentStore) Close() bool {
 // if content document id is missing from content, then system generated id will be used as key for storage.
 // returns error if content with same ID already exists in store.
 // For replacing already existing content, use 'Remove() + Add()'.
-func (cs *contentStore) Save(auth string, ct ContentType, content []byte, options ...AddContentOptions) error {
-	opts := &addContentOpts{}
+func (cs *contentStore) Save(auth string, ct ContentType, content []byte, options ...AddContentOptions) error { //nolint:lll,gocyclo
+	opts := &addContentOpts{
+		validateDataModel: true,
```
sudeshrshetty, line 192: by default validate data model should be disabled

pkg/wallet/options.go
```diff
@@ -281,6 +284,13 @@ func AddByCollection(collectionID string) AddContentOptions {
 	}
 }
 
+// WithDataModelValidation enables or disables data validations of adding content.
+func WithDataModelValidation(validateDataModel bool) AddContentOptions {
```
sudeshrshetty, line 288: In case of boolean option, you don't have to pass any argument. 
Follow [this](https://github.com/hyperledger/aries-framework-go/blob/d5971cedf7bac70db39048faf96c087ba1fbcc40/pkg/doc/verifiable/credential.go#L590-L594) example.

Something like `ValidateContent()`, by default it false and when someone uses this option then it becomes true.

pkg/wallet/contents.go
```diff
@@ -178,13 +187,20 @@ func (cs *contentStore) Close() bool {
...
 // For replacing already existing content, use 'Remove() + Add()'.
-func (cs *contentStore) Save(auth string, ct ContentType, content []byte, options ...AddContentOptions) error {
+func (cs *contentStore) Save(auth string, ct ContentType, content []byte, options ...AddContentOptions) error { //nolint:lll,gocyclo
 	opts := &addContentOpts{}
 
 	for _, option := range options {
 		option(opts)
 	}
 
+	if opts.validateDataModel && ct != DIDResolutionResponse {
```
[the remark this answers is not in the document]
vkubiv: Oh sorry, I did not add files with changes to git.

sudeshrshetty (approved):

PR #3264, opened 2022-06-27
refactor: separate didcomm from wallet.

vkubiv:

**Title:**
Separate DidComm code from Wallet

**Description:**
To make it possible to use Wallet with our DidComm functionality and required services, DidComm was separated from wallet


pkg/controller/command/vcwallet/command.go
```diff
@@ -888,7 +909,14 @@ func (o *Command) ProposeCredential(rw io.Writer, req io.Reader) command.Error {
 		return command.NewExecuteError(ProposeCredentialErrorCode, err)
 	}
 
-	msg, err := vcWallet.ProposeCredential(request.Auth, request.Invitation,
+	didComm, err := wallet.NewDidComm(vcWallet, o.ctx)
```
sudeshrshetty, line 912: __follow up task__: can you check if `wallet.NewDidComm()` is expensive in terms of store initialization, kms operaton etc? if yes then there will be performance implications initializing this instance everytime inside this wallet function.
Work around can be `client.NewWallet() //for regular wallet & client.NewDIDCommWallet() // for didcomm wallet`.

sudeshrshetty (approved):

pkg/controller/command/vcwallet/command.go
vkubiv: Only NewLookup service inside wallet.NewDidComm is accessing storage. It just calls SetStoreConfig, this shouldn't be expensive. And no KMS operations are called in NewDidComm.

fqutishat (approved):

PR #3272, opened 2022-07-11
refactor: enable usage of legacy Anoncrypt packer. Make nested forwards while creating forward message. Add legacyForward model

Abdulbois:
**Title:**
Enable legacy Anoncrypt packer. Make nested packed forwards while creating forward message

**Summary:**

- Add legacy Anoncrypt packer while creating packers
- Add ability to create nested packed forwards (one nested forward for each routing key)
- Check and convert msg field of Forward to Envelope in order to support DIDComm V1 Forward types
- Revert part of 04bfea80427b3d2ae374ddad2ced84b77c70a179 commit related to generating keys inside mediator service. Because with previous changes (having two types of generated keys belonging to the same router) it breaks Route Coordination protocol (while creating nested forwards it will pack two times to the same mediator then mediator cannot handle second forward message
- Remove test (which checks bad sender key)  due to forcing anoncrypt while calling createForwardMessage method


sudeshrshetty (approved):
thanks for your contribution @Abdulbois. 
PR looks good to me, but please also signoff your commits as per [guidelines](https://github.com/hyperledger/aries-framework-go/blob/main/.github/CONTRIBUTING.md#pull-request).

sudeshrshetty:
@Abdulbois can you please squash your commits.
https://github.com/hyperledger/aries-framework-go/blob/main/.github/CONTRIBUTING.md#pull-request

fqutishat:
@Abdulbois can you rebase

pkg/didcomm/dispatcher/outbound/outbound.go
```diff
@@ -382,26 +381,61 @@ func (o *Dispatcher) createForwardMessage(msg []byte, des *service.Destination)
...
+	packedMsg, err := o.createPackedNestedForwards(msg, senderKey, fwdKeys, forwardMsgType, mtProfile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create packed nested forwards: %w", err)
+	}
+
+	return packedMsg, nil
+}
+
+func (o *Dispatcher) createPackedNestedForwards(msg, senderKey []byte, routingKeys []string, fwdMsgType, mtProfile string) ([]byte, error) { //nolint: lll
+	for i, key := range routingKeys {
```
ashcherbakov, line 395: It turned out that DIDComm v1 and v2 use different order of routingKeys when preparing Forward messages:
- DIDComm v1: direct order (as currently implemented and expected by Verity): https://github.com/hyperledger/aries-rfcs/tree/main/features/0160-connection-protocol#request-transmission
- DIDComm v2: reverse order https://identity.foundation/didcomm-messaging/spec/#sender-process-to-enable-forwarding

It seems we need to support both versions (use the switch above?)

Abdulbois:
> @Abdulbois can you rebase

pls check it

troyronda (approved):

fqutishat (approved):

PR #3284, opened 2022-07-20
feat: Size computation for allocation may overflow

bhaskarvilles:
Performing calculations involving the size of potentially large strings or slices can result in an overflow (for signed integer types) or a wraparound (for unsigned types). An overflow causes the result of the calculation to become negative, while a wraparound results in a small (positive) number.

This can cause further issues. If, for example, the result is then used in an allocation, it will cause a runtime panic if it is negative, and allocate an unexpectedly small buffer otherwise.

Signed-off-by: Bhaskar <ram@hacker.ind.in>

### DIFF
--- a/pkg/vdr/fingerprint/fingerprint.go
+++ b/pkg/vdr/fingerprint/fingerprint.go
@@ -5,14 +5,17 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 package fingerprint
+package main
 
 import (
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/elliptic"
+	"encoding/json"
 	"encoding/binary"
 	"errors"
 	"fmt"
+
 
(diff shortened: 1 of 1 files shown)


pkg/vdr/fingerprint/fingerprint.go
```diff
@@ -145,6 +148,16 @@ func KeyFingerprint(code uint64, pubKeyValue []byte) string {
 	return fmt.Sprintf("z%s", base58.Encode(buf))
 }
 
+func encryptValue(v interface{}) ([]byte, error) {
```
sudeshrshetty, line 151: where is this function being used? also, I dont see any unit tests.

bhaskarvilles: Yes i have just added the function and writing new code now :)

PR #3303, opened 2022-07-26
feat: mediator supports didcomm v1/v2 per-connection

Moopli:


Moopli:
I disabled mediator BDD tests that use didcomm v2 keys but do didexchange between an agent and its router - this use case was always a hacky blend, and not required by specs, and we can't support it if we want agents to determine the didcomm version they need for their routing key based on the didcomm version of their mediator connection.

fqutishat (approved):

PR #3306, opened 2022-07-28
feat: Adding sqlite to mobile agent

afrancoc2000:
**Title:**
Adding sqlite to mobile agent

**Description:**
I created a component in the mobile agent that allows a user to use a storage provider that connects with SQLite.

**Summary:**

Using https://github.com/hyperledger/aries-framework-go-ext/blob/main/component/storage/mysql/store.go as reference I created a store provider that connects with SQLite. 
Also, I implemented the TotalItems() method inside the storage wrapper.

@DRK3 What do you think?


DRK3:
Hi @afrancoc2000! Thanks again for the contribution.

I think this new storage provider implementation should be in https://github.com/hyperledger/aries-framework-go-ext/tree/main/component/storage alongside the other SQL storage provider implementations, as I think it could be useful to non-mobile applications as well. Let me know what you think.

DRK3:
Hi again @afrancoc2000, after I wrote that comment, I realized that your storage provider implementation here returns the interface types defined in https://github.com/hyperledger/aries-framework-go/blob/main/cmd/aries-agent-mobile/pkg/api/storage.go rather than the ones in https://github.com/hyperledger/aries-framework-go/blob/main/spi/storage/storage.go, which is why I'm assuming you didn't put your implementation in aries-framework-go-ext.

I think what you can do to allow your storage provider implementation to be usable with non-mobile bindings is this:
1. Move your implementation to aries-framework-go-ext and change the interface types to the ones in https://github.com/hyperledger/aries-framework-go/blob/main/spi/storage/storage.go
2. Create a wrapper type that "converts" between the interface types, similar but opposite to the wrapper in https://github.com/hyperledger/aries-framework-go/blob/main/cmd/aries-agent-mobile/pkg/wrappers/storage/storage.go.

Now your new SQLite Aries storage implementation is usable across non-mobile and mobile (using the wrapper) implementations.

Let me know what you think or if you have any questions!

afrancoc2000:
Hi, @DRK3 great idea! I agree, I will do the change, I have a question, what do you think about adding a getter with the wrapper that imports the https://github.com/hyperledger/aries-framework-go-ext/tree/main/component/storage classes for the mobile agent? this, because when I tried importing 2 different aar files generated using go mobile, they conflicted with each other as both have one library that is the same for both inside, and gradle doesn't know how to solve this conflict. What do you think?

Thanks!

DRK3:
> Hi, @DRK3 great idea! I agree, I will do the change, I have a question, what do you think about adding a getter with the wrapper that imports the https://github.com/hyperledger/aries-framework-go-ext/tree/main/component/storage classes for the mobile agent? this, because when I tried importing 2 different aar files generated using go mobile, they conflicted with each other as both have one library that is the same for both inside, and gradle doesn't know how to solve this conflict. What do you think?
> 
> Thanks!

Hi @afrancoc2000!

For your getter idea, what would that look like? One thing we would want to avoid is importing `aries-framework-go-ext` into `aries-framework-go` as it creates a circular dependency.

Also, what is conflicting when you try to generate the Go mobile bindings? Is it a naming conflict?

afrancoc2000:
Hi @DRK3, The issue is that if I create an aar using go mobile for 2 different projects they will both generate an aar project with this folder structure:

```
> classes > go
          > META-INF
          > org/hyperledger....
```

And the `go` class gets duplicated. That wouldn't be an issue, if gradle could ignore one of the packages, but there's an issue inside gradle trying to exclude packages from an aar, as shown [here](https://stackoverflow.com/questions/53149129/could-not-find-method-exclude-for-arguments-on-object-of-type-org-gradle-api-i), we couldn't figure out how to exclude packages when they come from an aar. Also I'm not completely sure how would the classes from the aries-framework-go-ext get exported, hopefuly they would be ok, but I had to create an extra getter in the code I created for the Provider constructor to appear in the aar.

DRK3:
Hi @afrancoc2000, unfortunately my knowledge of how the mobile bindings work is somewhat limited, but is the issue basically that you don't have a way to create the mobile bindings with a storage provider from a different project/module (like `aries-framework-go-ext`?)

DRK3:
I can also ask around and see if someone else who knows the mobile binding might be able to help...

DRK3:
Hi @afrancoc2000, just wanted to check in on this. My apologies - I meant to send an update earlier, but unfortunately I couldn't find someone who knows how the mobile bindings work. Are you still stuck on this? If so, I think perhaps our best option is to just merge the sqlite implementation you have in as-is where you've originally put it and not worry about adding it to aries-framework-go-ext for now, since I realize that this might be blocking your progress.

afrancoc2000:
Hi @DRK3, sorry for the late response, I also was a little busy. For my project I generated the aar file and used it as it was, and also I created the sqlite component to the aries-framework-go-ext [here](https://github.com/hyperledger/aries-framework-go-ext/pull/282) it would be great, if you can give it a look. 
I started working on the wrapper idea [here](https://github.com/afrancoc2000/aries-framework-go/tree/feature/mobile_component_wrapper) but I haven't had the time to finish it, I want to review the code again, but the issue with not beeing able to import two aar files from gomobile into an android app remains, that's why I wanted to import the aries-framework-go-ext library from aries-framework-go only for the mobile app. 
So let me give it an extra look to the wrapper, but for now I think we can merge the sqlite implementation for the aries-framework-go-ext repo. What do you think?

DRK3:
Hi @afrancoc2000, I think what we can do is is merge this PR in as-is, and I'll create a follow-up issue for someone to investigate if it's possible to use the `aries-framework-go-ext` implementation. I have a minor comment to make on the `aries-framework-go-ext` implementation, and then I think that one will be good to merge too. Do you have any other changes you want to make to this PR? If not, I can get some of the maintainers to take a look (need their approval to merge). Also, can you take a look at that unit test failure?

afrancoc2000:
Hi @DRK3, I don't want to make other changes to the PR. I checked the test and it looks like it is failing because of a timeout, It looks ok on mine [here](https://github.com/afrancoc2000/aries-framework-go/actions/runs/3159325901). I will try to run it again changing comments

DRK3:
@afrancoc2000 I'm also seeing errors in Ursa/CL unit tests in my open PR, which doesn't make any changes to related code, so I suspect there's an unrelated issue going on... @sudeshrshetty @rolsonquadras Can we merge this PR?

DRK3:
@sudeshrshetty @rolsonquadras @fqutishat If you're ok with this PR, can we merge?

Issue #3297, opened 2022-07-25
Support for Presentation Exchange V2


sudeshrshetty:
Aries currently supporting presentation exchange [V1 ](https://identity.foundation/presentation-exchange/spec/v1.0.0/)

TASKS: 

1. Upgrade aries presentation exchange to support [V2](https://identity.foundation/presentation-exchange/)
2. Universal wallet to support both Presentation Exchange V1 & V2 for interoperability.
3. Make Presentation Exchange version in  Present Proof and other protocol consumers configurable (keep default to V1 for now)

sudeshrshetty:
@vkubiv

vkubiv:
Changes that need to be implemented to support V2:

1. Support an optional `format` property for `input_descriptor`.
2. Remove `schema` property, and instant use field constraint. 
3. Support `intended_to_retain`
4. JSON-LD Framing Feature

PR #3312, opened 2022-08-02
feat: support of presentation exchange v2.

vkubiv:

**Title:**
Presentation exchange v2

**Description:**
https://github.com/hyperledger/aries-framework-go/issues/3297

**Summary:**

Added support of changes from the latest presentation exchange draft https://identity.foundation/presentation-exchange/

### DIFF
--- a/pkg/doc/presexch/definition.go
+++ b/pkg/doc/presexch/definition.go
@@ -27,6 +27,7 @@ import (
 )
 
 const (
+
 	// All rule`s value.
 	All Selection = "all"
 	// Pick rule`s value.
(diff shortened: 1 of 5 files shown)


pkg/doc/presexch/definition.go
```diff
@@ -888,6 +962,24 @@ func (a byID) Len() int           { return len(a) }
...
+		return result
+	}
+
+	for _, credential := range credentials {
+		for _, proofType := range format.LdpVP.ProofType {
+			if hasProofWithType(credential, proofType) {
+				result = append(result, credential)
+			}
+		}
+	}
```
sudeshrshetty, line 978: how we are handling JWT VC here?
Do we have a follow up task on that?

sudeshrshetty (approved):

pkg/doc/presexch/definition.go
vkubiv: We can add JWT VC to the list of supported formats here.

PR #3325, opened 2022-08-11
feat: Add client, rest and command handlers for legacy-connection pro…

### LINKED ISSUE #3300: A Connection abstraction on top of didexchange/connection protocols
Create a Connection abstraction in af-go that allows to

- Use Connection protocol to establish connection to be used in other protocols (Issuance, Presentation, etc.)
- Be compatible with did exchange protocol (so either Connection or did exchange protocols can be used to establish a connection)

Cover by unit tests

Abdulbois:
**Title:**
Add client, rest and command handlers for legacy-connection protocol

**Summary:**

- Add client for legacy-connection protocol
- Move didexchange.Event to didcomm/common/event.go in order to make it common for didexchange and legacy-connection protocols
- Refactor HandleInboundEnvelope method of inbound message handler in order to handle messages of legacy-connection protocol
- Add rest and command handlers for legacy-connection protocol. Use them inside controller package
- Add creating legacy destination
- Add new LegacyConnection Error group
- Enable handling base58 keys while packing legacy media type envelopes
- Change media type profile while initializing legacy-connection service

Closes #3300

### DIFF
--- a/pkg/client/legacyconnection/client.go
+++ b/pkg/client/legacyconnection/client.go
@@ -1,0 +1,443 @@
+/*
+Copyright Avast Software. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package legacyconnection
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/google/uuid"
+
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/common/service"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/legacyconnection"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/mediator"
+	"github.com/hyperledger/aries-framework-go/pkg/didcomm/transport"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
+	"github.com/hyperledger/aries-framework-go/pkg/kms"
+	"github.com/hyperledger/aries-framework-go/pkg/store/connection"
+	"github.com/hyperledger/aries-framework-go/spi/storage"
+)
+
+// InvitationMsgType defines the connection invite message type.
+const InvitationMsgType = legacyconnection.InvitationMsgType
+
+// ErrConnectionNotFound is returned when connection not found.
+var ErrConnectionNotFound = errors.New("connection not found")
+
+type options struct {
+	routerConnections  []string
+	routerConnectionID string
+}
+
+func applyOptions(args ...Opt) *options {
+	opts := &options{}
+
+	for i := range args {
+		args[i](opts)
+	}
+
+	return opts
+}
+
+// Opt represents option function.
+type Opt func(*options)
+
+// InvOpt represents option for the CreateInvitation function.
+type InvOpt Opt
+
+// WithRouterConnectionID allows you to specify the router connection ID.
+func WithRouterConnectionID(conn string) InvOpt {
+	return func(opts *options) {
+		opts.routerConnectionID = conn
+	}
+}
+
+// WithRouterConnections allows you to specify the router connections.
+func WithRouterConnections(conns ...string) Opt {
+	return func(opts *options) {
+		for _, conn := range conns {
+			// filters out empty connections
+			if conn != "" {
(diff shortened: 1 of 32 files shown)


pkg/client/legacyconnection/client.go
```diff
@@ -0,0 +1,443 @@
...
+		for _, conn := range conns {
+			// filters out empty connections
+			if conn != "" {
+				opts.routerConnections = append(opts.routerConnections, conn)
+			}
+		}
+	}
+}
+
+// provider contains dependencies for the DID exchange protocol and is typically created by using aries.Context().
```
ashcherbakov, line 73: Please replace `DID Exchange` to `Connection` in comments

ashcherbakov (approved):

pkg/client/legacyconnection/client.go
Abdulbois: Changed to `legacy-connection` in order to not confuse readers with the definition of `connection` package

pkg/controller/rest/didexchange/models.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright SecureKey Technologies Inc. All Rights Reserved.
+Copyright Avast Software. All Rights Reserved.
```
sudeshrshetty, line 2: While updating existing code : I would suggest adding `Copyright Avast Software. All Rights Reserved.` as newline instead of replacing existing license .

Abdulbois: Ok

pkg/client/legacyconnection/client.go
```diff
@@ -0,0 +1,443 @@
...
+
+// WithRouterConnectionID allows you to specify the router connection ID.
+func WithRouterConnectionID(conn string) InvOpt {
+	return func(opts *options) {
+		opts.routerConnectionID = conn
+	}
+}
+
+// WithRouterConnections allows you to specify the router connections.
+func WithRouterConnections(conns ...string) Opt {
```
kgoncharov, line 62: what if we move all those helper methods like `WithRouterConnections` and `applyOptions` to separate file like `opts.go`? (also all `With..`s methods at the end of the file) 
I guess the same would go to `didexchange` and other clients, but still. It will reduce amount of code in one file.

pkg/didcomm/common/service/destination_default.go
```diff
@@ -70,7 +74,7 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
 	} else { // didcomm v1 service
 		didCommService, ok = diddoc.LookupService(didDoc, didCommServiceType)
 		if !ok {
-			return nil, fmt.Errorf("create destination: missing DID doc service")
+			return createLegacyDestination(didDoc)
```
kgoncharov, line 77: Maybe as a part of incremental refactoring we could split the whole `CreateDestionation` method onto `createV2`, `createV1` and `createLegacy`? To avoid long nested if's

pkg/didcomm/common/service/destination_default.go
```diff
@@ -106,3 +110,65 @@ func CreateDestination(didDoc *diddoc.Doc) (*Destination, error) {
...
+
+	return &Destination{
+		RecipientKeys:     convertAnyB58Keys(didCommService.RecipientKeys),
+		ServiceEndpoint:   model.NewDIDCommV1Endpoint(uri),
+		RoutingKeys:       convertAnyB58Keys(didCommService.RoutingKeys),
+		MediaTypeProfiles: didCommService.Accept,
+	}, nil
+}
+
+func convertAnyB58Keys(keys []string) []string {
```
kgoncharov, line 141: Move to some utils maybe? Jut a proposal, don't know if it fits Golang and af-go conventions

pkg/controller/rest/didexchange/models.go
Abdulbois: @sudeshrshetty I guess that after making your suggestion `repolint` job fails now or I have missed something else. Can you pls check it?

pkg/client/legacyconnection/client.go
Abdulbois: I think it will be better to do such kind of refactoring in another PR because `With` convention is used in a lot of places

kgoncharov (approved):

sudeshrshetty (approved):

PR #3336, opened 2022-08-15
fix: credential manifest omit optional properties

### LINKED ISSUE #3335: Optional credential manifest fields are not omitted
There are several properties of credential manifests, credential fulfillments, and credential applications which are optional as per DIF specs. These properties were defined in the structs with the `omitempty` tag but were not actually omitted because there are no default empty values for embedded structs. This affected some validate functions as the assumption that certain fields do not exist were incorrect.

HeidiHan0000:
Closes #3335 

**Summary:**

- `omitempty` tag for mandatory properties were removed
- optional properties (with the `omitempty` tag) of custom types were changed to pointers so they can be omitted with the `omitempty` tag
- additional validation checks which were missing and more unit tests

### DIFF
--- a/pkg/doc/cm/credentialmanifest.go
+++ b/pkg/doc/cm/credentialmanifest.go
@@ -27,28 +27,27 @@ const CredentialManifestAttachmentFormat = "dif/credential-manifest/manifest@v1.
 // CredentialManifest represents a Credential Manifest object as defined in
 // https://identity.foundation/credential-manifest/#credential-manifest-2.
 type CredentialManifest struct {
-	ID                     string                           `json:"id,omitempty"`
-	Version                string                           `json:"version,omitempty"`
-	Issuer                 Issuer                           `json:"issuer,omitempty"`
-	OutputDescriptors      []*OutputDescriptor              `json:"output_descriptors,omitempty"`
-	Format                 presexch.Format                  `json:"format,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/doc/cm/credentialmanifest.go
```diff
@@ -137,27 +136,68 @@ func (cm *CredentialManifest) UnmarshalJSON(data []byte) error {
...
+	err = ValidateOutputDescriptors(cm.OutputDescriptors)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
+func validateIssuer(issuer Issuer) error {
+	if issuer == (Issuer{}) {
```
DRK3, line 161: A better way to check this condition would be to make the `Issuer` field a pointer, rather than doing a struct value comparison like this. Strictly speaking, it doesn't make sense to say a non-pointer struct is "missing", since you don't know whether it was just defaulting to to the zero values or someone (for some reason) purposefully initialized it with zero values.

Alternatively, you might consider simply removing this check. Is there a meaningful difference between the entire `issuer` struct being missing and the issuer ID being missing?

DRK3 (changes requested):

pkg/doc/cm/credentialmanifest.go
HeidiHan0000: Oh that's true, I'll remove this check!

fqutishat (approved):

PR #3338, opened 2022-08-15
feat: KMS storage interface redesign

### LINKED ISSUE #3331: Change local KMS to not require Aries storage interface
This is part of an overall larger storage rework to switch Aries from using the current "one size fits all" approach to data storage to one that allows consumers of Aries-Framework-Go to define their storage layer according to their needs.

The goal of this task is to switch the local KMS implementation to not require an Aries Storage provider. To make the transition easier, a storage wrapper type can be created to allow for a caller to continue using the existing storage implementations while they are working on new ones.

DRK3:
* KMS storage interface now uses a minimal interface instead of requiring a full Aries storage provider implementation (from the spi package)
* With this new approach, the consumer of an Aries KMS no longer has to let the Aries storage interface/implementation design dictate how they store KMS-related data. One notable example of this is that the choice of database name is now decided by the consumer rather than by Aries.
* To ease the transition, I added a function (NewAriesProviderWrapper) to allow a consumer to continue using an Aries storage provider implementation with a KMS if they wish.
* When using Local KMS before, "k" was prepended to all keyset IDs being stored. This was a workaround for CouchDB, which would fail if someone tried storing a key using a keyset ID starting with "_". This prefixing was removed as part of the overall new philosophy of removing database-specific logic from Aries and letting the caller control their database design. Existing Aries implementations will need their databases updated to remove the "k" or, if using CouchDB and keyset IDs beginning with "_", then keyset IDs will need to be prefixed outside of Aries before trying to store/retrieve/delete from a KMS store.

closes #3331


rolsonquadras (approved):

aholovko (approved):

Baha-sk (approved):
lgtm,

just to double check, you made sure this change works with couchdb store for the kms? as discussed with relation to kid possibly starting with `_` as id?

DRK3:
> lgtm,
> 
> just to double check, you made sure this change works with couchdb store for the kms? as discussed with relation to kid possibly starting with `_` as id?

Spoke with @Baha-sk, and we decided to leave this PR as-is and keep the CouchDB-prefix workaround out of Aries, as it doesn't seem appropriate to have it here. I added a note to the commit message explaining why it was done and what a user might need to do to update their code when pulling this change.

fqutishat (approved):

PR #3342, opened 2022-08-18
refactor: A few refactoring of legacy-connection package

Abdulbois:
**Title:**
Refactor legacy-connection package due to interoperability issues

**Summary:**

- Add new legacy http header to handle legacy requests. Cover with unit tests
- Encode invitation Recipient key to base58
- Change embedded verification to referenced while creating verification methods
- Add legacy file into doc package to convert DIDdoc to raw doc. Cover with unit test
- Add handling DID id's where did method is not specified
- Add JSONBytes and ParseConnection methods to use them while handling connection data
- Change AckMsgType content to proper one
- Add method to convert array of did-keys to base58 keys
- Enable saving decryption key while handling envelope at connection-request state
- Add new InvitationRecipientKeys into connection.Record model
- Refactor requestMsgRecord, handleInboundRequest and prepareConnectionSignature methods and their unit-tests
- Save Recipient keys while handling connection request
- Use SaveDID instead SaveDIDByResolving

### DIFF
--- a/pkg/didcomm/protocol/legacyconnection/models.go
+++ b/pkg/didcomm/protocol/legacyconnection/models.go
@@ -7,6 +7,13 @@ SPDX-License-Identifier: Apache-2.0
 package legacyconnection
 
 import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+
 	"github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/decorator"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 )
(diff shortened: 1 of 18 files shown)


pkg/didcomm/protocol/legacyconnection/models.go
```diff
@@ -76,3 +83,76 @@ type Connection struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("JSON marshalling of connection raw failed: %w", err)
+	}
+
+	return byteConn, nil
+}
+
+// ParseConnection creates an instance of Connection by reading a JSON connection from bytes.
+func ParseConnection(data []byte) (*Connection, error) {
+	raw := &connectionRaw{}
```
ashcherbakov, line 135: I think the term `raw` here is a bit confusing. Can we use a term `legacy` instead?
So that we can have a connection and a legacy connection, a diddoc and a legacy diddoc.
Also we can have the methods
- `func (con *Connection) toLegacyJSONBytes() ([]byte, error)` instead of just JSONBytes
- `func ParseLegacyJSONBytes(data []byte) (*Connection, error)` instead of ParseConnection.

ashcherbakov:
Are we planning to write an integration acceptance test in af-go that will check the interoperability between af-go legacy connection protocol and other implementations (VCX in particular)? Will a BDD test do these checks? Or we need another one?

pkg/didcomm/protocol/legacyconnection/states.go
```diff
@@ -638,20 +655,8 @@ func (ctx *context) resolveDidDocFromConnection(con *Connection) (*did.Doc, erro
...
-	var method string
-
-	if parsedDID != nil && parsedDID.Method != "sov" {
-		method = parsedDID.Method
-	} else {
-		method = "peer"
-	}
 	// store provided did document
-	_, err = ctx.vdRegistry.Create(method, con.DIDDoc, vdrapi.WithOption("store", true))
+	_, err = ctx.vdRegistry.Create("peer", con.DIDDoc, vdrapi.WithOption("store", true))
```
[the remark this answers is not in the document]
Abdulbois: We use peer to save DIDdoc there because saving data into storage is not implemented inside key vdr

pkg/doc/did/legacy.go
```diff
@@ -0,0 +1,120 @@
...
+*/
+
+package did
+
+import (
+	"fmt"
+)
+
+// ToLegacyRawDoc converts document to raw doc.
+func (doc *Doc) ToLegacyRawDoc() (interface{}, error) {
```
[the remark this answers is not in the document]
Abdulbois: It uses private methods and vars from doc file

pkg/didcomm/protocol/legacyconnection/models.go
```diff
@@ -76,3 +83,80 @@ type Connection struct {
 	DID    string   `json:"DID,omitempty"`
 	DIDDoc *did.Doc `json:"DIDDoc,omitempty"`
 }
+
+type connectionRaw struct {
```
ashcherbakov, line 87: I would rename it to `legacyConnection`

pkg/didcomm/protocol/legacyconnection/models.go
```diff
@@ -76,3 +83,80 @@ type Connection struct {
...
+	DID    string  `json:"DID,omitempty"`
+	DIDDoc *rawDoc `json:"DIDDoc,omitempty"`
+}
+
+type rawDoc struct {
```
ashcherbakov, line 92: I would rename it to `legacyDoc`

sudeshrshetty (approved):

ashcherbakov (approved):

Baha-sk (approved):

Issue #3349, opened 2022-08-26
Can't reconnect mediator after connection terminated from server side


borancar:
**What I'm trying to do**

I'm using AFGO WASM and I'm trying to reconnect to all the mediators using mediator.reconnectAll() after one connection timed out from the server side and was reported as "Closing Abnormally" (I know, the server was not the friendliest and just sad "Going away" as the status, but that's AWS API GW for you...).

**Expected result**

The app should reconnect to the mediator.

**Actual result**

I get the following error
```
Error: aries wasm: code: 5006, message: send noop request: outboundDispatcher.Send: failed to send msg using outbound transport: websocket write message : failed to write: WebSocket closed: unclean connection close: status = StatusAbnormalClosure and reason = ""
```

**Root cause**

This was was actually difficult to spot! But the root cause is nothing cleans up the abnormally terminated connections. If you check the function below, verKeys always stays an empty array:

https://github.com/hyperledger/aries-framework-go/blob/6c0c224401eed96cc4ad6d8478664374885596fc/pkg/didcomm/transport/ws/pool.go#L82-L96

PR #3352, opened 2022-08-26
Clean up connections after websocket close

borancar:
**Title:**
Clean up connections after WebSocket close

**Description:**
Make sure we clean up connections by just going through the map and
cleaning all the verkeys that match the connection. This is a suboptimal
solution, but a connection close is not something happening often.

Fixes https://github.com/hyperledger/aries-framework-go/issues/3349

### DIFF
--- a/pkg/didcomm/transport/ws/pool.go
+++ b/pkg/didcomm/transport/ws/pool.go
@@ -79,9 +79,7 @@ func (d *connPool) remove(verKey string) {
 }
 
 func (d *connPool) listener(conn *websocket.Conn, outbound bool) {
-	verKeys := []string{}
-
-	defer d.close(conn, verKeys)
+	defer d.close(conn)
 
 	go keepConnAlive(conn, outbound, pingFrequency)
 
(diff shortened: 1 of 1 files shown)


pkg/didcomm/transport/ws/pool.go
```diff
@@ -79,9 +79,7 @@ func (d *connPool) remove(verKey string) {
 }
 
 func (d *connPool) listener(conn *websocket.Conn, outbound bool) {
-	verKeys := []string{}
```
borancar, line 82: These were never populated.

borancar:
If you like the approach, or if it's acceptable, then I can try to add tests.

borancar:
```
linting with js/wasm..
Error: pkg/didcomm/transport/ws/pool.go:74:20: func `(*connPool).remove` is unused (unused)
func (d *connPool) remove(verKey string) {
                   ^
make: *** [lint] Error 1
Makefile:37: recipe for target 'lint' failed
Error: Process completed with exit code 2.
```

The unused check is incorrect (or possibly incomplete as it doesn't take the regular server into account), it's used here https://github.com/hyperledger/aries-framework-go/blob/b1decdfed852057fe64660b40ac705953b4d6d0f/pkg/didcomm/transport/ws/upgrade_srv.go#L28-L35

Issue #3351, opened 2022-08-26
JSON LD Contexts in JWT VCs


sudeshrshetty:
Currently JSON LD contexts in JWT VCs are required, but that doesn't clearly enforces contexts beyond Verifiable Credential data model. In simple words `https://www.w3.org/2018/credentials/v1` is required but any additional contexts for custom properties can be optional.

Example: https://github.com/decentralized-identity/jwt-vc-presentation-profile/pull/52/files#diff-76909e03d654c1deb30b8258103972c9fa20ccb9411e72f0dc2436151c70ddc9R2-R4

**Objective of this task:**
For JWT VCs,
- [ ] make VC data model context as required and additional contexts can be optional
- [ ] If [strict validation option](https://github.com/hyperledger/aries-framework-go/blob/6c0c224401eed96cc4ad6d8478664374885596fc/pkg/doc/verifiable/credential.go#L578) enabled while parsing JWT VCs, then JSON LD  data model validation of JWT VC should be performed.

sudeshrshetty:
@mishasizov-SK

mishasizov-SK:
PR: https://github.com/hyperledger/aries-framework-go/pull/3357

PR #3357, opened 2022-08-30
JSON LD Contexts in JWT VCs

mishasizov-SK:
Signed-off-by: Mykhailo Sizov <mykhailo.sizov@securekey.com>

**Title:**
JSON LD Contexts in JWT VCs

**Description:**
Solves: https://github.com/hyperledger/aries-framework-go/issues/3351

**Summary:**

Added extra option to `ValidateJSONLD` function called `WithStrictContextURIPosition` that sets strict validation of URI position within context property. The index of uri in underlying slice represents the position of given URI in parsed context array.
Also, added this option as a default to `*verifiable.Credential.validateJSONLD(.... docjsonld.WithStrictContextURIPosition(baseContext))` function in order to make base context URI required.

### DIFF
--- a/pkg/doc/jsonld/validate.go
+++ b/pkg/doc/jsonld/validate.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/piprate/json-gold/ld"
 
@@ -20,6 +21,7 @@ type validateOpts struct {
 	strict               bool
 	jsonldDocumentLoader ld.DocumentLoader
 	externalContext      []string
+	contextURIPositions  []string
 }
 
 // ValidateOpts sets jsonld validation options.
@@ -43,6 +45,15 @@ func WithExternalContext(externalContext []string) ValidateOpts {
 func WithStrictValidation(checkStructure bool) ValidateOpts {
 	return func(opts *validateOpts) {
 		opts.strict = checkStructure
+	}
+}
+
+// WithStrictContextURIPosition sets strict validation of URI position within context property.
+// The index of uri in underlying slice represents the position of given uri in @context array.
+// Can be used for verifiable credential base context validation.
+func WithStrictContextURIPosition(uri string) ValidateOpts {
+	return func(opts *validateOpts) {
+		opts.contextURIPositions = append(opts.contextURIPositions, uri)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


fqutishat (approved):

pkg/doc/jsonld/validate.go
```diff
@@ -46,6 +48,15 @@ func WithStrictValidation(checkStructure bool) ValidateOpts {
...
 }
 
+// WithStrictContextURIPosition sets strict validation of URI position within context property.
+// The index of uri in underlying slice represents the position of given uri in @context array.
+// Can be used for verifiable credential base context validation.
+func WithStrictContextURIPosition(uri string) ValidateOpts {
+	return func(opts *validateOpts) {
+		opts.contextURIPositions = append(opts.contextURIPositions, uri)
+	}
+}
```
sudeshrshetty, line 58: why do we need this flag?
By default JWT VCs should check presence of credential contexts as mentioned in #3351 .

And I don't see usage of existing `WithStrictValidation` option to structurally validate VCs.

mishasizov-SK: >why do we need this flag?

This flag is used to strictly define the position of base context withing contexts list.
I added it [here](https://github.com/hyperledger/aries-framework-go/pull/3357/files#diff-953974a5ec9fe3293be8ffd004be86b23666847d300650428cb21673e78fa140R889) to satisfy the documentation requirement:
```
The value of the @context MUST be an ordered set where the first item is a URI with the value 

https://www.w3.org/2018/credentials/v1.
```

>By default JWT VCs should check presence of credential contexts as mentioned in https://github.com/hyperledger/aries-framework-go/issues/3351 .

The default validation model is [combinedValidation](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/verifiable/credential.go#L830), that implies validation the VC against json schema and json ld validation (and they includes checking context property presence). Adding the [explicit requirement on base context](https://github.com/hyperledger/aries-framework-go/pull/3357/files#diff-953974a5ec9fe3293be8ffd004be86b23666847d300650428cb21673e78fa140R889) brings the requirement "make VC data model context as required and additional contexts can be optional". Without this addition strict validation option disabled will not return an error in case VC without base context URI.

>And I don't see usage of existing WithStrictValidation option to structurally validate VCs.

Strict validation option [is enabled by default](https://github.com/hyperledger/aries-framework-go/blob/main/pkg/doc/jsonld/validate.go#L51). Do you mean that is should be added explicitly over [here](https://github.com/hyperledger/aries-framework-go/pull/3357/files#diff-953974a5ec9fe3293be8ffd004be86b23666847d300650428cb21673e78fa140R889)?

PR #3413, opened 2022-10-21
feat: Update Tink version, protobuf type

DRK3:
Updated the Tink version and our methods to use the newer protobuf type that it uses.

Reason for this change: the existing version of Tink makes use of a very large int in one particular place in the code, and this very large int prevents the gomobile tool from successfully generating Android bindings. This meant that aries-framework-go packages that used Tink could not be used in any code that you may want to generate Android bindings for. The updated version of Tink does not have that check anymore, which resolves the issue.

Due to a circular dependency between the main module and component/storage/edv (and the fact that both of them require Tink), there are some module/build issues with this commit. This will be resolved in the next commit.


pkg/crypto/tinkcrypto/primitive/secp256k1/secp256k1_signer_key_manager_test.go
```diff
@@ -278,14 +278,6 @@ func TestPublicKeyDataWithInvalidInput(t *testing.T) {
 
 	_, err = pkm.PublicKeyData(serializedKey)
 	require.Error(t, err, "expect an error when input is a modified serialized key")
-
-	// nil
-	_, err = pkm.PublicKeyData(nil)
```
DRK3, line 283: No longer returns an error with the updated Tink version (pub key bytes in returned struct are just empty now)

pkg/crypto/tinkcrypto/primitive/secp256k1/secp256k1_signer_key_manager_test.go
```diff
@@ -278,14 +278,6 @@ func TestPublicKeyDataWithInvalidInput(t *testing.T) {
...
-	require.Errorf(t, err, "expect an error when input is nil")
-
-	// empty slice
-	_, err = pkm.PublicKeyData([]byte{})
```
DRK3, line 287: No longer returns an error with the updated Tink version (pub key bytes in returned struct are just empty now)

pkg/crypto/tinkcrypto/unwrap_support_test.go
```diff
@@ -38,9 +38,6 @@ func Test_ExtractPrivKey(t *testing.T) {
 	require.EqualError(t, err, "extractPrivKey: can't extract unsupported private key 'type.googleapis.com/"+
 		"google.crypto.tink.AesGcmKey'")
 
-	_, err = extractPrivKey(&keyset.Handle{})
```
DRK3, line 41: Passing an empty keyset handle causes Tink to do a panic now instead of throwing an error

pkg/crypto/tinkcrypto/primitive/secp256k1/secp256k1_signer_key_manager_test.go
Baha-sk: you can remove the whole test function since the main test is not applicable anymore

pkg/crypto/tinkcrypto/unwrap_support_test.go
Baha-sk: ok, this is fine then

pkg/crypto/tinkcrypto/primitive/secp256k1/secp256k1_signer_key_manager_test.go
DRK3: @Baha-sk You mean remove `TestPublicKeyDataWithInvalidInput`? Are the other test cases in it not applicable?

Baha-sk: yes, the whole test function's purpose was to test empty/nil cases, it seems Tink now panics on purpose in these cases... you can probably recover the panic here and make sure it does actually panic instead of removing these calls? that would be a good test case.

Moopli:
Why do we need a second PR for the second commit?
One PR with two commits should allow tests to pass once the PR is merged.

pkg/crypto/tinkcrypto/primitive/secp256k1/secp256k1_signer_key_manager_test.go
DRK3: There are some other test cases in this function still that work the same as before above the code I removed). For these two particular test cases I removed, they just don't return an error anymore (i.e. are not an invalid test case anymore)

DRK3: Good point re: catching the panic - I'll update the other places in this PR where I removed those tests that panic to catch it

pkg/crypto/tinkcrypto/unwrap_support_test.go
DRK3: Updated to check for the panic now using require.PanicsWithValue

pkg/crypto/tinkcrypto/primitive/secp256k1/secp256k1_signer_key_manager_test.go
Baha-sk: > There are some other test cases in this function still that work the same as before (above the code I removed, in this same function). For these two particular test cases I removed, they just don't return an error anymore (i.e. are not an invalid test case anymore)

I think you can use `require.Panics()` in this case, without having to set a value.

DRK3: These two cases don't panic - they're not an error or panic

Baha-sk (approved):

Baha-sk:
> One PR with two commits should allow tests to pass once the PR is merged.

you can push a second commit within the same PR where you can use the commit hash of the first commit in go.mod of the second commit, @Moopli is right

Moopli (approved):

rolsonquadras (approved):

sudeshrshetty (approved):

PR #3436, opened 2022-11-29
refactor: Minimize interfaces required by didsignjwt.SignJWT

DRK3:
Minimized the interfaces required by the didsignjwt.SignJWT method by replacing the usage of the kms.KeyManager, crypto.Crypto, and vdr.Registry interfaces with minimized interfaces that define only the functionality needed for the SignJWT function. This makes it easier for consumers to inject in their own implementations. The method signatures of the new minimized interfaces still line up with the method signatures of the kms.KeyManager, crypto.Crypto, and vdr.Registry interfaces, so implementations of those interfaces can still be used without making any changes.

### DIFF
--- a/pkg/doc/util/didsignjwt/signjwt.go
+++ b/pkg/doc/util/didsignjwt/signjwt.go
@@ -9,9 +9,9 @@ package didsignjwt
 import (
 	"errors"
 	"fmt"
+
 	"strings"
 
-	"github.com/hyperledger/aries-framework-go/pkg/crypto"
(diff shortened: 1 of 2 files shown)


rolsonquadras (approved):

Moopli (approved):

pkg/doc/util/didsignjwt/signjwt.go
```diff
@@ -9,9 +9,9 @@ package didsignjwt
 import (
 	"errors"
 	"fmt"
+
```
Moopli, line 12: bonus newline?

PR #3447, opened 2022-12-16
refactor: Minimize VDResolver interface needed for did config validate

DRK3:
Currently, the DID config validate functionality has an option for injecting in a VDR, however most of the methods in the interface are not actually used.

This change switches the option to use a minimal version of that interface that is itself a subset of the full VDR interface. With this change, a caller can now use the VDR option with a more minimal implementation that only has a Resolve method. The minimal interface still lines up with the larger VDR interface, so another caller can still use a full VDR interface implementation without making any changes.

### DIFF
--- a/pkg/client/didconfig/didconfig.go
+++ b/pkg/client/didconfig/didconfig.go
@@ -17,6 +17,7 @@ import (
 	jsonld "github.com/piprate/json-gold/ld"
 
 	"github.com/hyperledger/aries-framework-go/pkg/common/log"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/did"
 	"github.com/hyperledger/aries-framework-go/pkg/doc/didconfig"
 	vdrapi "github.com/hyperledger/aries-framework-go/pkg/framework/aries/api/vdr"
 )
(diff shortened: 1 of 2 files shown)


sandrask (approved):

rolsonquadras (approved):

pkg/doc/didconfig/didconfig.go
```diff
@@ -39,10 +40,14 @@ const (
 	linkedDIDsProperty = "linked_dids"
 )
 
+type didResolver interface {
```
Baha-sk, line 43: 👍

Baha-sk (approved):

pkg/client/didconfig/didconfig.go
```diff
@@ -66,10 +67,14 @@ func WithJSONLDDocumentLoader(documentLoader jsonld.DocumentLoader) Option {
 	}
 }
 
+type didResolver interface {
+	Resolve(did string, opts ...vdrapi.DIDMethodOption) (*did.DocResolution, error)
+}
+
 // WithVDRegistry defines a vdr service.
-func WithVDRegistry(vdrRegistry vdrapi.Registry) Option {
+func WithVDRegistry(didResolver didResolver) Option {
```
sudeshrshetty, line 75: Should it still be called `WithVDRegistry`?

DRK3: `WithDIDResolver` would be a better name, but the consumers of this will need to update. @sandrask do you know where this is used?

pkg/client/didconfig/didconfig.go
```diff
@@ -66,10 +67,14 @@ func WithJSONLDDocumentLoader(documentLoader jsonld.DocumentLoader) Option {
 	}
 }
 
+type didResolver interface {
```
sudeshrshetty, line 70: private interface?

sudeshrshetty (approved):

sandrask:
https://github.com/hyperledger/aries-framework-go/blob/6fe06c571ddc923a943109aa6decffd9c092bf1b/pkg/doc/didconfig/didconfig.go#L72

sandrask:
https://github.com/hyperledger/aries-framework-go/blob/6fe06c571ddc923a943109aa6decffd9c092bf1b/pkg/doc/didconfig/didconfig.go#L439

PR #3450, opened 2022-12-19
feat: enable ed25519 in jwkkid.BuildJWK()

Moopli:

### DIFF
--- a/pkg/doc/util/jwkkid/kid_creator.go
+++ b/pkg/doc/util/jwkkid/kid_creator.go
@@ -51,15 +51,6 @@ func CreateKID(keyBytes []byte, kt kms.KeyType) (string, error) {
 		}
 
 		return x25519KID, nil
-	case kms.ED25519Type: // go-jose JWK thumbprint of Ed25519 has a bug, manually build it and build its resulting KID.
-		// TODO remove `case kms.ED25519Type` when go-jose fixes Ed25519 JWK thumbprint.
-		//  Also remove `createED25519KID(keyBytes []byte)` function further below.
-		ed25519KID, err := createED25519KID(keyBytes)
-		if err != nil {
-			return "", fmt.Errorf("createKID: %w", err)
-		}
-
-		return ed25519KID, nil
 	case kms.BLS12381G2Type: // BBS+ as JWK thumbprint.
 		bbsKID, err := createBLS12381G2KID(keyBytes)
 		if err != nil {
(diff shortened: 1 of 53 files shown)


rolsonquadras (approved):

pkg/doc/util/jwkkid/kid_creator.go
```diff
@@ -52,6 +52,8 @@ func CreateKID(keyBytes []byte, kt kms.KeyType) (string, error) {
 
 		return x25519KID, nil
 	case kms.ED25519Type: // go-jose JWK thumbprint of Ed25519 has a bug, manually build it and build its resulting KID.
+		// TODO remove `case kms.ED25519Type` when go-jose fixes Ed25519 JWK thumbprint.
```
Baha-sk, line 55: go jose already implemented the fix, need to update the dep and remove this case fully since j.Thumbprint() below will handle the correct ED25519 type

pkg/doc/util/jwkkid/kid_creator.go
```diff
@@ -194,13 +196,11 @@ func BuildJWK(keyBytes []byte, kt kms.KeyType) (*jwk.JWK, error) {
...
 			return nil, fmt.Errorf("buildJWK: failed to build JWK from ecdsa DER key: %w", err)
 		}
-		// TODO remove `case kms.ED25519Type` in CreateKID() and uncomment below case when go-jose fixes Ed25519
-		//      JWK thumbprint. Also remove `createED25519KID(keyBytes []byte)` function further below.
-	// case kms.ED25519Type:
-	//	j, err = jwksupport.JWKFromKey(ed25519.PublicKey(keyBytes))
-	//	if err != nil {
-	//		return nil, fmt.Errorf("buildJWK: failed to build JWK from ed25519 key: %w", err)
-	//	}
+	case kms.ED25519Type:
```
Baha-sk, line 199: this is good to uncomment, but go-jose needs to be updated to pick up the fixed thumbprint computation as mentioned in the previous comment.

pkg/doc/util/jwkkid/kid_creator.go
Moopli: What go-jose version added the fix? If we update, do we want to update to latest?

Baha-sk: It dosen't hurt to be at latest.. but I'll check the version when it was updated

Baha-sk: Weird, I could swear to have seen that they fixed it. It doesn't seem to be fixed yet: https://github.com/square/go-jose/blob/v2/jwk.go#L335

it still uses a bad ed25519 jwk string for thumbnail (missing `"` before `x`)

Baha-sk: ahhh... the repo name is different: github.com/go-jose/go-jose

see the fix here: https://github.com/go-jose/go-jose/blob/v3/jwk.go#L341

need to update the go-jose to this new repo instead, the newer v3 has the fix.
square/go-jose is the legacy repo

Baha-sk (approved):

sudeshrshetty (approved):

PR #3462, opened 2023-01-05
feat: Unit test for SD-JWT Flow (Issuer, Holder, Verifier)

### LINKED ISSUE #3461: Unit Test for SD-JWT Flow (Issuer, Holder, Verifier)
Add unit-test for SD-JWT Flow.

sandrask:
Unit test for SD-JWT Flow.

Closes #3461

### DIFF
--- a/pkg/doc/sdjwt/integration_test.go
+++ b/pkg/doc/sdjwt/integration_test.go
@@ -1,0 +1,75 @@
+/*
+Copyright SecureKey Technologies Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package sdjwt
+
+import (
+	"crypto/ed25519"
+	"crypto/rand"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	afjwt "github.com/hyperledger/aries-framework-go/pkg/doc/jwt"
+	"github.com/hyperledger/aries-framework-go/pkg/doc/sdjwt/holder"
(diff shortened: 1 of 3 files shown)


fqutishat (approved):

sudeshrshetty (commented):
From [Section 5.5.1](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-02.html#section-5.5.1) of spec

> If no Holder Binding JWT is included, the Combined Format for Presentation ends with the ~ character after the last Disclosure.

It means in case of holder disclosure, there will be a stray `~` at the end of disclosures list

pkg/doc/sdjwt/integration_test.go
```diff
@@ -0,0 +1,75 @@
...
+	t.Run("success", func(t *testing.T) {
+		// Issuer will issue SD-JWT for specified claims.
+		token, e := issuer.New(testIssuer, claims, nil, signer)
+		r.NoError(e)
+
+		// TODO: Should we have one call instead of two (designed based on JWT)
+		sdJWTSerialized, e := token.Serialize(false)
+		r.NoError(e)
+
+		fmt.Println(fmt.Sprintf("issuer SD-JWT: %s", sdJWTSerialized))
```
sudeshrshetty, line 51: is this `fmt.println()` leftover?

sandrask: No, I left print outs on purpose to show output formats.

sandrask:
Yes, according to spec if there is no holder binding it should end with ~. I will fix this today since we currently don't have holder binding. I will work on holder binding next week.

PR #3499, opened 2023-01-26
feat: method for SD-JWT VC to create non-SDJWT 'display credential'

Moopli:

### DIFF
--- a/pkg/doc/verifiable/credential_sdjwt.go
+++ b/pkg/doc/verifiable/credential_sdjwt.go
@@ -277,3 +277,116 @@ func makeSDJWT(vc *Credential, signer jose.Signer, signingKeyID string) (*issuer
 
 	return sdjwt, nil
 }
+
+type displayCredOpts struct {
+	displayAll   bool
+	displayGiven []string
+}
+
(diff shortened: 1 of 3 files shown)


pkg/doc/verifiable/credential_sdjwt.go
```diff
@@ -277,3 +278,181 @@ func makeSDJWT(vc *Credential, signer jose.Signer, signingKeyID string) (*issuer
...
+
+		disclosuresByHash[digest] = disclosure
+	}
+
+	return disclosuresByHash, nil
+}
+
+// repopulateSDClaims edits claims in-place, removing _sd_alg fields and
+// replacing _sd fields with the matching claims in disclosuresByHash.
+func repopulateSDClaims(claims map[string]interface{}, disclosuresByHash map[string]*common.DisclosureClaim) error {
```
sandrask, line 398: Is there a reason not to use sdjwt.common.GetDisclosedClaims?

sandrask: We should keep spec implementation in one place. It will be easier to maintain.

sudeshrshetty (approved):

PR #3521, opened 2023-02-08
feat: option to pass SD hash alg into Credential.MakeSDJWT()

Moopli:


sandrask (approved):
We should have had options from start. We should consider excluding ID as an option.

sudeshrshetty (approved):

PR #3590, opened 2023-05-18
refactor: move presexch to component/models, VDR to component/vdr.

Moopli:


rolsonquadras (approved):

component/vdr/mock/mock_vdr.go
```diff
@@ -0,0 +1,75 @@
...
+package mock
+
+import (
+	"github.com/hyperledger/aries-framework-go/component/models/did"
+	vdrapi "github.com/hyperledger/aries-framework-go/spi/vdr"
+)
+
+// MockVDR mock implementation of vdr
+// to be used only for unit tests.
+type MockVDR struct { // nolint: golint // ignore stutter
```
DRK3, line 16: Similar comment as above - can be changed to simply `VDR` to avoid the stutter

pkg/doc/presexch/models.go
```diff
@@ -0,0 +1,100 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package presexch
+
+import (
+	"github.com/hyperledger/aries-framework-go/component/models/presexch"
+)
+
+// PresentationSubmission is the container for the descriptor_map:
```
DRK3, line 13: Why did you redefine all these types?

sudeshrshetty (approved):

pkg/doc/presexch/models.go
Moopli: They're not redefined - this package is the original location of all these types, so it's the import that all consumers of the package are currently using. Rather than break all consumers of refactored packages, I've been doing this sort of thing in all my refactor PRs. Original names in original locations can continue to be used, and migrated to using the new packages/modules at our convenience.

component/vdr/mock/mock_vdr.go
Moopli: Done!

pkg/doc/presexch/models.go
DRK3: Makes sense, thanks!

DRK3 (approved):

Moopli:
Split the lint check out, with component linting pulled out into its own github check - timeouts were too tight.

PR #3598, opened 2023-06-08
refactor: cleanup messy code in pEx implementation

Moopli:
- pEx implementation no longer needs to replace Credential ID temporarily
- assorted cleanup

### DIFF
--- a/component/models/presexch/definition.go
+++ b/component/models/presexch/definition.go
@@ -580,15 +580,13 @@ func (pd *PresentationDefinition) matchRequirement(req *requirement, creds []*ve
 		var matchedVCs []*verifiable.Credential
 
 		if opts.applySelectiveDisclosure {
-			matchedVCs, err = limitDisclosure(filtered, opts.credOpts...)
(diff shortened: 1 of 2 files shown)


vkubiv (approved):

component/models/presexch/definition.go
```diff
@@ -693,10 +713,10 @@ func (pd *PresentationDefinition) applyRequirement(req *requirement, creds []*ve
 
 	exclude := map[string]struct{}{}
 
-	for k := range set {
-		if !req.isLenApplicable(len(set[k])) {
-			for desc, cID := range set[k] {
-				exclude[desc+cID] = struct{}{}
+	for _, descriptors := range satisfiedDescriptors {
```
Moopli, line 716: Something I noticed here: this code expects a single VC (and its BBS derived creds) to satisfy the correct number of input descriptors for a submission requirement that has nested submission requirements. As far as I know, this isn't quite right: the parent submission requirement should validate the number of satisfied nested submission requirements (not the number of input descriptors), and it shouldn't require that they be satisfied by a single credential.

PR #3605, opened 2023-06-30
fix: pEx excludes non-SD VCs when SD is required

Moopli:


rolsonquadras (approved):

component/models/presexch/definition.go
```diff
@@ -946,6 +943,11 @@ func limitDisclosure(filterResults []constraintsFilterResult,
 			}
 		}
 
+		if constraints.LimitDisclosure.isRequired() && !predicate &&
+			credential.SDJWTHashAlg == "" && !hasBBS(credential) && !subjectIsIssuer(credential) {
```
vkubiv, line 947: Maybe it will be good to have some function like `supportsSelectiveDisclosure` etc. And it will contain these conditions `credential.SDJWTHashAlg == "" && !hasBBS(credential)`. It will be more readable and can be reused in other places. and if we will have some other type of credentials with selective disclosure we can add additional condition only in one place.

vkubiv (approved):

component/models/presexch/definition.go
Moopli: That's a good point. The annoying thing I noticed is that the conditions aren't always exactly the same - sometimes we want specifically BBS, sometimes specifically SDJWT, sometimes the condition is something like "either predicate or selective disclosure" etc.

I could make a few helper functions like `supportsSelectiveDisclosure`, `isSDJWTCredential`, etc, that way the logic is made more explicit.

sudeshrshetty (approved):

PR #3608, opened 2023-07-25
feat: vc data integrity API and structure

Moopli:


component/models/dataintegrity/models/models.go
```diff
@@ -0,0 +1,51 @@
...
+Copyright Gen Digital Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package models
+
+import "time"
+
+// TODO integrate VerificationMethod model with the did doc VM model
```
Moopli, line 11: In my next PR I'll be taking a look at the TODOs added in this one. If y'all want, I could remove these TODOs for less clutter.

rolsonquadras (approved):

component/models/dataintegrity/models/models.go
DRK3: I'm neutral on whether the TODOs stay in or not, but thanks for the heads up/explanation!

component/models/dataintegrity/signer.go
```diff
@@ -0,0 +1,102 @@
...
+// suites to perform data integrity signing.
+func NewSigner(suites ...suite.SignerInitializer) (*Signer, error) {
+	signer := &Signer{
+		suites: map[string]suite.Signer{},
+	}
+
+	for _, initializer := range suites {
+		suiteType := initializer.Type()
+
+		if _, ok := signer.suites[suiteType]; ok {
```
DRK3, line 35: Is this check here to ensure there aren't duplicates? I'm wondering if maybe you should state this in the NewSigner comment docs or perhaps make the input argument a map to begin with? Or maybe that makes the parameter too complicated.... hmm. Something to think about

component/models/dataintegrity/signer.go
```diff
@@ -0,0 +1,102 @@
...
+// ErrProofGeneration.
+func (s *Signer) AddProof(doc []byte, opts *models.ProofOptions) ([]byte, error) { // nolint:gocyclo
+	signerSuite, ok := s.suites[opts.SuiteType]
+	if !ok {
+		return nil, ErrUnsupportedSuite
+	}
+
+	proof, err := signerSuite.CreateProof(doc, opts)
+	if err != nil {
+		return nil, ErrProofGeneration
```
DRK3, line 72: I see you're not wrapping the error here - is that intentional? (same question applies for the other error returns in this function that don't wrap lower-level errors)

component/models/dataintegrity/verifier.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if !proofRaw.Exists() {
+		return ErrMissingProof
+	}
+
+	proof := &models.Proof{}
+
+	err := json.Unmarshal([]byte(proofRaw.Raw), proof)
+	if err != nil {
+		return ErrMalformedProof
```
DRK3, line 93: Intentionally not wrapping errors?

DRK3 (approved):
Minor comments to consider

component/models/dataintegrity/signer.go
Moopli: Yes, essentially. Without a check, then multiple instances of the same suite might be initialized (if, for example, you have an API that initializes some defaults and then you also pass in a list of suites you want to include, they might overlap), but it doesn't make sense to have multiple versions of the same suite active in the same data-integrity signer/verifier. So, we'd still only have one instance of each suite available in the signer/verifier, meaning all the others were initialized for no reason. Since initialization might mean initializing some sort of key store etc, I thought it would be better to avoid letting duplicates initialize in the first place.

sudeshrshetty (approved):

component/models/dataintegrity/signer.go
Moopli: Yup!

component/models/dataintegrity/verifier.go
Moopli: Yup! Same with all the others.

Actually, particularly on the verification side, it might be worthwhile to ensure each error also has a constant execution path to avoid side-channel sniffing of the source of the error.

